PR #32, opened 2017-08-21
Redux saga

ggazzo:
refactoring to use redux and redux saga, better login screen and more (:


gdelavald:
Can we fix the conflicts and get this merged? :smile:

PR #34, opened 2017-08-31
Redux saga

ggazzo:


ggazzo:
@gdelavald

gdelavald:
@ggazzo can you fix the CodeFactor issue?

ggazzo:
@gdelavald I am perfectly able, but I dont want :), actually I would love fix this in a better moment, I will just ~fix...

PR #40, opened 2017-09-01
Components moved to 'containers' or 'presentation' folders.

gilmarsquinelato:
Extracted some components from Message component to have less logic in component.


app/containers/Avatar.js
```diff
@@ -15,27 +15,42 @@ const styles = StyleSheet.create({
...
+			fontSize: size / 2
+		};
+
+		const avatarStyle = {
+			width: size,
+			height: size
+		};
+
+		const uri = avatar || `${ baseUrl }/avatar/${ text }`;
+		const image = (avatar || baseUrl) && (
```
ggazzo: its wrong , we need the text behind the image. if the image is a svg type, the image cache will not render, in case of network error to.

.vscode/launch.json
```diff
@@ -0,0 +1,41 @@
+{
```
ggazzo: remove this file and ignore it on git :)

app/containers/Avatar.js
gilmarsquinelato: But the text is before the image, just in a variable to reduce code complexity.

PR #41, opened 2017-09-06
[NEW] Changed navigation library to react-navigation.

gilmarsquinelato:
Refactoring on sagas to remove navigation navigation flow do it in views.
Added new modal library.
Renamed view files to the same name of their classes.
Added support to redux-devtools using the chrome extension through Remote devtools menu.
Some components was refactored to reduce complexity and more legibility.


gilmarsquinelato:
@ggazzo I've added a drawer menu containing all added servers using realm, just access and check it out :).

PR #44, opened 2017-09-23
Register user

gilmarsquinelato:
Created a component with fields Name, Email and password.
Created actions to receive the user data and dispatch to rocketchat lib to create user.


gilmarsquinelato:
@ggazzo Thanks for the feedback, I'll make the changes :)

PR #68, opened 2017-11-13
Fix storybook and splash screen

rodrigok:


ggazzo:
I will close this PR... we will continue it here.

Issue #55, opened 2017-11-03
Messages and attachments types


diegolmello:
- [x] image

- [x] url

- [x] video

- [x] audio

PR #82, opened 2017-11-16
Messages actions

diegolmello:
- [x] delete
- [x] edit
- [x] copy
- [x] permalink
- [x] quote
- [x] reply
- [x] starmessage
- [x] pinmessage


rodrigok:
@diegolmello Lint is broken

diegolmello:
Tap and hold on message to open actions.

screen shot 2017-11-17 at 15 07 36
screen shot 2017-11-17 at 15 08 01

app/containers/message/Card.js
```diff
@@ -55,7 +55,7 @@ export default class Cards extends React.PureComponent {
 					</Card>
 				</TouchableOpacity>
 				<PhotoModal
-					title={this.props.data.title}
+					title={this.props.data.title || 'test'}
```
ggazzo: 🤔

diegolmello: Just a workaround to fix PropTypes.string.isRequired.
It will be fixed later on #55 .

PR #116, opened 2017-11-28
[NEW] Open room on tap(push notification)

ggazzo:


app/containers/routes/NavigationService.js
```diff
@@ -21,3 +21,22 @@ export function goBack() {
 		config.navigator.dispatch(action);
 	}
 }
+
+
+export function goRoom({ rid, name }, counter = 0) {
+	if (!rid || !name || counter > 10) {
```
diegolmello: Why `counter` must be higher than 10?

PR #130, opened 2017-12-05
[FIX] messagebox growing width and icons

ggazzo:
image
image


android/app/src/main/AndroidManifest.xml
```diff
@@ -1,63 +0,0 @@
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
```
diegolmello: Why did you empty it?

diegolmello:
screen shot 2017-12-06 at 08 12 46
@ggazzo iOS raised this error

PR #140, opened 2017-12-14
[FIX] regexp server validation

### LINKED ISSUE #137: Cannot add custom rocket.chat server
I have an adding server issue.
I downloaded application from Android Store and I cannot add an url to my own rocket chat server, which is perfectly working with regular - cordova application. I receive an error "is not a valid rocket chat instance". The only server that is correct is https://open.rocket.chat.

Maybe there is something that as an user I can do? Or maybe it is an issue in this app.

Regards.

ggazzo:
closes #137
image


sampaiodiego:
you should may consider using this one `^(?=.{1,253}\.?$)(?:(?!-|[^.]+_)[A-Za-z0-9-_]{1,63}(?<!-)(?:\.|$)){2,}$`. see https://regex101.com/r/d5Yd6j/1/tests

PR #192, opened 2018-01-13
Improves messageBox

### LINKED ISSUE #191: Improve message box

saketkumar95:
@RocketChat/ReactNative

Closes #191 

- [x] Enhanced the message box same as on web. 
If there is no text then the **+** icon will be shown and if there is any text then the **send** icon will be shown which is same on the web.


ggazzo:
image
image

PR #256, opened 2018-03-28
Password show/hide feature

### LINKED ISSUE #234: Show/Hide option in password input

abhishek71994:
@RocketChat/ReactNative

Closes #234 

screen shot 2018-03-29 at 2 21 39 am
So the Hide Button interchanges with Show. Need some advice on how to style it in. 
Also should we use text or some picture?

### DIFF
--- a/app/views/LoginView.js
+++ b/app/views/LoginView.js
@@ -30,7 +30,7 @@ class LoginView extends React.Component {
 		this.state = {
 			username: '',
 			password: '',
-			showPassword:false,
+			showPassword: false
 		};
 	}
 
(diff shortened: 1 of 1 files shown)


ggazzo:
@abhishek71994 thanks for contribute!!! I will make few changes, ok?

abhishek71994:
Yeah ok! Can you please let me know the errors that I'm committing? So that I won't make the same mistakes the next time I contribute?

ggazzo:
sure! I will do a review, and submit my commit (:

ggazzo:
@abhishek71994 could you fix the conflicts? I think you are using a old development

app/views/LoginView.js
```diff
@@ -103,18 +104,25 @@ class LoginView extends React.Component {
...
-								onChangeText={password => this.setState({ password })}
-								secureTextEntry
-								autoCorrect={false}
-								returnKeyType='done'
-								autoCapitalize='none'
-								underlineColorAndroid='transparent'
-								onSubmitEditing={this.submit}
-								placeholder={this.props.Accounts_PasswordPlaceholder || 'Password'}
-							/>
+							<View style={{ flexDirection: 'row', borderColor: 'black' }}>
```
ggazzo: - we have an component called TextInput, plz insert this logic ( `showPassword` ... etc ) inside of component and use it... 

- use an icon instead of text 

- use a position absolute and a wrap to put the icon inside of input

PR #333, opened 2018-06-22
Switch to react-native-navigation

diegolmello:
@RocketChat/ReactNative

Moving from react-navigation to react-native-navigation since we already do a lot of JS job.


diegolmello:
Ancestor of #346

PR #351, opened 2018-07-05
Added the events for unsubscribe calls

reyalpsirc:
@RocketChat/ReactNative

Closes #ISSUE_NUMBER

There was no code checking for when the unsubscription happened and so it was never removed.


reyalpsirc:
@diegodorgam I as using the DDP.js file alone on another project and noticed that there was no subscription to the "unsub"/"nosub" messages. However, it seems that when unsubscribing from "stream-room-messages", this error happens: [https://forums.rocket.chat/t/unsubscribe-from-stream-room-messages/1500](https://forums.rocket.chat/t/unsubscribe-from-stream-room-messages/1500)

I can't provide the project itself where I found the issue since I'm not allowed too but, any call to a unsubscribe method with a log on on the Promise resolve for the "unsubscribe" should be enough to see that it is never called without this

reyalpsirc:
@diegolmello I just updated this because the problem was not totally corrected. I basically have a TTS implementation of this file and noticed that when I called the unsubscribe method and them the subscribe again, the old events linked to the previous subscribe would be revived when they actually shouldn't.

diegolmello:
@reyalpsirc Lint didn't pass. Can you fix it?

app/lib/ddp.js
```diff
@@ -290,9 +296,12 @@ export default class Socket extends EventEmitter {
 				id,
 				name,
 				params,
-				unsubscribe: () => this.unsubscribe(id)
+				unsubscribe: () => {
+					// Also remove the events attached to the subscription of the collection
```
ggazzo: use `removeAllListeners` instead

app/lib/ddp.js
```diff
@@ -290,9 +296,12 @@ export default class Socket extends EventEmitter {
...
+					this.removeAllForEvent(name);
```
ggazzo: you should call this method `removeAllListeners` inside of `unsubscribe`

reyalpsirc: to use "removeAllListeners" there then we need to pass the name as an extra parameter on unsubscribe. Otherwise there's no way to link them together.

reyalpsirc:
@ggazzo I just updated the file

app/lib/ddp.js
```diff
@@ -39,6 +39,11 @@ class EventEmitter {
 			}
 		}
 	}
+	removeAllForEvent(event) {
```
diegolmello: We fixed some bugs at #379.
`removeAllForEvent` doesn't seem to be necessary anymore, since every subscribe closes its listener.

app/lib/ddp.js
```diff
@@ -113,6 +118,7 @@ export default class Socket extends EventEmitter {
 
 		this.on('result', data => this.ddp.emit(data.id, { id: data.id, result: data.result, error: data.error }));
 		this.on('ready', data => this.ddp.emit(data.subs[0], data));
+		this.on('nosub', data => this.ddp.emit(data.id, data));
```
diegolmello: This line is useful.

app/lib/ddp.js
reyalpsirc: Hmm, I made this patch from a version of ddp.js that I have in typescript (made for a specific React Native project). In my case, I needed it so that I could clear all events that were added while the stream 'stream-room-messages' was open

diegolmello: Ok. Sounds like it's a different behaviour than ours.
In our case, we would clear events one by one.
For us, `removeAllForEvent` could cause unpredictable behaviour in future.

PR #443, opened 2018-09-14
[FIX] Avoid errors on Audio/Image/Video

diegolmello:
@RocketChat/ReactNative


SteveUNSW:
Is upload video still live?

There isn't an option to upload video on mobile, just photos.

PR #558, opened 2018-12-04
Use Rest API calls

diegolmello:
@RocketChat/ReactNative


faizanzahiddeloitte:
i will test and give my findings, good work so far.

FaizanZahid:
i will test and give my findings, good work so far.

FaizanZahid:
My feedback:

1. Still when i type my message in  the message input box, it takes some seconds lag before the send button changes to send blue arrow button.

2. Long tapping on a message shows no popup for message options.

3. Overall the app looks a bit slow in doing actions as compared to the native kotlin android app. Like opening images, loading old chats , opening a room for first time. etc Why its not fast like fb messenger, when these both apps r on react native.

4. there should be a way when i am seeing the files attachments in full screen, i should be able to swipe left and right to go to next images instead of closing each image every-time to see next one.

5. Long and fast scroll down on files attachment page still shows a blank white page. Unable to fully scroll to the end of the files attachments. it shows rescrolls mid way on top.

6. i tested these changes on 1.6.1.5184

7. Is this the same app for iOS react native too? or the iOS one is more stable? Which i0S version should we deploy for production, the native swift one for iOS or the react native one for iOS.?

PR #583, opened 2019-01-14
[NEW] Add Bubble Messages 

fifou17:
screenshot_2019-01-14-20-36-17


diegolmello:
@fifou17 That's a nice work, but (as we talked on dm) this doesn't follow our design pattern.
We redesigned our Message component about 3 months ago.
We don't plan to support bubble messages in near future.
I'll close this PR for now.
Thanks for your work :)

PR #588, opened 2019-01-25
Change avatar image problem

AugustoAleGon:
@RocketChat/ReactNative

Closes #ISSUE_NUMBER
Closes#441 Partially

If you are in a low connection on your phone. And the upload of images files in the message screen and you want to delete it, the app is gonna crash.

Also the update image problem was partially solved. Changing the url anytime the Avatar component is mounted, that will reload the image.

### DIFF
--- a/app/containers/Avatar.js
+++ b/app/containers/Avatar.js
@@ -26,7 +26,7 @@ export default class Avatar extends PureComponent {
 		const {
 			text, size, baseUrl, borderRadius, style, avatar, type, children
 		} = this.props;
-
+		const timeStamp = Math.floor(Date.now() / 1000);
 		const avatarStyle = {
 			width: size,
 			height: size,
@@ -40,7 +40,7 @@ export default class Avatar extends PureComponent {
 		const room = type === 'd' ? text : `@${ text }`;
 		// Avoid requesting several sizes by having only two sizes on cache
 		const uriSize = size === 100 ? 100 : 50;
-		const uri = avatar || `${ baseUrl }/avatar/${ room }?format=png&width=${ uriSize }&height=${ uriSize }`;
+		const uri = avatar || `${ baseUrl }/avatar/${ room }?format=png&width=${ uriSize }&height=${ uriSize }&ts=${ timeStamp }`;
 
 		const image = (
 			<FastImage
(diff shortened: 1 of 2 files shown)


app/views/RoomView/UploadProgress.js
```diff
@@ -157,7 +158,7 @@ export default class UploadProgress extends Component {
 						<Text style={styles.tryAgainButtonText}>{I18n.t('Try_again')}</Text>
 					</TouchableOpacity>
 				</View>
-				<Icon name='close' size={20} color='#9EA2A8' onPress={() => this.deleteUpload(item)} />
+				<Icon name='close' size={20} color='#9EA2A8' onPress={() => this.deleteUpload(this.uploads)} />
```
diegolmello: Calling `deleteUpload` with `this.uploads` would remove all uploads instead of just the tapped item.

app/containers/Avatar.js
```diff
@@ -40,7 +40,7 @@ export default class Avatar extends PureComponent {
 		const room = type === 'd' ? text : `@${ text }`;
 		// Avoid requesting several sizes by having only two sizes on cache
 		const uriSize = size === 100 ? 100 : 50;
-		const uri = avatar || `${ baseUrl }/avatar/${ room }?format=png&width=${ uriSize }&height=${ uriSize }`;
+		const uri = avatar || `${ baseUrl }/avatar/${ room }?format=png&width=${ uriSize }&height=${ uriSize }&ts=${ timeStamp }`;
```
diegolmello: Adding `ts` to the url will break the cache and it's not an option.
Your solution will make the app download the same avatars over and over.

app/views/RoomView/UploadProgress.js
AugustoAleGon: I see. What would be for just the item? Because the `item` is not a valid object.

app/containers/Avatar.js
AugustoAleGon: I know. But the only way the image will update once is fetching from the url is that url change. I tried to change in the `FastImage` the way cache works using `FastImage.cacheControl.web` but I got the same result. No update.

AugustoAleGon:
[https://imgur.com/a/upNCq8v](Bug after cancel download)

app/containers/Avatar.js
SteveUNSW: Any other luck?

PR #687, opened 2019-03-07
Switch to react-navigation

diegolmello:
@RocketChat/ReactNative


alleniver:
awesome ! This is the component that more people use! i suggest changing redux-saga to redux-thunk as well, because redux-saga is very beautiful, but redux-thunk uses more people. This component should cater to more people.

PR #692, opened 2019-03-08
adding extra condition to handle invisible status

### LINKED ISSUE #691: Cannot set invisible status
- Your Rocket.Chat Experimental app version: 1.10.0.99
- Your Rocket.Chat server version: ####
- Device (or Simulator) you're running with: Redmi note 4 API 7

Try setting your status to invisible it won't set but if you check it on the web it is chaged by actions on the app, it seems that upcomming calls from sever are not handled correctly for this ,

pranavpandey1998official:
the server response for invisible or offline setting doesn't have fields parameter hence required action cannot be dispatched it has a cleared field so adding an extra condition  for it
fixes:#691

@RocketChat/ReactNative

Closes #691


diegolmello:
@pranavpandey1998official Nice! If everything is working, I'll merge it into 1.11.

PR #738, opened 2019-03-18
[Add] Add feature to read and unread messages 

### LINKED ISSUE #709: [NEW] Mark room as read/unread

pranavpandey1998official:
Add feature to read and unread messages and also favorite messages from roomListView

fixes:#709

@RocketChat/ReactNative

Closes #709

git commit


pranavpandey1998official:
@diegolmello does hide bottom archive the message or there is a different api for it

diegolmello:
> @diegolmello does hide bottom archive the message or there is a different api for it

@pranavpandey1998official  https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/rocketchat.js#L622

diegolmello:
@pranavpandey1998official You probably don't need `isRead` in `RoomItem`, since there's `unread` prop.

pranavpandey1998official:
there are other factors too which are deciding if unread or not as here https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/bf9e3d96a5e2e78295032f85939697ba9818bf1a/app/views/RoomsListView/index.js#L276 i have just copied this condition and make it a different prop

diegolmello:
@pranavpandey1998official subscriptions are updated automatically via DDP here: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/methods/subscriptions/rooms.js#L42

There's a delay until the data is received from the socket and maybe your solution has a better experience, but be aware of it anyway.
Instead of creating another prop or waiting for socket to receive the updates, maybe you should update Realm object.
Open a query and update the object.
RoomsListView should update the data automatically.

pranavpandey1998official:
> > @diegolmello does hide bottom archive the message or there is a different api for it
> 
> @pranavpandey1998official https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/rocketchat.js#L622

this is for archiving rooms, in the docs, I got different sections for hiding and archiving rooms plus I am getting an error as 'I am not authorized to archive this room' for some channels, are you sure this is the right api to call here

diegolmello:
@pranavpandey1998official Forget about hiding.
Archive requires `archive-room` permission.
You should check before calling it.

pranavpandey1998official:
> @pranavpandey1998official Forget about hiding.
> Archive requires `archive-room` permission.
> You should check before calling it.

So what should the button do if the user doesn't have 'archive-room' permission??

diegolmello:
@pranavpandey1998official You shouldn't show it.

pranavpandey1998official:
> @pranavpandey1998official subscriptions are updated automatically via DDP here: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/methods/subscriptions/rooms.js#L42
> 
> There's a delay until the data is received from the socket and maybe your solution has a better experience, but be aware of it anyway.
> Instead of creating another prop or waiting for the socket to receive the updates, maybe you should update Realm object.
> Open a query and update the object.
> RoomsListView should update the data automatically.

ok I tried it so the problem I am facing are 
1) we need that prop no matter what because unread prop alone is not enough for the task ,
2) I opened the query and updated the object but I don't know which property the API is changing in the Realm object  I could just change anyone and let socket do the correction but I don't think that will be the correct approach if you could tell me which property I should change it could be implemented

diegolmello:
@pranavpandey1998official Ok, do as you think is for the better.

app/views/RoomView/index.js
```diff
@@ -52,15 +52,12 @@ export default class RoomView extends LoggedView {
 		const rid = navigation.getParam('rid');
 		const t = navigation.getParam('t');
 		const f = navigation.getParam('f');
```
diegolmello: Remove this line and https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/738/commits/6e001300dd8e943b20fd0ba4427b9bf128bd231a#diff-31da83bd6996164d4ba4ea85b1e303ceR148

diegolmello:
@pranavpandey1998official lint isn't passing

pranavpandey1998official:
@diegolmello https://github.com/kmagiera/react-native-gesture-handler/issues/344 known issue

PR #767, opened 2019-03-28
Fullscreen ReactionPicker

### LINKED ISSUE #765: ReactionsPicker should be full screen
Currently, ReactionPicker is a square modal, but this is not a great ux:

Instead, we should open a full screen modal like below:

paulvazbad:
@RocketChat/ReactNative

Closes #765


diegolmello:
Thanks, @paulvazbad.
It's not just about window size, but ux as you can see here:


It should be a react-navigation's modal, scroll vertically, search, etc

PR #782, opened 2019-04-04
placeholder bug fixed

aryankoul:
@RocketChat/ReactNative

Input in upload modal was displaying full text instead of hint text


aryankoul:
> Thanks!
> If the user submit a file upload without a filename, it'll raise an error though.
> Can you prevent him to do it? It's related to the placeholder bug anyway.

@diegolmello i have put a check for filename in upload model

app/containers/MessageBox/UploadModal.js
```diff
@@ -119,6 +119,10 @@ export default class UploadModal extends Component {
 	submit = () => {
 		const { file, submit } = this.props;
 		const { name, description } = this.state;
+		if (name === '') {
+			ToastAndroid.show('Please enter a filename', ToastAndroid.SHORT);
```
diegolmello: What about iOS?

aryankoul: @diegolmello i have changes Toast to alert.

app/containers/MessageBox/UploadModal.js
```diff
@@ -119,6 +119,10 @@ export default class UploadModal extends Component {
 	submit = () => {
 		const { file, submit } = this.props;
 		const { name, description } = this.state;
+		if (name === '') {
+			Alert.alert('Oops!', 'Filename is required.');
```
diegolmello: Use I18n

app/containers/MessageBox/UploadModal.js
```diff
@@ -119,6 +119,10 @@ export default class UploadModal extends Component {
 	submit = () => {
 		const { file, submit } = this.props;
 		const { name, description } = this.state;
+		if (name === '') {
```
diegolmello:
```jsx
if (!name) {
```

PR #766, opened 2019-03-27
[FIX] Replies don't render images

aryankoul:
@RocketChat/ReactNative

Closes #704 

**The reply of an image now renders full image in the reply box**

Screenshot_1553721134

**The reply preview shows thumbnail of image in the component**

Screenshot_1553721171

PR #783, opened 2019-04-04
[FIX] reply bug fixed

aryankoul:
@RocketChat/ReactNative

Fixed a bug related to replyPreview component. If the replyPreview component was rendered already and we click on another message for reply, the props in the replyPreview component did not update. It showed the previous message only.


diegolmello:
@aryankoul Fix it in #766. Thanks.

PR #812, opened 2019-04-13
added dropdown to select between http and https

### LINKED ISSUE #802: Getting error while adding my server url in sign in your server page.
- Your Rocket.Chat Experimental app version: 1.11.0
- Your Rocket.Chat server version: 0.74.3
- Device (or Simulator) you're running with: Device

**The app isn't connecting to your server?**
Make sure your server supports WebSocket. These are the minimum requirements for Apache 2.4 and Nginx 1.3 or greater.

Getting error Oops! The URL you entered is invalid. Check it and try again, please!

pranavpandey1998official:
@RocketChat/ReactNative

Closes  #802

WhatsApp Image 2019-04-13 at 1 08 06 PM


app/views/NewServerView.js
```diff
@@ -128,21 +134,14 @@ export default class NewServerView extends LoggedView {
...
-
-		if (/^(https?:\/\/)?(((\w|[0-9])+(\.(\w|[0-9-_])+)+)|localhost)(:\d+)?$/.test(url)) {
-			if (/^localhost(:\d+)?/.test(url)) {
-				url = `http://${ url }`;
-			} else if (/^https?:\/\//.test(url) === false) {
-				url = `https://${ url }`;
-			}
-		}
-
+		url = protocolType + url;
```
IlarionHalushka: prefer using templates instead of concatenation

app/views/NewServerView.js
```diff
@@ -182,17 +181,30 @@ export default class NewServerView extends LoggedView {
...
-							containerStyle={styles.inputContainer}
-							placeholder={defaultServer}
-							value={text}
-							returnKeyType='send'
-							onChangeText={this.onChangeText}
-							testID='new-server-view-input'
-							onSubmitEditing={this.submit}
-							clearButtonMode='while-editing'
-						/>
+						<View
```
IlarionHalushka: might consider moving this View to a separate function for a better readability 🤔

app/views/NewServerView.js
```diff
@@ -182,17 +181,30 @@ export default class NewServerView extends LoggedView {
...
+							style={styles.inputContainer}
+						>
+							<Picker
+								mode='dropdown'
+								selectedValue={protocolType}
+								onValueChange={this.handlePickerValueChange}
+								style={styles.picker}
+							>
+								<Picker.Item label='https://' value='https://' />
```
IlarionHalushka: for repeating strings it's better to define constants `protocol.HTTP` and `protocol.HTTPS`

app/views/NewServerView.js
```diff
@@ -35,27 +35,25 @@ const styles = StyleSheet.create({
 		alignSelf: 'center'
 	},
 	inputContainer: {
+		flex: 1,
 		marginTop: 25,
-		marginBottom: 15
-	},
-	input: {
```
IlarionHalushka: not sure why you changed the styles and removed sharedStyles?

app/views/NewServerView.js
pranavpandey1998official: cool

pranavpandey1998official: maybe += is better

IlarionHalushka: not critical at all, I just prefer using templates when working with strings (and templates are used all around in this project)

app/views/NewServerView.js
pranavpandey1998official: input style was not being used previously, I just have created a new style here and named it input

diegolmello:
I don't think this is a good user experience.
And we shouldn't help users to use insecure servers.

pranavpandey1998official:
@diegolmello then I think we must inform them to use secure server

Tom-Development:
I believe this is still an issue, on the react mobile app we cannot still join servers without HTTPS. 

This is not any good forcing people to use HTTPS sites when in our case are on local testing environments. There should still be an option to change between the two.

PR #815, opened 2019-04-14
add email validation on RegisterView

IlarionHalushka:
@RocketChat/ReactNative

Currently Register screen doesn't display any validation messages - only Register button is disabled if some value is invalid.

This PR adds a validation to email field so that when email format is invalid it displays validation message and makes the border red.

Behavior before (no validation, only disabled button 'Register'): 
image

Behavior after this PR:
https://drive.google.com/file/d/1D365W1cGhLWlX9KAoUgJ6MDitnbYneCG/view?usp=sharing


app/views/RegisterView.js
```diff
@@ -142,9 +172,10 @@ export default class RegisterView extends LoggedView {
 							returnKeyType='next'
 							keyboardType='email-address'
 							iconLeft='mail'
-							onChangeText={email => this.setState({ email })}
+							onChangeText={email => this.onEmailInputChange(email)}
```
diegolmello: `onChangeText={this.onEmailInputChange}`

app/views/RegisterView.js
```diff
@@ -84,13 +103,24 @@ export default class RegisterView extends LoggedView {
 		const {
 			name, email, password, username
 		} = this.state;
-		return name.trim() && email.trim() && password.trim() && username.trim() && isValidEmail(email);
+		return name.trim() && email.trim() && password.trim() && username.trim();
+	}
+
+	validEmail = () => {
```
diegolmello: Can we change this function name?
This way we have `validEmail` and `isValidEmail` inside the same class.
Maybe `isValidEmailValidation` for the util.
What do you think?

app/views/RegisterView.js
```diff
@@ -58,6 +60,10 @@ export default class RegisterView extends LoggedView {
 	}
 
 	shouldComponentUpdate(nextProps, nextState) {
+		const { validationError } = this.state;
```
diegolmello: Instead of `validationError`, can we test a primitive?
Suggestion: add another state like `error: true/false` and add it to `shouldUpdateState` const

djorkaeffalexandre:
We're validating the email before enable the `submit` button.
Closing it.

PR #816, opened 2019-04-15
add precommit hooks to run: eslint fix, eslint, jest

IlarionHalushka:
@RocketChat/ReactNative

We all could notice that in our PRs usually the last commits are called 'fix lint errors' or 'update jest snapshot'. And the CI usually fails and needs to re-run job `lint-testunit`.
In order to avoid these irritating commits and to reduce the CI re-run this PR introduces pre-commit hooks.

Firstly `"eslint . --fix"` is run to format the code according to eslint rules.

After that commit won't be processed if any of the following commands fail:
-       "eslint ."
-       "jest ./__tests__/*"


package.json
```diff
@@ -13,6 +13,19 @@
...
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.js": [
+      "eslint . --fix",
+      "eslint .",
+      "jest ./__tests__/*",
+      "git add ."
```
pranavpandey1998official: I don't think it is a good idea to stage all files all the time

package.json
```diff
@@ -13,6 +13,19 @@
...
     "snyk-protect": "snyk protect"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.js": [
+      "eslint . --fix",
```
pranavpandey1998official: in windows this always fails due to 		https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/5aec0ec1860b32158f4c69bd16cca14fa34c70f6/.eslintrc.js#L121 , So I dont think its a good idea

package.json
IlarionHalushka: Probably you are right, never thought about it and always used this way on all projects. Thanks!

IlarionHalushka: But if I remove `eslint . fix` then there is no need to do `git add` so I will remove it anyway

package.json
IlarionHalushka: That's interesting about windows 🤔
Do you agree that running `"eslint .",` and `"jest ./__tests__/*",` in pre-commit is useful and will save lots of time and some nerves? 🙂

pranavpandey1998official: yes , it is but eslint --fix don't solve all the problems. and the one setting I mentioned cannot be solved In windows,  you can read about it (https://stackoverflow.com/questions/37826449/expected-linebreaks-to-be-lf-but-found-crlf-linebreak-style) if we can change this setting , which i don't think this PR is ok

IlarionHalushka: I removed `eslint --fix` and `git add` as you suggested. So now it's just running eslint and tests in precommit.

pranavpandey1998official: yaah, but still eslint is going to fail and person using windows will not be able to push commits

IlarionHalushka:
Final edition: hooks run only jest tests and `eslint .`
image

package.json
IlarionHalushka: Thanks for checking again!
Hopefully this one will be the last edition of this PR :)
I disabled linebreak-style rule in eslint as it's not important. More important is to not push 💩 code in repo...

package.json
```diff
@@ -13,6 +13,17 @@
     "android": "react-native run-android",
     "snyk-protect": "snyk protect"
   },
+  "husky": {
+    "hooks": {
+      "pre-commit": "lint-staged"
```
diegolmello: - Instead of running tests manually, we should separate `"ci": "eslint . && jest && codecov"` into `eslint . && jest` and `codecov`.
For example:
```
// scripts
"pre-commit": "eslint . && jest",
"ci": "npm pre-commit && codecov",

// hooks
"pre-commit": "npm pre-commit"
```

- We should run it everytime, instead of `*.js` only.
For example: RN may add a11y that will break jest tests, because nothing has changed in js, but RN update changed view structure.

package.json
```diff
@@ -13,6 +13,17 @@
...
+    }
+  },
+  "lint-staged": {
```
diegolmello: Do you agree we should remove it?

package.json
```diff
@@ -99,9 +110,11 @@
     "eslint-plugin-jsx-a11y": "^6.2.1",
     "eslint-plugin-react": "^7.11.1",
     "eslint-plugin-react-native": "^3.3.0",
+    "husky": ">=1",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^23.6.0",
     "jest-cli": "^23.6.0",
+    "lint-staged": ">=8",
```
diegolmello: Here as well

package.json
```diff
@@ -13,6 +13,17 @@
...
+      "pre-commit": "lint-staged"
+    }
+  },
+  "lint-staged": {
+    "*.js": [
+      "jest ./__tests__/*",
+      "eslint ."
+    ]
+  },
   "rnpm": {
```
diegolmello: Since you're here, can you check if we still need this?
I don't think so.

IlarionHalushka: deprecated, removed.

package.json
IlarionHalushka: yes I agree that it should run before every commit, not just *js files

PR #818, opened 2019-04-15
[NEW] Report message

### LINKED ISSUE #701: Report Message
Add a message action to report it.

https://rocket.chat/docs/developer-guides/rest-api/chat/reportmessage/

pranavpandey1998official:
@RocketChat/ReactNative

Closes #701 

WhatsApp Image 2019-04-18 at 2 25 25 AM


app/i18n/locales/en.js
```diff
@@ -199,6 +199,7 @@ export default {
 	Message_pinned: 'Message pinned',
 	Message_removed: 'Message removed',
 	Messages: 'Messages',
+	Message_Reported: 'Message reported',
```
IlarionHalushka: you should also add translations for other languages using google translate :)

app/containers/MessageActions.js
```diff
@@ -293,6 +297,12 @@ export default class MessageActions extends React.Component {
 		toggleReactionPicker(actionMessage);
 	}
 
+	handleReport = async() => {
+		const { actionMessage } = this.props;
+		await RocketChat.reportMessage(actionMessage._id);
```
IlarionHalushka: wrap it in try-catch block and add showToast(error)

app/i18n/locales/en.js
```diff
@@ -257,6 +258,7 @@ export default {
 	Register: 'Register',
 	Repeat_Password: 'Repeat Password',
 	Reply: 'Reply',
+	Report: 'Report',
```
IlarionHalushka: translations

app/i18n/locales/en.js
pranavpandey1998official: In my personal experience, no translation is better than bad translation, someone familiar with the language should do it

app/containers/MessageActions.js
pranavpandey1998official: sure

app/i18n/locales/en.js
IlarionHalushka: you can also look up here for translations https://github.com/RocketChat/Rocket.Chat.iOS/tree/develop/Rocket.Chat/Resources/de.lproj

IlarionHalushka: and as there are only a few words to be translated - google translate does this pretty well. and I agree that with large text strings google translate might make small mistakes

app/containers/MessageActions.js
diegolmello: @pranavpandey1998official Add an Alert instead of toast. (we're planning to remove it)
Also add try/catch.

app/i18n/locales/en.js
diegolmello: IMO no translation is better than bad translation as well.

diegolmello: `en` is fine

app/containers/MessageActions.js
pranavpandey1998official: done

Issue #697, opened 2019-03-12
null is not an object (evaluating 'U.startApp')


weijia-yu:
- Your Rocket.Chat Experimental app version: 
`develop branch`
` eb4b1553bf4bed65cce9ca201652143e8f778943`
- Your Rocket.Chat server version: 
`develop branch` 
`eb4b1553bf4bed65cce9ca201652143e8f778943`
- Device (or Simulator) you're running with: 
Simulator Emualtor Nexus 5X API 28 x86
Android Studio version: 3.2.1

**Detail**
My app can run on other simulators such as `Pixel 2 XL API 26` and `Pixel 2 XL API 26`.
But it shows null is not an object `(evaluating 'U.startApp')` on `Nexus 5X API 28 x86`

malivp3494:
Can you please provide error log?

didacroyo:
Is it related to issue #591 ?

pranavpandey1998official:
will be Fixed in #896

Exordian:
i still hit this bug using the latest 'develop' branch (currently on 1a6011e9d11b98281b559ed7dfe87d7e96f9d989 )

arminfelder:
I solved this issue by deleting the android/app/src/main/assets/index.android.bundle
then run react-native start and copy the new bundle file with:
curl "http://localhost:8081/index.android.bundle?platform=android" -o "android/app/src/main/assets/index.android.bundle"

PR #879, opened 2019-05-07
Fixed Bundler Issue on Android

RobotJINI:
…ersions block cleartext traffic by default.

@RocketChat/ReactNative

Closes #ISSUE_NUMBER

https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/697

Starting with Android 9.0 (API level 28), cleartext support is disabled by default. We need to enable it.

https://stackoverflow.com/questions/45940861/android-8-cleartext-http-traffic-not-permitted/50834600#50834600

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -15,7 +15,8 @@
 		android:label="@string/app_name"
 		android:icon="@mipmap/ic_launcher"
 		android:theme="@style/AppTheme"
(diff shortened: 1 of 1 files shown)


RobotJINI:
The metro build wasn't connecting to the android devices because: Starting with Android 9.0 (API level 28), cleartext support is disabled by default.

RobotJINI:
exampleerror

diegolmello:
Fixed in #896

PR #906, opened 2019-05-20
[FIX] App crash with backspace on input message

### LINKED ISSUE #902: [BUG] - App crash with backspace on input message.
- Your Rocket.Chat Experimental app version: 1.13.0.7547
- Device (or Simulator) you're running with: iPhone X (iOS 12.3)

**Steps to reproduce**
- Open any chat
- Press backspace
- Back to chat list

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #902.


app/containers/MessageBox/index.js
```diff
@@ -189,7 +189,7 @@ class MessageBox extends Component {
 		this.setInput(text);
 
 		if (this.component) {
-			const { start, end } = this.component._lastNativeSelection;
+			const { start, end } = this.component._lastNativeSelection || { start: 0, end: 0 };
```
diegolmello: Doesn't it looks like a workaround instead of a fix?
I mean, when this code run `this.component. _lastNativeSelection` is empty, right?
If it's empty, isn't it better if we test `text`?

Just change from:
```jsx
if (this.component) {
```

to:
```jsx
if (!isTextEmpty) {
```

What do you think?

djorkaeffalexandre: Nice!!

djorkaeffalexandre:
@diegolmello I fixed your review.

PR #907, opened 2019-05-20
[IMPROVEMENT] Add toggle markdown to settings

diegolmello:
@RocketChat/ReactNative


app/i18n/locales/en.js
```diff
@@ -328,7 +329,6 @@ export default {
 	Thread: 'Thread',
 	Threads: 'Threads',
 	Timezone: 'Timezone',
-	Toggle_Drawer: 'Toggle_Drawer',
```
djorkaeffalexandre: Remove that in all locales.

PR #908, opened 2019-05-20
[IMPROVEMENT] Share channel

### LINKED ISSUE #897: [NEW] Share channel
Similar to "Share message", it must open OS context menu with the same channel link from the web.
The button exists, but is disabled on RoomActionsView.

For example: https://open.rocket.chat/channel/general

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #897


app/containers/MessageActions.js
```diff
@@ -155,7 +155,7 @@ export default class MessageActions extends React.Component {
 
 	getPermalink = async(message) => {
 		try {
-			return await RocketChat.getPermalink(message);
+			return await RocketChat.getPermalink(message, 'message');
```
IlarionHalushka: in folder `app/constants` create file `permalinkMessageTypes.js` and move strings `'message'`, `'room'` there.

app/views/RoomActionsView/index.js
```diff
@@ -353,6 +353,22 @@ export default class RoomActionsView extends LoggedView {
 		}
 	}
 
+	getRoomlink = async(room) => {
+		try {
+			return await RocketChat.getPermalink(room, 'room');
+		} catch (error) {
+			return null;
```
IlarionHalushka: also log the error

app/containers/MessageBox/index.js
```diff
@@ -341,7 +341,7 @@ class MessageBox extends Component {
 
 	getPermalink = async(message) => {
 		try {
-			return await RocketChat.getPermalink(message);
+			return await RocketChat.getPermalink(message, 'message');
```
IlarionHalushka: same file line 615: `const permalink = await this.getPermalink(replyMessage);` - add second argument

app/lib/rocketchat.js
```diff
@@ -508,10 +508,10 @@ const RocketChat = {
 		}
 		return Promise.resolve(result);
 	},
-	async getPermalink(message) {
+	async getPermalink(obj, type) {
 		let room;
 		try {
-			room = await RocketChat.getRoom(message.rid);
+			room = await RocketChat.getRoom(obj.rid);
```
IlarionHalushka: make sure `await RocketChat.getRoom` is necessary when the `type` is `'room'` and first argument `obj` is basically the room itself

djorkaeffalexandre:
@IlarionHalushka Fixed your reviews.

app/lib/rocketchat.js
```diff
@@ -509,21 +510,25 @@ const RocketChat = {
 		}
 		return Promise.resolve(result);
 	},
-	async getPermalink(message) {
+	async getPermalink(obj, type) {
```
diegolmello: It's not consistent.
We're planning to add TypeScript and this is the kind of methods that would require an `any` type on `obj` param.
IMO this method should be split in two: on for message and other for channel.
What do you think?

app/constants/types.js
```diff
@@ -2,3 +2,5 @@ export const SET_CURRENT_SERVER = 'SET_CURRENT_SERVER';
 export const SET_ALL_SETTINGS = 'SET_ALL_SETTINGS';
 export const SET_CUSTOM_EMOJIS = 'SET_CUSTOM_EMOJIS';
 export const ADD_SETTINGS = 'ADD_SETTINGS';
+export const SHARE_MESSAGE = 'SHARE_MESSAGE';
```
diegolmello: We need to remove this `app/constants/types` file in future.
I just never had a chance to do it (I'm not asking you to do it in this PR).
It's breaking the pattern.
Anyway, if you still need to create these consts, do it elsewhere.

app/lib/rocketchat.js
djorkaeffalexandre: nice!

djorkaeffalexandre:
@diegolmello fixed your reviews.

PR #909, opened 2019-05-21
[NEW] In app notification

pranavpandey1998official:
@RocketChat/ReactNative

Closes #ISSUE_NUMBER

Simulator Screen Shot - iPhone X - 2019-05-30 at 16 17 29


app/views/AdminPanelView/index.js
```diff
@@ -32,6 +33,7 @@ export default class AdminPanelView extends React.Component {
 		return (
 			<SafeAreaView style={styles.container} testID='terms-view'>
 				<StatusBar />
+				<NotificationBadge />
```
diegolmello: It's working nicely, but you're repeating your self too much in this line.
You could do like this: https://reactnavigation.org/docs/en/custom-navigators.html#extending-navigators

Example overwriting `InsideStackModal` on `app/index`:
```jsx
class CustomNavigator extends React.Component {
	static router = InsideStackModal.router;

	render() {
		const { navigation } = this.props;
		return (
			<React.Fragment>
				<InsideStackModal navigation={navigation} />
				<NotificationBadge navigation={navigation} />
			</React.Fragment>
		);
	}
}
```

You'll have to change your notification style a little, because it's going to open from the top of the app instead of the bottom of the header.

app/reducers/notification.js
```diff
@@ -0,0 +1,19 @@
+import * as types from '../actions/actionsTypes';
```
diegolmello:
```jsx
import { NOTIFICATION } from '../actions/actionsTypes';
...
switch (action.type) {
	case NOTIFICATION.RECEIVED:
...
```

app/notifications/inApp/index.js
```diff
@@ -0,0 +1,181 @@
...
+		flex: 1,
+		flexDirection: 'row',
+		alignItems: 'center',
+		justifyContent: 'flex-start'
+	},
+	avatar: {
+		marginHorizontal: 10
+	},
+	roomName: {
+		color: COLOR_TEXT
```
diegolmello: Use text styles from shared styles.

app/notifications/inApp/index.js
```diff
@@ -0,0 +1,181 @@
...
+	},
+	message: {
+		maxWidth: MAX_WIDTH_MESSAGE
```
diegolmello: Use text styles from shared styles.

app/notifications/inApp/index.js
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	componentDidUpdate() {
+		const { navigation } = this.props;
+		if (navigation.isFocused()) {
+			this.show();
+		}
+	}
+
+	show = () => {
```
diegolmello: show and hide aren't animating everytime.

app/views/AdminPanelView/index.js
pranavpandey1998official: @diegolmello 
isTransitioning:false
index:0
key:InsideStack
routeName:InsideStack
params:undefined

this is what navigation.state is returning see the route name is always InsideStack , there is no way of knowing which view I am If I this logic , and because of that I cannot check for condition in which notification is not shown if your is inside the same room .. also it create difficulty in navigating ...

PR #920, opened 2019-05-24
Messagebox typing and buttons refactor

diegolmello:
@RocketChat/ReactNative

- Another Messagebox workaround to RN's slowness
Related: #658 

- Refactor Messagebox's buttons


app/containers/MessageBox/index.js
```diff
@@ -787,8 +780,8 @@ class MessageBox extends Component {
 			return (<Recording onFinish={this.finishAudioMessage} />);
 		}
 		return (
-			[
-				this.renderMentions(),
+			<React.Fragment>
```
lolimay: A good exemplar for me :+1:

diegolmello:
> Looking closely at files `/MessageBox/LeftButtons.android.js` + `/MessageBox/LeftButtons.ios.js` and `/MessageBox/RightButtons.android.js` + `/MessageBox/LeftButtons.android.js` we may notice that 90% of the code is duplicated and it can be avoided by just checking one condition `isAndroid - then render additional button`.
> What do you think?

That was my first experiment, but it looked messy to me.
Let's stick with this way and we can change it in future if we think it's bad DX, ok?

PR #928, opened 2019-05-27
[CHORE] Use react-native-firebase

diegolmello:
@RocketChat/ReactNative

We need to migrate from deprecated react-native-fabric to react-native-firebase.
This PR enables following Firebase features:
- [x] Analytics
- [x] Crashlytics
- [x] Performance

It also tracks screen view without the necessity of HOC.

Future work:
I won't do it in this PR because it's large enough, but we need to log more app events, like 'sent_message', 'open_admin', 'media_upload', etc.


app/index.js
```diff
@@ -244,6 +258,14 @@ export default class Root extends React.Component {
 					ref={(navigatorRef) => {
 						Navigation.setTopLevelNavigator(navigatorRef);
 					}}
+					onNavigationStateChange={(prevState, currentState) => {
```
IlarionHalushka: maybe separate function?

app/views/View.js
```diff
@@ -1,14 +1,14 @@
 import React from 'react';
```
IlarionHalushka: is this component still required?

PR #933, opened 2019-05-29
[CHORE] Make e2e pass on CircleCI

IlarionHalushka:
@RocketChat/ReactNative

- [x] iOS e2e passing on ci
- [x] add docs to readme

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,6 @@ jobs:
     steps:
       - checkout
 
-      - restore_cache:
-          key: node-modules-{{ checksum ".circleci/config.yml" }}-{{ checksum "yarn.lock" }}
-
       - run:
           name: Install Node 8
           command: |
@@ -90,11 +87,6 @@ jobs:
 
       - store_artifacts:
           path: /tmp/screenshots
-
-      - save_cache:
-          key: node-modules-{{ checksum ".circleci/config.yml" }}-{{ checksum "yarn.lock" }}
-          paths:
-            - ./node_modules
 
   android-build:
     <<: *defaults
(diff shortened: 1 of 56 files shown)


e2e/README.md
```diff
@@ -0,0 +1,44 @@
...
+##### 1.2. Set up detox
+* Install dependencies by following this [guide](https://github.com/wix/Detox/blob/master/docs/Introduction.GettingStarted.md#step-1-install-dependencies) (only Step 1).
+
+### 2. Prepare test data
+* Run Rocket Chat meteor app: `meteor npm start` (make sure you to run this command from project that you created on Step 1.1.).
+* Open `localhost:3000` in browser.
+* Sign up as admin.
+* Create public room `detox-public`.
+* Create user with role: `user`, username: `detoxrn`, email: `detoxrn@gmail.com`, password: `123`.
+* Create user with role: `user`, username: `diego.mello`, email: `diego.mello@rocket.chat`, password: `123`.
```
diegolmello: Update all credentials to be generic.
I don't want my email on all tests 😛
It should be something like `YOUR_TEST_EMAIL@EMAIL.COM` and then you ask to update `./e2e.data` with same info.

.circleci/config.yml
```diff
@@ -241,9 +241,9 @@ workflows:
   build-and-test:
     jobs:
       - lint-testunit
-      # - e2e-test:
-      #     requires:
-      #       - lint-testunit
+      - e2e-test:
```
diegolmello: Tests take to long to run.
Add a hold step similar to `ios-hold-testflight` and then we can choose when to run them.

PR #959, opened 2019-06-04
[IMPROVEMENT] Lazy fetch server info

diegolmello:
@RocketChat/ReactNative

Check if user is logged to hide splash screen first and fetch server info later.
Also, stop alerting "Wrong url" in case of server is down.
The user is logged in, so the url is correct.

### DIFF
--- a/app/sagas/selectServer.js
+++ b/app/sagas/selectServer.js
@@ -11,11 +11,13 @@ import database from '../lib/realm';
 import log from '../utils/log';
 import I18n from '../i18n';
 
-const getServerInfo = function* getServerInfo({ server }) {
+const getServerInfo = function* getServerInfo({ server, raiseError = true }) {
 	try {
 		const serverInfo = yield RocketChat.getServerInfo(server);
 		if (!serverInfo.success) {
-			Alert.alert(I18n.t('Oops'), I18n.t(serverInfo.message, serverInfo.messageOptions));
+			if (raiseError) {
+				Alert.alert(I18n.t('Oops'), I18n.t(serverInfo.message, serverInfo.messageOptions));
+			}
 			yield put(serverFailure());
 			return;
 		}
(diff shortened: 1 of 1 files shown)


app/sagas/selectServer.js
```diff
@@ -11,11 +11,13 @@ import database from '../lib/realm';
 import log from '../utils/log';
 import I18n from '../i18n';
 
-const getServerInfo = function* getServerInfo({ server }) {
+const getServerInfo = function* getServerInfo({ server, raiseError = true }) {
 	try {
 		const serverInfo = yield RocketChat.getServerInfo(server);
 		if (!serverInfo.success) {
-			Alert.alert(I18n.t('Oops'), I18n.t(serverInfo.message, serverInfo.messageOptions));
+			if (raiseError) {
```
IlarionHalushka: why did you add `raiseError` parameter?
I suppose that checking `if (!serverInfo.success) {` would be enough to show error Alert.
And besides that `getServerInfo` is called only once in `handleSelectServer`: `getServerInfo({ server, raiseError: false });` so `raiseError` will always be falsy...

app/sagas/selectServer.js
```diff
@@ -11,11 +11,13 @@ import database from '../lib/realm';
...
+				Alert.alert(I18n.t('Oops'), I18n.t(serverInfo.message, serverInfo.messageOptions));
```
IlarionHalushka: maybe better use `showToast`?

app/sagas/selectServer.js
diegolmello: `getServerInfo` is called on `handleSelectServer`, `handleServerRequest` and `deepLinking`.
I don't want to raise the error every time.

app/sagas/selectServer.js
diegolmello: I want an action from the user.

PR #972, opened 2019-06-09
[FOR REVIEW ONLY] made single-server app

pranavpandey1998official:
this is review PR for single server do not merge this 
defaults sever is https://ilarion.rocket.chat please change that from app.json for review


app.json
```diff
@@ -1,4 +1,5 @@
 {
   "name": "RocketChatRN",
-  "displayName": "RocketChatRN"
+  "displayName": "RocketChatRN",
+  "server": "https://ilarion.rocket.chat"
```
diegolmello: Change to open.

pranavpandey1998official: test will start failing failling I cannot do anything

diegolmello: You mean e2e? Don't worry.

pranavpandey1998official: done

diegolmello:
Merged on `single-server`

PR #981, opened 2019-06-12
 [i18n] Add missing strings in de.js

AnBo83:
@RocketChat/ReactNative


app/i18n/locales/de.js
```diff
@@ -277,36 +298,48 @@ export default {
 	saving_profile: 'Profil speichern',
 	saving_settings: 'Einstellungen speichern',
 	Search_Messages: 'Nachrichten suchen',
-	Search: 'Suche',
+	Search: 'Suchen',
```
IlarionHalushka: I think here if should be a noun `Suche`, not a verb

AnBo83: Thanks for your comment, I changed it

PR #982, opened 2019-06-12
[IMPROVEMENT] Share credentials with Rocket.Chat.iOS

### LINKED ISSUE #983: [IMPROVEMENT] Share credentials between native and react-native client.
Use UserDefaults instead AsyncStorage on iOS to get credentials from [Rocket.Chat.iOS](https://github.com/RocketChat/Rocket.Chat.iOS).

djorkaeffalexandre:
Use UserDefaults instead AsyncStorage on iOS to get credentials from [Rocket.Chat.iOS](https://github.com/RocketChat/Rocket.Chat.iOS).
AndroidOS continue using AsyncStorage on RNUserDefaults lib.

Closes #983 

### How to test:
* Login any server on Rocket.Chat.iOS.
* Open Rocket.Chat.ReactNative and select same server! :)


app/sagas/deepLinking.js
```diff
@@ -1,8 +1,9 @@
-import { AsyncStorage } from 'react-native';
+// import { AsyncStorage } from 'react-native';
```
lolimay: There seems that we can remove this comment. What do you think of it?

app/sagas/init.js
```diff
@@ -1,6 +1,7 @@
-import { AsyncStorage } from 'react-native';
+// import { AsyncStorage } from 'react-native';
```
lolimay: the same as above.

app/sagas/init.js
```diff
@@ -15,10 +16,16 @@ import database from '../lib/realm';
 const restore = function* restore() {
 	try {
 		const { token, server } = yield all({
-			token: AsyncStorage.getItem(RocketChat.TOKEN_KEY),
-			server: AsyncStorage.getItem('currentServer')
+			token: RNUserDefaults.get(RocketChat.TOKEN_KEY),
+			server: RNUserDefaults.get('currentServer')
 		});
 
+		// RNUserDefaults.setName('group.ios.chat.rocket').then(() => {
```
lolimay: Can you remove these debugging code to keep the repo clean?

app/sagas/login.js
```diff
@@ -1,7 +1,8 @@
-import { AsyncStorage } from 'react-native';
+// import { AsyncStorage } from 'react-native';
```
lolimay: the same as above.

app/lib/realm.js
```diff
@@ -370,6 +384,7 @@ class DB {
 		serversDB: new Realm({
 			path: 'default.realm',
 			schema: [
+				userSchema,
```
diegolmello: Can we move this schema to `activeDB`?

djorkaeffalexandre: Hi @diegolmello, We can't move this schema to ```activeDB``` because on ```handleSelectServer``` we need to access user before to ```setActiveDB```.

PR #1004, opened 2019-06-21
[CHORE] Add bugsnag

diegolmello:
@RocketChat/ReactNative

Firebase doesn't provide full stacktraces.
We'll use Firebase for analytics and Bugsnag for error tracking.

To do:
- [x] Basic config
- [ ] Upload source maps via CI
- [ ] Check if we still want Crashlytics
- [ ] See other features we could implement on docs
- [ ] Add Bugsnag key requirement to readme


IlarionHalushka:
Closed due to new PR #1096

PR #1009, opened 2019-06-24
[CI] create job with e2e-build and save workspace for e2e-test

IlarionHalushka:
@RocketChat/ReactNative

The old e2e-test job was doing detox ios build and test on simulator iphone 7. 

In order to make it possible to create one detox ios build and run it on iPhone 7 and iPhone Xr the old job was divided into three:
* e2e-build - makes detox ios e2e build and persists the build at workspace.
* e2e-test-ios-iphone7 - ataches workspace from job e2e-build and then runs tests on iPhone7
* e2e-test-ios-iphoneXr - ataches workspace from job e2e-build and then runs tests on iPhoneXr

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,44 @@ jobs:
           paths:
             - ./node_modules
 
-  e2e-test:
+
+  e2e-build:
+    macos:
+      xcode: "10.2.1"
+
+    environment:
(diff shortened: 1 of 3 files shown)


diegolmello:
We've been having trouble with Detox's keyboard behaving differently on iPhone 7 and iPhone X.
It's making tests to fail.
We should track this issue: https://github.com/wix/Detox/issues/1476

PR #1012, opened 2019-06-25
[NEW] Auto-translate

### LINKED ISSUE #854: [NEW] Auto-Translation
We need to support Auto-translation feature.
When enabled, the server returns messages translated to the selected language.

There should be two ways of using it:
- Translating a single message (hold message and tap "Auto-translate" on the bottom sheet)
- Translating all messages in a room (tap "Auto-translate" on room actions)
  - Note: Only new messages will be translated

Both options must navigate to a new view with all available languages supported by Google.

There are two settings we need to fetch from our server:
- `AutoTranslate_Enabled`: shows/hides the entire logic.
- `AutoTranslate_GoogleAPIKey`: is the API Key we will be using to fetch languages from Google.

diegolmello:
@RocketChat/ReactNative

Closes #854 

Jun-26-2019 11-20-52
Jun-26-2019 11-21-53


IlarionHalushka:
is it possible to keep markdown the same after applying translation?

diegolmello:
> is it possible to keep markdown the same after applying translation?

I don't think so.

PR #1013, opened 2019-06-25
[NEW] Custom fields on signup

### LINKED ISSUE #978: [NEW] Custom fields on signup
We need to verify if the server has `Accounts_CustomFields` setting and send them on signup.

pranavpandey1998official:
@RocketChat/ReactNative

Closes #978 

WhatsApp Image 2019-06-26 at 5 31 03 AM

### DIFF
--- a/app/views/RegisterView.js
+++ b/app/views/RegisterView.js
@@ -19,6 +19,7 @@ import { loginRequest as loginRequestAction } from '../actions/login';
 import isValidEmail from '../utils/isValidEmail';
 import { LegalButton } from '../containers/HeaderButton';
 import StatusBar from '../containers/StatusBar';
+import log from '../utils/log';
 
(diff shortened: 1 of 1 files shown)


app/views/RegisterView.js
```diff
@@ -35,15 +39,30 @@ export default class RegisterView extends React.Component {
...
-		email: '',
-		password: '',
-		username: '',
-		saving: false
+	constructor(props) {
+		super(props);
+		const customFields = {};
+		this.parsedCustomFields = {};
+		if (props.Accounts_CustomFields) {
+			this.parsedCustomFields = JSON.parse(props.Accounts_CustomFields);
```
diegolmello: `Accounts_CustomFields` isn't correctly configured, the parse may fail and it'll crash the app.
Can you add a `try/catch`?

pranavpandey1998official: done

PR #1029, opened 2019-07-02
[CI] Restore cache on CI

### LINKED ISSUE #1019: [CI] Restore cache on CI
We're not restoring cache on all builds.

### yarn
Here's the time it takes to install npm modules in android by restoring cache:
image

Because we don't do it on ios builds, here's the difference:
image

Restore cache: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/.circleci/config.yml#L18
Save cache: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/.circleci/config.yml#L41

### fastlane
Something similar should be done on fastlane by following Fastlane's best practices: https://docs.fastlane.tools/best-practices/continuous-integration/circle-ci/
image

IlarionHalushka:
@RocketChat/ReactNative
Closes #1019


IlarionHalushka:
Now ios-build is much faster

.circleci/config.yml
```diff
@@ -174,6 +174,15 @@ jobs:
     steps:
       - checkout
 
+      - restore_cache:
+          key: bundle-v1-{{ checksum "ios/Gemfile.lock" }}-{{ arch }}
```
diegolmello: Does `name` property works?
It would be better if it was `Restoring npm cache`, `Restoring gems cache`, etc

.circleci/config.yml
```diff
@@ -174,6 +174,15 @@ jobs:
...
+
+      - restore_cache:
+          key: yarn-v1-{{ checksum "yarn.lock" }}-{{ arch }}
```
diegolmello: why `arch`?

diegolmello: Can you share cache with other tasks, like `lint-test`?
https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1029/files#diff-1d37e48f9ceff6d8030570cd36286a61R18

IlarionHalushka: no, cache is shared within one job

.circleci/config.yml
IlarionHalushka: not sure, haven't seen it in docs, but can try

.circleci/config.yml
IlarionHalushka: because this is the way it's suggested in docs

IlarionHalushka: https://docs.fastlane.tools/best-practices/continuous-integration/circle-ci/

diegolmello: I guess we may not need it.
We should try to use same cache for all jobs, as all of them uses `node_modules`.

IlarionHalushka:
Reusing node_modules cache didn't work when sharing it between different platforms linux\macos. So jobs `lint-test`, `android-build` share cache on linux; `e2e-test`, `ios-build`, `ios-testflight` share cache on mac os.

.circleci/config.yml
diegolmello: Can you do this for all similar tasks?
I can see there are two confusing tasks on android, for example:
image

IlarionHalushka: added names

PR #1039, opened 2019-07-05
[NEW] Mark message as unread

### LINKED ISSUE #989: [NEW] Mark message as unread
Similar to #709, we should have an option to mark the message as unread.

You need to:
- Add "Mark as unread" option on MessageActions (fired on message long press)
- Call `subscriptions.unread` with `firstUnreadMessage._id`: https://rocket.chat/docs/developer-guides/rest-api/subscriptions/unread/
- Go to rooms list (otherwise you'll mark the rooms as read again)

weijia-yu:
@RocketChat/ReactNative

Closes #989

unread


weijia-yu:
@diegolmello @IlarionHalushka can someone review please?

weijia-yu:
@djorkaeffalexandre Hi Possible to review it or assign a reviewer?

app/containers/MessageActions.js
```diff
@@ -82,6 +82,10 @@ export default class MessageActions extends React.Component {
 			this.REPLY_INDEX = this.options.length - 1;
 		}
 
+		// Mark as unread
+		this.options.push(I18n.t('Mark_as_unread'));
```
diegolmello: Push it to `this.options` only if it's from the same user.

diegolmello: Move it to above `Auto-translate` option

app/lib/rocketchat.js
```diff
@@ -520,6 +520,9 @@ const RocketChat = {
 	reportMessage(messageId) {
 		return this.sdk.post('chat.reportMessage', { messageId, description: 'Message reported by user' });
 	},
+	markMessageAsUnread(firstUnreadMessageId) {
```
diegolmello: Use `Rockechat.toggleRead()` instead.
You'll have to add `firstUnreadMessageId` param.

app/containers/MessageActions.js
weijia-yu: how to check it? this.isRoomReadOnly()?

diegolmello: You can only mark as unread something you didn't write yourself.
So we need to check if the selected message's author agains current user.
```jsx
// Toggle Auto-translate
props.actionMessage.u && props.actionMessage.u._id !== props.user.id
```

weijia-yu:
unread
@diegolmello I updated the branch, also here is the gif showing:

1. original mark room id as unread is still working

2. mark msg as unread in room

3. unable to mark my own msg as unread

weijia-yu:
@diegolmello Gotcha, my main language is Chinese and English, so I will leave `pt-BR`, `en` and `zh-cn`, what do you think? In that case, if they are using a different language, sometimes it would mix the other language and english, that should be fine?

weijia-yu:
@diegolmello I removed other language and only kept english and chinese translation

weijia-yu:
@diegolmello is it good now? Can you review it please 🕺

diegolmello:
@weijia-yu Sure. Thanks!

app/lib/rocketchat.js
```diff
@@ -614,11 +614,17 @@ const RocketChat = {
 		// RC 0.64.0
 		return this.sdk.post('rooms.favorite', { roomId, favorite });
 	},
-	toggleRead(read, roomId) {
+	toggleRead(read, roomId, firstUnreadMessageId) {
 		if (read) {
-			return this.sdk.post('subscriptions.unread', { roomId });
+			if (roomId !== undefined) {
```
IlarionHalushka: just do `if (roomId) {`

app/lib/rocketchat.js
```diff
@@ -614,11 +614,17 @@ const RocketChat = {
...
+				return this.sdk.post('subscriptions.unread', { roomId });
+			}
+			if (firstUnreadMessageId !== undefined) {
```
IlarionHalushka: just do `if (firstUnreadMessageId) {`

app/containers/MessageActions.js
```diff
@@ -311,6 +317,16 @@ export default class MessageActions extends React.Component {
 		replyInit(actionMessage, true);
 	}
 
+	handleUnread = () => {
+		const { actionMessage } = this.props;
+		try {
+			RocketChat.toggleRead(true, undefined, actionMessage._id);
+			Navigation.navigate('RoomsListView');
```
IlarionHalushka: Not sure that navigation to RoomsListView should occur after message is marked as read\unread. 
@diegolmello what do you think?

diegolmello: It's the same behaviour of web version.
I think it's ok.

app/lib/rocketchat.js
```diff
@@ -614,11 +614,17 @@ const RocketChat = {
 		// RC 0.64.0
 		return this.sdk.post('rooms.favorite', { roomId, favorite });
 	},
-	toggleRead(read, roomId) {
+	toggleRead(read, roomId, firstUnreadMessageId) {
```
diegolmello: Can you switch this signature to `toggleRead({ read = true, rid, messageId })`?
This way is much more readable.

#### Before:
```jsx
RocketChat.toggleRead(true, undefined, actionMessage._id);
```

#### After:
```jsx
RocketChat.toggleRead({ messageId : actionMessage._id });
```

diegolmello: And on https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/views/RoomsListView/index.js#L392:

#### Before:
```jsx
const result = await RocketChat.toggleRead(isRead, rid);
```

#### After:
```jsx
const result = await RocketChat.toggleRead({ rid, read: isRead });
```

djorkaeffalexandre:
Was done by #1785.
Thanks.

PR #1059, opened 2019-07-18
[TESTS] Configure detox for android tests

IlarionHalushka:
@RocketChat/ReactNative


.circleci/config.yml
```diff
@@ -99,6 +100,145 @@ jobs:
       - store_artifacts:
           path: /tmp/screenshots
 
+  e2e-test-android:
```
diegolmello: Lots of commands here are not on the [Android Detox Guide](https://github.com/wix/Detox/blob/master/docs/Introduction.Android.md).
If you got them elsewhere, make sure everything is really needed on our case and comment wherever possible.
It's ok to learn from other repos, but not ok to have unnecessary workarounds on our own.

.circleci/config.yml
```diff
@@ -99,6 +100,145 @@ jobs:
...
+          command: |
+            source ~/.bash_profile
+            export BOOT=""
+            echo "Waiting for AVD to finish booting"
+            export PATH=$(dirname $(dirname $(command -v android)))/platform-tools:$PATH
+            until [[ "$BOOT" =~ "1" ]]; do
+              sleep 5
+              export BOOT=$(adb -e shell getprop sys.boot_completed 2>&1)
+            done
+            sleep 15
```
diegolmello: This is one example from my other comment... Why?

android/build.gradle
```diff
@@ -2,11 +2,13 @@
 buildscript {
     ext {
         buildToolsVersion = "28.0.3"
-        minSdkVersion = 21
+        minSdkVersion = 24
```
diegolmello: Is this really needed?

android/build.gradle
```diff
@@ -2,11 +2,13 @@
...
         compileSdkVersion = 28
         targetSdkVersion = 28
         supportLibVersion = "28.0.0"
+        kotlinVersion = "1.3.0"
         glideVersion = "4.9.0"
+        googlePlayServicesVersion = "16.+"
```
diegolmello: It seems you're not using this constant anywhere.

e2e/12-broadcast.spec.js
```diff
@@ -5,122 +5,144 @@ const OTP = require('otp.js');
...
-		await waitFor(element(by.id('room-actions-view'))).toBeVisible().withTimeout(2000);
-		await tapBack();
-		await waitFor(element(by.id('room-view'))).toBeVisible().withTimeout(2000);
-		// await tapBack();
-		// await waitFor(element(by.id('rooms-list-view'))).toBeVisible().withTimeout(2000);
-		// await waitFor(element(by.id(`rooms-list-view-item-broadcast${ data.random }`))).toExist().withTimeout(60000);
-		// await expect(element(by.id(`rooms-list-view-item-broadcast${ data.random }`))).toExist();
-	});
+		before(async () => {
+			if (device.getPlatform() === 'ios') {
```
diegolmello: Can we do a "global" `if iOS return` instead of one condition per test?

.circleci/config.yml
```diff
@@ -88,7 +88,8 @@ jobs:
       - run:
           name: Test
           command: |
-            detox test --configuration ios.sim.release --cleanup
+            detox clean-framework-cache && detox build-framework-cache
+            detox test --configuration ios.sim.release --cleanup --loglevel verbose
```
diegolmello: Why verbose?
Doesn't it makes everything slower on CI?

.circleci/config.yml
```diff
@@ -99,6 +100,145 @@ jobs:
...
+            adb shell settings put global transition_animation_scale 0
+            adb shell settings put global animator_duration_scale 0
+            echo "Android Virtual Device is now ready."
+          name: Wait for AVD to be ready
+          no_output_timeout: 5m
+      - run:
+          command: |
+            source ~/.bash_profile
+            /usr/local/share/android-sdk/platform-tools/adb -s emulator-5554 install -r -g -t "/Users/distiller/project/android/app/build/outputs/apk/release/app-release.apk"
+            detox test -c android.emu.release -l trace --headless
```
diegolmello: Why `trace --headless`?

e2e/helpers/roomListView.js
```diff
@@ -0,0 +1,16 @@
+const {
```
diegolmello: Rename this file to `rooms.js`

IlarionHalushka: done

.circleci/config.yml
IlarionHalushka: removed loglevel trace. Headless is suggested by detox cli `[Android Only] Launch Emulator in headless mode. Useful when running on CI.` and it should be a bit faster this way

.circleci/config.yml
IlarionHalushka: it's easier to debug why tests failed. I removed it.

e2e/12-broadcast.spec.js
IlarionHalushka: unfortunately no :( because detox's object `device` is undefined if I inside of it `describe` block

android/build.gradle
IlarionHalushka: it is used by some detox gradle dependencies (removing it fails build)

android/build.gradle
IlarionHalushka: reverted to `21`, but before building detox I will exchange `21` => `24` in circleCI job

.circleci/config.yml
IlarionHalushka: removed

.circleci/config.yml
IlarionHalushka: step by step I removed some commands and then run job. Now only required commands are left

PR #1061, opened 2019-07-19
[IMPROVEMENT] Image Viewer

### LINKED ISSUE #903: Improve image preview
Currently, we're using [react-native-image-zoom-viewer](https://github.com/magicwing/react-native-image-zoom-viewer) as image viewer.
Animations are too junky and it's not well maintained.

We need to refactor and split by OS:

### iOS
iOS supports zoom by default on ScrollView. We need to wrap the image in a ScrollView and use [maximumZoomScale](https://facebook.github.io/react-native/docs/scrollview#maximumzoomscale)

### Android
It doesn't support zoom in the same way as iOS, so we need to have a separate logic.
[react-native-reanimated](https://github.com/kmagiera/react-native-reanimated) is really awesome and have an ImagePreview example: https://github.com/kmagiera/react-native-reanimated/blob/master/README.md#100-declarative-gesture-interactions

pranavpandey1998official:
@RocketChat/ReactNative

Closes #903

removed Image viewer and made a custom one with gesture handler and react native reanimated


DoM1niC:
It will support saves Images too??

pranavpandey1998official:
> It will support saves Images too??

no ..maybe in the next PR

djorkaeffalexandre:
Closed by #1479 .

PR #1063, opened 2019-07-19
[FIX] Share Extension iOS memory issue

djorkaeffalexandre:
Share Extension iOS crashes with approximately 500 chats.


djorkaeffalexandre:
moved to #1064 .

PR #1064, opened 2019-07-19
[FIX] Share Extension issues

djorkaeffalexandre:
Share Extension iOS crashes with approximately 500 chats, we will go to slice chats with limit 50 and show the fifty most recents only.
This pr also adds header search and fixes some bugs on android because on android the store is shared between App and Share Extension when both are open.


app/views/ShareListView/index.js
```diff
@@ -308,7 +269,7 @@ export default class ShareListView extends React.Component {
 	search = (text) => {
 		const result = database.objects('subscriptions').filtered('name CONTAINS[c] $0', text);
 		const subscriptions = database.objects('subscriptions');
-		const data = result.length !== subscriptions.length ? result : [];
+		const data = result.length !== subscriptions.length ? result.slice(0, 50) : [];
```
diegolmello: If search length === 0, show a label.

djorkaeffalexandre: fixed.

app/lib/realm.js
```diff
@@ -424,6 +427,16 @@ class DB {
 						newServers[i].roomsUpdatedAt = null;
 					}
 				}
+				if (oldRealm.schemaVersion === 9 && newRealm.schemaVersion === 10) {
```
diegolmello: Not needed

app/views/ShareListView/SearchBar.js
```diff
@@ -0,0 +1,48 @@
...
+
+	const onCancelPress = () => {
+		Keyboard.dismiss();
+		onChangeText('');
+		setHasCancel(false);
+	};
+
+	const onFocus = () => {
+		setHasCancel(true);
+		LayoutAnimation.easeInEaseOut();
```
diegolmello: Wrap with `isIOS` condition

app/views/ShareView/index.js
```diff
@@ -112,8 +114,10 @@ export default class ShareView extends React.Component {
 
 	sendTextMessage = async() => {
 		const { value, rid } = this.state;
+		const { token } = this.props;
 		if (value !== '' && rid !== '') {
 			try {
+				await RocketChat.login({ resume: token });
```
diegolmello: Shouldn't you do the same for `sendMediaMessage`?
Why not login in `shareExtensionInit`?

app/views/ShareListView/index.js
```diff
@@ -22,31 +21,27 @@ import {
...
-import ServerItem, { ROW_HEIGHT as ROW_HEIGHT_SERVER } from '../../presentation/ServerItem';
+import ServerItem from '../../presentation/ServerItem';
 import { CloseShareExtensionButton, CustomHeaderButtons, Item } from '../../containers/HeaderButton';
-import SearchBar from '../RoomsListView/ListHeader/SearchBar';
+import SearchBar from './SearchBar';
 import ShareListHeader from './Header';
 
 import styles from './styles';
 
 const SCROLL_OFFSET = 56;
```
diegolmello: It never hides.
Can you place the searchbox at the header?

djorkaeffalexandre: done.

app/views/ShareListView/Header.js
```diff
@@ -84,6 +85,12 @@ class ShareListHeader extends PureComponent {
 					/>
 				</View>
 			);
+		} else if (isIOS) {
```
diegolmello: Why do we have a `SearchHeader` component for iOS and a TextInput on Android?
It should be the component, if necessary.

app/views/ShareListView/SearchHeader.js
```diff
@@ -0,0 +1,83 @@
...
+import { HEADER_BACKGROUND } from '../../constants/colors';
+
+import sharedStyles from '../Styles';
+
+const styles = StyleSheet.create({
+	container: {
+		backgroundColor: HEADER_BACKGROUND,
+		width: '100%',
+		flexDirection: 'row',
+		paddingTop: getStatusBarHeight() + (isNotch ? 20 : 0),
```
diegolmello: I didn't like this lib.
I think we probably don't need it.
I'll take a look.

app/views/ShareView/index.js
```diff
@@ -81,6 +85,11 @@ export default class ShareView extends React.Component {
 		navigation.setParams({ sendMessage: this._sendMessage, canSend: !(isReadOnly(room, { username }) || isBlocked(room)) });
 	}
 
+	componentDidUpdate() {
+		const { setSearch } = this.props;
+		setSearch('');
```
diegolmello: Bad practice.
You executing `setSearch('')` for every render.
Is this what you want to do here?

djorkaeffalexandre:
@diegolmello Apple News was fixed on `rn-extensions-share v2.3.8`.

app/lib/methods/sendMessage.js
```diff
@@ -5,8 +5,9 @@ import reduxStore from '../createStore';
 import log from '../../utils/log';
 import random from '../../utils/random';
 
-export const getMessage = (rid, msg = '', tmid) => {
+export const getMessage = (rid, msg = '', tmid, shareExtension) => {
```
diegolmello: This is not scalable.
You can send `userId` and `username` as params instead.

djorkaeffalexandre: Fixed.

PR #1066, opened 2019-07-22
[FIX] Moment locales

djorkaeffalexandre:
We have an error when use moment.locale because our language key is 'pt-BR', with 'BR' on uppercase and moment.locale not have this locale.

https://github.com/facebook/react-native/issues/16540

### DIFF
--- a/app/utils/moment.js
+++ b/app/utils/moment.js
@@ -1,5 +1,5 @@
 const localeKeys = {
-	en: 'en-au',
+	en: 'en',
 	ru: 'ru',
 	'pt-BR': 'pt-br',
 	'zh-CN': 'zh-cn',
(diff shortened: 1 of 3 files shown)


app/utils/moment.js
```diff
@@ -0,0 +1,11 @@
+const localeKeys = {
+	en: 'en-au',
```
diegolmello: Why australian?
`en-us` is probably the right one.

app/views/RoomsListView/index.js
```diff
@@ -153,7 +154,7 @@ export default class RoomsListView extends React.Component {
 	componentDidMount() {
 		this.getSubscriptions();
 		const { navigation, userLanguage } = this.props;
-		moment.locale(userLanguage);
+		moment.locale(toMomentLocale(userLanguage));
```
diegolmello: Why on `RoomsListView`? 🤔 
Will it work on `OnboardingView`, for example?

djorkaeffalexandre: fixed.

PR #1069, opened 2019-07-22
[NEW] Add "Send Crash Report" option

djorkaeffalexandre:
User can set `Send Crash Report`.
Default value is `false`.

### DIFF
--- a/app/views/SettingsView/index.js
+++ b/app/views/SettingsView/index.js
@@ -7,7 +7,7 @@ import { connect } from 'react-redux';
 import firebase from 'react-native-firebase';
 
 import { toggleMarkdown as toggleMarkdownAction } from '../../actions/markdown';
-import { toggleCrashlytics as toggleCrashlyticsAction } from '../../actions/crashlytics';
+import { toggleCrashReport as toggleCrashReportAction } from '../../actions/crashReport';
 import { SWITCH_TRACK_COLOR } from '../../constants/colors';
 import { DrawerButton } from '../../containers/HeaderButton';
 import StatusBar from '../../containers/StatusBar';
@@ -15,7 +15,7 @@ import ListItem from '../../containers/ListItem';
 import { DisclosureImage } from '../../containers/DisclosureIndicator';
 import Separator from '../../containers/Separator';
 import I18n from '../../i18n';
-import { MARKDOWN_KEY, CRASHLYTICS_KEY } from '../../lib/rocketchat';
+import { MARKDOWN_KEY, CRASH_REPORT_KEY } from '../../lib/rocketchat';
 import { getReadableVersion, getDeviceModel } from '../../utils/deviceInfo';
 import openLink from '../../utils/openLink';
 import scrollPersistTaps from '../../utils/scrollPersistTaps';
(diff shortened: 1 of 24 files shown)


filipedelimabrito:
Are you guys using analytics on the app too?

djorkaeffalexandre:
> Are you guys using analytics on the app too?

yes!

app/views/SettingsView/index.js
```diff
@@ -83,6 +98,17 @@ export default class SettingsView extends React.Component {
 		);
 	}
 
+	renderCrashlyticsSwitch = () => {
```
diegolmello: We should do it same way as iOS native does:

- "Send Crash Report"
- A description of it



Also, remove "Crashlytics" word from other places like methods, actions, etc.

djorkaeffalexandre: nice!

djorkaeffalexandre: fixed.

filipedelimabrito:
@djorkaeffalexandre Thanks for replying. I am not sure if you guys are getting the analytics data only when the `Send Crash Report` is enabled from now, but I'd like to say that from both native iOS and Android apps we are doing so. It [brings some questions with GDPR](https://github.com/RocketChat/Rocket.Chat.Android/issues/1757) too - jfyi.

That said, is this PR being applied only for Crashlytics or Analytics (or both of them)?

djorkaeffalexandre:
@filipedelimabrito only for crashlytics. 😁

filipedelimabrito:
@djorkaeffalexandre 😬

@diegolmello we need to start getting analytics data only after the user consent.

diegolmello:
> @diegolmello we need to start getting analytics data only after the user consent.

@filipedelimabrito I couldn't find `Send analytics` on iOS native or `Send crash report` on Android native.
I couldn't find these options on onboarding on native apps neither, as the link you sent.

I think this PR does what the title says and we could implement `Analytics collection` later: https://rnfirebase.io/docs/v5.x.x/analytics/reference/analytics#setAnalyticsCollectionEnabled

diegolmello:
By testing, we figured Firebase's API could be better: https://rnfirebase.io/docs/v5.x.x/crashlytics/reference/crashlytics#enableCrashlyticsCollection

It forces us to start the app disabled and enable inside the code with `enableCrashlyticsCollection` method.
What if the user wants to disable crash report?
Should we ask him to close the app and reopen?
That's not a good UX.

Let's close this PR and try Bugsnag in the future.
At least their API _seems_ to be better than this: https://docs.bugsnag.com/platforms/react-native/react-native/#disabling-error-reporting

Thanks anyway, @djorkaeffalexandre 👍

filipedelimabrito:
@diegolmello I can refer to you the code on native ones if you want to confirm it - actually the display text isn't good on iOS native (it only tells the users about the crash reports while it serves to opt in and out the analytics too), on Android you can see it (the analytics I started talking about) clearly on the settings screen (let me attach an image here).  

Sorry if it has disturbed you but I think that since this PR is implementing a collect/tracking option it worths bringing this **initial** discussion here - to you and @djorkaeffalexandre. 


Screen Shot 2019-07-23 at 18 29 29

diegolmello:
@filipedelimabrito You're right.
I just followed what the iOS label says.
Anyway, the lib API is not good and we're planning to use Bugsnag.

PR #1096, opened 2019-08-01
[CHORE] Bugsnag logging setup

### LINKED ISSUE #123: Add some error tracking like crashlytics or bugsnag
https://github.com/corymsmith/react-native-fabric
https://docs.bugsnag.com/platforms/react-native/

IlarionHalushka:
@RocketChat/ReactNative
Closes #123 

This PR aims to add error logging using Bugsnag.

Bugsnag provides ability to upload source maps for each build in order to be able to see an informative stack trace.

- [x] basic configuration of bugsnag 
- [x] upload iOS source maps in ios-build CI job
- [x] upload Android source maps in android-build CI job
- [x] exchange firebase error logging to bugsnag
- [x] add to Settings `Send crash reports` switch which enables\disables reports\analytics sending.
- [ ] change API key from Ilarion's to RocketChat's and test error reporting on release builds

Screenshots of error stack trace on Bugsnag: 
Screen Shot 2019-08-01 at 10 59 35 PM

Uploaded source maps (release via CI, debug from local machine): 
Screen Shot 2019-08-02 at 11 01 44 AM


.circleci/config.yml
```diff
@@ -114,10 +114,26 @@ jobs:
     steps:
       - checkout
 
+      - run:
```
diegolmello: Why?

.circleci/config.yml
```diff
@@ -148,13 +164,18 @@ jobs:
...
           command: |
-            cd android/app
             cp google-services.prod.json google-services.json
+          working_directory: android/app
+
+      - run:
+          name: Upload sourcemaps to Bugsnag
+          command: |
+            react-native bundle     --platform android     --dev false     --entry-file index.js     --bundle-output android-release.bundle     --sourcemap-output android-release.bundle.map
+            curl https://upload.bugsnag.com/react-native-source-map  -F apiKey=72a0364cb361fc8f0fdc6cbf605f9963  -F appVersionCode=$CIRCLE_BUILD_NUM  -F dev=false  -F platform=android  -F sourceMap=@android-release.bundle.map  -F bundle=@android-release.bundle  -F projectRoot=app.bugsnag.com/rocketchat-1
```
diegolmello: We need to change this API Key

.circleci/config.yml
```diff
@@ -224,6 +247,12 @@ jobs:
             cp GoogleService-Info.prod.plist GoogleService-Info.plist
           working_directory: ios
 
+      - run:
+          name: Upload sourcemaps to Bugsnag
+          command: |
+            react-native bundle     --platform ios     --dev false     --entry-file index.js     --bundle-output ios-release.bundle     --sourcemap-output ios-release.bundle.map
+            curl https://upload.bugsnag.com/react-native-source-map  -F apiKey=72a0364cb361fc8f0fdc6cbf605f9963  -F appBundleVersion=$CIRCLE_BUILD_NUM  -F dev=false  -F platform=ios  -F sourceMap=@ios-release.bundle.map  -F bundle=@ios-release.bundle  -F projectRoot=app.bugsnag.com/rocketchat-1
```
diegolmello: We need to change this API key

android/app/src/main/AndroidManifest.xml
```diff
@@ -52,6 +52,9 @@
 					<data android:mimeType="*/*" />
 				</intent-filter>
 		</activity>
+		<meta-data
```
diegolmello: API key

index.js
```diff
@@ -2,8 +2,12 @@ import 'react-native-console-time-polyfill';
 
 import './app/ReactotronConfig';
 import { AppRegistry } from 'react-native';
+import { Client } from 'bugsnag-react-native';
 import { name as appName, share as shareName } from './app.json';
 
+const bugsnag = new Client('72a0364cb361fc8f0fdc6cbf605f9963');
```
diegolmello: API Key.
Is it ok to use it here or should we hide from Github?

.circleci/config.yml
```diff
@@ -148,13 +164,18 @@ jobs:
...
           name: Set Google Services
           command: |
-            cd android/app
             cp google-services.prod.json google-services.json
+          working_directory: android/app
+
+      - run:
+          name: Upload sourcemaps to Bugsnag
+          command: |
+            react-native bundle     --platform android     --dev false     --entry-file index.js     --bundle-output android-release.bundle     --sourcemap-output android-release.bundle.map
```
diegolmello: You can use `react-native` from scripts. Same as `yarn android`, for example.

.circleci/config.yml
IlarionHalushka: node is required react-native-cli for generating source maps

index.js
IlarionHalushka: sensitive info of users can't be accessed using API key so I think it should stay

IlarionHalushka:
I had to remove the description of an error like `log('err_on_login', e);` like it used to be with firebase logging. The change is made because `bugsnag.notify` accepts only 1 parameter with type `Error`.

 Fortunately bugsnag groups errors by lines where they occurred like this:
Screen Shot 2019-08-03 at 9 51 15 PM

index.js
lorek123: I would move this to some separate config file just to not keep API Keys within source code. Also you should consider if you want to get bug reports from apps that are forks of this app especially when they won't share the sources.

IlarionHalushka:
Source maps are successfully uploaded to bugsnag during latest builds:
Screen Shot 2019-08-11 at 3 02 46 PM

app/utils/analytics.js
```diff
@@ -0,0 +1,3 @@
+import firebase from 'react-native-firebase';
```
diegolmello: Analytics is a type of log, so I think we could add this to `utils/log`.
We could even have a `logEvent` that sents data both to Crashlytics and Bugsnag.
That would help to analyze how our users are using the app and to track errors.

app/views/DirectoryView/index.js
```diff
@@ -100,7 +100,7 @@ class DirectoryView extends React.Component {
 				this.setState({ loading: false });
 			}
 		} catch (error) {
-			log('err_load_directory', error);
+			log(error);
```
diegolmello: Can you refactor all places to use `log(e);`?

config.js
```diff
@@ -0,0 +1,3 @@
+export default {
+	BUGSNAG_API_KEY: '72a0364cb361fc8f0fdc6cbf605f9963'
```
diegolmello: Can you add a task on CI to replace this key and other places?
What will happen if we commit with `BUGSNAG_API_KEY: ''`?
Will it compile?
I'm thinking in DX here.
We shouldn't enforce anyone to have a Bugsnag profile in order to compile our app, but it would be better if we left that empty and replace on build time.

IlarionHalushka: I cannot test it because my bugsnag account trial has expired :(
When I change BUGSNAG_API_KEY to `''` then build doesn't fail but I'm not sure if whether it still sends requests to bugsnag.

IlarionHalushka:
Closed due to new opened PR:  https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1145

PR #1103, opened 2019-08-05
[IMPROVEMENT] Unify search UX

### LINKED ISSUE #1010: [IMPROVEMENT] Unify search UX
Our search is different across OS:

### Android
Magnifier on header

### iOS
Swipe down to show search box

### Proposed change
- We compared the same feature on other apps in different OS' and we should keep only iOS' UX.
Let's make Android work the same way as iOS.

- We should also add "Cancel" button, to stop searching state and dismiss keyboard.

juanpetterson:
@RocketChat/ReactNative

Closes #1010

Searchbar is different across OS.

Searchbar now has the same UX functionality to show her in both OS.
Swipe down to show the bar.

screenshot-2019-08-05_07 06 58 016

### DIFF
--- a/app/views/RoomsListView/index.js
+++ b/app/views/RoomsListView/index.js
@@ -1,12 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
-	View, FlatList, BackHandler, ActivityIndicator, Text, ScrollView, Keyboard, LayoutAnimation, InteractionManager, Dimensions
+	View, FlatList, BackHandler, ActivityIndicator, Text, ScrollView, Keyboard, Animated, LayoutAnimation, InteractionManager, Dimensions
 } from 'react-native';
 import { connect } from 'react-redux';
 import { isEqual } from 'lodash';
 import { SafeAreaView } from 'react-navigation';
 import Orientation from 'react-native-orientation-locker';
+import { PanGestureHandler, State } from 'react-native-gesture-handler';
 
 import database, { safeAddListener } from '../../lib/realm';
 import RocketChat from '../../lib/rocketchat';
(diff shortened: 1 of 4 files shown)


diegolmello:
Thank you, @juanpetterson!
I'll review it soon.
Cheers.

app/views/RoomsListView/index.js
```diff
@@ -634,21 +625,28 @@ export default class RoomsListView extends React.Component {
...
-						/>
-					)
-					: null
-				}
-				{showServerDropdown ? <ServerDropdown /> : null}
+				<PanGestureHandler
+					onHandlerStateChange={this.onHandlerStateChange}
+					minDeltaY={20}
+				>
+					<Animated.ScrollView style={styles.scrollView}>
```
diegolmello: No need for this Animation logic here.
This offset already works as expected: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/views/RoomsListView/index.js#L602
If you remove iOS condition, it's probably going to work on Android.

juanpetterson: The Animation is needed only to detect the swipe, but I will send one improvement.

diegolmello: This search already works on iOS using offset.
Your PR doesn't.
Offset is not an optimal solution, but it's fine.
Wrapping everything in another ScrollView is error prone.
Can we talk on https://open.rocket.chat?
I'm `@diego.mello`.
Thanks :)

juanpetterson:
Closing this PR because contentOffset is only available on iOS.

Link to documentation.
https://facebook.github.io/react-native/docs/scrollview#contentoffset

Issue #428, opened 2018-09-06
authentication via SAML is missing


aba-rechsteiner:
- Your Rocket.Chat.Android app version: 1.1.1
- Your Rocket.Chat server version: 0.69.0

- Device model (or emulator) you're running with: OnePlus 5 - Android 8.1.0

- Steps to reproduce:
setup authentication via SAML

localguru:
Yes! +1

PR #1093, opened 2019-07-31
[NEW] Custom OAuth

jbuechele:
@RocketChat/ReactNative

Closes #1018

PR #1108, opened 2019-08-06
[NEW] SAML authentication support

filipedelimabrito:
@RocketChat/ReactNative

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/428

To be merged after https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1093.


app/views/SSOView.js
```diff
@@ -0,0 +1,101 @@
...
+			loading: false
+		};
+	}
+
+	dismiss = () => {
+		const { navigation } = this.props;
+		navigation.pop();
+	}
+
+	login = async(params) => {
```
filipedelimabrito: Seems not beautiful to me since we are using the same on `OAuthView.js`. Can we move to somewhere else and use it from there here and on `OAuthView.js`?

app/lib/rocketchat.js
```diff
@@ -776,15 +776,17 @@ const RocketChat = {
...
-			let loginServicesFilter = [];
+			let loginServices = [];
 			const loginServicesResult = await fetch(`${ server }/api/v1/settings.oauth`).then(response => response.json());
 
 			if (loginServicesResult.success && loginServicesResult.services.length > 0) {
 				const { services } = loginServicesResult;
-				loginServicesFilter = services.filter(item => item.custom !== undefined); // TODO: remove this after SAML and CAS
+				loginServices = services;
+
+				const loginServicesReducer = loginServices.reduce((ret, item) => {
```
djorkaeffalexandre: You can use only `(services || []).reduce` and not need to create `loginServices`.

app/lib/rocketchat.js
```diff
@@ -795,21 +797,25 @@ const RocketChat = {
 				}, {});
 				reduxStore.dispatch(setLoginServices(loginServicesReducer));
 			}
```
djorkaeffalexandre: `services && services.length` ?

app/lib/rocketchat.js
```diff
@@ -795,21 +797,25 @@ const RocketChat = {
...
-			return Promise.resolve(loginServicesFilter.length);
+			return Promise.resolve(loginServices.length);
```
djorkaeffalexandre: `services && services.length`?

app/views/LoginSignupView.js
```diff
@@ -329,16 +337,29 @@ class LoginSignupView extends React.Component {
...
 			default:
 				break;
 		}
+		name = name.charAt(0).toUpperCase() + name.slice(1);
+		let buttonText;
+		if (service.service === 'saml') {
+			buttonText = <Text style={styles.serviceName}>{name}</Text>;
+		} else {
+			buttonText = (
+				<>
```
djorkaeffalexandre: `<React.Fragment>`?

filipedelimabrito: @diegolmello I think it is related to your changes ˆˆ

diegolmello: @djorkaeffalexandre this is a shorthand for `<React.Fragment>` :)

djorkaeffalexandre: Nice! 🙀

app/lib/rocketchat.js
diegolmello: True! We had a filter here, but we don't anymore. I'll fix it. Thanks :)

diegolmello: After looking further into this, it's better this way.
`services` is a response from `settings.oauth` and we need to return its length.
This way we have one `return`... the other we would need two.
There's no right answer and both works.
I'll leave it as it is :)

ggazzo:
@diegolmello @filipedelimabrito should we update 
image

PR #1116, opened 2019-08-08
[NEW] CAS authentication

filipedelimabrito:
@RocketChat/ReactNative

To be merged after https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1108


app/views/AuthenticationWebView.js
```diff
@@ -58,37 +62,58 @@ class OAuthView extends React.PureComponent {
...
 
+	onNavigationStateChange = (webViewState) => {
+		const url = decodeURIComponent(webViewState.url);
+		if (this.authType === 'saml' || this.authType === 'cas') {
+			const { navigation } = this.props;
+			const ssoToken = navigation.getParam('ssoToken');
+			if (url.includes('ticket') || url.includes('validate')) {
+				let payload;
+				if (this.authType === 'saml') {
+					payload = `{ "saml": true, "credentialToken": "${ ssoToken }" }`;
```
djorkaeffalexandre: Why you not use a simple object?

app/views/LoginSignupView.js
```diff
@@ -329,16 +346,35 @@ class LoginSignupView extends React.Component {
...
+			case 'cas': {
+				onPress = () => this.onPressCas();
+				break;
+			}
 			default:
 				break;
 		}
+		name = name.charAt(0).toUpperCase() + name.slice(1);
+		let buttonText;
+		if (service.service === 'saml') {
```
djorkaeffalexandre: You can use `if (service.service === 'saml' || (service.service === 'cas' && this.CAS_enabled))`.

filipedelimabrito: True (and simple)! Let me change it!

app/views/AuthenticationWebView.js
filipedelimabrito: Is it more efficient @djorkaeffalexandre?

djorkaeffalexandre: @filipedelimabrito I think `object` is more clean that `string + json.parse`, but is only a suggestion hahaha.

djorkaeffalexandre: You can use:
```
const credentialToken = { credentialToken: ssoToken };
if (this.authType === ‘saml’) {
  payload = { ...credentialToken, saml: true };
} else {
  payload = { cas: credentialToken };
}
```

app/views/LoginSignupView.js
```diff
@@ -329,16 +346,33 @@ class LoginSignupView extends React.Component {
...
+			case 'cas': {
+				onPress = () => this.onPressCas();
+				break;
+			}
 			default:
 				break;
 		}
+		name = name.charAt(0).toUpperCase() + name.slice(1);
+		let buttonText;
+		if (service.service === 'saml' || (service.service === 'cas' && this.CAS_enabled)) {
```
diegolmello: Did you mean `this.props.CAS_enabled`?

app/views/AuthenticationWebView.js
```diff
@@ -58,37 +62,58 @@ class OAuthView extends React.PureComponent {
...
+				} else {
+					payload = `{ "cas": { "credentialToken": "${ ssoToken }" } }`;
+				}
+				// We need to set a timeout when the login is done with SSO in order to make it work on our side.
+				// It is actually due to the SSO server processing the response.
+				setTimeout(() => {
+					this.login(JSON.parse(payload));
+				}, 3000);
```
diegolmello: Does it needs to be whole 3 seconds?
What about a poor internet connection?
Would it work properly?

filipedelimabrito: Yes, unfortunately it need to be as it is - as explained on the comment. It is not an issue from our side (read the comment pls).

We've done lot's of tests when we had implemented the feature on the native apps. Refer to the delay [here](https://github.com/RocketChat/Rocket.Chat.Android/blob/e9821393c593bac04856b6bde4b3e03428362c68/app/src/main/java/chat/rocket/android/authentication/loginoptions/presentation/LoginOptionsPresenter.kt#L113) for the Android Native and [here](https://github.com/RocketChat/Rocket.Chat.iOS/blob/42b9857ef445a238fb915716a68bbdd5210b38a6/Rocket.Chat/Controllers/Auth/CASViewController.swift#L76) for the iOS Native.

About the question with a poor internet connection, then I advise implementing a kind of retry IO like we have done on native, but then it needs to be done not only for CAS, SAML, but for all supported authentication we have. This way we can ensure a kind of repeated callback to handle this kind of situation.

app/views/AuthenticationWebView.js
filipedelimabrito: Thanks @djorkaeffalexandre! It's more clean now!

PR #1121, opened 2019-08-09
[FIX] Endless "connecting" on notification tap on iOS

diegolmello:
@RocketChat/ReactNative

Turns out react-native-notifications v2 fires `notificationOpened` event even on cold start.
It was dispatching `deepLinkingOpen` twice causing `selectServerRequest` to never finish.


djorkaeffalexandre:
Nice! <3

PR #1132, opened 2019-08-13
[IMPROVEMENT] Use expo video player

IlarionHalushka:
@RocketChat/ReactNative

PR introduces new look to Video Player using expo-av library.

Video:


app/containers/FileModal.js
```diff
@@ -38,6 +44,23 @@ const styles = StyleSheet.create({
...
 	indicator: {
 		flex: 1
+	},
+	video: {
+		width: Dimensions.get('window').width,
+		height: Dimensions.get('window').height - STATUSBAR_HEIGHT
+	},
+	poster: {
+		backgroundColor: 'black'
+		// opacity: 0.1
```
djorkaeffalexandre: why?

app/views/MessagesView/index.js
```diff
@@ -225,6 +232,10 @@ class MessagesView extends React.Component {
 		}
 	}
 
+	setFileLoading = (fileLoading) => {
```
djorkaeffalexandre: `setFileLoading = fileLoading => this.setState({ fileLoading });` ?

app/views/RoomView/index.js
```diff
@@ -452,6 +455,10 @@ class RoomView extends React.Component {
 		}
 	}
 
+	setFileLoading = (fileLoading) => {
```
djorkaeffalexandre: `setFileLoading = fileLoading => this.setState({ fileLoading });` ?

index.js
```diff
@@ -9,3 +9,5 @@ AppRegistry.registerComponent(shareName, () => require('./app/share').default);
 
 // For storybook, comment everything above and uncomment below
 // import './storybook';
+
+console.disableYellowBox = true;
```
djorkaeffalexandre: remove it

app/containers/FileModal.js
```diff
@@ -38,6 +44,23 @@ const styles = StyleSheet.create({
 	},
 	indicator: {
 		flex: 1
+	},
+	video: {
```
diegolmello: No need to set width and height.
`flex: 1` works nicely for both portrait and landscape.

Tip: try not to use `Dimensions` directly on styles.
In this case, it works ok for portrait, but try to change your device orientation... just for fun ;)

app/containers/FileModal.js
```diff
@@ -75,19 +98,31 @@ const ModalContent = React.memo(({
...
-					disableVolume
+					rate={1.0}
+					volume={1.0}
+					isMuted={false}
+					resizeMode='cover'
+					shouldPlay
+					isLooping={false}
+					style={styles.video}
+					posterStyle={styles.poster}
+					usePoster
```
diegolmello: What `poster` means? Thumb?
If so, it appears not to be working.

app/containers/FileModal.js
```diff
@@ -75,19 +98,31 @@ const ModalContent = React.memo(({
 		const uri = formatAttachmentUrl(attachment.video_url, user.id, user.token, baseUrl);
 		return (
 			<SafeAreaView style={styles.safeArea}>
```
diegolmello: We can remove this `SafeAreaView`.
It's going to be like this:

app/containers/FileModal.js
```diff
@@ -75,19 +98,31 @@ const ModalContent = React.memo(({
...
+					resizeMode='cover'
+					shouldPlay
+					isLooping={false}
+					style={styles.video}
+					posterStyle={styles.poster}
+					usePoster
+					useNativeControls
+					onReadyForDisplay={() => setLoading(false)}
+					onLoadStart={() => setLoading(true)}
+					onError={Alert.alert}
```
diegolmello: Don't alert this to the user.
It may crash.
Just send to analytics.

app/views/RoomView/index.js
diegolmello: Why do we need this function?

app/containers/FileModal.js
```diff
@@ -97,24 +132,28 @@ const FileModal = React.memo(({
 		onSwipeComplete={onClose}
 		swipeDirection={['up', 'left', 'right', 'down']}
```
diegolmello: This is disabling video seek.
Leave just `swipeDirection={['up', 'down']}`

app/views/RoomView/index.js
IlarionHalushka: to start/stop loader

app/containers/FileModal.js
IlarionHalushka: lol, I did try to rotate it - doesn't look good :)

app/containers/FileModal.js
IlarionHalushka: no, poster is an image that is displayed before video has loaded. I removed it because in fact it looks better without it.

app/views/RoomView/index.js
IlarionHalushka: if I do it this way I get eslint:  error    `Unexpected parentheses around single function argument having a body with no curly braces  arrow-parens`

app/views/MessagesView/index.js
IlarionHalushka: if I do it this way I get eslint:  error    `Unexpected parentheses around single function argument having a body with no curly braces  arrow-parens`

app/views/RoomView/index.js
djorkaeffalexandre: I think you can use `setFileLoading = fileLoading => this.setState({ fileLoading });`

app/views/MessagesView/index.js
djorkaeffalexandre: Only if you put:
`setFileLoading = (fileLoading) => this.setState({ fileLoading });`
You need to remove parentheses on params :)

djorkaeffalexandre:
worked nice for me! really pretty! 👍

app/views/RoomView/index.js
IlarionHalushka: 😍 😍 😍

Issue #1075, opened 2019-07-25
New: hide oauth button option


ggazzo:
https://github.com/RocketChat/Rocket.Chat/pull/15053 ;)

diegolmello:
@filipedelimabrito Can you do it right after #1018? Thanks!

PR #1140, opened 2019-08-20
[IMPROVEMENT] Checks if custom OAuth button should be shown

filipedelimabrito:
Defaults setting to true (even more for not being available on most servers yet).

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/1075

### DIFF
--- a/app/lib/rocketchat.js
+++ b/app/lib/rocketchat.js
@@ -802,9 +802,11 @@ const RocketChat = {
 		}
 	},
 	_determineAuthType(services) {
-		const { name, custom, service } = services;
+		const {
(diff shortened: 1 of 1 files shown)


app/lib/rocketchat.js
```diff
@@ -802,9 +802,11 @@ const RocketChat = {
 		}
 	},
 	_determineAuthType(services) {
-		const { name, custom, service } = services;
+		const {
+			name, custom, showButton = true, service
```
djorkaeffalexandre: Why not default `false`?

PR #1162, opened 2019-08-29
[FIX] Avatar shown when useRealName is activated

### LINKED ISSUE #1138: Avatars not being shown when UseRealName is activated
When your server is configured to use Real Names, the user's avatars are not being shown (it shows the autogenerated first letter ones).
I've tracked down the issue to Avatar.js, when it creates the Avatar URL, it uses the Real Name which doesn't exists on the server.

	const room = type === 'd' ? text : `@${ text }`;

I'll try to fix (will need to send the username down to Avatar.js from RoomItem) and create a PR (if no one fixed it yet).
Thanks!

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #1138


wreiske:
We need this! ❤️

app/presentation/RoomItem/index.js
```diff
@@ -273,15 +274,7 @@ class RoomItem extends React.Component {
...
-								<Avatar
-									text={name}
-									size={avatarSize}
-									type={type}
-									baseUrl={baseUrl}
-									style={styles.avatar}
-									userId={userId}
-									token={token}
-								/>
+								<Avatar text={avatar} size={avatarSize} type={type} baseUrl={baseUrl} style={styles.avatar} userId={userId} token={token} />
```
diegolmello: Wrong indentation

PR #1182, opened 2019-09-08
[IMPROVEMENT] Group push notifications and show user avatar on it

filipedelimabrito:
A demo of this improvement can be seen [here](https://drive.google.com/file/d/1_5n7HBkgUNJnymdybvaRHFG__XIj0eja/view?usp=sharing).


ggazzo:
@filipedelimabrito  about the images... as far I could see at the pr... we dont have avatars like:
image

how difficult would be implement the avatars? thx :)

filipedelimabrito:
Guys, a demo was attached on the PR description showing the improvement done here. 
Added the user avatar on the notification.

Note: We need to get the user avatar by sending the current logged user id and token to securely get the that avatar - right now if the servers doesn't allow getting it without them no avatar will be show.

ggazzo:
:clap:

android/app/src/main/java/chat/rocket/reactnative/pushnotification/CustomPushNotification.kt
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        notificationManager =
+            (mContext.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager)
+
+        groupBuilder = NotificationCompat.Builder(mContext, ejson.host)
+            .setContentTitle(title)
+            .setContentText(message)
+            .setSmallIcon(smallIcon)
+            .setStyle(NotificationCompat.InboxStyle().setSummaryText(title))
```
diegolmello: How about https://developer.android.com/reference/android/support/v4/app/NotificationCompat.MessagingStyle.html?

djorkaeffalexandre:
Closed by #1448 .

PR #1206, opened 2019-09-18
[IMPROVEMENT] Use Reanimated's Transitions instead of LayoutAnimation

diegolmello:
@RocketChat/ReactNative


app/share.js
```diff
@@ -78,18 +79,37 @@ class Root extends React.Component {
 
 	render() {
 		const { isLandscape } = this.state;
+		if (isAndroid) {
```
djorkaeffalexandre: We really need this condition? I think we can use directly `<LayoutAnimation>`

diegolmello: 🤦‍♂ Fixed

PR #1221, opened 2019-09-23
implemented react.fragment shorthand

### LINKED ISSUE #1203: [CHORE] Use React Fragment shorthand
We've been using `<React.Fragment></React.Fragment>` in some places, but we should be using its shorthand for better readability: `<></>`.

Normal usage: https://reactjs.org/docs/fragments.html#usage
Short syntax: https://reactjs.org/docs/fragments.html#short-syntax

Prateek93a:
@RocketChat/ReactNative

Closes #1203 

As per the request to use the shorthand for React.Fragment, I implemented the shorthand in every place wherever React.Fragment was used.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -238,11 +238,11 @@ class CustomInsideStack extends React.Component {
 	render() {
 		const { navigation } = this.props;
 		return (
-			<React.Fragment>
+			<>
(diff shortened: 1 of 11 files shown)


diegolmello:
@Prateek93a Thanks, but there's an open PR here: #1212

PR #1224, opened 2019-09-24
[FIX] Close SortDropdown on sort select

### LINKED ISSUE #1222: [BUG] Close SortDropdown on sort select
- Your Rocket.Chat Experimental app version: latest
- Your Rocket.Chat server version: Official Rocket.Chat server instance
- Device (or Simulator) you're running with: Samsung Galaxy J6 Android 9

In the following gif, the sorting pop up does not close after selecting a sorting scheme, so is this behaviour normal? Would not it be better if the popup closes after selecting an option?

ezgif com-video-to-gif

Prateek93a:
@RocketChat/ReactNative

Closes #1222

With this bug fix, after hitting the sort by alphabet button or sort by activity button, the SortDropdown will also close along with sorting

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -238,11 +238,11 @@ class CustomInsideStack extends React.Component {
 	render() {
 		const { navigation } = this.props;
 		return (
-			<React.Fragment>
+			<>
 				<InsideStackModal navigation={navigation} />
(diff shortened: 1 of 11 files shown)


Prateek93a:
Ok, I will fix my code

PR #1229, opened 2019-09-24
[FIX] Unify upload check

### LINKED ISSUE #1167: Not able to upload images or files
- Your Rocket.Chat Experimental app version: 1.18.0
- Your Rocket.Chat server version: 1.3.2
- Device (or Simulator) you're running with: Moto G7 Power / Android 9

This may be a react issue but I cannot upload images or files because it believes the file is too large regardless of the server setting or the actual file size.

diegolmello:
@RocketChat/ReactNative

Closes #1167

### DIFF
--- a/app/containers/MessageBox/Recording.js
+++ b/app/containers/MessageBox/Recording.js
@@ -5,6 +5,7 @@ import {
 } from 'react-native';
 import { AudioRecorder, AudioUtils } from 'react-native-audio';
 import { BorderlessButton } from 'react-native-gesture-handler';
+import RNFetchBlob from 'rn-fetch-blob';
 
 import styles from './styles';
 import I18n from '../../i18n';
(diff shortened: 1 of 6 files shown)


app/containers/MessageBox/Recording.js
```diff
@@ -102,7 +105,8 @@ export default class extends React.PureComponent {
 			this.recording = false;
 			const filePath = await AudioRecorder.stopRecording();
 			if (isAndroid) {
-				this.finishRecording(true, filePath);
+				const data = await RNFetchBlob.fs.stat(decodeURIComponent(filePath));
```
djorkaeffalexandre: I think here is better to use `FileSystem.getInfoAsync(fileUri, options)`, I can't use it on Share Extension yet that's why I use `RNFetchBlob` but this code not is on Share Extension then you can use `Expo-File-System`.

PR #1254, opened 2019-09-27
[FIX] Reply preview

djorkaeffalexandre:
@RocketChat/ReactNative

Captura de Tela 2019-09-27 às 16 43 18

Captura de Tela 2019-09-27 às 16 43 37

Captura de Tela 2019-09-27 às 16 45 10

Captura de Tela 2019-09-27 às 16 45 27


djorkaeffalexandre:
We will make it on another pull request with logic to preview messages using only specific markdown renders.

PR #1302, opened 2019-10-11
[NEW] Delete server

### LINKED ISSUE #1087: Remove server directly from server's list
## Describe the bug
Current it is only possible to remove a server by logging out from it. But if the server is not available anymore (i.e.: the server has been turned off) I cannot log in into to it to log out.

## To Reproduce
1. Connect to a running server
2. Shutdown the server
3. Try remove the server from the app

## Expected behavior
A long press on the server item from server's list could show an option to remove it.

## Screenshots
image

There is no way no remove servers `192.168.1.x` as they are not running anymore.

## Devices and Versions

Your Rocket.Chat.Android version: 3.4.2 (2070)
Your Rocket.Chat Server version: 1.2.0-develop

Mobile device model and OS version: Mi A2 - Android 9

CalebeRios:
Fix function to delete server

Translate messages of alert for brazilian portugues and english

Co-authored-by: Bruno Dantas <oliveiradantas96@gmail.com>

@RocketChat/ReactNative

Closes #1087

ezgif com-video-to-gif

### DIFF
--- a/app/i18n/locales/en.js
+++ b/app/i18n/locales/en.js
@@ -441,5 +441,7 @@ export default {
 	Version_no: 'Version: {{version}}',
 	You_will_not_be_able_to_recover_this_message: 'You will not be able to recover this message!',
 	Change_Language: 'Change Language',
-	Crash_report_disclaimer: 'We never track the content of your chats. The crash report only contains relevant information for us in order to identify problems and fix it.'
+	Crash_report_disclaimer: 'We never track the content of your chats. The crash report only contains relevant information for us in order to identify problems and fix it.',
+	Caution: 'Caution!',
+	Delete_Server: 'Are you sure do you want remove this server?'
 };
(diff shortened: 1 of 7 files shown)


djorkaeffalexandre:
I think is more better if you remove **all data** about this server when delete it on ServerDropdown.. like the database, `RNUserDefaults` info and the `serverRecord` on `serversDB`.

CalebeRios:
Thanks for changes @diegolmello. I will see the changes you made and finish that issue. My problem is to delete the database, because in logout we delete the active database, and I need get the database of another server. Can you give me a light?

diegolmello:
@CalebeRios Ok.
You should do something similar to `setActiveDB`, but without changing the current state.
Think in something like `getDB`, that runs a similar logic to as `setActiveDB` but it doesn't change `this.databases.activeDB`.
https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/database/index.js#L54

From there, you can receive a database obj and run an action to reset it.

CalebeRios:
@djorkaeffalexandre @diegolmello  I made the changes you request. Sorry for the delay.

djorkaeffalexandre:
It'll be closed by #1975 .
Thanks.

PR #1306, opened 2019-10-15
[IMPROVEMENT] Respect "Hide counter" preference

### LINKED ISSUE #1287: [IMPROVEMENT] Respect "Hide counter" preference
If "Hide counter" is set, we shouldn't show alerts from the channel.

- Web behaviour (correct):
image

- App behaviour (incorrect):
Screen Shot 2019-10-08 at 17 08 51

Look how the app is showing information in bold and blue. We should behave as if the app has nothing new (everything is read).

CalebeRios:
@RocketChat/ReactNative

Closes #1287

ezgif com-video-to-gif

### DIFF
--- a/app/views/RoomsListView/index.js
+++ b/app/views/RoomsListView/index.js
@@ -367,7 +367,7 @@ class RoomsListView extends React.Component {
 
 			// unread
 			if (showUnread) {
-				unread = chats.filter(s => s.unread > 0 || s.alert);
+				unread = chats.filter(s => (s.unread > 0 || s.alert) && !s.hideUnreadStatus);
 			} else {
 				unread = [];
 			}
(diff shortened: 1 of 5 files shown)


app/views/RoomsListView/index.js
```diff
@@ -601,6 +601,7 @@ class RoomsListView extends React.Component {
 			<RoomItem
 				alert={item.alert}
 				unread={item.unread}
+				hideUnreadStatus={item.hideUnreadStatus}
```
djorkaeffalexandre: You need to add this value to `Subscription` model on Database or this value not persist if it is offline..

CalebeRios:
@djorkaeffalexandre I updated this PR adding the hideUnreadStatus to schema of Subscription

PR #1319, opened 2019-10-21
[FIX] Invalid version

djorkaeffalexandre:
@RocketChat/reactnative 

After we start `serverSelectRequest` the `RocketChat.connect` is started and call `login`, when login success we call `fetchCustomEmojis`, `fetchPermissions` and this access `redux.getState().server.version`, but if we get this before `serverSelectSuccess` we have a null version.. it can occur when `/api/info` is slower than `/login`..

Other possible approach is create a action `setVersion` and call it when `getServerInfo` 🤔 ...
What do you think?

### DIFF
--- a/app/lib/methods/getSettings.js
+++ b/app/lib/methods/getSettings.js
@@ -2,6 +2,7 @@ import { InteractionManager } from 'react-native';
 import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
 import { Q } from '@nozbe/watermelondb';
 
+import RocketChat from '../rocketchat';
 import reduxStore from '../createStore';
 import * as actions from '../../actions';
 import settings from '../../constants/settings';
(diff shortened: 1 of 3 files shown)


app/lib/methods/getCustomEmojis.js
```diff
@@ -84,14 +84,18 @@ export async function setCustomEmojis() {
 export function getCustomEmojis() {
 	return new Promise(async(resolve) => {
 		try {
-			const serverVersion = reduxStore.getState().server.version;
+			const { server } = reduxStore.getState().server;
+			const serversDB = database.servers;
+			const serversCollection = serversDB.collections.get('servers');
+			const serverInfo = await serversCollection.find(server);
+
```
rodrigok: Is it possible to create a helper to prevent duplicate this code everywhere we need to check the version?

djorkaeffalexandre: Yeah! 👍

djorkaeffalexandre: @rodrigok what you think about it?
> Other possible approach is create a action setVersion and call it when getApiInfo 🤔 ...

rodrigok: @djorkaeffalexandre I'm not sure how those actions work so can't opine on that matter :(

djorkaeffalexandre: Ok! Thx :)

djorkaeffalexandre:
I found another way to reproduce it and will work on it..
Steps:
* Login on open.rocket.chat
* Add server
* Login on stable.rocket.chat
* Go to open.rocket.chat
* Add server
* Write stable.rocket.chat
* It raises a `version = null` and the app will be connecting infinitely.

PR #1322, opened 2019-10-21
[FIX] Server with subpath

djorkaeffalexandre:
@RocketChat/ReactNative

When we have a server with `https://open.rocket.chat/rocketchat` [subpath] it raises an error because we try to create and read a database file with name `open.rocket.chat/rocketchat`.
To fix it after remove `http://` or `https://` I replace all `/` with `.`, it will create `open.rocket.chat.rocketchat` and you can use multiples servers with differents `subpaths.

### DIFF
--- a/app/lib/database/index.js
+++ b/app/lib/database/index.js
@@ -52,7 +52,7 @@ class DB {
 	}
 
 	setActiveDB(database = '') {
-		const path = database.replace(/(^\w+:|^)\/\//, '');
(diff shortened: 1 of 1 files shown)


JoshMcCullough:
Thanks for the fix, we ran into this issue as well. Any idea when/what version this fix will show up in?

diegolmello:
@JoshMcCullough It's already available on Rocket.Chat 4.0.2

JoshMcCullough:
Great! I can confirm it is working now. Thanks.

PR #1283, opened 2019-10-08
Fixed Channel Name Issue

Prateek93a:
@RocketChat/ReactNative

Closes #1169

This fixes the name inconsistency that arises after changing the Room name. Previously the changed room name was not displayed at the Room View. Now it does.

Issue #1169, opened 2019-09-02
Cannot properly rename channel


reetp:
- Your Rocket.Chat Experimental app version: 1.19.0.10977
- Your Rocket.Chat server version: 1.3.2
- Device (or Simulator) you're running with: Samsung S9

Create a channel using the Desktop.

eg testchannel

Go to React App, edit channel to say 'Messing About' and save.

Room Changed to Messing About

Top of screen it still shows 'testchannel' until you go back to the channel list.

Editing channel shows 'messing-about' as the channel name.

Prateek93a:
@diegolmello If nobody has taken up this issue, can I work on this?

Prateek93a:
@diegolmello Sir, I have fixed the issue, can I raise the PR?

diegolmello:
@Prateek93a Don't be shy. You can always open PRs helping us. You don't have to ask :)

Prateek93a:
@diegolmello  Ok Sir, thanks.

djorkaeffalexandre:
Can't reproduce on latest version, I'll close this.
Thanks.

PR #1335, opened 2019-10-29
fixed-channelname-change-issue

Prateek93a:
@RocketChat/ReactNative

Closes #1283  #1169

Due to the same previous mistake, #1283 got messed up too. So I am raising this new PR with all the required changes and fixes.


app/views/RoomView/index.js
```diff
@@ -58,7 +58,7 @@ class RoomView extends React.Component {
 	static navigationOptions = ({ navigation }) => {
 		const rid = navigation.getParam('rid');
 		const prid = navigation.getParam('prid');
-		const title = navigation.getParam('name');
+		const title = navigation.state.params.title || navigation.getParam('name');
```
djorkaeffalexandre: You not need to set two different params.
You can use only `navigation.getParam('name');`

app/views/RoomView/index.js
```diff
@@ -336,6 +336,10 @@ class RoomView extends React.Component {
 		const observable = room.observe();
 		this.subSubscription = observable
 			.subscribe((changes) => {
+				const { navigation } = this.props;
```
djorkaeffalexandre: Is better if you use only `name` param, like:
```
const name = navigation.getParam('name');
if (changes.name !== name) {
  navigation.setParams({ name: changes.name });
}
```

Prateek93a:
@djorkaeffalexandre Sorry for this long delay. Thanks for reviewing my code. Actually I thought initially the name parameter might not have any value so I used two variables with OR. But as you mentioned that in line 339 that it is better to use only name parameter instead of setting title, so the issue i was thinking will not arise. Thanks for this review Sir.

Prateek93a:
@djorkaeffalexandre Sorry Sir, I messed things up here. I will raise new PR for the issue. Sorry for the incovinience

Prateek93a:
@djorkaeffalexandre Sorry Sir, I did not update the branch earlier, and also forgot to remove the changes I made for myself. But I have fixed everything now.

app/views/RoomView/index.js
```diff
@@ -390,6 +390,11 @@ class RoomView extends React.Component {
...
+				const name = navigation.getParam('name');
+				if (changes.name !== name) {
+					navigation.setParams({ name: changes.name });
```
diegolmello: I think we're missing other title headers here.
If it's a discussion or Real Name is enabled, the header comes from `fname`.
Check it out: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/views/RoomsListView/index.js#L484

app/views/RoomView/index.js
Prateek93a: Ok Sir, made the changes accordingly

app/views/RoomView/index.js
Prateek93a: Thanks for the review Sir. I have made the changes.

app/views/RoomView/index.js
Prateek93a: Thanks for the review Sir. So first I should check whether name param is having some value or not, if there is no value, then I should use fname param, am I right on this?

djorkaeffalexandre:
Can't reproduce on latest version.
Closing.
Thanks.

PR #1367, opened 2019-11-05
[FIX] Improve RoomView's List performance by removing messages comparison and delaying update

diegolmello:
@RocketChat/ReactNative

Note: This is just an experiment.


diegolmello:
After testing this PR on release mode, it didn't achieve the results we were expecting.
I'll leave it open for now if any new ideas pop up.

PR #1379, opened 2019-11-12
[IMPROVEMENT] Error messages on connect

djorkaeffalexandre:
@RocketChat/ReactNative

Now when we get serverInfo we check if `websocket` is enabled and if this is a `Rocket.Chat` server.

Captura de Tela 2019-11-12 às 16 48 08
Captura de Tela 2019-11-12 às 16 48 29
Captura de Tela 2019-11-12 às 16 50 01


app/i18n/locales/en.js
```diff
@@ -257,6 +257,7 @@ export default {
 	No_Reactions: 'No Reactions',
 	No_Read_Receipts: 'No Read Receipts',
 	Not_logged: 'Not logged',
+	Not_is_rc: 'Apparently this is not a server Rocket.Chat - Contact your server admin',
```
diegolmello: I think it should be a little different.
`Not_RC_Server`, `This is not a Rocket.Chat server` and we should try to separate `Contact your server admin` in another sentence, so it appears this way...
```
This is not a Rocket.Chat server.
Contact your server admin.
```

app/i18n/locales/en.js
```diff
@@ -375,7 +376,7 @@ export default {
 	Tap_to_view_servers_list: 'Tap to view servers list',
 	Terms_of_Service: ' Terms of Service ',
 	Theme: 'Theme',
-	The_URL_is_invalid: 'The URL you entered is invalid. Check it and try again, please!',
+	The_URL_is_invalid: 'The URL you entered is invalid or unable to establish a secure connection. Check it and try again, please!',
```
diegolmello: I don't think `, please` is necessary.

diegolmello:
```
Invalid URL or unable to establish a secure connection.
Contact your server admin.
```

app/lib/rocketchat.js
```diff
@@ -90,7 +90,22 @@ const RocketChat = {
 	},
 	async getServerInfo(server) {
 		try {
-			const result = await fetch(`${ server }/api/info`).then(response => response.json());
+			let result = await fetch(`${ server }/api/info`)
+				.then(response => response.text());
+			try {
+				result = JSON.parse(result);
+				if (!{}.hasOwnProperty.call(result, 'success')) {
```
diegolmello: I'm lazy and I think this is too complicated.
What do you think of `!(result && result.success)`?

app/lib/rocketchat.js
```diff
@@ -90,7 +90,22 @@ const RocketChat = {
...
+					return {
+						success: false,
+						message: 'Not_is_rc'
+					};
+				}
+			} catch (e) { // if returns html on body
```
diegolmello: It could fail because of other reasons, so I don't think this comment is necessary.

app/lib/rocketchat.js
```diff
@@ -90,7 +90,22 @@ const RocketChat = {
 	},
 	async getServerInfo(server) {
 		try {
-			const result = await fetch(`${ server }/api/info`).then(response => response.json());
+			let result = await fetch(`${ server }/api/info`)
+				.then(response => response.text());
```
diegolmello: Can you explain why `response.text()`?

Issue #1370, opened 2019-11-06
Unable to upload avatar through the app


mar-v:
- Your Rocket.Chat app version: 4.0.3 for official and 1.20.1-jitsi-bb12048 for Experimental
- Your Rocket.Chat server version: 2.0.0 (same with 2.2.0)
- Device (or Simulator) you're running with: HomTom HT70

Uploading a picture in the profile settings seems to do nothing. Steps to reproduce
1. Open the app -> Profile
2. Tap on the cloud icon to upload a new pic.
3. Select a pic form local directory, tap the check mark to approve the upload. 
4. Back to Profile page - nothing has changed, the button "Save changes" is greyed out.
5. Uploading using the link works just fine. 

Expected behaviour: after step 3 the button "Save changes" should be clickable and the new picture should appear as avatar, as it does on the desktop app, for example. 

Same thing with both apps (official and Experimental one)

rasos:
This would not fix #441, but would make it less relvant.

Aaryan-kapur:
May I work on this?

djorkaeffalexandre:
Looks like it's working fine now, when you select a new image, the button to save is enabled. :)
Thanks @mar-v. 🤗

Issue #1409, opened 2019-11-21
Profile picture upload is not working


ani4aniket:
I have attached a screen recording which I've faced during running the app on iOS simulator.
![Screen Recording 2019-11-21 at 5 23 41 PM](https://user-images.githubusercontent.com/31184591/69336576-cf382100-0c84-11ea-9a79-573c467f9b57.gif)

diegolmello:
@ani4aniket There're two issues apparently.
First one is not showing the image you've selected at the top of the view.
The other is the avatar cache not refreshing, because your avatar seems to be changed on the web.
<img src="https://user-images.githubusercontent.com/804994/69338943-0b4a9200-0c43-11ea-9c08-1ecfbfcb5614.png" width="350" />

ani4aniket:
Yeah, You are right as in my phone app. The image is not updated. 
![Screenshot_2019-11-21-18-54-16-04_6c2112e95a10afaab76cfa27a6d42c7b](https://user-images.githubusercontent.com/31184591/69341961-8ab28280-0c90-11ea-8ee6-e7d0712becdc.png)

nccurry:
Seeing this issue on android as well

mahal:
As by my tests, this is related to the caching issue #441 

But by my tests I encountered following when uploading avatars from both Android and iOS:

Steps to reproduce:
- Start RC-RN client 
- Login to open.rocket.chat
- Tap Menu > Profile
- Choose image from picture roll 
- Save Changes

Actual result:
- App shows "There was an error while changig avatar!"

Environment:
RC-RN client: 4.6.4.295 iOS, 4.6.4.2118 (Android)

The same happens on our own rocket.chat server: 2.4.11 
The log file shows
```
[...]

Issue #441, opened 2018-09-13
Avatar images doesn't refresh


skhizerabass:
Please can you test this app on Android devices. Avatar uploading doesn't work on android. I have installed the latest version from playstore but it is not uploading my avatar.

diegolmello:
@skhizerabass It's a known bug.
Currently this is a limitation in the image caching lib we're using: see https://github.com/DylanVann/react-native-fast-image/issues/197#issuecomment-385333467

We're open to PRs or even to react-native-fast-image PR fixing this error.
I'll leave this issue open until it's fixed.

skhizerabass:
I believe it never uploads the avatar on server. Because i never get any response from server Nor it changes even after uninstalling and reinstalling the app.
Can you look into this issue ?

diegolmello:
@skhizerabass Sure. I'll take a look as soon as react-native-fast-image is fixed too.

SteveUNSW:
Hey team,

Any updates on this? Avatar doesn't refresh and I can't change it to anything.

saadqbal:
We can resolve this for the time being by setting priority to high on FastImage in Avatar.js


`FastImage
		style={[styles.avatar]}
		source={{
		uri,
		priority: FastImage.priority.high
		}}
/>`

saadqbal:
@diegolmello our code is way too changed. Won't be possible for you to merge.

AugustoAleGon:
The best way for me was to add a ts which is always different...

diegolmello:
@AugustoAleGon Can you explain more or open a PR?

FreewheelingLive:
I chose the easy way. Enter any number after the URL. Restart the app. Then the photo has been changed.

mahal:
I did a deep investigation into the issue and would like to share following result:
RC-RN client does not reflect any changes to the Avatar. 

RC-RN client: 4.6.4.295 iOS, 4.6.4.2118 (Android)
RC server: 2.4.11

Things I tried and tested:

- Setting the Avatar thru API on server
- Setting the Avatar thru Admin-Panel on server
- Setting an external Avatar provider

The "Avatar cache time" is correctly reflected in the http-response of the Avatar, but the RC-RN client does not respect it. Also using the apps "Clear local server cache" will not refresh the avatar. 

[...]

PR #1413, opened 2019-11-22
 Fix Unable to upload avatar through the app

ani4aniket:
@RocketChat/ReactNative

Closes Issue #1370  #1409  #441 

Screen Recording 2019-11-23 at 4 49 44 AM

### DIFF
--- a/app/containers/Avatar.js
+++ b/app/containers/Avatar.js
@@ -1,103 +1,86 @@
-import React from "react";
-import PropTypes from "prop-types";
-import { View } from "react-native";
-import FastImage from "react-native-fast-image";
+import React from 'react';
+import PropTypes from 'prop-types';
+import { View } from 'react-native';
+import FastImage from 'react-native-fast-image';
 
-var date = new Date();
-var timestamp = date.getTime();
+const date = new Date();
+const timestamp = date.getTime();
 
-const formatUrl = (url, baseUrl, uriSize, avatarAuthURLFragment) =>
-  `${baseUrl}${url}?format=png&width=${uriSize}&height=${uriSize}${avatarAuthURLFragment}&t=${timestamp}`;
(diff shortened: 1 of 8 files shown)


app/containers/Avatar.js
```diff
@@ -1,80 +1,103 @@
...
-const Avatar = React.memo(({
-	text, size, baseUrl, borderRadius, style, avatar, type, children, userId, token
-}) => {
-	const avatarStyle = {
-		width: size,
-		height: size,
-		borderRadius
-	};
+const formatUrl = (url, baseUrl, uriSize, avatarAuthURLFragment) =>
+  `${baseUrl}${url}?format=png&width=${uriSize}&height=${uriSize}${avatarAuthURLFragment}&t=${timestamp}`;
```
diegolmello: By using a timestamp on the link, you're breaking the entire avatar cache.
Avatar will start to load properly, but they will be requested from network all the time.

ani4aniket: I have fixed the lint errors which is caused by Prettier. Please review it.

TDHTTTT:
Thanks a lot for the PR. Any news for merging or anything I can help with?

djorkaeffalexandre:
Hey guys, I'll just close this because of inactivity.
This can't be merged because it's not a good solution, if we use this, the avatars will be requested again at each time that appears at the screen.
We've tried to control the cache better but didn't find a good solution, so, we're waiting for a solution that will be implemented at our server side, with dynamics avatar urls.
Thanks anyway. 😁

danfelbm:
This issue can be traced back to 2018! It's been a long time since this issue was first encountered. I wonder how this hasn't been addressed (I'd have thought people would complain more about not being able to change their avatars)

shaheem-khanzada:
@djorkaeffalexandre @diegolmello I think its not an image cache problem because as I test it and find out that we are using React.memo in avatar component so that why our avatar is not getting re-render because previous and next username is same

diegolmello:
@shaheem-khanzada Can you open a PR fixing it then?
Thanks!

PR #1419, opened 2019-11-26
Feature/sol 821

zesinho:
Added Agenda and Community link to SideBarView


zesinho:
was not supposed to compare against base repo.

PR #1425, opened 2019-11-27
[IMPROVEMENT] Tap app and server version to copy to clipboard

### LINKED ISSUE #1415: [IMPROVEMENT] Tap app and server version to copy to clipboard
We should turn `Version` and `Server version` into touchable and copy the content to clipboard.
It's a nice way to get feedback.

Prateek93a:
@RocketChat/ReactNative

Closes #1415

Implemented the save to clipboard on tap feature. I used Alert.alert to notify the user that the text has been copied.


Prateek93a:
The code worked without /src as well. Now I didnot get the error. So I have removed /src part from import statement in the latest commit.

app/views/SettingsView/index.js
```diff
@@ -112,6 +112,11 @@ class SettingsView extends React.Component {
 		Share.share({ message: isAndroid ? PLAY_MARKET_LINK : APP_STORE_LINK });
 	}
 
+	saveToClipboard = (content) => {
+		Clipboard.setString(content);
+		Alert.alert('Copied to clipboard');
```
diegolmello: Can you show a toast instead?
Here's an example: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/containers/MessageActions.js#L259

app/views/SettingsView/index.js
```diff
@@ -221,10 +226,15 @@ class SettingsView extends React.Component {
...
-					<ListItem title={I18n.t('Version_no', { version: getReadableVersion })} testID='settings-view-version' />
+					<ListItem
+						title={I18n.t('Version_no', { version: getReadableVersion })}
+						onPress={() => this.saveToClipboard(getReadableVersion)}
+						testID='settings-view-version'
+					/>
 					<Separator />
 					<ListItem
 						title={I18n.t('Server_version', { version: server.version })}
+						onPress={() => this.saveToClipboard(server.version)}
```
diegolmello: I think we should try to prevent that anonymous function, like this:
```jsx
onPress={this.copyServerVersion}
```

And then on the new `copyServer` function, you get the content you want like:
```jsx
const { server } = this.props;
this.saveToClipboard(server.version)
```

And the same for the app version.

Prateek93a: Thanks for this review. I made the changes. Sir, so should we try to avoid using anonymous functions whenever possible?

app/views/SettingsView/index.js
Prateek93a: Actually even I didn't like the Alert box but I could not recall the word 'toast' even though I was thinking of toast box itself. Then I thought of 'modal' and 'alert'. So I mentioned that I used Alert so that a better alternative could be suggested  . Thanks for the review

PR #1465, opened 2019-12-12
[I18N] German translations

phriedrich:
@RocketChat/ReactNative

(Second attempt.)

This adds some missing German translations.


phriedrich:
The latest commit adds a space behind the colon to fix this problem:
IMG_6821

app/i18n/locales/de.js
```diff
@@ -383,20 +429,38 @@ export default {
...
 	Yes_action_it: 'Ja, {{action}}!',
 	Yesterday: 'Gestern',
 	You_are_in_preview_mode: 'Sie befinden sich im Vorschaumodus',
 	You_are_offline: 'Sie sind offline',
 	You_can_search_using_RegExp_eg: 'Sie können mit RegExp suchen. z.B. `/ ^ text $ / i`',
-	You_colon: 'Sie:',
+	You_colon: 'Sie: ',
 	you_were_mentioned: 'Sie wurden erwähnt',
-	you: 'sie',
+	you: 'Sie',
```
djorkaeffalexandre: I think we prefer this with lower case.

djorkaeffalexandre:
Thanks! We really appreciate your contribution!! <3

app/i18n/locales/de.js
phriedrich: Why? The formal „Sie“ is always written with uppercase „S“. 

The informal „du“ can be written with „d“ (newer style) or „D“ (traditional).

PR #1502, opened 2019-12-27
[FIX] Profile fields automatically reset

### LINKED ISSUE #1496: The field is resetting
- Your Rocket.Chat Experimental app version: 4.2.0
- Your Rocket.Chat server version: 2.4.0-rc.0
- Device (or Simulator) you're running with: Android & iOS

For the profile part, after I keyed-in value to the field, then it seems like is resetting automatically in awhile. 
And also for the image, i try to select image from picker and debug using Reactotron it showed there is value in the avatar object after selected, but it also seems like resetting and became empty right after.

reset field

ssbing99:
@RocketChat/ReactNative

Closes #1496


app/views/ProfileView/index.js
```diff
@@ -81,7 +82,7 @@ class ProfileView extends React.Component {
 
 	componentWillReceiveProps(nextProps) {
 		const { user } = this.props;
-		if (user !== nextProps.user) {
+		if (isEqual(user, nextProps.user)) {
```
djorkaeffalexandre: You can use `equal` from `deep-equal` and not need to import `lodash/isEqual` :)

ssbing99: Oops , sorry my bad, will resubmit again

ssbing99: I have pushed another commit

app/views/ProfileView/index.js
```diff
@@ -81,7 +81,7 @@ class ProfileView extends React.Component {
 
 	componentWillReceiveProps(nextProps) {
 		const { user } = this.props;
-		if (user !== nextProps.user) {
+		if (equal(user, nextProps.user)) {
```
djorkaeffalexandre: `if (!equal(user, nextProps.user))`

ssbing99: argh .... have pushed another commit. Thank you

PR #1510, opened 2020-01-02
[FIX] Admin Panel

### LINKED ISSUE #1476: Admin Panel is not visible in the Android app.
- Your Rocket.Chat app version: 4.1.0.2095
- Your Rocket.Chat server version: 2.3.0
- Device (or Simulator) you're running with: Galaxy J7 SM-J700H (Android 6.0.1)

Steps to reproduce: 
1. Log into your server via an Android app (being an Admin on this server). 
2. Open the left-side menu. 
3. "Chats" "Profile" "Settings" "Logout" options are visible. 

Expected behavior: having the "Admin Panel" option visible in the left-side menu.

Ritesh-Ag:
@RocketChat/ReactNative

Closes #1476 

I have added componentwillmount() method and updated the state for admin access under this method.

### DIFF
--- a/app/views/SidebarView/index.js
+++ b/app/views/SidebarView/index.js
@@ -60,6 +60,10 @@ class Sidebar extends Component {
 		};
 	}
 
+	componentWillMount() {
+		this.setIsAdmin();
+	}
+
 	componentDidMount() {
 		this.setStatus();
 		this.setIsAdmin();


app/views/SidebarView/index.js
```diff
@@ -60,6 +60,10 @@ class Sidebar extends Component {
 		};
 	}
 
+	componentWillMount() {
```
djorkaeffalexandre: Hello @Ritesh-Ag , I think this method isn't the best approach to resolve this issue, `componentWillMount` [will be deprecated](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount) soon.
I think this problem is related to database asynchronous, we try to read permissions from database before this was insert.

Ritesh-Ag: @djorkaeffalexandre Okay Thank you so much sir. Can you please suggest me what should be the best approach to solve the issue?

PR #1542, opened 2020-01-10
[I18N] Gallery Permissions DE

phriedrich:
@RocketChat/ReactNative


phriedrich:
Is "license/cla Expected — Waiting for status to be reported" holding this back?

I had signed it before (#1465) but it seems that it's not automatically applied here, while providing me no option to do it again.

PR #1551, opened 2020-01-12
[FIX] Hide unread counter

### LINKED ISSUE #1443: Muted channels show up in unread counter
I have some channels muted (no notifications and no unread counter) but they increment the unread counter on the mobile app.

aashaypalliwar:
@RocketChat/ReactNative

Closes #1443

### DIFF
--- a/app/presentation/RoomItem/UnreadBadge.js
+++ b/app/presentation/RoomItem/UnreadBadge.js
@@ -8,10 +8,7 @@ import { themes } from '../../constants/colors';
 const UnreadBadge = React.memo(({
 	theme, unread, userMentions, type, hideUnreadStatus
 }) => {
-	if (hideUnreadStatus) {
-		return;
-	}
-	if (!unread || unread <= 0) {
+	if (hideUnreadStatus || !unread || unread <= 0) {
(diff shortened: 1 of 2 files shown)


app/presentation/RoomItem/UnreadBadge.js
```diff
@@ -6,8 +6,11 @@ import styles from './styles';
 import { themes } from '../../constants/colors';
 
 const UnreadBadge = React.memo(({
-	theme, unread, userMentions, type
+	theme, unread, userMentions, type, hideUnreadStatus
 }) => {
+	if (hideUnreadStatus) {
+		return;
+	}
 	if (!unread || unread <= 0) {
```
djorkaeffalexandre: Add the condition here, you don't need to create other if case. :)
```
if (hideUnreadStatus || !unread || unread <= 0) {
  return;
}
```

djorkaeffalexandre:
Hello @aashaypalliwar, I looked it up on our web client and counter it showed when you are mentioned even if `hideUnreadStatus` is `true`.
Try to do a similar behavior of the web client.
Thanks

Issue #1544, opened 2020-01-10
No confirmation message is displayed on successful registration 


m1-key:
- Your Rocket.Chat Experimental app version: 4.2
- Your Rocket.Chat server version: 2.5.0-develop open.rocket.chat
- Device (or Simulator) you're running with: Android (P)

No confirmation message is displayed when a user successfully register on this app. There is no way user can know that registration is done successfully unless the app is restarted.
![ss](https://user-images.githubusercontent.com/39923784/72157628-66e10280-33de-11ea-829c-54c152f2980b.jpeg)

m1-key:
Can I work on this issue ?

diegolmello:
@m1-key Sure. The app needs to redirect the user to rooms list after that screen.

msisskin:
Any progress on this issue? We have users reporting the same behavior and it's causing user acceptance difficulties. They aren't always in front of a computer to register that way. Thanks!

diegolmello:
@msisskin There's a fix on the way: #1728

msisskin:
@diegolmello Thank you for the quick update

djorkaeffalexandre:
Hello @msisskin, we’re working on this, probably it'll be launched in the next release, `4.5`.
Thanks.

PR #1567, opened 2020-01-12
[FIX] Email verification message and navigation to login portal

### LINKED ISSUE #1548: No message is displayed to verify the email-id

There is an issue and it is that when we create an account ( not using the pre-described methods like via google, github etc.) then it needs email-verification , so what I am planning is that when we create an account using above way then it must re-direct us to login portal after registration so that after email-id verification we can login.  When we click on Register then an alert screen should pop up and informs us to verify our email-ids and a button on that screen which will take us to login page. Then after verifying the mail we can login.

m1-key:
@RocketChat/ReactNative

Closes #1548 

In this I have created an Alert Screen , which pop up when a user press the register button ( when he/she is registering through ```Create an account``` option, not by any pre-described methods like ```Continue via Google```, GitHub etc. ) . On that alert screen Congrats will be written for registering on Rocket Chat and then there will be a message Verify your email and a close button. Now user will have information that they have to verify their email which was not mentioned earlier. And on pressing the close button, user will be redirected to LoginView i.e Login Portal, where after successfully verifying the mail , user can login. 

I have removed the ```login request``` prop, so user will only be able to login once the email is verified. 

Login via Google , GitHub etc issue is resolved the pull request I made in #1544 , so if both the pull requests are merged then whole register issue will be resolved. i.e
When we register via google, github then on pressing the register button , we will be redirected to Room List (#1544 ) and when we choose ```Create an Account``` and press register button then we will be redirected to login portal and message will be displayed which will say to verify the email.

Here is view of that.

ESe_Hf

One thing more is that I have provided the message 'Congrats' and 'Verify your email' in all the languages supported by the app.


app/i18n/locales/zh-CN.js
```diff
@@ -333,6 +334,7 @@ export default {
 	Username: '用户名',
 	Username_or_email: '用户名或邮箱',
 	Validating: '正在验证',
+	Verify_your_email: '验证您的电子邮件',
```
diegolmello: If you don't know the language, you shouldn't use Google Translate.

app/views/RegisterView.js
```diff
@@ -119,13 +119,20 @@ class RegisterView extends React.Component {
...
 			name, email, password, username, customFields
 		} = this.state;
-		const { loginRequest } = this.props;
 
 		try {
 			await RocketChat.register({
 				name, email, pass: password, username, ...customFields
 			});
-			await loginRequest({ user: email, password });
+			const { navigation } = this.props;
```
diegolmello: There's a setting on the server enforcing email verification.
You should do this only if that setting is enabled.

diegolmello:
Will be fixed by #1725

PR #1568, opened 2020-01-12
[FIX] Added navigation from Register to Room List

### LINKED ISSUE #1544: No confirmation message is displayed on successful registration

m1-key:
@RocketChat/ReactNative

Closes #1544 

I added just 2 line code into the RegisterView.js and SetUsername.js , first line extracted the navigation from the props and then in the second line I navigated from above two mentioned screens to the RoomsListView. So earlier when we click on Register button , nothing happens which can be seen on screen. Now on pressing the Register Button , we will be directed to the RoomsListView.

Here is view of that :

vSUosn

### DIFF
--- a/app/views/RegisterView.js
+++ b/app/views/RegisterView.js
@@ -126,12 +126,12 @@ class RegisterView extends React.Component {
 				name, email, pass: password, username, ...customFields
 			});
 			await loginRequest({ user: email, password });
+			const { navigation } = this.props;
+			navigation.navigate('RoomsListView');
 		} catch (e) {
 			Alert.alert(I18n.t('Oops'), e.data.error);
 		}
 		this.setState({ saving: false });
-		const { navigation } = this.props;
-		navigation.navigate('RoomsListView');
 	}
 
 	renderCustomFields = () => {
--- a/app/views/RegisterView.js
+++ b/app/views/RegisterView.js
@@ -130,6 +130,8 @@ class RegisterView extends React.Component {
 			Alert.alert(I18n.t('Oops'), e.data.error);
 		}
 		this.setState({ saving: false });
+		const { navigation } = this.props;
+		navigation.navigate('RoomsListView');
 	}
 
 	renderCustomFields = () => {
--- a/app/views/RegisterView.js
+++ b/app/views/RegisterView.js
@@ -119,15 +119,13 @@ class RegisterView extends React.Component {
 		const {
 			name, email, password, username, customFields
 		} = this.state;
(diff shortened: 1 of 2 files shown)


app/views/RegisterView.js
```diff
@@ -130,6 +130,8 @@ class RegisterView extends React.Component {
 			Alert.alert(I18n.t('Oops'), e.data.error);
 		}
 		this.setState({ saving: false });
+		const { navigation } = this.props;
```
djorkaeffalexandre: On both cases you can't redirect to RoomsList if some error occurred on `catch`. 😁
I think it's better you can put it after `await loginRequest({ user: email, password });`.

app/views/SetUsernameView.js
```diff
@@ -98,6 +98,8 @@ class SetUsernameView extends React.Component {
 			log(e);
 		}
 		this.setState({ saving: false });
+		const { navigation } = this.props;
```
djorkaeffalexandre: Same of comment above.

app/views/RegisterView.js
m1-key: So basically I have to put ```await loginRequest({ user: email, password });``` before ```const { navigation } = this.props``` , is it @djorkaeffalexandre  ?

app/views/SetUsernameView.js
m1-key: I have made the changes, now you can verify that sir .

djorkaeffalexandre:
Hello @m1-key , how you can see here: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/sagas/login.js#L118, we make this navigation on `appStart` after login success.
With your PR, we had navigate two times to RoomsList. 
Can you provide an example about when the navigation doesn't work?
Thanks.

m1-key:
@djorkaeffalexandre  What if I change the navigation ```Create an Account``` from there to login portal and for registration via Google, GitHub takes us to RoomList as there is no need of any verification. Something similar I have done in #1567 . So when user create an account using ```Create an Account``` method then user will be re-directed to login portal where he/she can login after verifying the email. And when the user create the account via Google or GitHub , then he/she will be redirected to Room list .

m1-key:
@djorkaeffalexandre  any changes required in this ?

PR #1584, opened 2020-01-15
[FIX] Change username causes weird behaviour on avatar

### LINKED ISSUE #1583: Avatar changes as username is changed in Profile View
- Your Rocket.Chat Experimental app version: 4.2.2.9999999

This is a bug issue. As the gif shows, when we are changing the username, the avatar image gets changed simultaneously. 
ezgif com-crop(8)

Prateek93a:
@RocketChat/ReactNative

Closes #1583

This Pr solves the changing avatar issue. Actually to set avatar, username is sent as a prop, and when the username is being changed, the avatar also gets changed when the component rerenders, so I set an initial username separately to prevent this from happening.
ezgif com-crop(9)

### DIFF
--- a/app/views/ProfileView/index.js
+++ b/app/views/ProfileView/index.js
@@ -59,7 +59,7 @@ class ProfileView extends React.Component {
 		saving: false,
 		name: null,
 		username: null,
-		initialUsername: null,
+		avatarText: null,
 		email: null,
 		newPassword: null,
 		currentPassword: null,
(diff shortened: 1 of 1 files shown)


app/views/ProfileView/index.js
```diff
@@ -59,6 +59,7 @@ class ProfileView extends React.Component {
 		saving: false,
 		name: null,
 		username: null,
+		initialUsername: null,
```
djorkaeffalexandre: IMO is better if you use something like `avatarText` instead `initialUsername`.

Prateek93a:
@djorkaeffalexandre Thanks for the reivew Sir, I have made the change.

djorkaeffalexandre:
Will be fixed by https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2311/commits/ff73068bdd515bb92bfe23411c72b80fb9eb1b25, thanks for your contribution. 😁 
I've added you as co-author. :)

PR #1586, opened 2020-01-15
[FIX] Channels doesn't load

djorkaeffalexandre:
@RocketChat/ReactNative

We need to catch problems on `db.batch` for rooms get and not update `roomsUpdatedAt` if save to db raises any error.


app/sagas/rooms.js
```diff
@@ -72,23 +68,15 @@ const handleRoomsRequest = function* handleRoomsRequest() {
...
-				log(e);
-			}
-			return allRecords.length;
-		});
-
-		yield serversDB.action(async() => {
-			try {
-				await serverRecord.update((record) => {
-					record.roomsUpdatedAt = newRoomsUpdatedAt;
+				await serversDB.action(async() => {
```
diegolmello: This block should run if `!subscriptions.length`

PR #1587, opened 2020-01-15
[FIX] Blank lines in messages should be rendered as in browser view

### LINKED ISSUE #1513: Blank lines in messages are not rendered
- Your Rocket.Chat Experimental app version: Not Experimental but stable 4.1.0 on Android
- Your Rocket.Chat server version: 2.1.3
- Device (or Simulator) you're running with: Apple Phones or Motorola One _(Same issue)_

How to reproduce:
 - Write a message like this
   ```
   Whatever
   <BLANK LINE>
   Whatever
   ```

PC:
image

Phone:
image

aKn1ghtOut:
@RocketChat/ReactNative

Closes #1513

Write a message like this

> hijk
> 
> 
> jk

Browser Render:

Current app render:

FIXED app render:

The markdown library was removing new lines as they did not recognize whitespace to be a legitimate character denoting any specific message. I replaced all newlines with just whitespace in them to new lines with the special whitespace character that is not escaped by the markdown renderer. It solves the problem and makes it future proof in case markdown decides to solve this themselves later.


aKn1ghtOut:
I am not able to do anything about the license/cla problem(not getting a "details" link next to it)
On the "Rocket.Chat" repository, a bot provided the link to sign, which I did. Any suggestions on how to solve this?

app/containers/markdown/index.js
```diff
@@ -381,6 +381,7 @@ class Markdown extends PureComponent {
 		// Ex: '[ ](https://open.rocket.chat/group/test?msg=abcdef)  Test'
 		// Return: 'Test'
 		m = m.replace(/^\[([\s]]*)\]\(([^)]*)\)\s/, '').trim();
+		m = m.replace(/\n(?=\n)/g, '\n&nbsp;');
```
diegolmello: I didn't had the chance to investigate the markdown, but it looks like this shouldn't be done in the component.
The problem is most likely to be on the markdown lib not parsing correctly.
Can you check?

aKn1ghtOut: Yes, the problem definitely resided in the markdown library, in that it doesn't render empty lines, following from HTML standards. The problem here is that, even though that is acceptable behavior for the markdown library by its own standards, it introduces inconsistencies in Rocket Chat, being that these empty lines are acknowledged on the browser, but not in the app.
The change I made puts an empty space character in the empty lines so that markdown acknowledges them as would be, on the browser client. This also doesn't introduce any possible future problems, in case they change their policy on this.
One other solution could be to replace these particular **\n** breaks with <br> tags, instead of adding a whitespace character.

aKn1ghtOut: @diegolmello What should I do?

localguru:
Great! :+1:

PR #1593, opened 2020-01-16
minimum password length set to 6

### LINKED ISSUE #1592: There should be a minimum password length.

Right now app is supporting even a password of length 1. There should be a minimum length for this, so that password can be somewhat strong.

m1-key:
@RocketChat/ReactNative

Closes #1592 

Now user have to enter atleast 6 characters ( trailing whitespaces are not considered as characters here ) then only the register button will be active.

Here it is :

AgyI2Z

### DIFF
--- a/app/views/RegisterView.js
+++ b/app/views/RegisterView.js
@@ -106,6 +106,9 @@ class RegisterView extends React.Component {
 				requiredCheck = requiredCheck && customFields[key] && Boolean(customFields[key].trim());
 			}
 		});
+		if (password.trim().length < 6) {
+			return false;
+		}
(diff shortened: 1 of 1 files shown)


m1-key:
@diegolmello  Sir, since it will be useful for users to know that they need to input atleast 6 characters, so the placeholder which is currently showing ```Password``` can be changed into ```Password(Minimum six characters)```. Since this change should be displayed into all seven supported languages so that all type of user understands that they require atleast 6 characters to setup the password.  For this I have to set up a custom message for all the languages.
Should I change the placeholder info from ```Password``` to ```Password(Minimum six characters)``` or leave it as it ? Please help

PR #1595, opened 2020-01-16
[FIX] Check permission to pin message

### LINKED ISSUE #1594: Hide the pin message option
- Your Rocket.Chat Experimental app version: 4.2.2.999999

It is a minor improvement. As the gif shows, the Pin message feature is not working on direct message room. I checked the web client, and it doesn't provide the Pin message feature on direct message so on mobile client also, the option should be hidden for Direct message rooms.
ezgif com-crop(10)

Prateek93a:
@RocketChat/ReactNative
Removed pin message option

Closes #1594

Added condition to check if the room is direct message room and removed the pin message option accordingly
ezgif com-crop(11)

### DIFF
--- a/app/containers/MessageActions.js
+++ b/app/containers/MessageActions.js
@@ -46,8 +46,6 @@ class MessageActions extends React.Component {
 			Message_AllowStarring, Message_AllowPinning, Message_Read_Receipt_Store_Users, user, room, message, isReadOnly
 		} = this.props;
 
-		const isDirectMessage = room.t === 'd';
-
 		// Cancel
 		this.options = [I18n.t('Cancel')];
 		this.CANCEL_INDEX = 0;
@@ -89,7 +87,7 @@ class MessageActions extends React.Component {
 		}
 
 		// Pin
-		if (Message_AllowPinning && !isDirectMessage) {
+		if (Message_AllowPinning && this.hasPinPermission) {
 			this.options.push(I18n.t(message.pinned ? 'Unpin' : 'Pin'));
 			this.PIN_INDEX = this.options.length - 1;
 		}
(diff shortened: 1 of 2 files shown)


app/views/RoomActionsView/index.js
```diff
@@ -271,6 +266,15 @@ class RoomActionsView extends React.Component {
 				testID: 'room-actions-auto-translate'
 			});
 		}
+		if (!isDirectRoom) {
```
diegolmello: You can remove it.
Even if the user doesn't have access to pin, he always has access to see what's pinned.

app/views/RoomActionsView/index.js
```diff
@@ -250,18 +251,12 @@ class RoomActionsView extends React.Component {
...
-					name: I18n.t('Pinned'),
-					route: 'MessagesView',
-					params: { rid, t, name: 'Pinned' },
-					testID: 'room-actions-pinned'
 				}
 			],
 			renderItem: this.renderItem
 		}];
 
+
```
diegolmello: Can you remove it?

app/containers/MessageActions.js
```diff
@@ -87,7 +89,7 @@ class MessageActions extends React.Component {
 		}
 
 		// Pin
-		if (Message_AllowPinning) {
+		if (Message_AllowPinning && !isDirectMessage) {
```
diegolmello: That's not quite right.
Pin is should work on direct messages.

We should check for pin-message permission instead of the room type.
https://github.com/RocketChat/Rocket.Chat/blob/develop/app/message-pin/client/actionButton.js#L57

Prateek93a:
@diegolmello Thanks for the review Sir, I will make the changes

Prateek93a:
@diegolmello I made the changes. Actually, initially I thought that there was  no feature of pinning messages for direct messaging room hence I implemented that code.But ,it is related to permissions as you mentioned. Thanks for the review Sir.

app/views/RoomActionsView/index.js
```diff
@@ -238,6 +238,13 @@ class RoomActionsView extends React.Component {
 					params: { rid, t, name: 'Starred' },
 					testID: 'room-actions-starred'
 				},
+				{
```
djorkaeffalexandre: Why did you make these changes? Revert this, please. Thanks a lot.

Prateek93a:
@djorkaeffalexandre Sorry, I made ordering mistake there. Now I have fixed it. Thanks for the review

djorkaeffalexandre:
Hey @Prateek93a, I'm using your logic at #2114 to check permission to pin, thanks about your contribution.
I'll close this. 😁

PR #1598, opened 2020-01-16
fixed minimum length for new Password field of Profile View to minimum six characters

### LINKED ISSUE #1597: Minimum password length for New Password field of Profile View is 1

On clicking the profile, we get the option for new password. And if we only enter a single character then the ```Save Changes``` button becomes active, which is not efficient. Atleast 6 characters must be present in the password so that it is a bit secure and then only the ```Save Changes``` button becomes active.

m1-key:
@RocketChat/ReactNative

Closes #1597 

Now minimum 6 characters will be required to make the ```Save Changes``` button active.

Earlier it was like this : 
choe-K

Now it is like this : 

Y7nfzo

### DIFF
--- a/app/views/ProfileView/index.js
+++ b/app/views/ProfileView/index.js
@@ -118,6 +118,12 @@ class ProfileView extends React.Component {
 		});
 	}
 
+	validPassword = (password) => {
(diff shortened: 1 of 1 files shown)


diegolmello:
Thanks, but you should be following the password settings as I said on https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/1592#issuecomment-575223389

PR #1603, opened 2020-01-19
Fix flickering avatar by showing pre image

### LINKED ISSUE #1572: Show background view until Avatar is loaded
- Your Rocket.Chat Experimental app version: 4.2.1.999999

As the following gif shows, when avatar has not loaded yet, there is no pre image like thing to indicate that there is an image here. After loading the avatar just pops into the display. This is more noticeable on slow internet. It would be helpful to have some background image like, to show until the actual image has loaded.
ezgif com-crop(5)

Prateek93a:
@RocketChat/ReactNative

Closes #1572 

This Pr fixes the issue.
ezgif com-crop(13)


app/views/ProfileView/index.js
```diff
@@ -109,6 +110,7 @@ class ProfileView extends React.Component {
 		this.setState({
 			name,
 			username,
+			intialUsername: username,
```
diegolmello: These `initialUsername` are not needed.

Prateek93a: Sorry, that was related to another issue, I made a mistake here. But now I have fixed it

PR #1607, opened 2020-01-21
When we listen to any audio message then it will be good if the time is displayed

### LINKED ISSUE #1557: When we listen to any audio message then it will be good if the timer is displayed.
when we listen to any audio message then the timer should start until the audio ends so that we can see how much audio we have listened

devyaniChoubey:
…

@RocketChat/ReactNative

Closes #1557 

Peek 2020-01-14 01-36


app/containers/message/Audio.js
```diff
@@ -197,7 +197,7 @@ class Audio extends React.Component {
 						onValueChange={this.onValueChange}
 						thumbImage={isIOS && { uri: 'audio_thumb', scale: Dimensions.get('window').scale }}
 					/>
-					<Text style={[styles.duration, { color: themes[theme].auxiliaryText }]}>{this.duration}</Text>
+					{paused ? <Text style={[styles.duration, { color: themes[theme].auxiliaryText }]}>{this.duration}</Text> : <Text style={[styles.duration, { color: themes[theme].auxiliaryText }]}>{formatTime(currentTime)}</Text>}
```
diegolmello: The whole `<Text style={[styles.duration, { color: themes[theme].auxiliaryText }]}>...</Text>` is repeated twice.
Why not creating a new variable and use it like this?
```
<Text style={[styles.duration, { color: themes[theme].auxiliaryText }]}>
    {newVariable}
</Text>
```

devyaniChoubey:
@diegolmello sorry sir for late reply I was busy in my college exams.
I will do it

devyaniChoubey:
@sarthakpranesh  I will try to do the changes today only.Please let me do it. I was busy in my college exams.

djorkaeffalexandre:
Refer to #2073.

PR #1610, opened 2020-01-21
[FIX] Reply preview not updating

### LINKED ISSUE #1609: [BUG] Reply Message preview not updating
- Your Rocket.Chat Experimental app version: 4.2.4.9999999999

This is a bug issue. As the following gif shows, when we already have reply component on the screen, and then try to reply on some other message, the reply component is not updated, it still shows the previous message on which we used reply action. Moreover when we write a reply, it gets sent to the latest message on which we used the reply action.
ezgif com-crop (10)

Prateek93a:
@RocketChat/ReactNative

Closes #1609

 I don't know if how I solved it is the best way to do it but it works now.
ezgif com-crop (11)

### DIFF
--- a/app/views/RoomView/index.js
+++ b/app/views/RoomView/index.js
@@ -162,6 +162,7 @@ class RoomView extends React.Component {
 			lastOpen: null,
 			reactionsModalVisible: false,
 			selectedMessage: selectedMessage || {},
+			replyMessage: {},
 			canAutoTranslate: false,
 			loading: true,
 			showActions: false,
@@ -441,18 +442,14 @@ class RoomView extends React.Component {
 		}
 	}
 
-	onReplyInit = async(message, mention) => {
-		const { replying } = this.state;
-		if (replying) {
-			await this.setState({ replying: false });
-		}
+	onReplyInit = (message, mention) => {
 		this.setState({
-			selectedMessage: message, replying: true, showActions: false, replyWithMention: mention
+			replyMessage: message, replying: true, showActions: false, replyWithMention: mention, selectedMessage: {}
 		});
 	}
 
 	onReplyCancel = () => {
-		this.setState({ selectedMessage: {}, replying: false });
+		this.setState({ replyMessage: {}, replying: false });
 	}
 
 	onReactionInit = (message) => {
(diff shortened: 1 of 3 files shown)


Prateek93a:
@djorkaeffalexandre Sir, I tried the solution but I found some bugs.
 The reply preview got updated before we even pressed the reply message option, hence even we don't press reply message option after long pressing some message, the reply preview gets updated. 
Also, after opening a reply preview , then long pressing on another message but not selecting any option, the message action popup stops doesn't open, similar to #1612

djorkaeffalexandre:
Hello @Prateek93a.
So, you need to try use another logic, like, instead `replying` as a `boolean`, you set this as a message that you want to show in reply preview.
About the problem with the message actions popup, you have another PR to fix this, then you not need to consider this here.

Prateek93a:
@djorkaeffalexandre Ok Sir, on it.

Prateek93a:
@djorkaeffalexandre Thanks for the review Sir, I made the changes you suggested. 
Actually `if (!equal(nextProps.message, message)) {
  return true;
}` this caused the reply preview component to rerender every time we selected a new message by long pressing on it, but without it, the reply preview updates only when we have selected reply option on a new message. Now, only one more bug is left i.e. #1612. I will make changes to the Pr there. Sorry for the bad code, I will try my best to give better PRs. Thanks

Prateek93a:
@djorkaeffalexandre  Yes Sir,my previous solution was not good, it more like a work around. Now  the problem is fixed, Here I used your idea of having a separate variable for reply message. And then made the changes accordingly. Also I found that edit message also had the same bug, i.e. it did not update as well. So I added a commit to fix that issue as well instead of raising a new Pr, because the solution was almost similar. Now only #1612 is causing the bug. I will try to fix it as soon as possible.  Thanks for review Sir!

app/views/RoomView/index.js
```diff
@@ -417,23 +419,27 @@ class RoomView extends React.Component {
...
 	}
 
 	onErrorActionsHide = () => {
 		this.setState({ selectedMessage: {}, showErrorActions: false });
 	}
 
 	onEditInit = (message) => {
-		this.setState({ selectedMessage: message, editing: true, showActions: false });
+		this.setState({
+			editMessage: message, selectedMessage: {}, editing: true, showActions: false
```
Prateek93a: Here, I thought after selecting edit or reply option, I should make the `selectedMessage` as empty object for selecting new messages, but I can change it if needed.

PR #1613, opened 2020-01-22
[FIX] Message Actions not opening

### LINKED ISSUE #1612: [BUG] Message Action not opening
- Your Rocket.Chat Experimental app version: 4.2.4.999999999

Its a bug issue. As the following gif shows, when we select `Permalink` wtih the reply or the edit box open, after that, Message actions are not shown on long pressing any message. 
To reproduce the issue, 
-open message actions on some message and select edit or reply.
-without closing edit or reply box, open message actions on any message again and select Permalink.
- now try to open message actions again by long pressing on any message. It does not open now. 
ezgif com-crop(16)

Prateek93a:
@RocketChat/ReactNative

Closes  #1612

This Pr fixes the issue. Actually the issue was with the logical error. When replying or editing, the `showAction` is set to true. Now when we  press `Permalink` without closing reply or edit box, the `editing` or `replying` booleans are set to `true`. So the `onActionsHide` function returns without setting `showActions` to false. Hence after that, when we try to longPress any message, since `showActions` is still set to true, the state does not update. So I added the logic as given in the code, however I am not sure if its the best way to solve this logic error.
ezgif com-crop(17)

### DIFF
--- a/app/views/RoomView/index.js
+++ b/app/views/RoomView/index.js
@@ -412,9 +412,9 @@ class RoomView extends React.Component {
 		this.setState({ selectedMessage: message, showErrorActions: true });
 	}
 
-	onActionsHide = () => {
+	onActionsHide = (hideAction = false) => {
 		const { editing, replying, reacting } = this.state;
-		if (editing || replying || reacting) {
+		if ((editing || replying || reacting) && !hideAction) {
 			return;
 		}
 		this.setState({ selectedMessage: {}, showActions: false });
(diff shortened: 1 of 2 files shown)


app/views/RoomView/index.js
```diff
@@ -412,9 +412,9 @@ class RoomView extends React.Component {
 		this.setState({ selectedMessage: message, showErrorActions: true });
 	}
 
-	onActionsHide = () => {
+	onActionsHide = (hideAction = false) => {
 		const { editing, replying, reacting } = this.state;
-		if (editing || replying || reacting) {
+		if ((editing || replying || reacting) && !hideAction) {
```
djorkaeffalexandre: You is never consider `editing || replying || reacting`.
It isn't the correct solution, you need to make the changes requested #1610, and then you can remove this `if` case.

Prateek93a: Sir, I have made the changes in #1610, and I wanted to know that should I close this Pr and push commit to fix this issue in #1610 itself, since the solution relies on the code there?
https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1610#issuecomment-580195161

djorkaeffalexandre:
It was fixed at #2114.
Thanks a lot. 😁

PR #1616, opened 2020-01-23
[IMPROVEMENT] Show toast when a message is starred/unstarred

devyaniChoubey:
@RocketChat/ReactNative

Closes #1615 

Peek 2020-01-23 22-14


djorkaeffalexandre:
Hello @devyaniChoubey.
I think we not need a toast about message pinned, because we have an info message about it.
Captura de Tela 2020-01-27 às 17 52 12
You can focus your PR only in show a Toast when a message was `Starred/Unstarred`.
Thanks.

app/containers/MessageActions.js
```diff
@@ -271,6 +271,7 @@ class MessageActions extends React.Component {
 		const { message } = this.props;
 		try {
 			await RocketChat.toggleStarMessage(message.id, message.starred);
+			EventEmitter.emit(LISTENER, { message: message.starred ? I18n.t('Message_starred') : I18n.t('Message_unstar') });
```
djorkaeffalexandre: Change to `Message_unstarred`.

app/i18n/locales/en.js
```diff
@@ -234,6 +234,8 @@ export default {
 	Message_actions: 'Message actions',
 	Message_pinned: 'Message pinned',
 	Message_removed: 'Message removed',
+	Message_starred: 'Message starred',
+	Message_unstar: 'Message unstarred',
```
djorkaeffalexandre: Same of comment above.

devyaniChoubey: @djorkaeffalexandre sir is it fine now?

PR #735, opened 2019-03-17
Add support for scroll down when reading up old messages

pranavpandey1998official:
fixes:#675

@RocketChat/ReactNative

Closes #675 

commit

PR #1620, opened 2020-01-24
Scroll to bottom button Implemented

### LINKED ISSUE #1619: [FEATURE] Scroll To bottom button
- Your Rocket.Chat Experimental app version: 4.2.4.99999999

Its a feature issue. It would be helpful to have a scroll to bottom button on chat room view and scroll to top button on directory view, thread view, room list view, search views like members list view and other screens to save time when scrolling over large number of messages or search results.

Prateek93a:
@RocketChat/ReactNative

Closes #1619

I have implemented initial version of the button. It works fine as it should. I have used dimmer colors to avoid getting distracted by button while reading message. 
Right now I am using fixed offset to show the button, but if needed, we might use height of first message from bottom as in case of other apps.
If the feature seems useful, maybe we could implement it on ThreadView and SearchViews as well.

ezgif com-crop(18)

Screenshot_20200124-160734_ DEVELOP  RocketChatRN


diegolmello:
@Prateek93a Nice job. Can you resolve conflicts?

Prateek93a:
@diegolmello I have resolved the conflicts

app/views/RoomView/index.js
```diff
@@ -783,6 +817,16 @@ class RoomView extends React.Component {
 		return message;
 	}
 
+	renderScrollButton = () => {
+		const { scrollButtonTranslateY, scrollToBottom } = this;
+		const { theme } = this.props;
+		return (
+			<AnimatedTouch theme={theme} onPress={scrollToBottom} style={[styles.scrollButton, { backgroundColor: themes[theme].bannerBackground, transform: [{ translateY: scrollButtonTranslateY }] }]}>
```
diegolmello: Add `borderColor: themes[theme].borderColor`

app/views/RoomView/index.js
```diff
@@ -783,6 +817,16 @@ class RoomView extends React.Component {
...
+				<CustomIcon name='arrow-down' size={30} style={{ color: themes[theme].auxiliaryTintColor }} />
```
diegolmello: Can you use `tintColor` instead?

app/views/RoomView/styles.js
```diff
@@ -46,5 +46,16 @@ export default StyleSheet.create({
...
+	scrollButton: {
+		position: 'absolute',
+		bottom: 70,
+		right: 20,
+		height: 50,
+		width: 50,
+		borderRadius: 25,
+		alignItems: 'center',
+		justifyContent: 'center',
+		zIndex: 0
```
diegolmello: Add `borderWidth: 1`

app/views/RoomView/styles.js
```diff
@@ -46,5 +46,16 @@ export default StyleSheet.create({
 	previewMode: {
 		fontSize: 16,
 		...sharedStyles.textMedium
+	},
+	scrollButton: {
+		position: 'absolute',
+		bottom: 70,
```
diegolmello: This position should be based on device.
Take a look how it is hidden behind the keyboard on iPhone 11 Pro:

app/views/RoomView/index.js
```diff
@@ -783,6 +817,16 @@ class RoomView extends React.Component {
 		return message;
 	}
 
+	renderScrollButton = () => {
```
diegolmello: After the changes, it's going to look like this:

app/views/RoomView/List.js
```diff
@@ -251,6 +252,9 @@ class List extends React.Component {
 					onEndReachedThreshold={5}
 					maxToRenderPerBatch={5}
 					windowSize={10}
+					scrollEventThrottle={20}
```
diegolmello: I'm unsure about performance here.
`scrollEventThrottle={20}` means it's going to execute `toggleScrollToBottomButton` every 20ms.
We already have some performance issues on low-end devices.
This implementation is much better than https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/735, but we still have to think about it.

Prateek93a:
Ok Sir, on it.

Prateek93a:
I added listContainer because while typing, as the textbox height grew, the scroll button would get hidden behind the texbox since the `bottom` was set with respect to the bottom of the screen and not the textbox. So I added these changes.
Also I reduced the height of the textbox because since I had used listContainer, so now when textbox was at max height, the scrollbutton would get hidden behind header.
before-

after-

app/views/RoomView/index.js
```diff
@@ -912,8 +956,11 @@ class RoomView extends React.Component {
 					room={room}
 					renderRow={this.renderItem}
 					loading={loading}
+					animated={this.beginAnimating}
```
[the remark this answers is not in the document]
Prateek93a: I have moved the component and logic inside list component

app/views/RoomView/List.js
Prateek93a: @diegolmello  I have changed the number to 8

PR #1623, opened 2020-01-26
[FIX] Not allow to send messages to archived room

### LINKED ISSUE #1577: When we archive a room we should not able to send messages in it.
Your Rocket.Chat Experimental app version: 4.2.1.999999999
Your Rocket.Chat server version: 2.5.0-develop open.rocket.chat
Device (or Simulator) you're running with: Android 9

Peek 2020-01-14 17-34

devyaniChoubey:
Previously when we were archiving a room we could still send messages in it.
It was because there was no condition in RoomView.js to check if the room has been archived and hide the messagebox.In this pr I have that if check

@RocketChat/ReactNative

Closes #1577

Peek 2020-01-26 12-15


app/views/RoomView/index.js
```diff
@@ -798,6 +798,13 @@ class RoomView extends React.Component {
 				</View>
 			);
 		}
+		if (room.archived) {
```
djorkaeffalexandre: Instead of create a new case, you can add `room.archived` [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/cd7ab896bbcad9b599e65789e670f4293013c4d8/app/views/RoomView/index.js#L794) and show the same message of web client.
Thanks.

app/i18n/locales/en.js
```diff
@@ -390,6 +390,7 @@ export default {
 	There_was_an_error_while_action: 'There was an error while {{action}}!',
 	This_room_is_blocked: 'This room is blocked',
 	This_room_is_read_only: 'This room is read only',
+	This_room_has_been_archived: 'This room has been archived',
```
djorkaeffalexandre: Remove.

PR #1627, opened 2020-01-27
[NEW] Request review

djorkaeffalexandre:
@RocketChat/ReactNative

We now want to request reviews to users when they has a successfully use of the app.
This PR add one modal that will show when the user do 5 actions with successfully.
* When user click on `Cancel` it never will be shown again.
* When user click on `Ask me later` it will be shown again after 7 days. 
* When user click on `Sure!` it never will be shown again.

Actions that handle a Positive Event:
* Send a message
* Add a Reaction
* Send a File Message
* Create a channel


app/utils/review.js
```diff
@@ -0,0 +1,76 @@
+import { Alert, Linking } from 'react-native';
+import RNUserDefaults from 'rn-user-defaults';
```
diegolmello: We shouldn't be using RNUserDefaults here as it is shared between both apps.
This way the user would maybe review experimental instead official.
If we user AsyncStorage here, we're going to ask them to review twice (official and experimental) and can be annoying, but it's probably the best solution.

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+
+const store = isIOS ? 'App' : 'Play';
+
+const reviewKey = 'reviewKey';
+const popupDelay = 2000;
+const numberOfDays = 7;
+const numberOfPositiveEvent = 5;
+
+// handle official and experimental app
+const id = getBundleId.includes('reactnative') ? '1272915472' : '1148741252';
```
diegolmello: Can we add these links building to https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/constants/links.js?
It's easier for us to maintain and for forks.

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+	// if review wasn't done yet or wasn't cancelled
+	if (daysBetween(lastReviewDate, new Date()) >= numberOfDays && !doneReview) {
+		setTimeout(handlePositiveEvent, popupDelay);
+	}
+};
+
+class ReviewTheApp {
+	positiveEventCount = 0;
+
+	pushEvent = () => {
```
diegolmello: Can we name it `pushPositiveEvent`?
`pushEvent` feels like any event and we're pushing only positives.

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+			Linking.openURL(link);
+		}
+	} catch (e) {
+		showErrorAlert(I18n.t('Unable_to_open_store', { store }));
+	}
+};
+
+const onAskMeLaterPress = () => RNUserDefaults.setObjectForKey(reviewKey, { lastReview: new Date().getTime() });
+
+const handlePositiveEvent = () => Alert.alert(
```
diegolmello: Can we name it `askReview`?
We're not handling events here.
We're opening an alert asking for the user to review.

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+	const lastReviewDate = new Date(lastReview);
+
+	// if ask me later was pressed, we only can show again after {{numberOfDays}} days
+	// if review wasn't done yet or wasn't cancelled
+	if (daysBetween(lastReviewDate, new Date()) >= numberOfDays && !doneReview) {
+		setTimeout(handlePositiveEvent, popupDelay);
+	}
+};
+
+class ReviewTheApp {
```
diegolmello: Can we change this name to `ReviewApp`?

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+		{ text: I18n.t('Sure'), onPress: onReviewPress }
+	]
+);
+
+const tryReview = async() => {
+	const reviewData = await RNUserDefaults.objectForKey(reviewKey) || {};
+	const { lastReview = 0, doneReview = false } = reviewData;
+	const lastReviewDate = new Date(lastReview);
+
+	// if ask me later was pressed, we only can show again after {{numberOfDays}} days
```
diegolmello:
```suggestion
	// if ask me later was pressed earlier, we can ask for review only after {{numberOfDays}} days
```

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+	// if review wasn't done yet or wasn't cancelled
```
diegolmello:
```suggestion
	// if there's no review and it wasn't dismissed by the user
```

app/utils/review.js
```diff
@@ -0,0 +1,76 @@
...
+	if (daysBetween(lastReviewDate, new Date()) >= numberOfDays && !doneReview) {
+		setTimeout(handlePositiveEvent, popupDelay);
+	}
+};
+
+class ReviewTheApp {
+	positiveEventCount = 0;
+
+	pushEvent = () => {
+		this.positiveEventCount += 1;
```
diegolmello: I think we should test if `this.positiveEventCount > numberOfPositiveEvent` and `return`.
It would prevent us from having a huge value inside `this.positiveEventCount` and having to test it against `numberOfPositiveEvent` on every positive event.

app/utils/review.js
```diff
@@ -51,26 +61,30 @@ const handlePositiveEvent = () => Alert.alert(
...
 	}
 };
 
-class ReviewTheApp {
+class ReviewApp {
 	positiveEventCount = 0;
 
-	pushEvent = () => {
+	pushPositiveEvent = () => {
+		if (this.positiveEventCount >= numberOfPositiveEvent) {
```
diegolmello:
```suggestion
		if (this.positiveEventCount > numberOfPositiveEvent) {
```

`>=` would never call `tryReview`

PR #1673, opened 2020-02-06
[BUG-FIX] Mentions Bug fixed

### LINKED ISSUE #1580: [BUG] Mentions autocomplete does not work on the mobile app.
Mention autocomplete which allows the user to see a list of users in the current channel after typing @... does not pop up the users list on mobile devices, it works fine for the same user on desktop app / native web browser.

It does work on mobile only if the user has the view-d-room & view-outside-room permission but then he can see all the users on the server not only the users in the relevant channel.

How to reproduce, verify your user does not have the view-d-room & view-outside-room permission, join a channel and try to mention other users by typing @ + first letter
On desktop app/web it works as expected

Screen Shot 2020-01-14 at 10 17 02

Try the same on mobile, does not work
Screen Shot 2020-01-14 at 10 18 50

sarthakpranesh:
1. Updated the regex to look for other mention categories too.
2. gave new meaningful names to the variables
3. Updated the params passed to identifyMentionKeyword()

@RocketChat/ReactNative
Closes #1580 

What was the bug?
In the web site, whenever we used "@" a list of possible mentions would appear whereas this wasn't the case in the mobile app. We had to enter one extra character after "@", "#" and ":" to give the possible mentions.

Is the bug fixed?
YES, now all the symbols "@", "#", ":" and "/" have the expected behavior. Showcased in the below screenshots.

### DIFF
--- a/app/containers/MessageBox/index.js
+++ b/app/containers/MessageBox/index.js
@@ -292,17 +292,16 @@ class MessageBox extends Component {
 				const { start, end } = this.component._lastNativeSelection;
 				const cursor = Math.max(start, end);
 				const lastNativeText = this.component._lastNativeText || '';
-				// matches if text either starts with '/' or have (@,#,:) then it groups whatever comes next of mention type
 				const regexp = /(#|@|:|^\/)([a-z0-9._-]+)$/im;
-				const result = lastNativeText.substr(0, cursor).match(regexp);
-				if (!result) {
-					const slash = lastNativeText.match(/^\/$/); // matches only '/' in input
-					if (slash) {
-						return this.identifyMentionKeyword('', MENTIONS_TRACKING_TYPE_COMMANDS);
+				const mentionWithSearch = lastNativeText.substr(0, cursor).match(regexp);
+				if (!mentionWithSearch) {
+					const mentionWithOutSearch = lastNativeText.match(/(#|@|:|^\/)$/);
+					if (mentionWithOutSearch) {
+						return this.identifyMentionKeyword('a', mentionWithOutSearch[0]);
 					}
 					return this.stopTrackingMention();
 				}
-				const [, lastChar, name] = result;
+				const [, lastChar, name] = mentionWithSearch;
 				this.identifyMentionKeyword(name, lastChar);
 			} catch (e) {
 				log(e);


app/containers/MessageBox/index.js
```diff
@@ -292,17 +292,16 @@ class MessageBox extends Component {
...
-				const result = lastNativeText.substr(0, cursor).match(regexp);
-				if (!result) {
-					const slash = lastNativeText.match(/^\/$/); // matches only '/' in input
-					if (slash) {
-						return this.identifyMentionKeyword('', MENTIONS_TRACKING_TYPE_COMMANDS);
+				const mentionWithSearch = lastNativeText.substr(0, cursor).match(regexp);
+				if (!mentionWithSearch) {
+					const mentionWithOutSearch = lastNativeText.match(/(#|@|:|^\/)$/);
+					if (mentionWithOutSearch) {
+						return this.identifyMentionKeyword('a', mentionWithOutSearch[0]);
```
djorkaeffalexandre: Hello @sarthakpranesh, very thanks about your contribution.
For first, you solution not is about the issue #1580, but is another problem that we needs to resolve! :)
But, this `'a'` char isn't the better solution to this problem, how you can see [here](https://github.com/sarthakpranesh/Rocket.Chat.ReactNative/blob/Sarthak-MentionFix/app/lib/rocketchat.js#L543), we have a control to turn off promises when you're searching for something, you need to change this logic to consider a blank string in the search, and only add others mentions track on `lastNativeText.match`, it'll fix the problem.

sarthakpranesh: Cool, I'll make the required changes asap :+1:

sarthakpranesh:
Updated app/lib/rocketchat.js search function
1. Removed unnecessary if condition which was not allowing the search for the blank string for any mention without any search text
2. Removed 'a' char from this.identifyMentionKeyword function and added blank string

Let me know if there are any further changes that are required to be made for this problem. I would definitely open another PR soon for issue #1580 

Warm Regards

app/lib/rocketchat.js
```diff
@@ -540,11 +540,6 @@ const RocketChat = {
 			this.oldPromise('cancel');
 		}
 
-		if (searchText === '') {
```
djorkaeffalexandre: You can't remove this without add another logic to `delete` the `oldPromise` when search was finished.

sarthakpranesh: Ok, I'll look into it and will see what is required

app/containers/MessageBox/index.js
```diff
@@ -295,17 +295,16 @@ class MessageBox extends Component {
...
-				// matches if text either starts with '/' or have (@,#,:) then it groups whatever comes next of mention type
 				const regexp = /(#|@|:|^\/)([a-z0-9._-]+)$/im;
-				const result = lastNativeText.substr(0, cursor).match(regexp);
-				if (!result) {
-					const slash = lastNativeText.match(/^\/$/); // matches only '/' in input
-					if (slash) {
-						return this.identifyMentionKeyword('', MENTIONS_TRACKING_TYPE_COMMANDS);
+				const mentionWithSearch = lastNativeText.substr(0, cursor).match(regexp);
+				if (!mentionWithSearch) {
+					const mentionWithOutSearch = lastNativeText.match(/(#|@|:|^\/)$/);
```
diegolmello: Can you explain what's `mentionWithSearch` and `mentionWithOutSearch`?

diegolmello: Emojis didn't fire with `:`.

sarthakpranesh: mentionWithSearch is basically a variable to check if the characters #, @, : and / are followed by an alphabet or not. If they are followed by an alphabet then we need to separate those alphabets and use it in the search. Whereas the mentionWithOutSearch is to verify if the user just started with @, #, : and / but has entered any alphabet after that, in this case, we just want to make a search with an empty string.

sarthakpranesh: I'll look into this asap

sarthakpranesh:
FIX
All mentions ( `@`, `#`, `:`, `/` ) show the expected behavior similar to that of the web client.

Please let me know of any further changes.
mentions

ankar84:
@sarthakpranesh @diegolmello @djorkaeffalexandre 
Great changes!
Wait them to be merged a lot!

PR #1675, opened 2020-02-06
[CHORE] Memory leaks investigation

diegolmello:
@RocketChat/ReactNative

- [x] Update WatermelonDB@0.16.0-9
- [x] Update all pods
- [x] Instatiate markdown Parser once
- [x] Fix room in/out navigation animation


app/views/RoomView/index.js
```diff
@@ -183,7 +183,6 @@ class RoomView extends React.Component {
 		}
 
 		this.beginAnimating = false;
```
djorkaeffalexandre: Can be removed.

app/views/RoomView/index.js
```diff
@@ -917,6 +914,7 @@ class RoomView extends React.Component {
 					renderRow={this.renderItem}
 					loading={loading}
 					animated={this.beginAnimating}
```
djorkaeffalexandre: Can be removed.

PR #1685, opened 2020-02-08
[FIX] No Invalid Char is allowed in request

### LINKED ISSUE #1682: username change: no invalid chars should be allowed as input, improve error message
- Your Rocket.Chat Experimental app version: 4.3.1.13914
- Your Rocket.Chat server version: 2.3.2
- Device (or Simulator) you're running with: 13.3

When changing the profile settings, it is possible to enter invalid chars in the username field. 

The app complains with something like: "error occured on saving profile"

Two things should be changed here:

- entering invalid chars in the username field should be suppressed on app side
- reasonable error message should be given to the user, at least that there is an issue in the username field

On server side it API complains with:

```Feb 08 09:50:52 XXX rocketchat[14023]: server.js:204 API ➔ debug POST: /api/v1/users.updateOwnBasicInfo
Feb 08 09:50:52 XXX rocketchat[14023]: server.js:204 Meteor ➔ method saveUserProfile -> userId: pT5QeM5Cc8nzWeXew, arguments: [{},{"XXXX":"XXXX"}]
Feb 08 09:50:52 XXX rocketchat[14023]: server.js:204 Meteor ➔ method setUsername -> userId: pT5QeM5Cc8nzWeXew, arguments: ["cygnus b"]
Feb 08 09:50:52 XXX rocketchat[14023]: server.js:204 API ➔ debug post users.updateOwnBasicInfo threw an error: Error: cygnus b is not a valid username, use only letters, numbers, dots, hyphens and underscores [username-invalid]```

### PR DESCRIPTION
**Pull request by sarthakpranesh**
1. updated submit() to check for any invalid character in the username
2. if found, the request to update profile is not made and an alert is displayed

***Please verify, but the Server Version: 3.0.0-develop does not allow the username to have dot.

@RocketChat/ReactNative
Closes #1682

### CODE REVIEW

File: app/views/ProfileView/index.js [hunk 1]
```diff
@@ -180,7 +185,26 @@ class ProfileView extends React.Component {
 
 		// Username
 		if (user.username !== username) {
-			params.username = username;
+			if (!username.match(/[$&+,:;=?@#|'.<>^* ()%!]/g)) {
+				params.username = username;
```
djorkaeffalexandre: You can't use this regex here, because on our servers we have a configuration about regex usernames and channel names, you need to get this setting from server and try to match.
To find this setting on your own server go to `Administration -> General -> UTF8`.
Screen Shot 2020-02-08 at 14 48 30

sarthakpranesh: Ok cool, thanks for letting me know about this. I will make the required changes.
:+1:

sarthakpranesh:
1. Used server settings for username validation
2. Added necessary text to i18n and updated ProfileView alert to use it

app/i18n/locales/en.js
```diff
@@ -496,5 +496,8 @@ export default {
 	New_line: 'New line',
 	You_will_be_logged_out_of_this_application: 'You will be logged out of this application.',
 	Clear: 'Clear',
-	This_will_clear_all_your_offline_data: 'This will clear all your offline data.'
+	This_will_clear_all_your_offline_data: 'This will clear all your offline data.',
+	Incorrect_username: 'Incorrect Username',
+	Username_can_only_have_letters_numbers_dots_hyphens_and_underscores: 'Username can only have letters, numbers, dots, hyphens and underscores',
```
djorkaeffalexandre: We can't always confirm that, I think you can use just: `Username has chars not allowed by server.`

app/i18n/locales/en.js
```diff
@@ -496,5 +496,8 @@ export default {
...
+	Okay: 'Okay'
```
djorkaeffalexandre: If you follow the bellow comment you can remove this.

app/views/ProfileView/index.js
```diff
@@ -1,6 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { View, ScrollView, Keyboard } from 'react-native';
+import {
+	View,
+	ScrollView,
+	Keyboard,
+	Alert
```
djorkaeffalexandre: Remember to remove these changes when you switch from `Alert` to `showErrorAlert`.

app/views/ProfileView/index.js
```diff
@@ -181,7 +187,27 @@ class ProfileView extends React.Component {
 
 		// Username
 		if (user.username !== username) {
-			params.username = username;
+			const regEx = new RegExp(`^${ UTF8_Names_Validation }$`);
+			if (regEx.test(username)) {
```
djorkaeffalexandre: instead of make this logic here, I think it's better if you create a function like: `validUsername`.
```javascript
validUsername = () => {
  const { username } = this.state;
  const { UTF8_Names_Validation } = this.props;
  const usernameRegex = new RegExp(`^${ UTF8_Names_Validation }$`);
  if (usernameRegex.test(username)) {
    return true;
  }
  showErrorAlert(yourMessage, yourTitle);
  return false;
}
```
so, you can use only: 
```javascript
if (this.validUsername()) {
  params.username = username;
} else {
  this.setState({ saving: false });
  return;
}
```

app/views/ProfileView/index.js
```diff
@@ -181,7 +187,27 @@ class ProfileView extends React.Component {
...
+				params.username = username;
+			} else {
+				this.setState({
+					saving: false,
+					currentPassword: null
+				});
+				Alert.alert(
```
djorkaeffalexandre: Use `showErrorAlert` instead `Alert` to show error alerts.

app/i18n/locales/en.js
sarthakpranesh: Ok, yeah sure

app/views/ProfileView/index.js
sarthakpranesh: Yeah I agree on that, it'll be cleaner than

app/views/ProfileView/index.js
sarthakpranesh: Ok sure

sarthakpranesh:
Updates
1. Removed Alert from import and reverted changed
2. Updated the message text `Username has chars not allowed by server`
3. Cleaned the code and made another function for username validation

sarthakpranesh:
FIXED
All three RegistrationView, SetUsernameView and ProfileView check for the entered `username` to make sure it matches the settings on the server.

Please let me know of any further changes :smile: 
RegistrationView
profileView

PR #1709, opened 2020-02-12
[FIX] Small Fixes on UiKit

djorkaeffalexandre:
@RocketChat/ReactNative

- [X] Themed Overflow
- [X] KeyboardAwareScrollView ModalBlocksView
- [X] UIKit errors
- [X] Markdown Table
- [X] Wrong name on notification from Apps/Discussions
- [X] MarginBottom on MultiSelect

### DIFF
--- a/app/notifications/inApp/index.js
+++ b/app/notifications/inApp/index.js
@@ -165,7 +165,7 @@ class NotificationBadge extends React.Component {
 			return;
 		}
 		const name = type === 'd' ? payload.sender.username : payload.name;
-		// if not has sub on database, title will be null, so we use payload from notification
+		// if sub is not on local database, title will be null, so we use payload from notification
 		const { title = name } = notification;
 		await navigation.navigate('RoomsListView');
 		navigation.navigate('RoomView', {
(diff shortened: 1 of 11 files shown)


app/notifications/inApp/index.js
```diff
@@ -158,15 +158,18 @@ class NotificationBadge extends React.Component {
...
 	goToRoom = async() => {
-		const { notification: { payload }, navigation, baseUrl } = this.props;
+		const { notification, navigation, baseUrl } = this.props;
+		const { payload } = notification;
 		const { rid, type, prid } = payload;
 		if (!rid) {
 			return;
 		}
 		const name = type === 'd' ? payload.sender.username : payload.name;
+		// if not has sub on database, title will be null, so we use payload from notification
```
diegolmello:
```suggestion
		// if sub is not on local database, title will be null, so we use payload from notification
```

app/notifications/inApp/index.js
```diff
@@ -178,6 +181,8 @@ class NotificationBadge extends React.Component {
 		const { message, payload } = notification;
 		const { type } = payload;
 		const name = type === 'd' ? payload.sender.username : payload.name;
+		// if not has sub on database, title and avatar will be null, so we use payload from notification
```
diegolmello:
```suggestion
		// if sub is not on local database, title and avatar will be null, so we use payload from notification
```

PR #1725, opened 2020-02-14
[IMPROVEMENT] Alert verify email when enabled

### LINKED ISSUE #1548: No message is displayed to verify the email-id

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #1548.


app/i18n/locales/en.js
```diff
@@ -450,6 +450,8 @@ export default {
 	Username: 'Username',
 	Username_or_email: 'Username or email',
 	Validating: 'Validating',
+	Verify_email_title: 'Registration Succeded!',
```
diegolmello:
```suggestion
	Verify_email_title: 'Registration Succeeded!',
```

PR #1726, opened 2020-02-14
[IMPROVEMENT] Room announcements

### LINKED ISSUE #1718: [IMPROVEMENT] Display channel announcement banner prominently inside channel
It would be great if a channel has an announcement set, it gets displayed as prominently as in the web version of Rocket.Chat on top of the channel.

#### Web

Bildschirmfoto 2020-02-13 um 18 57 05

#### ReactNative

Ohne Titel

Prateek93a:
@RocketChat/ReactNative

Closes #1718

As the following Gif shows, room announcements are now displayed. I could not come up with color combination, so colors can be changed if needed. Also I noticed that the announcement did not get updated in the `RoomView`  when I changed it. I found that when `roomUpdate` would get updated while `RoomView` is not focused, and when it came to focus, since `roomUpdate` was already chaged so the view did not rerender as `nextState.roomUpdate` was same as `roomUpdate` so I added a `hasRoomChanged` variable.


Prateek93a:
Before-



Now-

frdmn:
Well done @Prateek93a 👍

Prateek93a:
@frdmn @diegolmello Thanks a lot for the review

PR #1729, opened 2020-02-14
[FIX]Add empty space before custom emoji if needed

### LINKED ISSUE #1717: Add empty space when custom emoji is added
- Your Rocket.Chat Experimental app version: 4.4.0.9999999999
- Your Rocket.Chat server version: 3.0.0-rc.8
As the following gif shows, custom emojis are not displayed when there is no space between text and emoji
ezgif com-crop(1)

aKn1ghtOut:
@RocketChat/ReactNative

Closes #1717 

If there is no space before custom emojis, then that does not get rendered. Therefore, changed the default keyboard action for adding custom emojis to add a space right before, if it wasn't there already. This ensures the custom emoji, when added through the keyboard, is always rendered by default.

Before FIX:

After FIX:

### DIFF
--- a/app/containers/MessageBox/index.js
+++ b/app/containers/MessageBox/index.js
@@ -367,17 +367,17 @@ class MessageBox extends Component {
 		const { text } = this;
 		const { emoji } = params;
 		let newText = '';
-		// Add space if not before emoji if not already there
+		// Add space before custom emoji if not already there
 		let addSpace = '';
 
 		// if messagebox has an active cursor
 		if (this.component && this.component._lastNativeSelection) {
 			const { start, end } = this.component._lastNativeSelection;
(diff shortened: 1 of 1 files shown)


aKn1ghtOut:
@diegolmello What do you think?

diegolmello:
@aKn1ghtOut PR description is broken. Did you intent to upload some images?

aKn1ghtOut:
Yes. I don't know how I missed that.
Attached below:

BEFORE:
smiley-space2

AFTER:
smiley-space1

I now think though that this might not have been the best approach since the web version still renders non-space-separated custom emoticons/emojis perfectly while it's only the mobile client that has this issue. 
Looking into this further

PR #1734, opened 2020-02-15
[NEW] Open RoomInfo Avatar

### LINKED ISSUE #1731: [IMPROVEMENT] Open AttachmentView on Avatar of RoomInfo
Use our `AttachmentView` to open the avatar of `RoomInfo`.

sarthakpranesh:
@RocketChat/ReactNative
Closes #1731 

If the user clicks on the Avatar in RoomInfo, AttachmentView is used to open the Avatar on the screen

Used `BorderlessButton` for handling onPress and `AttachmentView` for show casing the Avatar.


sarthakpranesh:
1. Added new prop to Avatar called canView.
2. When it's true the user can click on the Avatar in RoomInfo view, to see it full screen

app/containers/Avatar.js
```diff
@@ -40,6 +41,9 @@ const Avatar = React.memo(({
 		uri = formatUrl(`/avatar/${ room }`, baseUrl, uriSize, avatarAuthURLFragment);
 	}
 
+	function openView() {
+		Navigation.navigate('AttachmentView', { attachment: { image_url: uri, image_type: 'png' } });
```
diegolmello: We're sending `width=100&height=100` here.
When we want to see an image on a modal, we want it as bigger as possible.

sarthakpranesh: So the `AttachmentView` opens the image using the full width of the device but due to avatar being square does not cover the whole height.
I'll still change the size to make sure the image rendered is not pixelated.

sarthakpranesh:
FIXED
Clicking on Avatar in RoomInfo View opens it using the AttachmentView, with no fixed size defined in the URL to get the maximum size of the avatar from the server.

RoomInfo

yash-rajpal:
@diegolmello Is this PR active. If you permit I will make a new PR with any new changes and resolving conflicts.

PR #1754, opened 2020-02-19
[FIX] Watermelon throwing "Cannot update a record with pending updates"

diegolmello:
@RocketChat/ReactNative

This PR fixes an issue that would make Watermelon to throw "Cannot update a record with pending updates".
That happens when a record is set to update (`prepareUpdate`), but not yet added to `database.batch`, as mentioned here: https://github.com/Nozbe/WatermelonDB/issues/612#issuecomment-577723861

On this app, a Message may belong to a Thread.
If a thread receives several messages in a row, `handleMessageReceived` will be fired each time.
In this case, a message makes it to query for the Thread and try to update it, but it may be triggered by a previous received message already.

This PR creates a window of CRUD to the batch and tries to update one thread only once.

Let's test it on beta for now and see how it behaves.


andrewlorenz:
I'm not sure about this change. It seems to have made the UI updates slow/jittery/out of sequence - in our testing messages between just 2 users on an otherwise empty service are all over the place.  I'm referring to the fix for Watermelon here, not the other bits n bobs within this PR.
Has it gone through any kind of testing?

diegolmello:
@andrewlorenz We're testing it for 20 days already and it seems ok to me.
It was merged because `develop` branch is for our Experimental app.
It wasn't released yet on the official app. It's on beta.
Can you open an issue and add more details (gif/video)?
Thanks.

PR #1758, opened 2020-02-19
[FIX] Use RealName when necessary

djorkaeffalexandre:
@RocketChat/ReactNative

There are some places where we need to use Real Name when it's enabled.
- [X] Mentions
- [X] Sidebar
- [X] Last Message
- [X] Search on RoomsList
- [X] Opened notification


app/presentation/RoomItem/index.js
```diff
@@ -110,7 +110,7 @@ const RoomItem = React.memo(({
 							ellipsizeMode='tail'
 							numberOfLines={1}
 						>
-							{name}
+							{getName({ name, username })}
```
diegolmello: You're only considering DMs.
If `Real Name` is set, all channel/groups/livechat will be `username` 🤷‍♂

PR #1786, opened 2020-02-27
[IMPROVEMENT] Log server version

djorkaeffalexandre:
@RocketChat/ReactNative

In order to have more information about crash reports, let's log server version.


app/sagas/selectServer.js
```diff
@@ -112,6 +112,9 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
 			serverInfo = yield getServerInfo({ server, raiseError: false });
 		}
 
+		// we'll set serverVersion as metadata for bugsnag
+		bugsnagServerVersion((serverInfo && serverInfo.version) || version);
```
diegolmello: There's a duplicated use of `(serverInfo && serverInfo.version) || version` here.
Can we create a const?

app/sagas/selectServer.js
```diff
@@ -112,6 +112,9 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
...
+
 		// Return server version even when offline
 		yield put(selectServerSuccess(server, (serverInfo && serverInfo.version) || version));
```
djorkaeffalexandre:
```suggestion
		// Return server version even when offline
		const serverVersion = (serverInfo && serverInfo.version) || version;

		bugsnagServerVersion(serverVersion);
		yield put(selectServerSuccess(server, serverVersion));
```

PR #1805, opened 2020-03-01
Sarthak audio one

### LINKED ISSUE #1654: [Bug] Play multiple voice message at a time
There is a bug that it will play multiple voice messages at a time. It should only play one at a time. Also it would be neat if it will automatically play the next voice message in the view.

sarthakpranesh:
@RocketChat/ReactNative

Closes #1654 

As mentioned in the Issue ( #1654 ), it was not correct for multiple audio messages to play at the same time. This PR blocks from multiple audio messages to be played together or at the same time.

### DIFF
--- a/app/views/RoomView/index.js
+++ b/app/views/RoomView/index.js
@@ -179,7 +179,8 @@ class RoomView extends React.Component {
 			replyWithMention: false,
 			reacting: false,
 			showAnnouncementModal: false,
-			announcement: null
+			announcement: null,
+			audioPlaying: false
 		};
 
 		if (room && room.observe) {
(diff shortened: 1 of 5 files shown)


app/views/RoomView/index.js
```diff
@@ -724,6 +725,20 @@ class RoomView extends React.Component {
 		}
 	});
 
+	getLock = (wasPlaying) => {
```
diegolmello: Let's say we want to allow to watch a video without having to fullscreen.
IMO we could have functions `playMedia`, `pauseMedia` and `stopMedia`.
All of them would control media being played on `RoomView`, being audio or video.

sarthakpranesh: Here my only doubt is that both the pauseMedia and stopMedia will at the end perform the same task i.e should I create just one or have both

diegolmello: You're right. Play and stop then.

sarthakpranesh:
Updates
1. playMedia - lets the component ( such as Audio or Video component ) play the audio/video
2. pauseMedia - lets the component pause the audio/video

Please let me know of any further changes necessary

sarthakpranesh:
Fixed
Whenever we play audio when another audio is already playing, the app now stops the previously playing audio and then plays the new one.

I figured out that it's better to use Events for this purpose rather than passing a series of props essentially doing the same. 
The same can be applied to the video component in the future.

roomVIew

PR #1813, opened 2020-03-02
[IMPROVEMENT] Request user presence on demand

djorkaeffalexandre:
@RocketChat/ReactNative

- [X] Batch usersPresence when loginSuccess


app/lib/methods/getUsersPresence.js
```diff
@@ -0,0 +1,60 @@
...
+		this.activeUsersSubTimeout = setTimeout(() => {
+			this.sdk.subscribe('activeUsers');
+		}, 5000);
+	} else {
+		let params = {};
+
+		// is server is greather than or equal 3.0.0
+		if (serverVersion && !semver.lt(semver.coerce(serverVersion), '3.0.0')) {
+			// Request userPresence on demand
+			params = { ids: ids.length ? ids.join(',') : [] };
```
djorkaeffalexandre: instead this, try to not call `users.presence` if `length === 0`.

app/lib/methods/getUsersPresence.js
```diff
@@ -0,0 +1,65 @@
...
+			this.sdk.subscribe('activeUsers');
+		}, 5000);
+	} else {
+		let params = {};
+
+		// if server is greather than or equal 3.0.0
+		if (serverVersion && !semver.lt(semver.coerce(serverVersion), '3.0.0')) {
+			// if not have any id, only subscribe to changes
+			if (!ids.length) {
+				this.sdk.subscribe('stream-notify-logged', 'user-status');
```
djorkaeffalexandre: We're subscribing a lot of times, always that have more one batch of users. It's wrong.

PR #1819, opened 2020-03-03
[FIX] Pop from room which you have been removed

djorkaeffalexandre:
@RocketChat/ReactNative

- [X] Pop to RoomsList when you are in a room from which you have been removed.
- [X] Pop to RoomsList when you are on some thread in a room from which you have been removed.


app/views/RoomView/index.js
```diff
@@ -551,6 +553,13 @@ class RoomView extends React.Component {
 		EventEmitter.removeListener('connected', this.handleConnected);
 	}
 
+	handleRemoved = ({ rid }) => {
+		const { navigation } = this.props;
+		if (rid === this.rid) {
+			navigation.pop();
```
diegolmello: We should tell the user what just happened.
Otherwise, he'll have no idea.

```js
You were removed from ${CHANNEL_NAME}
```

PR #1823, opened 2020-03-04
[FIX] Invalid URL Port

djorkaeffalexandre:
@RocketChat/ReactNative

Sometimes we're receiving some crashes like `Invalid URL Port: chat.domain.com`, maybe it's caused by some mistake about Basic Auth, to handle this case I added a condition that will check if have some `:` on the host, if yes, this url is invalid.

### DIFF
--- a/app/views/NewServerView.js
+++ b/app/views/NewServerView.js
@@ -201,6 +201,11 @@ class NewServerView extends React.Component {
 			url = parsedUrl.host;
 		}
 
+		// host never can contain a ':'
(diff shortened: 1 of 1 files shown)


app/views/NewServerView.js
```diff
@@ -201,6 +201,11 @@ class NewServerView extends React.Component {
 			url = parsedUrl.host;
 		}
 
+		// host never can contain a ':'
```
diegolmello, line 204: We should return an error, actually.
So the user can know what was wrong.

PR #1824, opened 2020-03-04
[FIX] Unable to logout from oauth

### LINKED ISSUE #1818: Logout from custom oauth does not log user out
- Your Rocket.Chat Experimental app version: 4.4.0
- Your Rocket.Chat server version: 3.0.0
- Device (or Simulator) you're running with: Android (9) and IOS (11.4)

When authenticated through custom oauth, logging out from the server does not completely log user out. 

Steps to reproduce:
- Authenticate through custom oauth
- Logout
- Log back in --> app redirects back in to the previous signed in account without user having to enter a new credentials.

xtilyn:
@RocketChat/ReactNative

Closes #1818

Clears cookies on android webview and disables shared proc pool for ios webview


diegolmello:
Easiest way of doing this is setting [incognito](https://github.com/react-native-community/react-native-webview/blob/master/docs/Reference.md#incognito) on [WebView](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/views/AuthenticationWebView.js#L112).

PR #1833, opened 2020-03-05
[NEW] Join mentioned channel

### LINKED ISSUE #1826: Join to a channel that's mentioned in a message when tapping on it
- Your Rocket.Chat Experimental app version: 4.5.0.15171
- Your Rocket.Chat server version: 3.0.0
- Device (or Simulator) you're running with: Huawei P20 Lite

When someone mentions a channel, tapping on it opens a preview that only shows the channel info and a avatar.

It would be nice to emulate the same behaviour of the web app (show a channel messages preview and a `Join` button) or at least a `Join` button.

That would make users life easier, because we currently need to go to the `Chat` list, open `Directory`, type the channel name, hit `Search`, tap in the channel name and then it behaves like the web app (messages preview and `Join` button).

Cheers!

youssef-md:
@RocketChat/ReactNative

Closes #1826
In this PR was done:
* Enter a private or public channel that the user has already joined.
* Block user from entering a not allowed channel.
* `navigation.push()`  for the user to keep track of where he was.

The 'You are in preview mode' page was already implemented.
ezgif com-gif-maker (1)


app/views/RoomView/index.js
```diff
@@ -681,6 +681,19 @@ class RoomView extends React.Component {
...
+	navToRoomPreview = async({ rid, name }) => {
+		const { user: { id: uid } } = this.props;
+		try {
+			const room = await RocketChat.canAccessRoom(rid, uid);
+			if (room) {
+				await Navigation.navigate('RoomsListView');
+				Navigation.navigate('RoomView', { rid, name, t: room.t });
+			}
+		} catch {
+			EventEmitter.emit(LISTENER, { message: I18n.t('error-not-allowed') });
```
diegolmello: You're always telling the user he's not allowed to access the room.
What if he's without internet connection?

app/lib/rocketchat.js
```diff
@@ -1123,6 +1123,9 @@ const RocketChat = {
 	useInviteToken(token) {
 		// RC 2.4.0
 		return this.sdk.post('useInviteToken', { token });
+	},
+	canAccessRoom(rid, uid) {
```
diegolmello: We have a similar logic [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/methods/canOpenRoom.js).
Is it the same?

youssef-md: With what I've understood, no. The api call in `canAccessRoom()` returns true if the user is already in the room(being private or public) and if the user could enter in the room(being public) and this cover hidden rooms too.

app/views/RoomView/index.js
youssef-md: Could I use something like [this example](https://reactnativeforyou.com/how-to-check-internet-connectivity-in-react-native-android-and-ios/) or there is already some code for this?

mahal:
@youssef-md Would you be motivated to work on this PR again? Please contact me on https://open.rocket.chat/direct/mahal

Issue #1842, opened 2020-03-07
[BUG] Random id is shown in notification preferences of discussions


devyaniChoubey:
![Screenshot_20200307-222104_RocketChat](https://user-images.githubusercontent.com/52153085/76147477-8c2e7c00-60c2-11ea-96f5-0748f6b5e431.jpg)
I am working on it

jeevanpuchakay:
I think this issue has been resolved, I'm unable to reproduce this

TC5022:
@diegolmello is this issue still open? I would like to work on it

diegolmello:
@TC5022 haven't heard of it for sometime. Closing. Thanks!

PR #1843, opened 2020-03-07
[FIX] Notification preferences of discussions 

devyaniChoubey:
After fix discussion name is showing
solve issue #1842 
Screenshot_20200307-224449_ DEVELOP  RocketChatRN

### DIFF
--- a/app/views/NotificationPreferencesView/index.js
+++ b/app/views/NotificationPreferencesView/index.js
@@ -252,7 +252,7 @@ class NotificationPreferencesView extends React.Component {
 						theme={theme}
 					/>
 					<Separator theme={theme} />
-					<Info info={I18n.t('Receive_notifications_from', { name: room.prid ? room.fname : room.name })} theme={theme} />
+					<Info info={I18n.t('Receive_notifications_from', { name: RocketChat.getRoomTitle(room) })} theme={theme} />
 					<SectionSeparator theme={theme} />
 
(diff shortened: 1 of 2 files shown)


app/views/NotificationPreferencesView/index.js
```diff
@@ -252,7 +252,7 @@ class NotificationPreferencesView extends React.Component {
 						theme={theme}
 					/>
 					<Separator theme={theme} />
-					<Info info={I18n.t('Receive_notifications_from', { name: room.name })} theme={theme} />
+					<Info info={I18n.t('Receive_notifications_from', { name: room.prid ? room.fname : room.name })} theme={theme} />
```
djorkaeffalexandre: Instead it, use `RocketChat.getRoomTitle`.

devyaniChoubey: @djorkaeffalexandre done

devyaniChoubey: @djorkaeffalexandre Please review

Prateek93a:
@devyaniChoubey  The same random name shows when we try to leave the discussion, can you please add a commit to solve that also, Thanks

devyaniChoubey:
> @devyaniChoubey The same random name shows when we try to leave the discussion, can you please add a commit to solve that also, Thanks
> 

@Prateek93a ok

PR #1849, opened 2020-03-09
Random discussion room name when opened from push notification

### LINKED ISSUE #1671: Random discussion room name when opened from push notification
- Your Rocket.Chat app version: 4.3.1.2102
- Your Rocket.Chat server version: 3.0.0.rc2
- Device (or Simulator) you're running with: HT70 

Push notification from discussions reflect random room name in the notification, and in the app when you open the room from notification. If you find it in the channel and open from here, the room name is OK (as set in the channel). 

Discussion_room_name

Expected behaviour: reflecting correct discussion room name both in the notification and in the app when you open the room from the Push.

devyaniChoubey:
@RocketChat/ReactNative
I was not able to test this. Because I am not getting push notification on [develop]rocket.chat.ReactNative . But on logging "notification prop" to the console I found that "payload.name" field is having a random string in case of "dissussions" rather than having name of the discussion.And "title" field is having the required name of the discussion.So, I think we should use "title" in place of "payload.name"
Peek 2020-03-09 12-15

Closes #1671

### DIFF
--- a/app/notifications/inApp/index.js
+++ b/app/notifications/inApp/index.js
@@ -159,17 +159,17 @@ class NotificationBadge extends React.Component {
 
 	goToRoom = async() => {
 		const { notification, navigation, baseUrl } = this.props;
-		const { payload } = notification;
+		const { payload, title } = notification;
 		const { rid, type, prid } = payload;
 		if (!rid) {
 			return;
 		}
-		const name = type === 'd' ? payload.sender.username : payload.name;
+		const name = type === 'd' ? payload.sender.username : title;
 		// if sub is not on local database, title will be null, so we use payload from notification
-		const { title = name } = notification;
+		// const { title = name } = notification;
 		await navigation.navigate('RoomsListView');
 		navigation.navigate('RoomView', {
-			rid, name: title, t: type, prid, baseUrl
+			rid, name, t: type, prid, baseUrl
 		});
 		this.hide();
 	}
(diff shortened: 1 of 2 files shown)


app/notifications/inApp/index.js
```diff
@@ -159,17 +159,17 @@ class NotificationBadge extends React.Component {
...
+		const { payload, title } = notification;
 		const { rid, type, prid } = payload;
 		if (!rid) {
 			return;
 		}
-		const name = type === 'd' ? payload.sender.username : payload.name;
+		const name = type === 'd' ? payload.sender.username : title;
 		// if sub is not on local database, title will be null, so we use payload from notification
-		const { title = name } = notification;
+		// const { title = name } = notification;
```
djorkaeffalexandre: This line do the same thing are you doing, you can revert your changes on this file.

app/notifications/push/index.js
```diff
@@ -10,13 +10,13 @@ export const onNotification = (notification) => {
 		if (data) {
 			try {
 				const {
-					rid, name, sender, type, host
+					rid, title, sender, type, host
```
djorkaeffalexandre: Title isn't a param from push notification..
https://github.com/RocketChat/Rocket.Chat/blob/develop/app/lib/server/functions/notifications/mobile.js#L57

djorkaeffalexandre:
Your PR doesn't seems fix anything and the main issue was fixed by #1758.
Closing for now.
Thanks.

PR #1857, opened 2020-03-10
fix  Can't open attachments on Tablet

Badal1234:
Screenshot_1583852819

@RocketChat/ReactNative

Closes #ISSUE_NUMBER

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -493,7 +493,7 @@ class CustomModalStack extends React.Component {
 			</View>
 		);
 
-		if (isAndroid) {
+		if (isAndroid && !isTablet) {
 			content = (
 				<ScrollView overScrollMode='never'>
 					{content}


app/index.js
```diff
@@ -493,7 +493,7 @@ class CustomModalStack extends React.Component {
 			</View>
 		);
 
-		if (isAndroid) {
+		if (isAndroid && !isTablet) {
```
djorkaeffalexandre: Hey @Badal1234! It's not correct 😅.
If we're showing this component we have certain that is a tablet, then it'll always evaluates to `true` and never enter on this condition, we're needing this `ScrollView` because we have some modals that have inputs and the keyboard is opened, it was added by #1742 and we can't do this condition here because it'll broken this fix.
Maybe you can use `pageSheet` value to handle the `ScrollView` and only add the `ScrollView` when `isAndroid && !pageSheet`.

Badal1234: Thanks @djorkaeffalexandre ,for your review and suggestion.

Badal1234:
You are right @djorkaeffalexandre, this change does not affect the previous fix.
Thanks

djorkaeffalexandre:
Hey @Badal1234, I tested it here and works fine with the proposal, but can you try to fix the error about the navigation on tablet?
I think we are navigating to `RoomActionsView` before navigate to `AttachmentView` and it raises some errors when you're opening some image on chat.
It maybe related to `tablet.js` and `app/index.js` and how the stack are configured when the app is splited.
Thanks in advance.

Badal1234:
Sorry, @djorkaeffalexandre I didn't get any such error. When we close the modal it navigates to `RoomActionView`.when we open any image it navigates to `AttachmentView`.Correct me if I am wrong.
Thanx

djorkaeffalexandre:
@Badal1234 check the logs of the app. :) It's trying to access data from room when you open some attachment on RoomView, because are navigating to RoomActionsView before go to AttachmentView. :(

Badal1234:
I have resolved the navigation error.The error is at `RoomActionsStack`.@djorkaeffalexandre review it and corrects any error.
Thank you

Badal1234:
Due to some issue I have closed this PR.
Refer to
[fix] tablet attachment and navigation #1986

PR #1858, opened 2020-03-10
[IMPROVEMENT] Create App Group property on Info.plist

djorkaeffalexandre:
@RocketChat/ReactNative


ios/RocketChatRN/Info.plist
```diff
@@ -110,5 +110,7 @@
 		<string>firefox</string>
 		<string>brave</string>
 	</array>
+	<key>SuiteName</key>
```
diegolmello: `SuiteName` is a bad name.
It means nothing on this context and it's hard to explain in the docs.
I'm thinking on whitelabel apps here.
Can we change it to `App Group` or something like that?

PR #1861, opened 2020-03-11
Show channel's avatar on discussions

### LINKED ISSUE #809: Show channel's avatar on discussions
We should be looking for channel avatar instead of generating a letter for the discussion.
This is better for ux and it'll be helpful when we implement "Change channel's avatar" feature.

Reference: https://github.com/RocketChat/Rocket.Chat/pull/14053

devyaniChoubey:
@RocketChat/ReactNative

Closes #809

### DIFF
--- a/app/lib/rocketchat.js
+++ b/app/lib/rocketchat.js
@@ -1109,7 +1109,7 @@ const RocketChat = {
 		return ((room.prid || useRealName) && room.fname) || room.name;
 	},
 	getRoomAvatar(room) {
-		return room.prid ? room.fname : room.name;
+		return room.prid ? room.topic : room.name;
 	},
 
 	findOrCreateInvite({ rid, days, maxUses }) {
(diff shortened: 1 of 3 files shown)


app/lib/rocketchat.js
```diff
@@ -1109,7 +1109,7 @@ const RocketChat = {
 		return ((room.prid || useRealName) && room.fname) || room.name;
 	},
 	getRoomAvatar(room) {
-		return room.prid ? room.fname : room.name;
+		return room.prid ? room.topic : room.name;
```
diegolmello: Why topic? It work in some cases, but not in all of them.

devyaniChoubey: I will fix it by tomorrow

djorkaeffalexandre: It's wrong, if you open a discussion by some direct message room you'll see that image is incorrect because topic is `null`.

devyaniChoubey: Ok @djorkaeffalexandre I will correct it

djorkaeffalexandre:
It'll be fixed by #2504.
Thanks.

PR #1863, opened 2020-03-11
[FIX] Change database location to Experimental Apps

djorkaeffalexandre:
@RocketChat/ReactNative

We're using an AppGroup on iOS to share credentials between our apps, app groups allow to share things between apps on iOS, it's needed to use at our Share Extension, because on iOS the Share Extension is as other App and can't access the data of main app, so, we use app groups to share credentials and database between the App and Share Extension, this also allow us to share credentials between our iOS Apps (Experimental/Official).
The sharing created a problem for us, because the databases of servers and the main database (where we're saving the credentials) are saved in the same location between the Official and Experimental App.
The main objective of this pr is to separate the databases but continue to share credentials between our apps and make the Share Extension works fine.

I followed these steps to do this:
1. Use a new form to share credentials between apps, now we're checking the `UserDefaults.get('token-userId')`, it'll return the token of user saved by the same app or other app with the same configuration of AppGroup.
2. Changed all database locations of the Experimental App to other folder of App Group. It needs to be on an App Group folder because the Share Extension need to access this.
3. Copied the content of the serversDB to the new location, it'll allow us to migrate the credentials of users to the new location without logout nobody. To do this I needed to reset `roomsUpdatedAt` value after done the copy.

Some considerations:
Nothing about this code is change the behavior of this on Android.

Test Plan:
- [ ] Check if the credentials are shared between the Official and Experimental App.
(You can try to build it and install on your device with the Official App installed, it needs to get you logged instantly)
- [ ] Check if clear cache'll reset something on the other app, it needs to be done only on one of the apps.
- [ ] Opening a notification with the app closed.

### DIFF
--- a/app/lib/database/index.js
+++ b/app/lib/database/index.js
@@ -25,23 +25,14 @@ import migrations from './model/migrations';
 
 import { isIOS } from '../../utils/deviceInfo';
 
-const appGroupPath = isIOS ? `${ RNFetchBlob.fs.syncPathAppGroup('group.ios.chat.rocket') }/` : '';
+const appGroupPath = isIOS ? `${ RNFetchBlob.fs.syncPathAppGroup('group.chat.rocket.experimental') }/` : '';
 
 if (__DEV__ && isIOS) {
 	console.log(appGroupPath);
 }
 
 class DB {
-	databases = {
-		serversDB: new Database({
-			adapter: new SQLiteAdapter({
-				dbName: `${ appGroupPath }default.db`,
-				schema: serversSchema
-			}),
-			modelClasses: [Server, User],
-			actionsEnabled: true
-		})
-	}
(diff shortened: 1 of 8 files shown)


app/lib/database/index.js
```diff
@@ -59,6 +50,17 @@ class DB {
 		return this.databases.serversDB;
 	}
 
+	setServersDB() {
```
djorkaeffalexandre: Maybe I don't need to change this.
The new instance will be created only after the first import of this file, so, I can make the copy of the database file before the `restore` on `app/index`.

djorkaeffalexandre: Unfortunately it's not possible :(

djorkaeffalexandre:
Refer to #2483

PR #1890, opened 2020-03-17
[FIX] Investigate app losing connection issues

djorkaeffalexandre:
@RocketChat/ReactNative

Sometimes the app loses some streams connections: it may be something on the app or on the JS SDK.
After reopen the app sometimes after 1 minute (time necessary to close the websocket connection), it never returns to open a new connection with the socket, to fix it I tried to always try to reopen the connection when the app state back to `foreground`.

Unfortunately I can't create a reproducible steps to this problem, but with some tries we can get the error following these steps:
- Open the app
- Minimize the app
- Wait 1 minute
- Reopen the app
- Do it sometimes (more than 3)

Some things that are fixed by this pr:
- Duplicated subscribes to same stream when reopen the websocket.
- Status of user staying offline after start the app.
- Status of the user don't change to online again after reopen the app when the websocket was closed. (after 1 minute minimized).
- Subscribe on some streams never are resolved
  - Sometimes when you reopen your app after the websocket was closed, you could open a room and this never subscribe correctly.
- Now the reopen not have a timeout to realize the open, instead this, I created a debounced function that runs immediate the open when reopen is called.
- If you open a room without internet access and give access after opened it'll subscribe you to room.
- Fix a problem with serverVersion, now we're saving it to use when the app is started offline.
- When you have more than 1 server, and receive a notification from not currently server and open, this the other server will not be subscribed.


app/lib/rocketchat.js
```diff
@@ -62,13 +62,15 @@ const RocketChat = {
...
 	callJitsi,
 	async subscribeRooms() {
-		if (this.roomsSub) {
-			this.roomsSub.stop();
-		}
-		try {
-			this.roomsSub = await subscribeRooms.call(this);
-		} catch (e) {
-			log(e);
+		// if (this.roomsSub) {
```
djorkaeffalexandre: We can remove this, this is done on our `connect` and `logout`.

patches/@rocket.chat+sdk+1.0.0-alpha.41.patch
```diff
@@ -1,63 +0,0 @@
-diff --git a/node_modules/@rocket.chat/sdk/lib/api/api.js b/node_modules/@rocket.chat/sdk/lib/api/api.js
```
djorkaeffalexandre: I need to add it again when finished the pr.

PR #1906, opened 2020-03-20
[FIX] Correct message for manual approval user Registration

### LINKED ISSUE #1891: [BUG] Registration with manually user approval
- Your Rocket.Chat Experimental app version: 3.5.1.15552
- Your Rocket.Chat server version: 3.0.4
- Device (or Simulator) you're running with: iPhone XS iOS 13.3.1

I changed the registration-setting "manually approve news users" to true. So if a new user is registered, an administrator have to activate the user manually.

The problem is that the user gets this error message in case of a successful registration:

IMG_6374

**I think the problem is that Rocket.Chat tries to log in the new user immediately, but the account is disabled by default.**

If I touch the registration button a second time, this error message appears:

IMG_6375

sarthakpranesh:
@RocketChat/ReactNative
Closes #1891 

Bug: Whenever a new user registered and the server had `Manually Approve New Users` turned on a wrong message was shown to the user even when a user was registered.

Fix: Now when a user registers a proper message is shown informing the user that he/she has been registered and the account is waiting for the approval of the Admin.


app/i18n/locales/en.js
```diff
@@ -524,5 +524,7 @@ export default {
 	You_will_be_logged_out_of_this_application: 'You will be logged out of this application.',
 	Clear: 'Clear',
 	This_will_clear_all_your_offline_data: 'This will clear all your offline data.',
-	Mark_unread: 'Mark Unread'
+	Mark_unread: 'Mark Unread',
+	Manual_Authentication_Title: 'Waiting for user to be activated',
```
djorkaeffalexandre: Try to use the same title of `Verify_email` [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/a9e66edc38a6940f73bac00635afbf91d770f20d/app/i18n/locales/en.js#L477), create a new translation and use on both, `Verify email` and `Manual Auth` title.
:)

app/i18n/locales/en.js
```diff
@@ -524,5 +524,7 @@ export default {
...
+	Manual_Authentication_Desc: 'User registered, waiting for admin to approve the new user'
```
djorkaeffalexandre: Use same message of our web client.
`Before you can login, your account must be manually activated by an administrator.`

app/i18n/locales/en.js
sarthakpranesh: Done

app/i18n/locales/en.js
sarthakpranesh: Done

PR #1925, opened 2020-03-23
[FIX] Deeplinking without RoomId

djorkaeffalexandre:
@RocketChat/ReactNative

Now we'll have the same behavior of web client, allow users to use deeplink without a `rid`, like this:

Channel
`xcrun simctl openurl booted rocketchat://room\?host\=open.rocket.chat\&path\=channel/public-dm`

Group you're not in
`xcrun simctl openurl booted rocketchat://room\?host\=open.rocket.chat\&path\=group/test-djorkaeff`

Dm
`xcrun simctl openurl booted rocketchat://room\?host\=open.rocket.chat\&path\=direct/diego.mello`


app/lib/methods/canOpenRoom.js
```diff
@@ -4,15 +4,51 @@ const restTypes = {
...
-		// RC 0.61.0
-		await this.sdk.post(`${ restTypes[type] }.open`, { roomId: rid });
-		return true;
-	} catch (e) {
-		if (e.data && /is already open/.test(e.data.error)) {
-			return true;
+		const params = rid ? { roomId: rid } : { roomName: name };
+
+		// if it's a direct link without rid we'll create a new dm
+		// if the dm alread exists it'll return the existent
```
diegolmello, line 12: Small typo
```suggestion
		// if the dm already exists it'll return the existent
```

PR #1931, opened 2020-03-24
[FIX] Make in-app action menus abide by the theme preferences on android

### LINKED ISSUE #1926: [BUG] The ActionSheet menus don't respect theme preferences
- Your Rocket.Chat Experimental app version: 4.5.0
- Your Rocket.Chat server version: 3.1.0-develop
- Device (or Simulator) you're running with: OnePlus 6 and Google Pixel 3 XL Emulator

The "ActionSheet" dialogs do not respect the theming preferences on Android. While the iOS Action Sheets are part of the system and get their colours from the OS preferences, they're a part of the application on Android and do not change colour even when the system is in "dark mode" since the module suggests changing the module code itself for those that want to change the styles.

The screenshot is shown below: 
Screenshot_20200323-225246

Even while the theme is dark, the dialog box is white, which interrupts the whole UI

Solution: The react-native-action-sheet module needs to be replaced with the more well-maintained [Expo/react-native-action-sheet](https://github.com/expo/react-native-action-sheet) or a forked and edited version of the current module. The expo-maintained module exposes the same options(plus customization) and is almost completely interchangeable with the currently used module while only having to add a HOC.

I'm working on a PR

aKn1ghtOut:
* Added the color keys to each of the themes in Colors.js
* Refactored the files in question to replace react-native-action-sheet usages with equivalent @expo/react-native-action-sheet functions
* Made sure everything still works as was intended
* No change in iOS implementation since theme colors for Action Sheet in iOS are system-wide

Closes #1926 

The current *react-native-action-sheet* module doesn't allow customization through JS, and therefore breaks compatibility with the theme. The module doesn't even respect the system theme preferences, and hence doesn't feel like the app uses a system functionality either. Hence, I replaced it with the expo module @expo/react-native-action-sheet. The Expo module is *probably* built on top of the react-native-action-sheet module but allows much greater customizability while still being a drop-in replacement for the currently used module.

Screenshots/GIFs are attached below

Before: 
Screenshot_20200324-173835
Screenshot_20200324-173841

Updated:
ezgif-3-4b0ddf84e840
ezgif-3-ed1411c3e388
ezgif-3-e53a13668c9c


djorkaeffalexandre:
Hey @aKn1ghtOut, now we'll use our own action sheet to show actions.
You can see at #2114.
So, I'll close this.
But thanks anyway. 😁

PR #1753, opened 2020-02-19
[FIX] HTTP Basic Auth

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #1355.

### How to configure a server (using Nginx):
1. put following into `server{}`
```
      auth_basic "closed site";
      auth_basic_user_file htpasswd;
```
2. Create a new file `htpasswd` with
```
testuser:{PLAIN}testpassword
```
3. Reload nginx

### Test plan:
1. Test if multiple servers are working properly.
2. Test login into a server using `http://testuser:testpassword@localhost`.
3. Test if all places that use connection with server are working properly.

PR #1937, opened 2020-03-24
[FIX] Add headers wherever necessary

regalstreak:
@djorkaeffalexandre 

Closes #1753 

More instructions for testing are mentioned on the issue. Added headers wherever necessary. Images, Audio, Video and Files now work with a reverse proxy with nginx basic http auth now

### DIFF
--- a/app/containers/message/Audio.js
+++ b/app/containers/message/Audio.js
@@ -8,6 +8,7 @@ import Slider from '@react-native-community/slider';
 import moment from 'moment';
 import equal from 'deep-equal';
 import { activateKeepAwake, deactivateKeepAwake } from 'expo-keep-awake';
+import { settings as RocketChatSettings } from '@rocket.chat/sdk';
 
 import Touchable from './Touchable';
 import Markdown from '../markdown';
@@ -127,7 +128,10 @@ class MessageAudio extends React.Component {
 		this.setState({ loading: true });
 		try {
 			await Audio.setAudioModeAsync(mode);
-			await this.sound.loadAsync({ uri: `${ url }?rc_uid=${ user.id }&rc_token=${ user.token }` });
+			await this.sound.loadAsync({
+				uri: `${ url }?rc_uid=${ user.id }&rc_token=${ user.token }`,
+				headers: RocketChatSettings.customHeaders
+			});
 		} catch {
 			// Do nothing
 		}
--- a/app/views/AdminPanelView/index.js
+++ b/app/views/AdminPanelView/index.js
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { WebView } from 'react-native-webview';
 import { SafeAreaView } from 'react-navigation';
 import { connect } from 'react-redux';
+import { settings as RocketChatSettings } from '@rocket.chat/sdk';
 
 import I18n from '../../i18n';
 import StatusBar from '../../containers/StatusBar';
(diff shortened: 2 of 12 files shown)


app/containers/message/Audio.js
```diff
@@ -175,7 +176,10 @@ class Audio extends React.Component {
 				>
 					<Video
 						ref={this.setRef}
-						source={{ uri }}
+						source={{
+							uri,
+							headers: RocketChatSettings.customHeaders
```
djorkaeffalexandre: Hey @regalstreak, how you can see [here](https://github.com/react-native-community/react-native-video#headers), headers on iOS doesn't work without a patch, can you try to change this component to use `Audio` from `Expo`?
https://docs.expo.io/versions/latest/sdk/audio/
The other solution (that I think is most difficult), is change the `RCTVideo.m` to use an allowed approach to headers.

app/views/AdminPanelView/index.js
```diff
@@ -35,7 +36,10 @@ class AdminPanelView extends React.Component {
 			<SafeAreaView style={[styles.container, { backgroundColor: themes[theme].backgroundColor }]} testID='terms-view'>
 				<StatusBar theme={theme} />
 				<WebView
-					source={{ uri: `${ baseUrl }/admin/info?layout=embedded` }}
+					source={{
+						uri: `${ baseUrl }/admin/info?layout=embedded`,
+						headers: RocketChatSettings.customHeaders
+					}}
```
djorkaeffalexandre: This ins't work on iOS, maybe you can check the [documentation](https://github.com/react-native-community/react-native-webview/blob/master/docs/Guide.md#working-with-custom-headers-sessions-and-cookies) and follow something that you have forgotten.

app/views/AuthenticationWebView.js
```diff
@@ -110,7 +111,10 @@ class AuthenticationWebView extends React.PureComponent {
 			<>
 				<StatusBar theme={theme} />
 				<WebView
-					source={{ uri }}
+					source={{
+						uri,
+						headers: RocketChatSettings.customHeaders
+					}}
```
djorkaeffalexandre: It'll open a third-party url, can you check if it's needed?

app/containers/message/Audio.js
regalstreak: Yes, I'm aware of that as I had read the documentation. Skeptical about `RCTVideo.m` as we'd have to edit the library. Also, that is not there in the public API. Will look into the expo audio player for now. My suggestion would be to have minimum changes as possible here, as the audio component is going to be revamped during the GSoC any way.
@djorkaeffalexandre

app/views/AdminPanelView/index.js
regalstreak: @djorkaeffalexandre I found this [issue](https://github.com/react-native-community/react-native-webview/issues/4), it has still not been addressed. They suggest to [patch](https://github.com/react-native-community/react-native-webview/issues/4#issuecomment-427671845) `RCTWebView.m` as a workaround.

app/views/AuthenticationWebView.js
regalstreak: Will look into this

app/containers/message/Audio.js
regalstreak: Expo-av will be done in a separate PR

regalstreak:
@djorkaeffalexandre have a look, rebased and added headers to the new audio component. Works for me with basic auth now.

PR #1947, opened 2020-03-26
[FIX] Logout when token expires

### LINKED ISSUE #1944: Users will not logout automat when Login Expiration in Days is set
- Your Rocket.Chat Experimental app version: 4.5.0
- Your Rocket.Chat server version: 2.4.11  
- Device (or Simulator) you're running with: Device

Hey guys,

currently we set our Login Expiration in Days to 7 on Rocket.Chat self hosted server.
Users are LDAP Based.

Now we got the Problem thats the User dont see if the Token is expired.
So the App itself just looked like it work but not sync anymore.
Should it not logout the Users on the App or ask about the user and password again like on the OSX App? Anything we could do?

best and stay healthy,
Dennis

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #1944 .

We have a problem when token expires and app is opened, if you don't close and reopen the app you don't will logged out.
I created a new stream on our sdk to emit a `unauthorized` when some request fails with status 401, we can use it to handle on our app and logout the user when this was emitted.
If it makes sense, I'll merge this changes on our sdk soon.

### DIFF
--- a/patches/@rocket.chat+sdk+1.0.0-alpha.41.patch
+++ b/patches/@rocket.chat+sdk+1.0.0-alpha.41.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@rocket.chat/sdk/lib/api/api.js b/node_modules/@rocket.chat/sdk/lib/api/api.js
-index 5b7dc21..49f1af5 100644
+index 5b7dc21..b57b497 100644
 --- a/node_modules/@rocket.chat/sdk/lib/api/api.js
 +++ b/node_modules/@rocket.chat/sdk/lib/api/api.js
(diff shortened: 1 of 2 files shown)


patches/@rocket.chat+sdk+1.0.0-alpha.41.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@rocket.chat/sdk/lib/api/api.js b/node_modules/@rocket.chat/sdk/lib/api/api.js
-index 5b7dc21..49f1af5 100644
+index 5b7dc21..b57b497 100644
```
diegolmello: Missing PR to `@rocket.chat/sdk`

PR #1956, opened 2020-03-28
[FIX] Preview url stop flash when add or remove reactions

### LINKED ISSUE #1955: [BUG] Url preview flash when add or remove reactions
- Your Rocket.Chat Experimental app version: 4.6.0.999999999
- Your Rocket.Chat server version: 3.1.0-develop
- Device you're running with: Asus zenfone 3 (ZE552KL) Android 8

**I'm working on it**

EzequielDeOliveira:
@RocketChat/ReactNative

Closes #1955


app/lib/methods/subscriptions/rooms.js
```diff
@@ -140,6 +140,8 @@ const createOrUpdateSubscription = async(subscription, room) => {
 
 			if (tmp.lastMessage) {
 				const lastMessage = buildMessage(tmp.lastMessage);
+				// This last Message have a different format when merge two streams results in a error caused by the urls object
```
djorkaeffalexandre: The error is relative to the server, this update the last message with a message before parse the urls with OEmbed.

app/lib/methods/subscriptions/rooms.js
```diff
@@ -140,6 +140,8 @@ const createOrUpdateSubscription = async(subscription, room) => {
...
+				delete lastMessage.urls;
```
diegolmello: Mar-30-2020 09-25-17

EzequielDeOliveira:
Right, I'm solving this problem in the server !

EzequielDeOliveira:
@diegolmello  @djorkaeffalexandre  This problem was solved in the server then i did some changes in the app to not parse the same message twice and solve this problem, But this changes depends this PR  RocketChat/Rocket.Chat#17167  and this problem will be fixed.

app/lib/methods/helpers/normalizeMessage.js
```diff
@@ -21,6 +21,11 @@ export default (msg) => {
 	if (!msg) {
 		return null;
 	}
+
```
djorkaeffalexandre, line 24: Try only removing `normalizeMessage` [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/methods/helpers/mergeSubscriptionsRooms.js#L16) instead of this changes.

EzequielDeOliveira: Done! works so well.

Issue #1444, opened 2019-12-08
Error when accessing threads of a channel that one has not joined


Prateek93a:
- Your Rocket.Chat Experimental app version: 1.25.0.9999...
- Your Rocket.Chat server version: 3.0.0-develop open.rocket.chat
- Device (or Simulator) you're running with: Android P Sumsung Galaxy J6

![ezgif com-video-to-gif(6)](https://user-images.githubusercontent.com/44807945/70388421-92379280-19d7-11ea-8e5e-f1ee40020a6f.gif)

As the gif shows, the app shows shows error, when one try to access threads of a channel that he has not joined.

Prateek93a:
@diegolmello 
app > views > ThreadMessagesView> index.js > line no 100 inside subsribeData() function. when I comment out the code on line 100, the error disappears, and everything works fine, but I am not able to understand why this is so. 


![Screenshot from 2019-12-26 12-21-50](https://user-images.githubusercontent.com/44807945/71462947-ecf42b00-27da-11ea-9276-4eb2034a0e0c.png)

djorkaeffalexandre:
@Prateek93a try change `findAndObserve` to two functions like:
```
try {
  const db = database.active;
  const subCollection = await db.collections.get('subscriptions');
  const subscription = await subCollection.find(this.rid);
  const observable = subscription.observe();
  this.subSubscription = observable
    .subscribe(changes => console.log(changes));
catch (e) {
  console.log('subscription not found');
}
```
Unfortunately the function `findAndObserve` not works fine..

Prateek93a:
@djorkaeffalexandre Yes Sir your solution worked, since the `findAndObserve` was not throwing error, the app was crashing. Now after splitting the function, the error got thrown and logged out in the console, without crashing the app. So the solution is working.

djorkaeffalexandre:
;)

djorkaeffalexandre:
hey @Prateek93a, can you confirm if it's happening yet?
I followed your gif [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1494#issuecomment-587361708) and can't reproduce any error.

Prateek93a:
@djorkaeffalexandre I can't reproduce it. I think it was fixed by #1997. Closing it.

PR #1971, opened 2020-04-01
Fixed threadview error

### LINKED ISSUE #1970: [BUG] When accessing the threads of a channel or discussion
- Your Rocket.Chat Experimental app version: 4.6.0.9
- Your Rocket.Chat server version: 3.1.0-rc.6
- Device (or Simulator) you're running with: Android P Samsung Galaxy J6

As the Gif shows, when we enter thread view of channel or discussion, we get error message.
ezgif com-resize(1)

Prateek93a:
@RocketChat/ReactNative

Closes #1970 

As the Gif shows, the error is fixed. It was similar to #1444 .

ezgif com-resize(2)


app/views/ThreadMessagesView/index.js
```diff
@@ -89,12 +89,13 @@ class ThreadMessagesView extends React.Component {
...
-	subscribeData = () => {
+	subscribeData = async() => {
 		try {
 			const db = database.active;
-			this.subObservable = db.collections
+			const subscription = await db.collections
 				.get('subscriptions')
-				.findAndObserve(this.rid);
+				.find(this.rid);
+			this.subObservable = subscription.observe();
```
diegolmello, line 98: `const observable` works?
I don't think we need `this.subObservable` elsewhere.

diegolmello:
@Prateek93a It doesn't seem to fix the issue.

app/views/ThreadMessagesView/index.js
Prateek93a: removed const declaration and this.subObservable.

Prateek93a:
@diegolmello Yes Sir, the issue is not solved properly. Actually on rooms, which we have joined, the thread view does not show error now and shows all the threads and updates under changes properly, but on rooms we have not joined the thread view always shows empty screen even if there are threads in the room as the subscription record for that room is not found in the local db.

diegolmello:
It's going to be fixed by #1997

PR #2000, opened 2020-04-06
[FIX] Change notifications preferences

djorkaeffalexandre:
@RocketChat/ReactNative


app/containers/Picker.js
```diff
@@ -0,0 +1,64 @@
...
+Item.propTypes = {
+	item: PropTypes.object,
+	selected: PropTypes.bool,
+	onItemPress: PropTypes.func,
+	theme: PropTypes.string
+};
+
+const Picker = React.memo(({
+	data,
+	value = data[0].value,
```
diegolmello, line 37: can you add a condition to prevent crashes?

PR #2038, opened 2020-04-15
[FIX] Avatar of message as an emoji

### LINKED ISSUE #2015: emoji ignored in POST to /api/v1/chat.postMessage
hi!

POSTing a message to `/api/v1/chat.postMessage` and including an Emoji/avatar for the message, the emoji is visible in web interface but not from mobile - I see the original avatar of the user/bot the token used to authenticate the HTTP request are associated with.

- Your Rocket.Chat Experimental app version: 4.6.0 / android 9
- Your Rocket.Chat server version
```json
  "version": "3.0.10",
  "os": {
    "type": "Linux",
    "platform": "linux",
    "arch": "x64",
    "release": "4.19.0-8-cloud-amd64",
    "uptime": 236054,
    "loadavg": [
      0.07763671875,
      0.18310546875,
      0.2373046875
    ],
    "totalmem": 10482647040,
    "freemem": 7498309632,
    "cpus": [
      {
        "model": "Common KVM processor",
        "speed": 3400,
        "times": {
          "user": 44628000,
          "nice": 0,
          "sys": 27278200,
          "idle": 2244832800,
          "irq": 0
        }
      },
      {
        "model": "Common KVM processor",
        "speed": 3400,
        "times": {
          "user": 41160800,
          "nice": 16100,
          "sys": 27354100,
          "idle": 2257292000,
          "irq": 0
        }
      },
      {
        "model": "Common KVM processor",
        "speed": 3400,
        "times": {
          "user": 40843800,
          "nice": 17800,
          "sys": 26948400,
          "idle": 2249069600,
          "irq": 0
        }
      }
    ]
  },
  "process": {             
    "nodeVersion": "v12.14.0",
    "pid": 1,                 
    "uptime": 235876.26382856       
  },                           
  "deploy": {                
    "method": "docker-official",
    "platform": "selfinstall"
  },
```

- Device (or Simulator) you're running with: kubernetes KVM worker

Prateek93a:
@RocketChat/ReactNative

Closes #2015

This PR fixes the issue #2015. Actually there was no check to see if emoji field in the message was having value or not and hence the Avatar component would not show the emoji. I added the check for the field and added code to display emoji if it passed in the POST request. Also I made changes to schema to include column of emoji, because the information about the emoji used as avatar was not saved locally. Hence if we reopen the app, the emoji field would be undefined and the original avatar would get displayed instead. I followed the steps mentioned [here](https://nozbe.github.io/WatermelonDB/Advanced/Migrations.html) in given order for writing migration and everything worked fine.

Before->

After->


app/containers/Avatar.js
```diff
@@ -23,7 +24,14 @@ const Avatar = React.memo(({
...
 	});
 
-	let image = (
+	let image = emoji ? (
+		<Markdown
+			msg={emoji}
+			baseUrl={baseUrl}
+			getCustomEmoji={getCustomEmoji}
+			theme={theme}
+		/>
```
diegolmello, line 33: This could be a FastImage directly

Prateek93a: Actually, `CustomEmoji` is giving blank output for `emojiUnicode`. It is giving output for custom emojies

app/containers/Avatar.js
```diff
@@ -23,7 +24,17 @@ const Avatar = React.memo(({
...
+		<Emoji
+			theme={theme}
+			baseUrl={baseUrl}
+			getCustomEmoji={getCustomEmoji}
+			isMessageContainsOnlyEmoji
+			customEmojis
+			literal={emoji}
+			emojiName={emoji.replace(/:/g, '')}
+		/>
+	) : (
```
djorkaeffalexandre, line 38: Hey @Prateek93a, as it's related to `Message`, I think you can do this logic inside `MessageAvatar` instead of here.

app/containers/Avatar.js
```diff
@@ -23,7 +24,17 @@ const Avatar = React.memo(({
...
 	});
 
-	let image = (
+	let image = emoji ? (
+		<Emoji
+			theme={theme}
+			baseUrl={baseUrl}
+			getCustomEmoji={getCustomEmoji}
+			isMessageContainsOnlyEmoji
+			customEmojis
```
djorkaeffalexandre, line 34: Do this as `true` default value inside `Markdown/Emoji`.
`{ customEmojis = true }`.

app/containers/Avatar.js
```diff
@@ -23,7 +24,17 @@ const Avatar = React.memo(({
...
+			literal={emoji}
+			emojiName={emoji.replace(/:/g, '')}
```
djorkaeffalexandre, line 36: I think you can change `Markdown/Emoji` to do this logic inside there, receive only the `literal` prop and add `:` to start and end when you try to `getCustomEmoji`.
Remember to change this inside `Markdown/index`.
Thanks.

app/containers/message/index.js
```diff
@@ -241,64 +241,69 @@ class MessageContainer extends React.Component {
...
+					author={u}
+					ts={ts}
+					type={t}
+					attachments={attachments}
+					blocks={blocks}
+					urls={urls}
+					reactions={reactions}
+					alias={alias}
+					avatar={avatar}
+          emoji={emoji}
```
djorkaeffalexandre, line 271: Pay attention to your code editor pattern, sometimes this changes to `Spaces` instead `Tabs` automatically and we don't percept.
:)

Prateek93a: Thanks, I will keep that in mind.

app/containers/Avatar.js
Prateek93a: So, I have passed only `literal` prop and then used `literal.replace(/:/g, '')` in `getCustomEmoji` to get `emoji` and made the changes in the `Markdown/index` to only pass `literal`

app/containers/Avatar.js
Prateek93a: @djorkaeffalexandre Thanks for the review, Sir. Also, I am not sure if I fully understood this. Should I move the logic to compute `image` inside `MessageAvatar`? The `Avatar` component is used in other views as well, so should I remove `image` logic?

PR #2050, opened 2020-04-20
[FIX] Send totp-code to meteor call

### LINKED ISSUE #2028: Error showing the members of a channel or group.
- Your Rocket.Chat Experimental app version: ####
- Your Rocket.Chat server version: 3.1.0
- Device (or Simulator) you're running with: Android 8.0.0

The option to show the members of a channel or group does not work, Those who are online are not shown and neither are all the channel members

Screenshot_20200412-200632_RocketChat

### LINKED ISSUE #2051: facebook , google login ( internal server error )
- Your Rocket.Chat Experimental app version: 4.6.4
- Your Rocket.Chat server version: 2.4.9
- Device (or Simulator) you're running with: huawei nova3i

after return from facebook  or google login its open use-name page then internal server error
any help please ?
 register  work normal  ( google, fb) from web
register 
Screenshot_20200420_184942_chat rocket reactnative
 with user , pass , mail work normal

Prateek93a:
@RocketChat/ReactNative

Closes #2028
Closes #2051

This PR fixes the issue. The value of `this.code` [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/5698701a14b886838150739a8bdd8ccf4409982a/app/lib/rocketchat.js#L912) was undefined and hence the error was thrown, but since `e.error` didn't satisfy the conditions in the `if` statement, hence `else` block would get executed and the request would be rejected. I added code to use empty string if `this.code` is `undefined`.
Also, I added an additional commit to remove the errors in the variable names such as `details?.method`

Gif -
ezgif com-resize (3)


djorkaeffalexandre:
Hey @Prateek93a, your PR fixes the main error, but remove the optional of `details` property throws errors with TOTP on older servers.
Screen Shot 2020-04-20 at 13 49 55
As you can see on the image above, on older servers, like `https://stable.rocket.chat`, details doesn't exists and doesn't have any method, so, we need to cast using optional and use the default method (totp).
[Here](https://dev.to/pichardoj/optional-null-safe-in-javascript-1b7k) you can learn more about `optional` on JS.
Can you revert your latest commit?
Thanks.

app/lib/rocketchat.js
```diff
@@ -334,11 +334,11 @@ const RocketChat = {
 				const result = await this.login(params);
 				return resolve(result);
 			} catch (e) {
-				if (e.data?.error && (e.data.error === 'totp-required' || e.data.error === 'totp-invalid')) {
+				if (e.data.error && (e.data.error === 'totp-required' || e.data.error === 'totp-invalid')) {
 					const { details } = e.data;
 					try {
-						const code = await twoFactor({ method: details?.method || 'totp', invalid: e.data.error === 'totp-invalid' });
```
djorkaeffalexandre, line 340: Revert optional operators.

Prateek93a: Reverted the last commit. Sorry, I didn't know about the optional operator but now I know. Thanks for the review.

Issue #2053, opened 2020-04-21
Unable to login to server using latest app


Endareth:
- Your Rocket.Chat app version: 4.6.4.295
- Your Rocket.Chat server version: 2.4.11
- Device (or Simulator) you're running with: iPhone XS (13.4.1), iPhone 6S (11.2.6), iPhone 11 Pro (13.4.5)

After installing fresh versions of the app, I'm unable to log into my Rocket.Chat server.

Steps:
1. Install latest app from App Store and launch
2. Hit "Join a workspace"
3. Enter \<my server url\> and hit Connect

Expected:
Login screen displayed

Actual:
Empty screen, Back button and "Your workspace" title at top

This is happening on multiple devices, OS versions, and connectivity types. Existing logins seem to be maintained okay on other devices, it seems something specific to completely fresh installs, or to logging out and attempting to log back in.

diegolmello:
@Endareth Can you ping me at https://open.rocket.chat?
I'm `@diego.mello`.
Thanks.

Endareth:
Just upgraded the server to 3.1.1, no change. Tried to ping @diegolmello , but I appear to be blocked :\

Endareth:
With help from Diego, I was able to track this issue down to an overly zealous firewall configuration, which was detecting the initial `/api/v1/settings.public` call as a possible SQL injection attack, and dropping the connection. Resolved by tweaking the firewall profile. Probably be worth throwing an error in the client if this initial call fails, but this issue can now be closed, thanks.

PR #2075, opened 2020-04-29
[FIX] Encode public settings URI

djorkaeffalexandre:
@RocketChat/ReactNative

Related to #2053.

I've encoded uri that request settings to pass on firewall and show an error alert when `public settings` can't be read, this will prevent wrong navigations to `WorkspaceView` when we can't get the server settings.


app/sagas/selectServer.js
```diff
@@ -145,6 +145,7 @@ const handleServerRequest = function* handleServerRequest({ server, certificate
 			yield put(selectServerRequest(server, serverInfo.version, false));
 		}
 	} catch (e) {
+		showErrorAlert(I18n.t('Cant_get_public_settings', { contact: I18n.t('Contact_your_server_admin') }), I18n.t('Oops'));
```
diegolmello, line 148: If `getServerInfo` (or something else) throws an error, it's going to raise an alert with `Can't get public settings`, which may cause confusion.

PR #2098, opened 2020-05-09
[CHORE] Create DimensionsContext

EzequielDeOliveira:
@RocketChat/reactnative

This pr removes react-native-responsive-ui and creates middleware to listen the dimensions of the window and provide it through redux.


app/lib/dimensionsMiddleware.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default () => createStore => (...args) => {
+	const store = createStore(...args);
+
+	const handleDimensionsChange = (nextDimensionState) => {
+		if (nextDimensionState) {
+			if (nextDimensionState.window) {
+				store.dispatch(dimensionsWindow(nextDimensionState.window));
+			}
+			if (nextDimensionState.screen) {
```
diegolmello, line 13: We may not need `screen`.
It seems to return same data as `window`.
image

app/lib/dimensionsMiddleware.js
```diff
@@ -0,0 +1,23 @@
+import { Dimensions } from 'react-native';
+
+import { dimensionsWindow, dimensionsScreen } from '../actions/dimensions';
+
+export default () => createStore => (...args) => {
+	const store = createStore(...args);
+
+	const handleDimensionsChange = (nextDimensionState) => {
```
diegolmello, line 8: Apparently, it's firing twice each `change` event.
It's not an error on your PR, but on RN.
Can you create an if condition to compare previous vs next?
Example: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/appStateMiddleware.js#L12

app/reducers/dimensions.js
```diff
@@ -0,0 +1,23 @@
+import { DIMENSIONS } from '../actions/actionsTypes';
+
+const initialState = {
```
diegolmello, line 3: We're about to remove `window` and `screen` objects, so you can create these values as the keys returned from `Dimensions.get('window')`.

app/containers/MessageBox/UploadModal.js
```diff
@@ -252,4 +252,8 @@ class UploadModal extends Component {
 	}
 }
 
-export default responsive(withTheme(withSplit(UploadModal)));
+const mapStateToProps = state => ({
+	window: state.dimensions.window
+});
+
+export default connect(mapStateToProps)(withTheme(withSplit(UploadModal)));
```
diegolmello, line 259: You should avoid to add `connect`.
`UploadModal`, for example, has a parent container (`Messagebox`) which declares a `connect`, so it's better to add this `window` props there and pass here as a normal prop.

PR #2106, opened 2020-05-11
[FIX] Autocomplete position on Android

djorkaeffalexandre:
@RocketChat/ReactNative


EzequielDeOliveira:
FYI The cursor can lose the right position with speed

But I think that, for the main purpose of this issue, it has been fixed.
Thanks.

PR #2107, opened 2020-05-11
[FIX] Don't autoCorrect if click a mention

djorkaeffalexandre:
@RocketChat/ReactNative

We'll check:
* If text was not pasted
* If autoCorrection is disabled
* If text.length > 1


patches/react-native+0.62.2.patch
```diff
@@ -308,3 +331,10 @@ index c5c1ec3..1bf60b6 100644
  
  - (BOOL)isKeyCommandRegisteredForInput:(NSString *)input
                           modifierFlags:(UIKeyModifierFlags)flags
+diff --git a/node_modules/react-native/scripts/.packager.env b/node_modules/react-native/scripts/.packager.env
```
diegolmello, line 334: I think you can remove this, right?

PR #2114, opened 2020-05-12
[NEW] Action Sheet

### LINKED ISSUE #759: Use reanimated-bottom-sheet
We switched action sheet lib months ago, but it's still not what we're looking for.
While native look and feel is great, the current lib isn't customizable.

https://github.com/osdnk/react-native-reanimated-bottom-sheet aims to implement [Material bottom sheet](https://material.io/design/components/sheets-bottom.html#standard-bottom-sheet) and it's using react-native-gesture-handler and react-native-reanimated to do it. Gestures and animations are great!

We should apply it on message actions and in messabox autocomplete container (for emojis, usernames, etc).

djorkaeffalexandre:
@RocketChat/ReactNative

Depends of #2146.
Closes #759, #1612, #1609, #1594, #1926.

Screen Shot 2020-05-29 at 14 39 09

Screen Shot 2020-06-02 at 11 44 21

Screen Shot 2020-06-02 at 11 43 35


app/containers/ActionSheet/index.js
```diff
@@ -0,0 +1,216 @@
...
+/*
+ * At Android height needs to be adjusted
+ * because of the StyleSheet has a difference of
+ * +2 dp between Android and iOS
+ */
+const ANDROID_ADJUST = isAndroid ? 2 : 0;
+
+const ActionSheet = forwardRef(({ children, theme }, ref) => {
+	const bottomSheetRef = useRef();
+	const [data, setData] = useState({});
```
diegolmello, line 56: Can we not use `setData` or manage re-render?
Cus maybe this is re-rendering more than it should.
```js
{} === {} // false
```

patches/react-native-scroll-bottom-sheet+0.6.1.patch
```diff
@@ -0,0 +1,33 @@
+diff --git a/node_modules/react-native-scroll-bottom-sheet/src/index.tsx b/node_modules/react-native-scroll-bottom-sheet/src/index.tsx
```
diegolmello, line 1: I'm not sure it's the case, but maybe we should create a fork

app/containers/ActionSheet/index.js
djorkaeffalexandre: I think it's not a problem, since we call `setData` just at `show`, we want to show this ActionSheet even the data is equal to the latest.

patches/react-native-scroll-bottom-sheet+0.6.1.patch
djorkaeffalexandre: I've opened a PR at the library because it's a bug fix.
I think we can maintain the patch while it's not merged.
https://github.com/rgommezz/react-native-scroll-bottom-sheet/pull/10

app/containers/ActionSheet/styles.js
```diff
@@ -0,0 +1,61 @@
...
+		width: '50%',
+		alignSelf: 'center',
+		left: '25%'
+	},
+	button: {
+		marginHorizontal: 16,
+		paddingHorizontal: 14,
+		justifyContent: 'center',
+		height: 48,
+		borderRadius: 2,
```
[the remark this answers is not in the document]
djorkaeffalexandre: I have copied this from [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/containers/Button/index.js#L15) to follow our default button.
Do you think we should do this different here?

PR #2132, opened 2020-05-18
[FIX] Handle no attached activity when try biometry

djorkaeffalexandre:
@RocketChat/ReactNative

Apparently, activities on Android has an extra time to be attached when the app is resumed, this causes some errors because the JS code is loaded before this be attached and some functions are called before the activity be attached, so some native modules are sending Errors like "The app needs to be at foreground", for now I'm doing only a handle of this errors, but I'll try to study how we can wait the activity be attached before try to do native things that needs a activity attached.


djorkaeffalexandre:
Seems to be fixed by #2130.

PR #2156, opened 2020-05-29
[Feature] Full Screen composer

EzequielDeOliveira:
@RocketChat/ReactNative

**Tasks**
  -  [x] Component of the full screen composer
  -  [x] Using a button to open and close the full screen composer
  -  [x] Open the full screen composer from the current composer.
  -  [x] Using the full screen composer with the current component.
  -  [x] Creating a modal with the full screen composer
  -  [x] Using react-native-modal in the full screen composer
  -  [x] Componentization of the MessageBox
  -  [x] Check the behavior on the tablet
  -  [x] Storybook for messageBox

**Phone**
ezgif com-video-to-gif(12) *Presentation*

ezgif com-video-to-gif(13) *Switching between composers*

ezgif com-video-to-gif(14) *Replying message*

**Tablet**

ezgif com-video-to-gif(28)


app/containers/MessageBox/index.js
```diff
@@ -797,13 +857,38 @@ class MessageBox extends Component {
...
+	changeComposerState = () => {
+		const { fullScreenUp } = this.state;
+
+		if (fullScreenUp) {
+			this.setState({ translateY: BOTTOM });
+		} else {
+			this.setState({ translateY: TOP });
+		}
+
+		this.setState({ fullScreenUp: !fullScreenUp });
```
rowanwork, line 869: How about combining them all to simplify it by reducing the branching and lines of code?
```
		this.setState({ 
			translateY: this.state.fullScreenUp ? BOTTOM : TOP, 
			fullScreenUp: !fullScreenUp 
		});
```

app/containers/MessageBox/index.js
```diff
@@ -111,7 +149,9 @@ class MessageBox extends Component {
 			},
 			commandPreview: [],
 			showCommandPreview: false,
-			command: {}
+			command: {},
+			fullScreenUp: false,
```
rowanwork, line 153: Perhaps we could have a slightly more boolean name for `fullScreenUp` e.g. `isFullScreenUp` or even just `isFullScreen` or `isOpen`.

app/containers/MessageBox/index.js
```diff
@@ -248,6 +288,12 @@ class MessageBox extends Component {
 		if (!equal(nextState.file, file)) {
```
rowanwork, line 288: Can you think of a better pattern for this than just appending `if` statements? Is there a slight performance improvement by putting all the deep equals calls (`equal(`)) at the very end?

app/containers/MessageBox/index.js
```diff
@@ -596,6 +642,11 @@ class MessageBox extends Component {
 	}
 
 	showUploadModal = (file) => {
+		const { fullScreenUp } = this.state;
+
+		if (fullScreenUp) {
```
rowanwork, line 647: If we just have the one value from the state and we are just using it once, then I think it's a bit cleaner to just using `this.state.fullScreenUp` in the `if` statement rather than assigning it to a variable.

app/containers/MessageBox/index.js
```diff
@@ -651,7 +702,11 @@ class MessageBox extends Component {
 		const {
 			rid, tmid, baseUrl: server, user
 		} = this.props;
+		const { fullScreenUp } = this.state;
```
rowanwork, line 705: Same comment here about destructuring

app/containers/MessageBox/index.js
```diff
@@ -674,6 +729,7 @@ class MessageBox extends Component {
 		const {
 			onSubmit, rid: roomId, tmid
 		} = this.props;
+		const { fullScreenUp } = this.state;
```
rowanwork, line 732: And destructuring here as well.

app/containers/MessageBox/index.js
```diff
@@ -797,13 +857,38 @@ class MessageBox extends Component {
...
+		}
+
+		this.setState({ fullScreenUp: !fullScreenUp });
+	}
+
+	renderTopButton = () => {
+		const { fullScreenUp } = this.state;
+		const { theme, editing } = this.props;
+		return (
+			<TouchableOpacity onPress={() => this.changeComposerState()} style={[stylez.topButton, { backgroundColor: themes[theme].messageboxBackground }, editing && { backgroundColor: themes[theme].chatComponentBackground }]}>
```
rowanwork, line 876: The value for `style` is quite long and complicated. Could it be extracted to a variable?

app/containers/MessageBox/index.js
```diff
@@ -818,56 +903,120 @@ class MessageBox extends Component {
...
 				<CommandsPreview commandPreview={commandPreview} showCommandPreview={showCommandPreview} />
 				<Mentions mentions={mentions} trackingType={trackingType} theme={theme} />
-				<View style={[styles.composer, { borderTopColor: themes[theme].separatorColor }]}>
-					<ReplyPreview
-						message={message}
-						close={replyCancel}
-						username={user.username}
-						replying={replying}
-						getCustomEmoji={getCustomEmoji}
+				{fullScreenUp ? (
```
rowanwork, line 906: This now feels a bit complicated with logic and a few fragment blocks. Could it be split into different components? It feels like `renderFullScreenComposer` could be converted into one

app/containers/MessageBox/index.js
diegolmello: We could identify state and props on arrays and compare them, but that's not related to this PR.

About `equal()`, yes, it makes a deep compare, so it should be at the bottom end.
When `shouldComponentUpdate` is triggered, it executes all `if` statements in order to find a `return`.
Primitive values cost a lot less than objects.
It makes no sense to have expensive comparisons at the top if a primitive was changed.

app/containers/MessageBox/index.js
diegolmello: Yep. `isFullscreen` is better.

app/containers/MessageBox/index.js
diegolmello: We have `react/destructuring-assignment` enabled.
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/destructuring-assignment.md

app/containers/MessageBox/index.js
diegolmello: I agree.
It's a good practice to trigger the least re-renders as possible.

app/containers/MessageBox/index.js
rowanwork: To me converting the `if` to concatenated `||` is more clear in terms or intentions and readability. But I guess you are right that it might be out of scope of this PR. Moving the `equals` to the bottom sounds good though.

app/containers/MessageBox/index.js
```diff
@@ -823,9 +849,152 @@ class MessageBox extends Component {
 		}
 	}
 
+	changeComposerState = () => {
```
rowanwork, line 852: maybe this would be better names `toggleFullScreen`. The current name isn't clear what it is actually doing

app/containers/MessageBox/index.js
EzequielDeOliveira: Can be extracted, I'll do it

app/containers/MessageBox/index.js
EzequielDeOliveira: I was thinking about it, thanks for the idea I'll try to convert this function to component.

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,975 @@
...
+	FileUpload_MediaTypeWhiteList: state.settings.FileUpload_MediaTypeWhiteList,
+	FileUpload_MaxFileSize: state.settings.FileUpload_MaxFileSize,
+	Message_AudioRecorderEnabled: state.settings.Message_AudioRecorderEnabled
+});
+
+const dispatchToProps = ({
+	typing: (rid, status) => userTypingAction(rid, status)
+});
+
+export default connect(mapStateToProps, dispatchToProps, null, { forwardRef: true })(MessageBox);
```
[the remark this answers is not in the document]
EzequielDeOliveira: Probably this component will be smaller with componentization!

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,975 @@
...
+						theme={theme}
+						{...isAndroidTablet}
+					/>
+				</Animated.View>
+			</>
+		);
+	}
+
+	return (
+		<>
```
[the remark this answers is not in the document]
EzequielDeOliveira: It's not required and will be improved!

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,975 @@
...
+									/>
+								</View>
+							</View>
+						</>
+					)}
+			</>
+		);
+	}
+
+	function renderFullScreenComposer() {
```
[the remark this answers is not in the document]
EzequielDeOliveira: Right, I'm trying separate in components.

app/views/FullScreenComposerView/index.js
```diff
@@ -0,0 +1,974 @@
...
+			if (this.tracking && this.tracking.resetTracking) {
+				this.tracking.resetTracking();
+			}
+		});
+		this.unsubscribeBlur = navigation.addListener('blur', () => {
+			this.component?.blur();
+		});
+	}
+
+	UNSAFE_componentWillReceiveProps(nextProps) {
```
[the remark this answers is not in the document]
EzequielDeOliveira: Right, I'll avoid this lifecycle.

app/views/FullScreenComposerView/index.js
```diff
@@ -0,0 +1,974 @@
...
+		if (slashCommand) {
+			const [, name, params] = slashCommand;
+			const commandsCollection = db.collections.get('slash_commands');
+			try {
+				const command = await commandsCollection.find(name);
+				if (command.providesPreview) {
+					return this.setCommandPreview(command, name, params);
+				}
+			} catch (e) {
+				console.log('Slash command not found');
```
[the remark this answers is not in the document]
EzequielDeOliveira: I'll investigate this. this is used in the small composer.

app/views/FullScreenComposerView/index.js
```diff
@@ -0,0 +1,974 @@
...
+				this.identifyMentionKeyword(name, lastChar);
+			} catch (e) {
+				log(e);
+			}
+		} else {
+			this.stopTrackingMention();
+		}
+	}, 100)
+
+	onKeyboardResigned = () => {
```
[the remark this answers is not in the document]
EzequielDeOliveira: Yeah. we can do it!

app/containers/MessageBox/index.js
```diff
@@ -290,6 +293,12 @@ class MessageBox extends Component {
 		if (!equal(nextProps.message, message)) {
 			return true;
 		}
+		if (nextState.isFullScreen !== isFullScreen) {
```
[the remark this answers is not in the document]
EzequielDeOliveira: Ok

app/containers/MessageBox/index.js
EzequielDeOliveira: Good idea thanks!

app/containers/MessageBox/index.js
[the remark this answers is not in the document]
EzequielDeOliveira: Ok

app/views/FullScreenComposerView/index.js
```diff
@@ -0,0 +1,974 @@
...
+		FileUpload_MaxFileSize: PropTypes.number,
+		Message_AudioRecorderEnabled: PropTypes.bool,
+		isMasterDetail: PropTypes.bool
+	}
+
+
+	constructor(props) {
+		super(props);
+		const { route } = props;
+		this.state = {
```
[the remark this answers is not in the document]
EzequielDeOliveira: This constructor is not used with the new architecture.

app/containers/MessageBox/index.js
```diff
@@ -804,13 +820,107 @@ class MessageBox extends Component {
 		}
 	}
 
+	changeComposerState = () => {
+		const { isFullscreen } = this.state;
+
+		this.setState({
+			isFullscreen: !isFullscreen
+		});
```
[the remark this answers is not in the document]
EzequielDeOliveira: This is very useful, I didn't know thanks!

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,286 @@
...
+		user: PropTypes.shape({
+			id: PropTypes.string,
+			username: PropTypes.string,
+			token: PropTypes.string
+		}),
+		innerRef: PropTypes.object
+	};
+
+	constructor(props) {
+		super(props);
```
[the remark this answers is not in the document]
EzequielDeOliveira: Cool! This can remove unneeded verbosity :)

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,286 @@
...
+				onFinish={finishAudioMessage}
+				onPress={() => this.startRecordingAudio()}
+			/>
+		);
+
+		return (
+			<>
+				<CommandsPreview commandPreview={commandPreview} showCommandPreview={showCommandPreview} />
+				<Mentions mentions={mentions} trackingType={trackingType} theme={theme} />
+				<View style={[styles.bottomBarButtons, { backgroundColor: themes[theme].messageboxBackground }, editing && { backgroundColor: themes[theme].chatComponentBackground }]}>
```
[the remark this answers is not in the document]
EzequielDeOliveira: Right I'll extract!

app/containers/MessageBox/MainComposer.js
```diff
@@ -0,0 +1,292 @@
...
+		user: PropTypes.shape({
+			id: PropTypes.string,
+			username: PropTypes.string,
+			token: PropTypes.string
+		}),
+		innerRef: PropTypes.object
+	};
+
+	constructor(props) {
+		super(props);
```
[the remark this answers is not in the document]
EzequielDeOliveira: Removed! really wasn't necessary

app/containers/MessageBox/index.js
```diff
@@ -942,20 +872,88 @@ class MessageBox extends Component {
...
-					trackInteractive
-					// revealKeyboardInteractive
-					requiresSameParentToManageScrollView
-					addBottomView
-					bottomViewColor={themes[theme].messageboxBackground}
-					iOSScrollBehavior={iOSScrollBehavior}
-				/>
+				{isFullScreen
+					? (
+						<FullScreenComposer
```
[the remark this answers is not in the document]
EzequielDeOliveira: Good idea I'll do it

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,975 @@
...
+										recordAudioMessage={recordAudioMessage}
+										recordAudioMessageEnabled={Message_AudioRecorderEnabled}
+										showMessageBoxActions={showMessageBoxActions}
+									/>
+								</View>
+							</View>
+						</>
+					)
+					: (
+						<>
```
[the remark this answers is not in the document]
EzequielDeOliveira: Not necessary.

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,975 @@
...
+	},
+	topButton: {
+		width: '100%',
+		alignItems: 'center',
+		justifyContent: 'center'
+	}
+});
+
+
+const MessageBox = React.memo(({
```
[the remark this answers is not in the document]
EzequielDeOliveira: It is no longer used.

app/views/FullScreenComposerView/index.js
```diff
@@ -0,0 +1,974 @@
...
+			log(e);
+		}
+
+		if (msg) {
+			this.setInput(msg);
+			this.setShowSend(true);
+		}
+
+		if (isAndroid) {
+			require('../../containers/MessageBox/EmojiKeyboard');
```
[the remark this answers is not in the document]
EzequielDeOliveira: I changed to await import() and works very well.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -64,6 +67,15 @@ export default class RecordAudio extends React.PureComponent {
 		};
 	}
 
+	componentDidMount() {
+		const { recordStartState, toggleRecordAudioWithState } = this.props;
+
+		if (recordStartState) {
```
rowanwork, line 73: Could there be a scenario where is mounts with `recordStartState` being false, but then the prop later changes to true? If so, would `this.startRecordingAudio();` also need to be called somewhere else as well?

app/containers/MessageBox/buttons/OpenFullScreenButton.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import BaseButton from './BaseButton';
+
+const OpenFullScreenButton = React.memo(({ theme, onPress }) => (
```
rowanwork, line 6: Is memo actually needed here?

app/containers/MessageBox/index.js
```diff
@@ -322,6 +323,15 @@ class MessageBox extends Component {
 		if (isTablet) {
 			EventEmiter.removeListener(KEY_COMMAND, this.handleCommands);
 		}
+		if (isAndroid) {
+			BackHandler.removeEventListener('backPress');
```
rowanwork, line 327: You should pass in the reference to your back function

app/containers/MessageBox/FullScreenComposer.js
```diff
@@ -0,0 +1,291 @@
...
+				isVisible={isFullScreen}
+				useNativeDriver
+				hideModalContentWhileAnimating
+				coverScreen={false}
+				backdropOpacity={backdropOpacity}
+				swipeDirection='down'
+				onSwipeComplete={() => this.closeModal()}
+			>
+				<View style={{ backgroundColor, flex: 1 }}>
+					<TouchableOpacity onPress={() => this.closeModal()} style={styles.fullScreenComposerCloseButton}>
```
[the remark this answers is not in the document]
EzequielDeOliveira: Ok!

app/containers/MessageBox/index.js
EzequielDeOliveira: Right, I changed to `this.backHandler = BackHandler.addEventListener('backPress', this.backPress);` and `this.backHandler.remove()`

rowanwork: Not sure that you need to add another field `this.backHandler`, `BackHandler.removeEventListener('backPress', this.backPress);` should work.

EzequielDeOliveira: Right work fine, I passed the reference!

app/containers/MessageBox/buttons/OpenFullScreenButton.js
EzequielDeOliveira: This memo is to improve the performance, these props don't often change then we can win with this memo. I created this button based in all other buttons.

rowanwork: I understand, but is there actually any measured performance gains, or is it just in theory? BaseButton also uses `React.memo`, so do we still need it at this level?
Is this component actually required since it is only used in one place and doesn't wrap anything complicated?

app/containers/MessageBox/buttons/OpenFullScreenButton.js
```diff
@@ -0,0 +1,21 @@
...
+	<BaseButton
+		onPress={onPress}
+		testID='messagebox-full-screen'
```
rowanwork, line 9: In general you want to avoid things like `testID` and test more from the users point of view e.g. find by `accessibilityLabel` instead.

app/containers/MessageBox/buttons/OpenFullScreenButton.js
EzequielDeOliveira: Got it. make sense because it was just a theory but this `React.memo` is not necessary in this button. thanks :)

app/containers/MessageBox/buttons/OpenFullScreenButton.js
EzequielDeOliveira: Right. I just use this props because I want to use the `BaseButton` and this props is required here https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/9dbe10bcf871392ed09b512152bc921578150e9f/app/containers/MessageBox/buttons/BaseButton.js#L27

app/containers/MessageBox/RecordAudio.js
EzequielDeOliveira: It was wrong because the changes to keep the keyboard while composers are switching, but that has already been resolved.

diegolmello:
Too outdated.

PR #2157, opened 2020-05-30
Removed server toggle dropdown if showServerDropdown is false

BizClimb:
@RocketChat/ReactNative

I am using the single server branch and noticed that server drop-down in the android header is still toggling. I added an IF statement that will check if **showServerDropdown** is true, if so it will act normally, if it's false I removed the toggle functionality along with the toggle image from the header. 

NOTE: I did not adjust the iOS header but I imagine the same problem exists there and I will get around to it eventually.

### DIFF
--- a/app/views/RoomsListView/Header/Header.android.js
+++ b/app/views/RoomsListView/Header/Header.android.js
@@ -59,29 +59,41 @@ const Header = React.memo(({
 			</View>
 		);
 	}
-	return (
-		<View style={styles.container}>
-			<TouchableOpacity
-				onPress={onPress}
-				testID='rooms-list-header-server-dropdown-button'
-				disabled={connecting || isFetching}
-			>
-				{connecting ? <Text style={[styles.updating, titleColorStyle]}>{I18n.t('Connecting')}</Text> : null}
-				{isFetching ? <Text style={[styles.updating, titleColorStyle]}>{I18n.t('Updating')}</Text> : null}
-				<View style={styles.button}>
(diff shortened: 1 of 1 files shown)


app/views/RoomsListView/Header/Header.android.js
```diff
@@ -59,29 +59,41 @@ const Header = React.memo(({
...
-							showServerDropdown && styles.upsideDown,
-							{ tintColor: themes[theme].headerTitleColor }
-						]}
-						source={{ uri: 'disclosure_indicator_server' }}
-					/>
-				</View>
-			</TouchableOpacity>
-		</View>
-	);
+        if (showServerDropdown) {
```
djorkaeffalexandre, line 62: Hey @BizClimb, it's not the right way to do this.
You can just use `disabled` at `TouchableOpacity` to achieve the same behavior.
And remove the image with the `disclosure`.

BizClimb:
> I think your intention is to merge this at the `single-server` branch instead of `develop`..

Thanks, yes, it was, for some reason it would only let me merge into develop. My first attempt at GitHub, thanks for the feedback.

app/views/RoomsListView/Header/Header.android.js
BizClimb: Ahh, thanks for the clarification and example.

diegolmello:
Closing as the header is a lot different now.
Let me know if the issue persists.
Thanks for your contribution.

PR #2162, opened 2020-06-03
[NEW] Send multiple attachments

### LINKED ISSUE #1101: [Improvement] Multiple Uploads
More then One Image or file selects from Gallery or FileBrower to upload.

djorkaeffalexandre:
@RocketChat/ReactNative

Closes #1101.


app/views/ShareView/index.js
```diff
@@ -1,307 +1,360 @@
...
+					<Item
+						title={I18n.t('Send')}
+						onPress={this.send}
+						buttonStyle={[styles.send, { color: themes[theme].previewTintColor }]}
+					/>
+				</CustomHeaderButtons>
+			);
+		}
+
+		options.headerBackground = () => <View style={[styles.container, { backgroundColor: themes[theme].previewBackground }]} />;
```
djorkaeffalexandre, line 93: you can use:
`headerStyle: { backgroundColor }`

app/views/ShareView/index.js
```diff
@@ -1,307 +1,360 @@
...
+		const { loading, selected } = this.state;
 		if (loading) {
 			return;
 		}
 
-		this.setState({ loading: true });
-		if (isMedia) {
-			await this.sendMediaMessage();
+		// update state
+		await this.selectFile(selected);
```
djorkaeffalexandre, line 154: This should be done only when it's a media share.

djorkaeffalexandre: `selectFile` doesn't return a promise, so we can't wait for this.
maybe it can be fixed solving a promise at the `this.setState` callback function.

app/views/ShareView/index.js
```diff
@@ -1,307 +1,360 @@
...
-					<Text style={[styles.mediaText, { color: themes[theme].titleText }]} numberOfLines={1}>{fileInfo.name}</Text>
-					<Text style={[styles.mediaText, { color: themes[theme].titleText }]}>{this.bytesToSize(fileInfo.size)}</Text>
-				</View>
-			</View>
-		);
-	};
+	onChangeText = (text) => {
+		const { attachments } = this.state;
+		// we manage attachments on this state only when it's text only sharing
+		if (!attachments.length) {
```
djorkaeffalexandre, line 245: Correct me if I'm wrong, but I don't guess so we will call this function when this component has some attachment.

app/views/AttachmentView.js
```diff
@@ -61,13 +68,16 @@ class AttachmentView extends React.Component {
...
+		const { route, navigation, theme } = this.props;
 		const attachment = route.params?.attachment;
 		const { title } = attachment;
 		const options = {
-			headerLeft: () => <CloseModalButton testID='close-attachment-view' navigation={navigation} />,
+			headerLeft: () => <CloseModalButton testID='close-attachment-view' navigation={navigation} buttonStyle={{ color: themes[theme].previewTintColor }} />,
 			title: decodeURI(title),
-			headerRight: () => <SaveButton testID='save-image' onPress={this.handleSave} />
+			headerRight: () => <SaveButton testID='save-image' onPress={this.handleSave} buttonStyle={{ color: themes[theme].previewTintColor }} />,
+			headerBackground: () => <View style={{ flex: 1, backgroundColor: themes[theme].previewBackground }} />,
```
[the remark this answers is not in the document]
diegolmello: Tested. Doesn't work.

PR #2171, opened 2020-06-08
[NEW] Add to F-Droid

GOVINDDIXIT:
- [x] Divide the current app into Play and Foss build
- [x] Remove all google/proprietary libraries from foss build
- [x] Remove all tracking, monitoring and reporting services like Bugsnag
- [x] Make play build functional
- [x]  Make foss build functional
- [x] Use white icons instead of blue one and other resources related changes for foss build
- [x] Add support to easily generate builds using the terminal.
`npm run play` for Play build
`npm run foss` for Foss build


android/app/build.gradle
```diff
@@ -1,8 +1,21 @@
+def taskRequests = getGradle().getStartParameter().getTaskRequests().toString()
+def isPlay = !(taskRequests.contains("Foss") || taskRequests.contains("foss"))
+
 apply plugin: "com.android.application"
 apply plugin: 'kotlin-android'
-apply plugin: "io.fabric"
-apply plugin: "com.google.firebase.firebase-perf"
+
 apply plugin: 'com.bugsnag.android.gradle'
+apply plugin: 'com.google.gms.google-services'
```
djorkaeffalexandre, line 8: Is that supposed to be inside `isPlay`?

android/app/build.gradle
```diff
@@ -179,6 +192,19 @@ android {
...
+    productFlavors {
+        foss {
+            dimension = 'type'
+            buildConfigField 'boolean', 'PLAY_BUILD', "false"
+        }
+        play {
+            dimension = 'type'
+            buildConfigField 'boolean', 'PLAY_BUILD', "true"
+        }
+    }
```
djorkaeffalexandre, line 206: The default is to use `"` in this file.

android/app/build.gradle
GOVINDDIXIT: I earlier thought that disabling google Services would be a better option instead of disabling the plugin.  Will update it 👍

android/app/build.gradle
GOVINDDIXIT: Need more light on this. Pls explain more

djorkaeffalexandre: It's just to change the `single quotes (')` for `double quotes (")`. :)

GOVINDDIXIT: Oh, Thanks will update soon :)

android/build.gradle
```diff
@@ -17,16 +17,18 @@ buildscript {
             url 'https://maven.fabric.io/public'
         }
     }
+    
+    def taskRequests = getGradle().getStartParameter().getTaskRequests().toString()
+    def isPlay = !(taskRequests.contains("Foss") || taskRequests.contains("foss"))
```
djorkaeffalexandre, line 22: I guess is better if you use `toString().toLowerCase()` on `getTaskRequests()` , so any variant of `Foss` will be considered a foss build.
Change this on all necessary places.

djorkaeffalexandre:
You changed the location of `google-services.json`, we're doing an instruction with this on our ci [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/.circleci/config.yml#L271).
`google-services-prod.json` should be considered too.

GOVINDDIXIT:
> You changed the location of `google-services.json`, we're doing an instruction with this on our ci [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/.circleci/config.yml#L271).
> `google-services-prod.json` should be considered too.

Sure @djorkaeffalexandre Will made change related to this on CI PR. Thanks 😀

app/utils/review.js
```diff
@@ -75,7 +76,7 @@ const tryReview = async() => {
 
 	// if ask me later was pressed earlier, we can ask for review only after {{numberOfDays}} days
 	// if there's no review and it wasn't dismissed by the user
-	if (daysBetween(lastReviewDate, new Date()) >= numberOfDays && !doneReview) {
+	if (daysBetween(lastReviewDate, new Date()) >= numberOfDays && !doneReview && !isFDroidBuild) {
```
diegolmello, line 79: `tryReview` execution should be completely dismissed.

app/constants/links.js
```diff
@@ -3,6 +3,7 @@ import { getBundleId, isIOS } from '../utils/deviceInfo';
 const APP_STORE_ID = '1272915472';
 
 export const PLAY_MARKET_LINK = `https://play.google.com/store/apps/details?id=${ getBundleId }`;
+export const FDROID_MARKET_LINK = 'https://f-droid.org/repo/chat.rocket.android_2057.apk'; // Once link is finalised, Update this
```
diegolmello, line 6: Change this before merging

android/fastlane/Fastfile
```diff
@@ -16,21 +16,26 @@
...
+  lane :playBuild do
+    gradle(task: "assemblePlayDebug")
   end
 
-  desc "Build App for release"
-  lane :release do
-    gradle(task: "bundleRelease")
+  desc "Foss build for development"
+  lane :fossBuild do
+    gradle(task: "assembleFossDebug")
```
diegolmello, line 26: It needs to be release

.circleci/config.yml
```diff
@@ -337,13 +390,16 @@ workflows:
           requires:
             - ios-hold-testflight
 
-      - android-build:
+      - android-play-build:
+          requires:
+            - lint-testunit
+      - android-foss-build:
```
diegolmello, line 396: IMO we should only trigger foss build in releases.
Otherwise we might be spending too many machines.

app/utils/log/index.js
```diff
@@ -1,6 +1,7 @@
 import { Client } from 'bugsnag-react-native';
 import analytics from '@react-native-firebase/analytics';
 import crashlytics from '@react-native-firebase/crashlytics';
+import { isFDroidBuild } from '../../constants/environment';
```
diegolmello, line 4: These libs shouldn't be loaded on F-Droid.

app/utils/log/index.js
```diff
@@ -42,7 +47,9 @@ export default (e) => {
 				}
```
diegolmello, line 47: Add bugsnag to this if

app/utils/review.js
```diff
@@ -92,7 +93,9 @@ class ReviewApp {
 		}
 		this.positiveEventCount += 1;
 		if (this.positiveEventCount === numberOfPositiveEvent) {
-			tryReview();
+			if (!isFDroidBuild) {
```
diegolmello, line 96: IMO you should test for the build and return right from the beginning of this function.

PR #2169, opened 2020-06-08
[NEW] Log events from Onboarding, NewServer, Login and Register screens

youssef-md:
### What was done in this PR:
* Creation of `logEvent` method to isolate to logic to improve future maintenance.
*  File `/utils/log/events.js` to define all the current events, to easily find them and to prevent typo if an event would be used in multiple places.

#### OnboardingView: 
* Join a workspace.
* Create new Workspace(with fail event).

#### NewServerView:
* Connect to typed workspace URL(with fail event).
* Join the open workspace.

#### LoginServices:
* Log all the coded OAuth options.

#### LoginView:
* Log the default login(with fail event).

#### RegisterView:
* Log the default sign up(with fail event).

Since Firebase Analytics Events can take a while to appear to the dashboard, I'll be using the [Analytics Debug View](https://firebase.google.com/docs/analytics/debugview) to test the event sending.

Here is a screenshot of Analytics Debug View with all the tracked events in this PR:
Screenshot from 2020-06-08 00-38-57

**user_engagement** and **screen_view** are default Firebase events, in this screenshot I selected the **join_open_workspace** event to show all the default payloads, the best one is **firebase_screen** which references the screen that the selected event was triggered.

PR #2190, opened 2020-06-14
[NEW] Log events from RoomsList, SideDrawer and Profile

youssef-md:
@RocketChat/ReactNative

Depends on #2169

- All events now start with the View responsible for that action.

#### Login/Register View:
- Changed all LOGIN_WITH_... to ENTER_WITH_...
- Enter with Apple(handle fail event);

#### RoomsListView:
- Toggle server dropdown;
- Add server;
- Change server;
- Navigate to `NewMessageView`;
- RoomsList search;
- Navigate to `DirectoryView`;
- Toggle sort dropdown;
- Sort by Name;
- Sort by Activity;
- Group by type;
- Group favorites;
- Unread on top;
- Go to room;
- Favorite/Unfavorite a channel (handle fail event);
- Read/Unread a channel (handle fail event);
- Hide channel (handle fail event);

#### DirectoryView:
- Show channels;
- Show users;

#### NewMessage:
- Chat with user;
- Create channel;
- Create group chat;
- Create discussion;

#### SelectedUsersView:
- Add user;
- Remove user;
- Create group(handle fail event);

#### CreateChannelView:
- Create a channel(handle fail event);
- Toggle type;
- Toggle read only;
- Toggle broadcast;
- Remove user;

#### CreateDiscussionView:
- Create a discussion(handle fail event);
- Select channel;
- Select users;

#### SibebarView:
- Navigate to Status;
- Navigate to Chats;
- Navigate to Profile;
- Navigate to Settings;
- Navigate to AdmPanel;

#### ProfileView:
- Pick profile avatar (handle fail event);
- Pick profile avatar with URL;
- Save profile avatar (handle fail event);
- Save profile changes (handle fail event);


app/views/SidebarView/index.js
```diff
@@ -180,21 +183,30 @@ class Sidebar extends Component {
 				<SidebarItem
 					text={I18n.t('Chats')}
 					left={<CustomIcon name='message' size={20} color={themes[theme].titleText} />}
-					onPress={() => this.sidebarNavigate('ChatsStackNavigator')}
+					onPress={() => {
+						logEvent(events.NAVIGATE_TO_ROOMSLIST);
```
diegolmello, line 187: All of these calls should be done inside of the function.
It's odd to add rules inside of the render function, since we already have a function to do it.

app/views/NewMessageView.js
```diff
@@ -133,11 +133,13 @@ class NewMessageView extends React.Component {
 	}
 
 	createChannel = () => {
+		logEvent(events.CREATE_CHANNEL_START);
```
diegolmello, line 136: It should identify the screen where the action started: `NEW_MESSAGE_CREATE_CHANNEL`

For example, we have tons of ways to create discussions and we want to know from where the action was triggered.
I saw on the upcoming PRs you're reusing actions, which is wrong.

app/views/SidebarView/index.js
youssef-md: I placed a logic to log the event inside `sidebarNavigate()` based on the passed route, but in this component `Navigation.navigate()` is being used twice, should I place the same logic there or change to use `sidebarNavigate()`?

youssef-md:
> * Change all events to identify the component/screen
> * Make sure you're not reusing events
> * Log trivial events like:
>   
>   * Open/close sort dropdown
>   * Open/close server dropdown

I've refactored the events and added events that I should have logged. It's way better now.

app/views/SidebarView/index.js
diegolmello: Check if it works and try to remove the other one.

youssef-md: It works fine using `this.sidebarNavigate()` instead of `Navigation.navigate()`

PR #2195, opened 2020-06-17
Audio-Enhancements: Expo-av audio recorder

regalstreak:
## Description
* Refactor MessageBox to accommodate recording button even when recording
* Implement audio recorder with `expo-av`


app/containers/MessageBox/index.js
```diff
@@ -818,63 +819,95 @@ class MessageBox extends Component {
 			returnKeyType: 'send'
 		} : {};
 
+		const recordAudio = showSend || !Message_AudioRecorderEnabled ? null : (
```
djorkaeffalexandre, line 822: I can't understand the main purpose of these changes, since, you can re-order the buttons inside the `MessageBox/buttons`. Why did you bring these logic here?

app/containers/MessageBox/index.js
```diff
@@ -818,63 +819,95 @@ class MessageBox extends Component {
...
+				theme={theme}
+				recordingCallback={this.recordingCallback}
+				onFinish={this.finishAudioMessage}
+			/>
+		);
+
+		let safeAreaViewStyle = null;
+		let commandsPreviewAndMentions = null;
+		let replyPreview = null;
+		let content = null;
```
djorkaeffalexandre, line 833: It's not a good practice, since you is removing the components and turning these into variables, it'll re-render all these componentes, even them not change.

app/containers/MessageBox/RightButtons.ios.js
```diff
@@ -1,26 +1,19 @@
...
-const RightButtons = React.memo(({
-	theme, showSend, submit, recordAudioMessage, recordAudioMessageEnabled
-}) => {
+const RightButtons = React.memo(({ theme, showSend, submit }) => {
 	if (showSend) {
 		return <SendButton theme={theme} onPress={submit} />;
 	}
-	if (recordAudioMessageEnabled) {
-		return <AudioButton theme={theme} onPress={recordAudioMessage} />;
-	}
```
djorkaeffalexandre, line 14: Why? This was already at the right of the screen.

app/containers/MessageBox/RightButtons.android.js
```diff
@@ -1,31 +1,21 @@
...
 		return <SendButton onPress={submit} theme={theme} />;
 	}
-	if (recordAudioMessageEnabled) {
-		return (
-			<>
-				<AudioButton onPress={recordAudioMessage} theme={theme} />
-				<ActionsButton onPress={showMessageBoxActions} theme={theme} />
-			</>
-		);
-	}
```
djorkaeffalexandre, line 19: Why not just change here to something like:
```
if (showSend) {
  return <SendButton />;
}
return (
  <>
    {!recording ? <ActionsButton /> : null}
    <AudioButton />
  </>
)
```

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+	});
+
+	/* Effects */
+	useEffect(() => {
+		recordingCallback(recordingStatus.isRecording);
+	}, [recordingStatus.isRecording]);
+
+	/* UI */
+	const recordingButton = (
+		<TouchableOpacity
```
djorkaeffalexandre, line 125: Why not `BordelessButton`?

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+		>
+			<CustomIcon name='mic' size={23} color={recordingStatus.isRecording ? themes[theme].focusedBackground : themes[theme].tintColor} />
+		</TouchableOpacity>
+	);
+
+	return (
+		<>
+			{recordingButton}
+		</>
+	);
```
djorkaeffalexandre, line 148: Why? You can `return` directly the `recordingButton`.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+			isRecording: false,
+			durationMillis: 0,
+			isDoneRecording: true
+		});
+		return onFinish && onFinish(false);
+	}
+};
+
+const RecordAudio = ({ theme, recordingCallback, onFinish }) => {
+	/* Refs */
```
djorkaeffalexandre, line 107: IMO these comments can be removed.
Comments should be relevant.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+		sampleRate: 22050,
+		bitRate: 128000,
+		outputFormat: Audio.RECORDING_OPTION_IOS_OUTPUT_FORMAT_MPEG4AAC,
+		numberOfChannels: 2
+	}
+};
+
+const startRecordingAudio = async(instance, setRecordingStatus) => {
+	try {
+		const permissions = await Audio.getPermissionsAsync();
```
djorkaeffalexandre, line 45: You don't need this logic, you can just call `await Audio.requestPermissionsAsync();` directly, if it was previously granted, the status returns granted and the prompt of permission not will be showed.  :)

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+			});
+
+			await instance.prepareToRecordAsync(RECORDING_SETTINGS);
+			await instance.startAsync();
+			activateKeepAwake();
+		} else {
+			await Audio.requestPermissionsAsync();
+		}
+	} catch (error) {
+		console.log('startrecordingerror', error);
```
djorkaeffalexandre, line 60: Remember to remove this.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+			type: 'audio/aac',
+			store: 'Uploads',
+			path: fileURI,
+			size: fileData.size
+		};
+
+		deactivateKeepAwake();
+		return onFinish && onFinish(fileInfo);
+	} catch (err) {
+		await Audio.setAudioModeAsync(playbackMode);
```
djorkaeffalexandre, line 94: Why you is doing this when an error happens?

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+		deactivateKeepAwake();
+		setRecordingStatus({
+			canRecord: true,
+			isRecording: false,
+			durationMillis: 0,
+			isDoneRecording: true
+		});
+		return onFinish && onFinish(false);
+	}
```
djorkaeffalexandre, line 103: You can call `deactivateKeepAwake` outside of the `try/catch` block.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,157 @@
...
+		const fileData = await RNFetchBlob.fs.stat(uriToPath(fileURI));
+
+		await Audio.setAudioModeAsync(playbackMode);
+
+		setRecordingStatus({
+			canRecord: true,
+			isRecording: false,
+			durationMillis: 0,
+			isDoneRecording: true
+		});
```
djorkaeffalexandre, line 80: Create a function that does this, so you don't need to duplicated this piece of code.

djorkaeffalexandre:
@regalstreak update this PR description to have only the main purpose from it self.
Try to focus your changes at this only at the use of `expo-av` at `Recording` component.

app/containers/MessageBox/RecordAudio.js
regalstreak: Was previously doing that. When I click on the button and it asks for permission on android, and after the permission is granted, even when I'm not touching anymore, it starts recording. Didn't want the audio to record when this happens.

app/containers/MessageBox/RecordAudio.js
regalstreak: Yes will do

app/containers/MessageBox/RecordAudio.js
regalstreak: It won't be run if no errors are caught as we're returning in the try block right?

app/containers/MessageBox/RecordAudio.js
regalstreak: If an error happens, expo-av will get stuck in recording mode and we would not be able to play audio

app/containers/MessageBox/RecordAudio.js
regalstreak: Have used `LongPressGestureHandler` from #2199 as we need those states, refer https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2200/files#diff-a254d347231b9ac3e1778f89da0eda28R216

app/containers/MessageBox/RecordAudio.js
regalstreak: Needed more `GestureHandlers` in future PRs

djorkaeffalexandre: Sure, but for now you can just return this directly? since maybe the other PRs can be merged at a considerable time after that.

app/containers/MessageBox/RecordAudio.js
djorkaeffalexandre: sure!

app/containers/MessageBox/RecordAudio.js
djorkaeffalexandre: Can you show me a GIF?

app/containers/MessageBox/RightButtons.ios.js
regalstreak: We don't need the `AudioButton` now. The `RecordAudio` component itself holds the UI and the Gestures

djorkaeffalexandre: sure

app/containers/MessageBox/RightButtons.android.js
regalstreak: One of the main reasons was to not pass the `recordingCallback` and the `recording` states through the `RightButtons` component.  Also, `finishAudioMessage` will have to be passed twice

app/containers/MessageBox/RecordAudio.js
regalstreak: MP4_20200619_225013

regalstreak: ezgif com-video-to-gif

app/containers/MessageBox/index.js
regalstreak: Related to https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2195/files/d2f5cffb24342c86009fc8513690d16f3ea123a0#r442894161, resolving

app/containers/MessageBox/RecordAudio.js
djorkaeffalexandre: Sure! I'll take a look when it is more done.
You can maintain with your logic. 😁

app/containers/MessageBox/RightButtons.android.js
djorkaeffalexandre: ok, you can get on with your logic.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+		if (recordingInstance.current) {
+			cancelAudioMessage(recordingInstance.current, setRecorderBusy);
+		}
+	}, []);
+
+	useEffect(() => {
+		recordingCallback(recordingStatus.isRecording);
+	}, [recordingStatus.isRecording]);
+
+	const recordingContent = recordingStatus.isRecording ? (
```
diegolmello, line 130: This is a bad pattern.
It's a huge `if` condition to attribute a constant and then returning it.
Why not
```
if (condition) {
  return ...
}

return ...
```
?

app/containers/MessageBox/index.js
```diff
@@ -818,62 +816,78 @@ class MessageBox extends Component {
...
+					close={replyCancel}
+					username={user.username}
+					replying={replying}
+					getCustomEmoji={getCustomEmoji}
+					theme={theme}
+				/>
+			</>
+		) : null;
+
+		const content = !recording ? (
```
diegolmello, line 842: Please be careful while refactoring.
You removed some styles.
You undo some of this code and change only change what you really need to change.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+		canRecord: false,
+		isRecording: false,
+		durationMillis: 0,
+		isDoneRecording: false
+	});
+	const [recorderBusy, setRecorderBusy] = useState(false);
+
+	useEffect(() => () => {
+		if (recordingInstance.current) {
+			cancelAudioMessage(recordingInstance.current, setRecorderBusy);
```
diegolmello, line 122: Passing instances back and forth messy here.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+	try {
+		await instance.stopAndUnloadAsync();
+	} catch (error) {
+		// Do nothing
+	}
+	deactivateKeepAwake();
+	setRecorderBusy(false);
+};
+
+const RecordAudio = ({ theme, recordingCallback, onFinish }) => {
```
diegolmello, line 109: Are you sure it needs to be a hook?
Don't write hooks just because of the hype.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+	const recordingInstance = useRef(null);
+
+	const [recordingStatus, setRecordingStatus] = useState({
```
diegolmello, line 112: Are you sure `setRecordingStatus` is always going to update these 4 values?
Otherwise, you're going to mess your state.

diegolmello, line 112: Are you sure `setRecordingStatus` is always going to update these 4 values?
Otherwise, you're going to mess your state.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+		canRecord: false,
+		isRecording: false,
+		durationMillis: 0,
+		isDoneRecording: false
+	});
+	const [recorderBusy, setRecorderBusy] = useState(false);
```
diegolmello, line 118: Keep your state consistent.
You wrote `isRecording`, `isDoneRecording` and `recorderBusy`.
All of them are boolean.
Don't you think the proper name is `isRecorderBusy`?

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+				<Text
+					key='currentTime'
+					style={[styles.recordingCancelText, { color: themes[theme].titleText }]}
+				>
+					{_formatTime(Math.floor(recordingStatus.durationMillis / 1000))}
+				</Text>
+			</View>
+			<View style={styles.recordingContentFinish}>
+				<BorderlessButton
+					onPress={() => {
```
diegolmello, line 158: Can you remove this inline function?

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,201 @@
...
+
+const _formatTime = function(seconds) {
+	let minutes = Math.floor(seconds / 60);
+	seconds %= 60;
+	if (minutes < 10) { minutes = `0${ minutes }`; }
+	if (seconds < 10) { seconds = `0${ seconds }`; }
+	return `${ minutes }:${ seconds }`;
+};
+
+const startRecordingAudio = async(instance, setRecordingStatus, setRecorderBusy) => {
```
diegolmello, line 51: I get it you're trying to make abstractions here, but think on the next dev who is going to do maintenance here.
`startRecordingAudio` receives an `instance` of what?
I have to read this whole file in order to be sure it needs to receive an instance of `new Audio.Recording()`.
Why not making this whole code a React.Component and make it a class function?

app/containers/message/Audio.js
```diff
@@ -225,6 +224,7 @@ class MessageAudio extends React.Component {
 			if (paused) {
 				await this.sound.pauseAsync();
 			} else {
+				await Audio.setAudioModeAsync(mode);
```
diegolmello, line 227: Is this change needed?

regalstreak: Yes, `setAudioModeAsync` needs to be called everytime we need to play audio

app/containers/MessageBox/index.js
```diff
@@ -818,62 +816,78 @@ class MessageBox extends Component {
...
+					showSend={showSend}
+					submit={this.submit}
+					showMessageBoxActions={this.showMessageBoxActions}
+				/>
+			</>
+		) : null;
+
+		return (
+			<>
+				{commandsReplyMentions}
```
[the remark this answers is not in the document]
regalstreak: The `RecordAudio` component has to be in the below `View`. What do you think could be a better way to do that?

app/containers/MessageBox/RecordAudio.js
regalstreak: How could I improve this? Move this inside the function component?

app/containers/MessageBox/RecordAudio.js
regalstreak: Wouldn't renaming the variable or moving the function into the component be a better option? Is it really necessary to make it a class component as we're not using lifecycle hooks which I think could be one of the main reasons to make the switch

app/containers/MessageBox/RecordAudio.js
regalstreak: Similar to https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2195/files#r443857101, resolving

app/containers/MessageBox/RecordAudio.js
regalstreak: Yes, `recordingStatus` is essentially a dictionary with these key-value pairs, as set by `expo-av`

regalstreak: Discussed here https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2195/files#r443854448

app/containers/MessageBox/RecordAudio.js
diegolmello: That's not the only reason to use class components.
It'd make your code much more readable if those function could access a class variable: `this.recording = new Audio.Recording()`

This way you wouldn't be relying on passing instances as params to functions.

app/containers/MessageBox/index.js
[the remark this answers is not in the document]
regalstreak: Actually @diegolmello, if we don't do this, we might have to pass `recordingCallback` to `RightButtons` and pass the state back here. Had a discussion about the same above too https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2195#discussion_r442894161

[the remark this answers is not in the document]
regalstreak: Have moved all Recording UI to `RecordAudio`, including the button

app/containers/MessageBox/index.js
```diff
@@ -818,61 +816,83 @@ class MessageBox extends Component {
...
+					placeholder={I18n.t('New_Message')}
+					onChangeText={this.onChangeText}
+					underlineColorAndroid='transparent'
+					defaultValue=''
+					multiline
+					testID='messagebox-input'
+					theme={theme}
+					{...isAndroidTablet}
+				/>
+
```
djorkaeffalexandre, line 871: remove this blank line

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,227 @@
...
+
+	startRecordingAudio = async() => {
+		const { isRecorderBusy } = this.state;
+
+		if (!isRecorderBusy) {
+			this.setState({ isRecorderBusy: true });
+			try {
+				const permissions = await Audio.getPermissionsAsync();
+
+				if (permissions.status === 'granted') {
```
djorkaeffalexandre, line 94: Can you try to extract this permission logic to other function?

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,227 @@
...
+					recording.setOnRecordingStatusUpdate((status) => {
+						this.setState({
+							isRecording: status.isRecording,
+							recordingDurationMillis: status.durationMillis
+						});
+					});
+
+					this.recording = recording;
+
+					await this.recording.startAsync();
```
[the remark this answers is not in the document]
regalstreak: Thought the same but that way was recommended in the example, maybe to prepare the instance first properly and then set it?

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,227 @@
...
+						style={styles.actionButton}
+					>
+						<CustomIcon
+							size={22}
+							color={themes[theme].dangerColor}
+							name='Cross'
+						/>
+					</BorderlessButton>
+					<Text
+						key='currentTime'
```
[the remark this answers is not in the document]
regalstreak: `key` was from the old component

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,227 @@
...
+export default class RecordAudio extends React.PureComponent {
+	static propTypes = {
+		theme: PropTypes.string,
+		recordingCallback: PropTypes.func,
+		onFinish: PropTypes.func
+	}
+
+	constructor(props) {
+		super(props);
+		this.recording = null;
```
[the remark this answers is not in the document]
regalstreak: Was recommended in the example. Will remove

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,227 @@
...
+		this.state = {
+			isRecording: false,
+			isRecorderBusy: false,
```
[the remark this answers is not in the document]
regalstreak: On Android, the recorder will crash if we don't use a busy state. Try to click on a discussion and then immediately spamming the button. Was trying to fix this for like 2-3 days. The example used `isLoading` and I tried something similar to make it work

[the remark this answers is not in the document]
djorkaeffalexandre: What about a class variable like `this.isRecorderBusy`, wouldn't it do for that?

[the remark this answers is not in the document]
regalstreak: Should work. Will try

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,227 @@
...
+
+	finishRecordingAudio = async() => {
+		const { isRecorderBusy } = this.state;
+
+		if (!isRecorderBusy) {
+			const { onFinish } = this.props;
+
+			this.setState({ isRecorderBusy: true });
+			try {
+				await this.recording.stopAndUnloadAsync();
```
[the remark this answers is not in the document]
regalstreak: What did you mean by this?

app/containers/MessageBox/RecordAudio.js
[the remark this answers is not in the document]
regalstreak: Removing it

app/containers/MessageBox/RecordAudio.js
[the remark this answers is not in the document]
djorkaeffalexandre: top of the `await this.recording.stopAndUnloadAsync()`, to simulate a fail at this function.

app/containers/MessageBox/RecordAudio.js
```diff
@@ -0,0 +1,215 @@
...
+			isRecording: status.isRecording,
+			recordingDurationMillis: status.durationMillis
+		});
+	}
+
+	startRecordingAudio = async() => {
+		if (!this.isRecorderBusy) {
+			this.isRecorderBusy = true;
+			try {
+				if (await this.isRecordingPermissionGranted()) {
```
djorkaeffalexandre, line 101: no use `await` inner a conditional.
My main idea was do something like that:
```javascript
isRecordingPermissionGranted = () => {
  try {
    const permission = await Audio.getPermissionsAsync();
    if (permission.status === 'granted') {
      return true;
    }
    await Audio.requestPermissionsAsync();
  } catch {
    // Do nothing
  }
  return false;
}
```
and here:
```javascript
const canRecord = await this.isRecordingPermissionGranted();
if (canRecord) {
...
}
```

app/containers/MessageBox/RecordAudio.js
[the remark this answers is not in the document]
diegolmello: @regalstreak If I understood correctly, if this line throws an error, we'll never stop recording.
You can test this bug by throwing an error before `await this.recording.stopAndUnloadAsync();`.
`catch` should prevent it.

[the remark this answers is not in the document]
regalstreak: So you mean that `isRecorderBusy` will remain true always? I'll have to handle it too!

PR #2200, opened 2020-06-18
Audio-Enhancements: Audio recorder persist gesture

regalstreak:
* Persist audio recording when swiped up
* Button grow animation
* Depends on #2199 #2195

PR #2199, opened 2020-06-18
Audio-Enhancements: Audio recorder cancel gesture

regalstreak:
* Swipe left to cancel recording, along with basic slide to cancel animation
* Depends on #2195

PR #2201, opened 2020-06-18
Audio-Enhancements: Audio recorder tooltip and min duration

regalstreak:
* Show recording tooltip
* Set minimum recording duration
* Depends on #2199 #2195 #2200


regalstreak:
Moved to #2199

PR #2213, opened 2020-06-23
[NEW] Push notification data privacy

djorkaeffalexandre:
@RocketChat/ReactNative

Depends on https://github.com/RocketChat/Rocket.Chat/pull/18373

## Last updated
Now we're sending a fallback message from servers, so, when the message can't be fetched from the server we'll show a notification with a generic message like: `Someone send you a message`.
This fallback notification can't be replied or grouped, this is how it looks like on android devices (Android 9.0)
Screen Shot 2020-07-24 at 16 42 23
When someone click on this fallback notification will be redirected to the server where the message came from, since we have the host on the notification payload.

## Main purpose:
Push notifications should be sent to Apple/Google push servers to be sent to devices, this can be a privacy fault, so, we'll remove the notification content that is send to Apple/Google push servers and will send only the messageId, we can fetch the message content before show this at your device, enjoying the protection of your server.

## Testplan:
- [ ] Check if everything is working ok at servers that doesn't have this capability.
- [ ] Check if grouped notifications are working properly.
- [ ] Check if reply notifications is working fine.

### Testing with a +3.5.0 Rocket.Chat server
1. Go to Admin Panel
2. Push
3. Privacy section
4. Toggle `Fetch full message content from the server on receipt`
Screen Shot 2020-07-23 at 10 13 27

### Test on iOS without a server
1. clone https://github.com/djorkaeffalexandre/node-push-notification/tree/push-privacy
2. set your own environment
3. run `yarn start`


AgentTNT:
Is there a server side change/feature that needs to go along with this to properly submit the new payload?

djorkaeffalexandre:
@AgentTNT, we will work at the server side soon, I'll link this PR with the server PR when opened.
Stay tuned.
Thanks.

AgentTNT:
Sounds good, thanks! I hope this isn't an Enterprise only feature once landed on the server side. :/

PR #2247, opened 2020-07-06
[NEW] Encrypt user credentials and preferences

djorkaeffalexandre:
@RocketChat/ReactNative

### Motivation
Currently, we're saving user credentials (userId/userToken) using `UserDefaults` and `SharedPreferences`, these are the native modules provided by iOS and Android for `key–value` store.
[MMKV](https://github.com/Tencent/MMKV) provides an API to `key–value` store with the capability to encrypt this store ([iOS](https://github.com/Tencent/MMKV/wiki/iOS_advance#encryption)/[Android](https://github.com/Tencent/MMKV/wiki/android_advance#encryption)), so, we'll use them to store and encrypt user credentials at Rocket.Chat.
[MMKV](https://github.com/Tencent/MMKV) promise to be more fast than the default `key-value` store of iOS and Android, you can check their benchmarks here [iOS](https://github.com/Tencent/MMKV/wiki/iOS_benchmark)/[Android](https://github.com/Tencent/MMKV/wiki/android_benchmark), we hope that will improve our perfomance too.

### Test plan
- [x] Check if data is encrypted
- [x] Credentials migration from our latest app release
- [x] Reply notifications (Android/iOS)
- [x] SSL Pinning (iOS)
- [x] Share Extension (Android/iOS)

### Docs
We should add a section about the new configuration key `KeychainGroup` at our whitelabel docs.
`KeychainGroup` should have the `$(AppIdentifierPrefix)` at the start always.
It should be changed at the `RocketChatRN/Info.plist` and `ShareRocketChatRN/Info.plist`.
`KeychainGroup` should should have the same string of [Keychain Access Groups](https://developer.apple.com/documentation/bundleresources/entitlements/keychain-access-groups) capatibility.
[Keychain Access Groups](https://developer.apple.com/documentation/bundleresources/entitlements/keychain-access-groups) capatibility should be changes at Main App target and Share Extension and should have the same value at both.
More info about this can be found [here](https://developer.apple.com/documentation/security/keychain_services/keychain_items/sharing_access_to_keychain_items_among_a_collection_of_apps).


app/sagas/init.js
```diff
@@ -29,71 +22,20 @@ export const initLocalSettings = function* initLocalSettings() {
...
-			try {
-				yield AsyncStorage.setItem('hasMigration', '1');
-			} catch (e) {
-				log(e);
-			}
+		let token; let server;
+		try {
+			({ token, server } = yield all({
+				token: MMKV.getStringAsync(RocketChat.TOKEN_KEY),
+				server: MMKV.getStringAsync('currentServer')
```
diegolmello, line 29: I think it's time for us to refactor this `currentServer` key already.
`RocketChat.CURRENT_SERVER`
WDYT?

app/sagas/login.js
```diff
@@ -127,8 +127,8 @@ const handleLoginSuccess = function* handleLoginSuccess({ user }) {
 			}
 		});
 
-		yield RNUserDefaults.set(`${ RocketChat.TOKEN_KEY }-${ server }`, user.id);
-		yield RNUserDefaults.set(`${ RocketChat.TOKEN_KEY }-${ user.id }`, user.token);
+		yield MMKV.setStringAsync(`${ RocketChat.TOKEN_KEY }-${ server }`, user.id);
```
diegolmello, line 130: Same as `currentServer`.
It could be a function instead of having to know the pattern.
But both could be on a separate PR.
I'm not sure.

djorkaeffalexandre: I'll do this on a separate PR, soon, create functions on UserPreferences specific to get/set user credentials based on server :)

app/sagas/init.js
```diff
@@ -29,71 +22,15 @@ export const initLocalSettings = function* initLocalSettings() {
...
-
-			try {
-				yield AsyncStorage.setItem('hasMigration', '1');
-			} catch (e) {
-				log(e);
-			}
-		}
-
 		if (!token || !server) {
 			yield all([
```
diegolmello, line 31: MMKV is throwing an error if you try to remove something that isn't there.

#### Steps:
- open a fresh app
- go to open
- close and open again
- it crashes

image

app/index.js
```diff
@@ -112,6 +113,23 @@ export default class Root extends React.Component {
...
 	}
 
+	// It should run only once
+	encryptMigratedData = async() => {
+		try {
+			const encryptMigration = await UserPreferences.getBoolAsync('encryptMigration');
+
+			if (!encryptMigration) {
+				// Encrypt the migrated data
+				await UserPreferences.encryption.encrypt();
```
diegolmello, line 123: Can you make this call inside `UserPreferences` and export a new function called `encrypt`?

app/sagas/init.js
```diff
@@ -29,71 +22,15 @@ export const initLocalSettings = function* initLocalSettings() {
...
-		let hasMigration;
-		if (isIOS) {
-			hasMigration = yield AsyncStorage.getItem('hasMigration');
-		}
-
-		let { token, server } = yield all({
-			token: RNUserDefaults.get(RocketChat.TOKEN_KEY),
-			server: RNUserDefaults.get('currentServer')
+		const { token, server } = yield all({
+			token: UserPreferences.getStringAsync(RocketChat.TOKEN_KEY),
```
diegolmello, line 26: I'm not sure if it's related to the error below, but sometimes the app never resolves this function and the app hangs on splash screen.

app/sagas/init.js
djorkaeffalexandre: I made some changes to the `react-native-mmkv-storage` that fixes this, I've tested on 4 different simulators using this same steps and it's not crashing anymore.
Let's see if it's working fine when you test, if yes, I'll open a PR to `react-native-mmkv-storage`. :)
`react-native-mmkv-storage` have an indexer, and it has an array with all the keys of each type (string/number/map), sometimes it's doing something like that:
#### Pseudo code
```haxe
let a = [1];
a.pop();
mmkv.save(a);
```
and it's crashing mmkv since we're trying to save a null/empty array, and it thrown a lot of errors like that you have found.

app/sagas/init.js
djorkaeffalexandre: Yeah, after my changes, it doesn't happen again for me.

PR #2249, opened 2020-07-07
[IMPROVEMENT] Native sort and limit queries

diegolmello:
@RocketChat/ReactNative

This PR makes usage of new WatermelonDB's experimental APIs for `sort`, `take` and `skip`.

### Comparison
- 30~40 MB less memory consumption loading a RoomView with 3k messages on local db
- 10 MB less memory loading RoomsListView with 500 channels
- 8 MB less memory on ShareListView with 500 channels

#### A
- RoomsListView should load 50 channels at time
- When grouping (by type or whatever), it should load all channels and sort native

#### B
- RoomView should load 50 message at time

#### C
- ShareListView should load 50 channels at time
- Search should by limited to 50 channels


app/views/ShareListView/index.js
```diff
@@ -7,7 +7,7 @@ import ShareExtension from 'rn-extensions-share';
 import * as FileSystem from 'expo-file-system';
 import { connect } from 'react-redux';
 import * as mime from 'react-native-mime-types';
-import { isEqual, orderBy } from 'lodash';
+import { isEqual } from 'lodash';
```
djorkaeffalexandre, line 10: WDYT about use `isEqual` from `react-fast-compare`, since this is already installed

app/presentation/RoomItem/index.js
```diff
@@ -16,82 +16,78 @@ import { themes } from '../../constants/colors';
 export { ROW_HEIGHT };
 
 const attrs = [
-	'name',
-	'unread',
-	'userMentions',
-	'showLastMessage',
```
djorkaeffalexandre, line 22: We're removing `showLastMessage` because it'll change only when the `getSettings` run (only once when the app open), right?

app/views/RoomsListView/index.js
```diff
@@ -800,40 +815,30 @@ class RoomsListView extends React.Component {
...
-				userMentions={item.userMentions}
-				isRead={this.getIsRead(item)}
-				favorite={item.f}
-				avatar={this.getRoomAvatar(item)}
-				lastMessage={item.lastMessage}
-				name={this.getRoomTitle(item)}
-				_updatedAt={item.roomUpdatedAt}
-				key={item._id}
 				id={id}
+				type={item.t}
```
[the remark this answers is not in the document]
diegolmello: It's needed on `mapStateToProps` :(

PR #2250, opened 2020-07-08
Run tests in docker

Fishbowler:
@RocketChat/ReactNative

Now runs tests in Docker rather than against an existing Rocket.Chat server, using the existing Docker Compose file for Rocket.Chat plus some overrides, and populated using Mongo's inbuilt init scripts for their docker image.

Also includes:
* Ability to run the "demo" RC env without running the tests (hattip to conversations on https://open.rocket.chat/)
* Detox upgrade (although not to v17 as there are breaking changes)
* Some test fixes & improvements

image


e2e/tests/assorted/02-broadcast.spec.js
```diff
@@ -28,8 +28,9 @@ describe('Broadcast room', () => {
 		await sleep(1000);
 		await waitFor(element(by.id('create-channel-view'))).toExist().withTimeout(5000);
 		await element(by.id('create-channel-name')).replaceText(`broadcast${ data.random }`);
-		await sleep(1000);
-		await element(by.id('create-channel-broadcast')).tap();
+		await sleep(2000);
+		await element(by.id('create-channel-broadcast')).tap(); //FLAKY
+		await expect(element(by.id('create-channel-broadcast'))).toHaveValue('1');
```
Fishbowler, line 33: There's a dedicated `expect` in here since the action in the line above fails for me in about 10% of runs. (This is flipping the "broadcast" toggle for the channel whilst creating it)

I dedicated some time to fixing it, and failed, so resorted to failing earlier and making the reason more obvious.

diegolmello: It's ok. Detox (and most of the e2e testing libs as well) is sadly not completely reliable.

e2e/docker/docker-entrypoint-initdb.d/users.js
```diff
@@ -0,0 +1,4 @@
...
+db.getCollection("users").insert({"_id":"nM6vXyDLGGzSPsLNy","createdAt":new Date(1584022531608),"services":{"password":{"bcrypt":"$2b$10$fXL9kVkWeKA7TbP2skwau.Xu3V52q1x/YPfZQ4oYjHhCRPdXwZOQ6"},"email":{"verificationTokens":[{"token":"YadCnp4E2o8lD8ZEyzn320qjoW9QEIGSU1jsaR2840J","address":"admin@example.com","when":new Date(1584022531675)}]},"resume":{"loginTokens":[]}},"emails":[{"address":"admin@example.com","verified":false}],"type":"user","status":"offline","active":true,"_updatedAt":new Date(1589468245023),"roles":["admin"],"name":"Admin","lastLogin":new Date(1589465335818),"statusConnection":"offline","username":"admin","utcOffset":NumberInt(1),"statusDefault":"online","statusText":"Lunch"});
```
diegolmello, line 3: What's the default password for the admin here?

e2e/data.js
```diff
@@ -1,16 +1,16 @@
 const random = require('./helpers/random');
 const value = random(20);
 const data = {
-    server: 'https://ilarion.rocket.chat',
+    server: 'http://127.0.0.1:3000',
```
diegolmello, line 4: This change makes tests to fail on CircleCI.
Can you run the docker on CI tests as well?

e2e/docker/docker-entrypoint-initdb.d/users.js
Fishbowler: `password`. I'll add it as a comment to the file, unless you think there's somewhere better?

diegolmello: On the file is ok and only for the admin, so anyone can easily change some setting on this awesome docker :)

PR #2272, opened 2020-07-09
[CHORE] Improve reply notification code (iOS)

djorkaeffalexandre:
@RocketChat/ReactNative

I studied `Method Swizzling` for some period and it looks like a great solution to turn some of our patches better, turn these patches a real part of our code.
Method swizzling is the process of changing the implementation of an existing selector.
Using method swizzling we can improve/change some native implementation that was done by some library, overriding methods.
Bringing the code to our main app, instead of use a patch, give us more maintainability, and turn the code more concisely; we can stop of duplicate some native codes at iOS side. :)
At some moment I'll try to move all the iOS land patch of the React Native core libraries to Method Swizzling.

A really good explanation about Method swizzling:
https://nshipster.com/method-swizzling/


diegolmello:
Let's hold this until we update [react-native-notifications](https://github.com/wix/react-native-notifications/releases), which have a couple of breaking changes.

djorkaeffalexandre:
refer to #2476

PR #2281, opened 2020-07-12
[FIX] Focusing Text Input for writing description in ShareView after uploading an attachment causes chat messages to be hidden under MessageBox in RoomView

### LINKED ISSUE #2276: (BUG) Sending Image/File with description is hidden on iOS upon sent
- Your Rocket.Chat Experimental app version: ####
4.8.0
- Your Rocket.Chat server version: ####
3.4.1
- Device (or Simulator) you're running with: ####
iPhone 11 Pro Max device

As you can see in the picture below, we downloaded the official RocketChat from iOS store with the latest version 4.8.0 and when we upload a new photo with description, the description became hidden below the MessageBox.

We have spent a lot of time trying to debug this and realised that the KeyboardAccessoryView in
app/containers/MessageBox/index.js has a param "requiresSameParentToManageScrollView" and this seems to cause
issue as it modifies the same flatlist that displays the messages in RoomView. Not sure whether resetTracking() is for what purpose but that did not helps in setting the correct content offset for the flatlist that contains the messages.

We also noticed that as long as after we uploaded an image/file and tap on the MessageBox input and even we close the ShareView without uploading any files, the flatlist displaying the messages will have the bottom hidden under the MessageBox.

This bug is introduced after the [NEW] Send multiple attachments (#2162) commit which changes to use MessageBox component.

IMG_2089

hengsok:
@RocketChat/ReactNative

Closes #2276

As long as we tap on the Text Input of MessageBox in ShareView after we have uploaded an attachment, no matter we send the attachment or not, the chat messages in RoomView will become hidden under the MessageBox component on iOS irregularly on different devices. On my devices, it seems to happen every time. 

This problem was traced all the way to the module "react-native-keyboard-tracking-view".
Please see Issue https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/2276.

IMG_2089

### DIFF
--- a/app/containers/MessageBox/styles.js
+++ b/app/containers/MessageBox/styles.js
@@ -15,7 +15,7 @@ export default StyleSheet.create({
 	},
 	composer: {
 		flexDirection: 'column',
-		borderTopWidth: StyleSheet.hairlineWidth
+		borderTopWidth: 1
 	},
 	textArea: {
 		flexDirection: 'row',


Fishbowler:
I'd seen this issue whilst working on the e2e tests, and it was really getting in my way. I've tested this change and it's fixed! Thanks so much!

diegolmello:
Thanks for your PR, @hengsok! :)

hengsok:
@Fishbowler @diegolmello you're welcome!

PR #2282, opened 2020-07-13
[NEW] Sign in with Apple

diegolmello:
@RocketChat/ReactNative


app/containers/LoginServices.js
```diff
@@ -285,9 +303,26 @@ class LoginServices extends React.PureComponent {
...
 			}
+			case 'apple': {
+				onPress = () => this.onPressAppleLogin();
+				break;
+			}
 			default:
 				break;
 		}
+
+		if (name === 'apple') {
```
diegolmello, line 314: Hide on Android

djorkaeffalexandre: I'm adding `apple` as a supported service only if it's an iOS here: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2282/files#diff-6e616dfd576f12234f426a580b065f20R1080

PR #2294, opened 2020-07-16
[FIX] MessagesView title not working

EzequielDeOliveira:
@RocketChat/ReactNative

This pr fix the way to set the title in RoomActionsView items that are wrong after update of react navigation.

**Problem**
Screenshot from 2020-07-16 00-02-59 *Files*

Screenshot from 2020-07-16 00-03-51 *Pinned*

**Resolved**

Screenshot from 2020-07-16 00-05-34

Screenshot from 2020-07-16 00-05-51


app/views/MessagesView/index.js
```diff
@@ -46,6 +42,7 @@ class MessagesView extends React.Component {
 
 	componentDidMount() {
 		this.load();
+		this.setHeader();
```
djorkaeffalexandre, line 45: Should be called on constructor or maybe can cause a header blink.

app/stacks/MasterDetailStack/index.js
```diff
@@ -139,7 +139,6 @@ const ModalStackNavigator = React.memo(({ navigation }) => {
 				<ModalStack.Screen
 					name='MessagesView'
 					component={MessagesView}
-					options={MessagesView.navigationOptions}
```
djorkaeffalexandre, line 142: Why did you remove this from the tablet mode but keep it on the regular inside stack?

app/views/MessagesView/index.js
EzequielDeOliveira: Right, I called this in the constructor now.

app/stacks/MasterDetailStack/index.js
EzequielDeOliveira: Removed.

PR #2300, opened 2020-07-16
[IMPROVEMENT] Clear image cache when clear cache

djorkaeffalexandre:
@RocketChat/ReactNative

~Let's stay tuned on https://github.com/DylanVann/react-native-fast-image/pull/425 to see if it get merged soon and update the library when got it.~
We're using our own [fork](https://github.com/RocketChat/react-native-fast-image/) now 😅 .

## Test plan
1. Clear local cache/data on Settings Screen.
2. All images should be loaded again.


patches/react-native-fast-image+8.1.5.patch
```diff
@@ -0,0 +1,174 @@
+diff --git a/node_modules/react-native-fast-image/android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java b/node_modules/react-native-fast-image/android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
```
diegolmello, line 1: Too many code on a patch and the PR is from early 2019.
IMO we should do a fork.

djorkaeffalexandre: Agree

PR #2311, opened 2020-07-17
[FIX] Avatar cache invalidation

### LINKED ISSUE #441: Avatar images doesn't refresh

### LINKED ISSUE #1583: Avatar changes as username is changed in Profile View

djorkaeffalexandre:
@RocketChat/ReactNative

### Screenshots

Closes #441.
Closes #1583.

Screen Shot 2020-08-05 at 14 52 23
Screen Shot 2020-08-05 at 14 52 19
Screen Shot 2020-08-05 at 14 52 14
Screen Shot 2020-08-05 at 14 52 10


app/containers/Avatar/index.js
```diff
@@ -0,0 +1,57 @@
...
+
+	const subscribeAvatar = async() => {
+		if (type === 'd') {
+			const db = database.active;
+			const usersCollection = db.collections.get('users');
+			try {
+				const [userRecord] = await usersCollection.query(Q.where('username', text)).fetch();
+				if (userRecord) {
+					const observable = userRecord.observe();
+					observable.subscribe((u) => {
```
djorkaeffalexandre, line 21: How we can handle the unsubscribe here?

app/presentation/RoomItem/index.js
```diff
@@ -49,14 +51,35 @@ const RoomItemContainer = React.memo(({
...
 	}, [connected]);
 
+	const subscribeUser = async() => {
+		const db = database.active;
+		const usersCollection = db.collections.get('users');
+		try {
+			const [user] = await usersCollection.query(Q.where('username', Q.eq(username))).fetch();
+			if (user) {
+				const observable = user.observe();
+				observable.subscribe((u) => {
```
djorkaeffalexandre, line 69: unsubscribe?

app/containers/Avatar/index.js
```diff
@@ -0,0 +1,59 @@
...
+import database from '../../lib/database';
+import { getUserSelector } from '../../selectors/login';
+import Component from './Avatar';
+
+const Avatar = React.memo(({ type, text, ...props }) => {
+	const [avatarETag, setAvatarETag] = useState();
+
+	useEffect(() => {
+		let subscription;
+		(async() => {
```
djorkaeffalexandre, line 15: Open to suggestions about how we can handle the unsubscribe without IIFE

app/containers/message/index.js
```diff
@@ -69,14 +70,29 @@ class MessageContainer extends React.Component {
...
 			const observable = item.observe();
 			this.subscription = observable.subscribe(() => {
 				this.forceUpdate();
 			});
 		}
+
+		const db = database.active;
+		const usersCollection = db.collections.get('users');
+		try {
+			const user = await usersCollection.find(item.u._id);
```
djorkaeffalexandre, line 87: I guess we can do a relationship between messages and users to don't need this find, but I don't know if it's better for us right now.
WDYT?

app/presentation/RoomItem/index.js
```diff
@@ -57,13 +60,33 @@ const RoomItemContainer = React.memo(({
 	}, [connected]);
 
 	useEffect(() => {
+		let userSub;
+		if (item.t === 'd') {
+			(async() => {
+				const db = database.active;
+				const usersCollection = db.collections.get('users');
+				try {
+					const [user] = await usersCollection.query(Q.where('username', Q.eq(username))).fetch();
```
djorkaeffalexandre, line 69: WDYT about do a relationship between sub/user using username as a FK?

app/lib/database/model/Users.js
```diff
@@ -0,0 +1,14 @@
+import { Model } from '@nozbe/watermelondb';
+import { field } from '@nozbe/watermelondb/decorators';
+
+export default class User extends Model {
```
djorkaeffalexandre, line 4: I've created this file as `Users`, since we already have a file called `User` on our models that refers to logged user.
I'm thinking about create a new folder to separate Models of the Default and Servers database.
WDYT?

app/containers/Avatar/index.js
diegolmello: There's an issue here that `subscription` won't have a `unsubscribe` method even using IIFE.
Correct me if I'm wrong, but as this unnamed function doesn't `await`, return is called right away.

This could be rewritten as a normal function and being called right after, but the problem would remain.
```js
// taken from https://github.com/facebook/react/issues/14326#issuecomment-441680293
useEffect(() => {
  async function fetchMyAPI() {
    let url = 'http://something';
    let config = {};
    const response = await myFetch(url);
    console.log(response);
  }  

  fetchMyAPI();
}, []);
```

The issue here is that we need `subscription` to return a value on `useEffect`, so the unmount cleans everything.
If you switch to React.Component, this issue is gone easily :)

app/lib/database/model/Users.js
diegolmello: I agree.

app/containers/message/index.js
diegolmello: Let's stick with this code as it doesn't hurt our users experience.

app/views/RoomView/Header/LeftButtons.js
```diff
@@ -32,30 +32,22 @@ const LeftButtons = React.memo(({
...
-				style={styles.avatar}
-				userId={userId}
-				token={token}
-				onPress={onPress}
-			/>
-		);
-	}
-	return null;
+
+	return (
```
diegolmello, line 36: IMO we should keep returning `null` here instead of relying to `Avatar` the responsibility

shaheem-khanzada:
@diegolmello whats the status for this branch

raza2022:
@djorkaeffalexandre @diegolmello any update on it, as I badly need this fix, let me also know if wanna any contribution in proposed solution/improvement

djorkaeffalexandre:
Hey guys, it's planned to be part of 4.12 release. :)
Thanks.

diegolmello (approved):
LGTM

PR #2335, opened 2020-07-23
[FIX] Minor i18n issues

EzequielDeOliveira:
## New labels translate to ptBr

image

image*Correct header title setting*

image*Correct header title setting*

@RocketChat/ReactNative


app/views/NotificationPreferencesView/index.js
```diff
@@ -171,6 +167,10 @@ class NotificationPreferencesView extends React.Component {
 	}
 
 	componentDidMount() {
+		const { navigation } = this.props;
```
diegolmello, line 170: Nice catch, man!
It looks like [this pattern](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2335/files#diff-6aa5f28f1b731385a70cab431cdecf1eL142-L145) doesn't work well with react-navigation v5 + i18n, because it's a static and it's never gonna be updated, because it's object and not a function 🤦‍♂️ 

There's an easier fix than the one you've proposed.
Check this out:

###### Code below won't update on language changed, because it's an object
```js
static navigationOptions = {	
	title: I18n.t('Notification_Preferences')	
}
```

###### Code below works nicely, because it's a function now :)
```js
static navigationOptions = () => ({	
	title: I18n.t('Notification_Preferences')	
})
```

I see there're other places with similar mistakes of mine, like DefaultBrowserView and CreateChannelView.
Can you fix them?
Thanks!

EzequielDeOliveira: Yes I can. I'll fix in the other places.
Thanks!

EzequielDeOliveira: I fixed this headers to update when the language change using the right pattern :)

PR #2349, opened 2020-07-28
[WIP][IMPROVEMENT] Enforce HTTPS on server URL

diegolmello:
## Proposed changes

In the current state of the app, we don't enforce secure connections, but that causes more harm than good, making users to believe it's an issue on the app when it's really a misconfiguration.

- Apple prevents insecure connections for a while now: https://developer.apple.com/documentation/security/preventing_insecure_network_connections
We do have the option to just disable this security feature, but having a less secure app is the opposite of we're trying to build here.
- Android prevents insecure connections by default since Android 9: https://developer.android.com/training/articles/security-config#CleartextTrafficPermitted

This PR starts to enforce secure connections on servers.

## Types of changes

## Checklist

### DIFF
--- a/app/views/NewServerView.js
+++ b/app/views/NewServerView.js
@@ -133,8 +133,8 @@ class NewServerView extends React.Component {
 	handleNewServerEvent = (event) => {
 		let { server } = event;
 		const { connectServer } = this.props;
-		this.setState({ text: server });
 		server = this.completeUrl(server);
+		this.setState({ text: server });
 		connectServer(server);
 	}
 
@@ -163,6 +163,7 @@ class NewServerView extends React.Component {
 		if (text) {
 			Keyboard.dismiss();
 			const server = this.completeUrl(text);
+			this.setState({ text: server });
 			await this.basicAuth(server, text);
 			connectServer(server, cert);
 		}
@@ -230,6 +231,8 @@ class NewServerView extends React.Component {
 				url = `http://${ url }`;
 			} else if (/^https?:\/\//.test(url) === false) {
 				url = `https://${ url }`;
+			} else if (/^http:\/\//.test(url)) {
+				url = url.replace('http://', 'https://');
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


app/views/NewServerView.js
```diff
@@ -225,11 +226,14 @@ class NewServerView extends React.Component {
 			url = `${ url }.rocket.chat`;
 		}
 
+		// Comment to test local server on Android device
```
djorkaeffalexandre, line 229: We can use `__DEV__` here, just to new devs don't worry about that.

Fishbowler:
Should this be a server concern rather than enforced by the app?
I feel like there are some valid use cases. If I'm testing changes to the app, or in any way throwing together any proof-of-concept that involves Rocket.Chat, should I _need_ to add HTTPS? Including to my local Docker RC?

grahamsmith:
I like the idea behind the intent here. 👍

That said I feel this sets developers up for failure as it hides the behaviour in rewriting URLs secretly in multiple places. While working on and using RocketChat this sort of change has cost lots of time and effort as a general theme. 

My suggestions are:

- Add a strong warning about the use HTTP rather than HTTPS in console or even on the client device.
- If HTTPS is important, then both iOS and Android have feature support around this, then we need to look beyond the React Native code. In the Android Network Security Config the setting for ClearTextTrafficPermitted is `true`. This could be moved to debug builds only as well as support for user certificates (which are allowed in production builds so I can bypass the HTTPS really easily by MITM if I knew what I was doing). There are similar concepts in iOS. Devs can also implement certificate pinning and alike when using HTTPS in production.
- If the decision has been taken to force HTTPS then make it configurable and easily changed and obvious (I can help update the docs if needed).

diegolmello:
@Fishbowler @grahamsmith Thanks for the comments, guys.
The intent here is to keep developer experience the same: it should work on insecure connections on development.

HTTP doesn't work on iOS production already and it works on Android just because we have `ClearTextTrafficPermitted` enabled (which we're going to fix in this PR).

I've created this PR as draft because we need a deeper discussion and testing :)

diegolmello:
Closed in favour of #2357 and #2449

PR #2399, opened 2020-08-14
[IMPROVEMENT] Add Enterprise modules on Omnichannel

diegolmello:
## Proposed changes

Look for `omnichannel-mobile-enterprise` and `livechat-enterprise` on omnichannel features.

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/enterpriseModules.js
+++ b/app/lib/methods/enterpriseModules.js
@@ -6,6 +6,7 @@ import log from '../../utils/log';
 import { setEnterpriseModules as setEnterpriseModulesAction, clearEnterpriseModules } from '../../actions/enterpriseModules';
 
 export const LICENSE_OMNICHANNEL_MOBILE_ENTERPRISE = 'omnichannel-mobile-enterprise';
+export const LICENSE_LIVECHAT_ENTERPRISE = 'livechat-enterprise';
 
 export async function setEnterpriseModules() {
 	try {
@@ -54,4 +55,9 @@ export function getEnterpriseModules() {
 export function hasLicense(module) {
 	const { enterpriseModules } = reduxStore.getState();
 	return enterpriseModules.includes(module);
+}
+
+export function isOmnichannelModuleAvailable() {
+	const { enterpriseModules } = reduxStore.getState();
+	return [LICENSE_OMNICHANNEL_MOBILE_ENTERPRISE, LICENSE_LIVECHAT_ENTERPRISE].every(module => enterpriseModules.includes(module));
 }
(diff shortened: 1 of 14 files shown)


app/lib/methods/enterpriseModules.js
```diff
@@ -0,0 +1,63 @@
...
+}
+
+export function hasLicense(module) {
+	const { enterpriseModules } = reduxStore.getState();
+	return enterpriseModules.includes(module);
+}
+
+export function isOmnichannelModuleAvailable() {
+	const { enterpriseModules } = reduxStore.getState();
+	return [LICENSE_OMNICHANNEL_MOBILE_ENTERPRISE, LICENSE_LIVECHAT_ENTERPRISE].every(module => enterpriseModules.includes(module));
```
rodrigok, line 62: Should be `some` rather than `every`?

diegolmello: If I get the rule, we need both of them to enable the feature.
If one module is enough to enable it, I agree with you.
Otherwise, we want `every`:

AgentTNT:
@diegolmello - Why is this feature now all of a sudden being converted to enterprise only?

app/lib/methods/enterpriseModules.js
rodrigok: One module only should enabled the feature.

diegolmello:
> @diegolmello - Why is this feature now all of a sudden being converted to enterprise only?


Yes, Omnichannel agent on mobile will be an enterprise feature.  We have just started to work on this.

AgentTNT:
> > @diegolmello - Why is this feature now all of a sudden being converted to enterprise only?
> 
> Yes, Omnichannel agent on mobile will be an enterprise feature. We have just started to work on this.

Well yes I can see that with this PR, but why after this feature has already been implemented for several releases? (v4.7.0, released 5/22/2020). Unless I am missing something here, this is directly taking away a feature for non enterprise servers. Wouldn't the appropriate time to have put this behind the enterprise license, been at the time of the feature implementation, and not 3 months after?

diegolmello:
> > > @diegolmello - Why is this feature now all of a sudden being converted to enterprise only?
> > 
> > 
> > Yes, Omnichannel agent on mobile will be an enterprise feature. We have just started to work on this.
> 
> Well yes I can see that with this PR, but why after this feature has already been implemented for several releases? (v4.7.0, released 5/22/2020). Unless I am missing something here, this is directly taking away a feature for non enterprise servers. Wouldn't the appropriate time to have put this behind the enterprise license, been at the time of the feature implementation, and not 3 months after?

You’re right @AgentTNT! We are going to include only the most recent work into Enterprise (see the changes in the PR). Thanks for the comment.

Issue #2025, opened 2020-04-10
Inline image data is displayed in web client, but not in mobile client


soundsc:
- Rocket.Chat server version: 2.4.8
- Device: iPad (also reported on iPhone)
- Client version 4.6.3.294 on iPadOS 13.4.1

**Bug Description**
Attached images are not displayed for messages with a certain format. Occurs on iPadOS but not in web client.

Screenshot (cropped):

![IMG_0408 PNG](https://user-images.githubusercontent.com/13485513/79027907-516ea680-7b8e-11ea-8b65-71ee1317bc67.jpeg)

**Expected behaviour**
The client should simply display the image (and its "title" field), as the web client does.

**How to reproduce**
This is observed for custom messages posted via the postMessage call of the REST API. Minimal message reproducing the behaviour:

`{"channel": "#test___", "text": "*mydescription*", "attachments": [{"image_url": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAkAAAAJCAYAAADgkQYQAAAABHNCSVQICAgIfAhkiAAAAAlwSFlzAAADMgAAAzIBfQlAtwAAABl0RVh0U29mdHdhcmUAd3d3Lmlua3NjYXBlLm9yZ5vuPBoAAACASURBVBiVfYyxDQMhDEXtHKK58nqYAckSDeuwCAzAULTeIP1tQEFjp4t0SuBJv/pPD1trbyI6YQEzD0NEZ0rpWkkAAK/d+VcqpRwxRlNrPR5W7/1W1Ski0zkn1lr13ouITFWdvff7W0JEyDlLCEFyzoKIv6XVHqUdhpnHTmDm8QFUtVR1lD2rSwAAAABJRU5ErkJggg==", "title": "mydescription"}]}`

(the image was base64-encoded with Python's `base64.b64encode()`)

**Additional remarks**
The message could be incomplete or malformatted. However,

* it is accepted by the server ("success": "True")
* it seems to conform to the [API documentation](https://rocket.chat/docs/developer-guides/rest-api/chat/postmessage/)
* it displays correctly in the web client

So it seems reasonable that the mobile client should display it, too.

Cheers!

soundsc:
It seems that the canonical way to upload files is via `/rooms.upload`. That explains why this kind of message is not supported: most clients simply would not create messages in this way.

[...]

PR #2527, opened 2020-10-06
[FIX] Display inline image

youssef-md:
## Proposed changes

The app isn't rendering an image when sending an image address as a message.

## Issue(s)	

#2025 

## How to test or reproduce

Copy an image address and send as a message.

## Screenshots
image

## Types of changes

## Checklist

### DIFF
--- a/app/containers/message/ImagePreview.js
+++ b/app/containers/message/ImagePreview.js
@@ -1,0 +1,38 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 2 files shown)


diegolmello:
I agree this PR fixes the issue, but I'm not sure it's something we really want to add.
What it does is adding another regex matching a couple of images formats and trying to fetch the image on render.
From a performance perspective, maybe keeping the issue is for the better and we can think on something else on server side.

I'm closing for now, but I'll keep it in mind.
Thanks.

PR #2619, opened 2020-11-09
[i18n] Add missing German strings

phriedrich:
This adds some missing German translation and updates some existing details to make them more consistent in the selection of thread selection view.


diegolmello (approved):
LGTM

PR #2685, opened 2020-12-01
[IMPROVEMENT] Ease white labelling for Android

grahamsmith:
## Proposed changes
Centralised application ID to the root project gradle file to simplify the white label process. Explicitly set the class names and packages declared in the manifest to account for the application ID and the root package not being the same.

## Issue(s)	
N/A - Discussed with team.

## How to test or reproduce

- Follow the existing guide to create a Firebase project and download the Google Service file.
- Update the project base gradle file where `ext` is set change applicationId to desired `applicationId` e.g. com.graham
- run `yarn android-whitelabel com.graham` to build and install the `playDev` and RUN the app.

## Screenshots
N/A

## Types of changes

## Checklist

## Further comments
Made the changes and tested in Android Studio - this complained about the package names being incorrect (they were) for the classes in the networking package.

Also tidied the various Manifests to prevent merge issues.


android/app/build.gradle
```diff
@@ -196,7 +196,7 @@ android {
     flavorDimensions "type"
     productFlavors {
         foss {
-            applicationId "chat.rocket.android"
```
grahamsmith, line 199: whoops - will fix this. Failed to notice the last segment was different. This seems an odd package name change but can discuss.

android/build.gradle
```diff
@@ -8,6 +8,7 @@ buildscript {
     def isPlay = !taskRequests.contains("foss")
 
     ext {
+        applicationId = "chat.rocket.reactnative"
```
diegolmello, line 11: On the white label docs, we ask them to set this app id (and other values) on the `gradle.properties`: https://docs.rocket.chat/guides/developer/mobile-apps/whitelabeling-mobile-apps#general-setup-1
This is different on the `single-server` vs `develop` branch, but I agree we can share the same behaviour.

IMO, it's easier if we ask the reader to change everything in one place, if possible.
Can we move this to `gradle.properties` as well?

android/app/build.gradle
diegolmello: We could remove it from here completely and make an additional step on the CI to change the appId.

package.json
```diff
@@ -10,6 +10,7 @@
     "ci": "npm run precommit && codecov",
     "ios": "npx react-native run-ios",
     "android": "npx react-native run-android --variant=playDebug",
+    "android-whitelabel": "react-native run-android --main-activity chat.rocket.reactnative.MainActivity --variant=playDebug --appId",
```
diegolmello, line 13: This is awesome! :)

diegolmello (commented):
Hey, @grahamsmith.
Left a few comments here.
Thank you! ❤️

android/app/build.gradle
grahamsmith: product flavours can change the suffix so ideally it would have been great if the package name for foss was something like `chat.rocket.reactnative.foss` as then the setup would be a typical gradle setup. I imagine now that the app has been submitted to FOSS it's a little late.

At this point I would vote for me reverting line 199 and we can deal with FOSS separately in a different piece of work. Hope that makes sense?

package.json
grahamsmith: 👍

android/build.gradle
grahamsmith: Sure - I'll push the change up

android/app/build.gradle
diegolmello: Sorry. Read this just now. I pushed some changes already.
FOSS points to the main app as if it's just the same app, but without proprietary code.
So both Google Play and F-Droid apps are `chat.rocket.android`.
I agree with you that it's not the ideal and FOSS could have another name, but it's like this for a while already and it'd break the updates if we changed the bundle id now.

This is what I did and it should work properly https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2685/commits/28fc754d64db65cf2861f8f3417f11f8c7b7c3e5

diegolmello (approved):
LGTM

PR #2690, opened 2020-12-04
[WIP] Inject Native Timer to Rocket.Chat.js.SDK

djorkaeffalexandre:
## Proposed changes

Currently the Rocket.Chat.js.SDK not allow us to inject a instance of Timer and is using a NodeJS Timer, the implementation of NodeJS Timer on react-native is limited to app state in foreground, using react-native-background-timer can be better for us to handle the SDK Websocket connection.
Refer to https://github.com/RocketChat/Rocket.Chat.js.SDK/commit/1f42d0b068bb8cc41bf9794b34c368d14da7ea2c to check SDK changes.
Now the SDK allow us to inject a instance of Timer and we're injecting a Native Timer [react-native-background-timer](https://github.com/ocetnik/react-native-background-timer).

## How to test or reproduce

Let's put these changes on beta and test during some days to see if it's better or not.

## Types of changes

- [X] Chore (non-breaking change which changes a current function)

## Checklist

### DIFF
--- a/app/lib/rocketchat.js
+++ b/app/lib/rocketchat.js
@@ -7,6 +7,7 @@ import {
 import { Q } from '@nozbe/watermelondb';
 import AsyncStorage from '@react-native-community/async-storage';
 import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
+import BackgroundTimer from 'react-native-background-timer';
 import RNFetchBlob from 'rn-fetch-blob';
 
 import reduxStore from './createStore';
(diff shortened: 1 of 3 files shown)


djorkaeffalexandre:
It's crashing the iOS app after 30 seconds.
In iOS, only specific app types are allowed to run in the background:
- Apps that play audible content to the user while in the background, such as a music player app
- Apps that record audio content while in the background
- Apps that keep users informed of their location at all times, such as a navigation app
- Apps that support Voice over Internet Protocol (VoIP).
- Apps that need to download and process new content regularly
- Apps that receive regular updates from external accessories

Some people did a workaround since iOS 4, that consists of play a no-sound audio track forever, so your app will be kept alive, but it's a bad hack, drain user battery.

This guy mention something like a "Permission from Apple" to run something on background https://stackoverflow.com/a/36708987, maybe this is what Whatsapp does to make the web client works.

The stability of the WebSocket is not related to use a native timer or not on our Rocket.Chat SDK. It was just a try of something to improve the current implementation that has some bugs and sometimes is not working properly, but it's a thing impossible to do due iOS limitations to protect battery energy and other things.

If do you want to help us with a implementation of WebSockets that alives on Background State of iOS devices, with a way that is allowed by Apple please feel free to contact us.
Thanks.

PR #2696, opened 2020-12-07
[i18n] Update arabic

faz1993:
## Types of changes

## Checklist


app/i18n/locales/ar.js
```diff
@@ -129,8 +129,8 @@ export default {
 	creating_invite: 'إنشاء دعوة',
 	Channel_Name: 'اسم القناة',
 	Channels: 'قنوات',
-	Chats: 'محادثات',
-	Call_already_ended: 'انتهت المكالمة بالفعل!',
+	Chats: 'الرسائل',
+	Call_already_ended: 'انتهت المكالمة!',
```
diegolmello, line 133: I'm reviewing with Google Translate, so it's not accurate, but it looks like the current translation is better.
The intention here is to tell the user "The call has already ended", which is precisely what the current string says.

app/i18n/locales/ar.js
```diff
@@ -101,7 +101,7 @@ export default {
 	Announcement: 'إعلان',
 	Apply_Your_Certificate: 'طبق شهادتك',
 	Applying_a_theme_will_change_how_the_app_looks: 'سيؤدي تطبيق السمة إلى تغيير شكل التطبيق',
-	ARCHIVE: 'أرشفة',
+	ARCHIVE: 'أرشيف',
```
diegolmello, line 104: This is a call to action to `Archive` a channel.
It's not a list of things.

diegolmello (changes requested):
Can you take a look? Thanks!

app/i18n/locales/ar.js
faz1993: sometimes google give wrong translation in arabic it well the call has already ended in first translation was the call is really ended but in the string is the call already ended so i corrected it to the same meaning

app/i18n/locales/ar.js
faz1993: ahhh i tought its list

diegolmello (approved):

PR #2715, opened 2020-12-15
[i18n] Add missing German strings

phriedrich:
added missing strings and fixed one (used translation from server)


phriedrich:
@diegolmello @djorkaeffalexandre please merge.

app/i18n/locales/de.js
```diff
@@ -318,7 +320,7 @@ export default {
 	Message_Reported: 'Nachricht gemeldet',
 	Microphone_Permission_Message: 'Rocket.Chat benötigt Zugriff auf das Mikrofon, damit du eine Audionachricht senden kannst.',
 	Microphone_Permission: 'Mikrofonberechtigung',
-	Mute: 'Stumm',
+	Mute: 'Diesem Benutzer das Chatten verbieten',
```
diegolmello, line 323: @phriedrich  Are you sure about this one?
It looks so verbose even for german :)
I wonder if it's going to be wrapped or cutted.
Does the previous version work?

phriedrich: Yes, this one looks especially cumbersome, but for the place where it is used, this is the actual term from the browser/server app. 

It's in a room's member list, please see the attached screenshots:

Browser-App

ReactNative

A shorter version would work, something like "Stumm schalten". 
But this one is more descriptive, and I tried to match the translations of the browser app, where possible. In this case, there's enough space, so I used the original, very verbose, form.

diegolmello: Got it. You're right.

diegolmello (approved):
LGTM

PR #2794, opened 2021-01-13
[IMPROVEMENT] Revamped Toast in place of EventEmitter

### LINKED ISSUE #2536: Export show toast function
### Current way
Out [Toast component](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/containers/Toast.js) works based on EventEmitter and [every time we need to call it, we emit an event](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/views/StatusView.js#L119).

### Proposed way
It'd be easier if we exported a `showFunction` receiving the message.

aKn1ghtOut:
## Proposed changes

Removed implementation of the toast functionality using EventEmitters. Instead, transferred the functionality to a react hook with a provider.

## Issue(s)	

Closes #2536 

## How to test or reproduce

1. Touch and hold down on any message
2. Select permalink
3. The toast works

## Types of changes

## Checklist


app/containers/Toast.js
```diff
@@ -20,15 +19,17 @@ const styles = StyleSheet.create({
...
 });
 
-export const LISTENER = 'Toast';
-
 class Toast extends React.Component {
 	static propTypes = {
 		theme: PropTypes.string
 	}
 
+	static mounted = null;
```
diegolmello, line 27: Do we really need `mounted` here?

app/containers/Toast.js
```diff
@@ -20,15 +19,17 @@ const styles = StyleSheet.create({
...
+
+	static show = (...args) => Toast.mounted && Toast.mounted.showToast(...args);
```
diegolmello, line 29: It's really odd how `showToast` is a child of `mounted`.

diegolmello, line 29: Instead of static, can we use refs instead? Maybe even migrate to hooks.

diegolmello (changes requested):
@aKn1ghtOut Thanks.
Left a few comments.
I also wonder if we have a better option by this point.
This lib is really easy to use, but it also isn't testable on Detox.

app/containers/Toast.js
aKn1ghtOut: I do agree that it's odd, yeah

aKn1ghtOut: Looking into it!

aKn1ghtOut:
@diegolmello Switched the implementation to use hooks. Feels much better. Please review.

PR #2805, opened 2021-01-15
[FIX] Spotlight returning duplicated entries

gerzonc:
## Proposed changes

The search method used to look for users in the RoomsListView requires to filter the users we already have a chat with. This is required, since this method looks for every user in the app, so when we're looking for a name, at search we will see the open chat with that person, the users with similar username and the username of the same person which we already have a chat open, and this is not the correct behavior for this.

## How to test or reproduce

In order to test this, you should be logged in a server.

- While being on the main screen, tap on the search button
- Type any username and open a chat with the one you looked for
- Go back to the main screen and search for the same username again

## Screenshots
<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</thead>
</tr>
<tbody>
<tr>
<td>
image
</td>

<td>
image
</td>
</tr>
</tbody>
</table>

## Types of changes

## Checklist


app/lib/rocketchat.js
```diff
@@ -647,13 +647,16 @@ const RocketChat = {
...
 				]);
 				if (filterUsers) {
-					data = data.concat(users.map(user => ({
-						...user,
-						rid: user.username,
-						name: user.username,
-						t: 'd',
-						search: true
-					})));
+					const usersFiltered = users.filter(user => !data.some(sub => user.username === sub.name));
```
diegolmello, line 650: Can you compare `ids` intead of `usernames`?

diegolmello (changes requested):
You should check for `atIndex(0)` on detox tests.
If your PR is correct, they aren't needed anymore.

diegolmello (approved):

PR #2772, opened 2021-01-07
[FEATURE] Show Thumbnails/Preview for Uploaded Videos

pratyaksh123:
## Proposed changes

Adds thumbnail support to Video files.

## Issue(s)	

Solves https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/2746

## Screenshots
   --->>>    
Before and After

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments
- Currently, it can take up to 6-7 seconds to fetch a thumbnail, depending upon the size of the video ( The URL of the video is passed to ` VideoThumbnails.getThumbnailAsync(uri);`. 
- IMO, this load time can only be improved if this was to be done on the server or we were caching the videos locally.
- Also, we can add something like this ( https://github.com/chramos/react-native-skeleton-placeholder) as an indicator that the thumbnail is loading.

Issue #2389, opened 2020-08-10
Feature: Video playback offline


zerobytes999:
- Your Rocket.Chat Experimental app version: 4.10.0.19667
- Your Rocket.Chat server version: 4.5.1

I think it will help alot when receiving videos they get downloaded (upon user tap on video box) and it is saved in the device instead of it being streamed. It will help greatly in re-viewing videos or while offline

Also video thumbnail will be an amazing step too.

Thank you very much for all the help and great server / applications.

diegolmello:
Thanks for opening.
Yeah, that's a great feature to have.

pratyaksh123:
Hey, I think I can work on this,
As a user opens the ```AttachmentView.js``` .We need to search the downloaded videos in the local directory ( with rn-fetch-blob) and if it's there load it and play, else we can stream it and also simultaneously download it ( like WhatsApp ) for later offline use.
And also a user can have an option to turn this auto-download off if he wants!

JASIM0021:
If the issue is still pending, I’d be happy to look into it and provide a fix.
Please confirm, @diegolmello .

diegolmello:
@JASIM0021 Sure! I'm happy to review your contributions!

PR #2809, opened 2021-01-18
[ IMPROVEMENT ] Support Caching of Video/Images

pratyaksh123:
## Proposed changes

Currently, there is no way to cache images/Videos. ( You can save it, but each time you open `AttachmentView.js` it will make a network request that makes the UX bad. This PR aims to solve this issue by storing these media assets in the user's Document Directory. This improves video load times to load almost instantaneously
## Issue(s)	
https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/2389

## How to test or reproduce

Play a video. ( After it is downloaded )

## Types of changes

## Checklist

## Further comments

We need to make a decision as to how the caching behavior should be when a user opens a video -
- [ ] Auto-download the video if it is not downloaded yet. ( maintain a global setting that users can toggle that decides if to auto download or not , similar to WhatsApp. )
- [ ] Use the existing `HandleSave` method to allow users to save the video explicitly.

This PR will also make loading of Video Thumbnails pretty instantaneous :slightly_smiling_face: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2772

### DIFF
--- a/app/views/AttachmentView.js
+++ b/app/views/AttachmentView.js
@@ -97,7 +97,7 @@ class AttachmentView extends React.Component {
 
 	getVideoRef = ref => this.videoRef = ref;
 
-	handleSave = async() => {
+	getAttachmentMetaData = async() => {
 		const { attachment } = this.state;
 		const { user, baseUrl } = this.props;
 		const {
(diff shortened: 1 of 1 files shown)


pratyaksh123:
@diegolmello what should be the caching behavior be like  ? ( check in the further comments above ) 
Please review when you get free.

app/views/AttachmentView.js
```diff
@@ -146,22 +147,57 @@ class AttachmentView extends React.Component {
...
+			const files = await FileSystem.readDirectoryAsync(FileSystem.documentDirectory);
+			files.forEach((t) => {
+				if (t === fileName) {
+					this.setState({ path: `${ FileSystem.documentDirectory + fileName }` });
+				}
+			});
+		} catch (e) {
+			EventEmitter.emit(LISTENER, { message: I18n.t(image_url ? 'error-save-image' : 'error-save-video') });
+		}
+	}
```
gerzonc, line 180: This function is pretty much like `handleSave`, can you refactor it so we can extract what we need for both?

app/views/AttachmentView.js
```diff
@@ -8,7 +8,7 @@ import RNFetchBlob from 'rn-fetch-blob';
 import { Video } from 'expo-av';
 import SHA256 from 'js-sha256';
 import { withSafeAreaInsets } from 'react-native-safe-area-context';
-
+import * as FileSystem from 'expo-file-system';
```
gerzonc, line 11: Why not `rn-fetch-blob`?

gerzonc (changes requested):

app/views/AttachmentView.js
pratyaksh123: `expo-FileSystem` had a more intuitive API and we were already using that so thought using this might be easier.

pratyaksh123:
Also, can you check the further comments I made in the PR above, we need to decide when to cache media, on handleSave, or on opening it for the first time?

raza2022:
@pratyaksh123 @gerzonc any update on this one as I just wanna it but some customize way but might finish it to here, instead of starting from scratch

pratyaksh123:
> @pratyaksh123 @gerzonc any update on this one as I just wanna it but some customized way so thinking to finish it here, instead of starting from scratch

You can pick it up , I'm no longer working on rocket.chat issues

PR #2857, opened 2021-02-01
[IMPROVEMENT] Check for focused rooms on in-app notifications

gerzonc:
## Proposed changes

When we are in a room, we still receive in-app notifications from that same room if we go deeper in the stack (e.g. we go to `RoomInfoView`, etc.) which is not right, because we didn't left that room yet, so we shouldn't receive in-app notifications while in it. I solved this issue by checking the room reducer instead of checking if we are on `RoomView`. 

## How to test or reproduce

In order to test this you should be logged in 2 devices with different users (this is not mandatory but it would make testing this easier). For verbal simplicity, we will refer to both users as user 1 and user 2.

### Test case 1
1. Get into a room between both users.
2. Send a message from user 1 to user 2 while both are in the room.
3. With user 1, go to room info and with user 2 send a message to that room.
4. User 1 shouldn't receive an in-app notification.

### Test case 2 
1. Get into a room between both users
2. With user 1, create a thread and then leave the thread
3. With user 2, send a message to the room.
4. User 1 shouldn't receive an in-app notification.

### Test case 3 
1. With user 1, create a room with a new user that is not user 2.
2. With user 2, send a message to user 1.
3. User 1 should receive an in-app notification.

## Screenshots
<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</thead>
</tr>
<tbody>
<tr>
<td>
image
</td>

<td>
image
</td>
</tr>
</tbody>
</table>

## Types of changes

## Checklist

## Further comments

I also changed how we unsubscribe from a room on the room reducer, which previously we filtered by `rid`, which is wrong, because if we use `rid !== action.rid` it won't only take away the current `rid`, but also any other that is the same `rid`, and that's problematic when we're inside a thread, because the thread has the same `rid` as the room you're focused on, so it will remove both of them. In order to solve this, I just remove the `rid` by index.


diegolmello (approved):
@gerzonc I've made a few changes and you should take a look.
Fixed an issue, removed ref and added appropriate prop types.

LGTM now :)

PR #2867, opened 2021-02-03
[FIX] Admin panel option is hidden after login

kresnaputra:
## Proposed changes

Fixing admin panel option hiding after login by adding componentDidUpdate function

## How to test or reproduce

Clickup task

## Types of changes

## Checklist

### DIFF
--- a/app/views/SidebarView/index.js
+++ b/app/views/SidebarView/index.js
@@ -45,15 +45,15 @@ class Sidebar extends Component {
 		loadingServer: PropTypes.bool,
 		useRealName: PropTypes.bool,
 		allowStatusMessage: PropTypes.bool,
-		isMasterDetail: PropTypes.bool,
-		isLogin: PropTypes.bool
+		isMasterDetail: PropTypes.bool
 	}
 
 	constructor(props) {
 		super(props);
 		this.state = {
 			showStatus: false,
-			isAdmin: false
+			isAdmin: false,
+			isSetAdminNeedCall: true
 		};
 	}
 
@@ -69,7 +69,7 @@ class Sidebar extends Component {
 	}
 
 	shouldComponentUpdate(nextProps, nextState) {
-		const { showStatus, isAdmin } = this.state;
+		const { showStatus, isAdmin, isSetAdminNeedCall } = this.state;
 		const {
 			Site_Name, user, baseUrl, state, isMasterDetail, useRealName, theme
 		} = this.props;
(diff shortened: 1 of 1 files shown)


app/views/SidebarView/index.js
```diff
@@ -103,9 +105,16 @@ class Sidebar extends Component {
 		if (nextState.isAdmin !== isAdmin) {
 			return true;
 		}
+		if (nextProps.state !== state && isSetAdminNeedCall === true) {
```
diegolmello, line 108: As you can see at the beginning of the block, `state` comes from `react-navigation` (it's track current state of the navigation) and it's an object.
`nextProps.state !== state` is an object comparison, so it'll never be equal.
image

app/views/SidebarView/index.js
```diff
@@ -103,9 +105,16 @@ class Sidebar extends Component {
...
+			return true;
+		}
 		return false;
 	}
 
+	componentDidUpdate() {
+		this.setIsAdmin();
```
diegolmello, line 115: Not a good practice to call something on `componentDidUpdate` without testing in what cases it should run.
`componentDidUpdate` runs every time the component updates: when state changes, when props change or when the component is forced updated.

Even if `shouldComponentUpdate` returns false, `this.setIsAdmin` is being called.
Add a `console.log` at the start of `setIsAdmin` and you'll see how many times it's called.

app/views/SidebarView/index.js
```diff
@@ -114,8 +123,11 @@ class Sidebar extends Component {
 			if	(roles) {
 				const permissionsCollection = db.collections.get('permissions');
 				const permissionsFiltered = await permissionsCollection.query(Q.where('id', Q.oneOf(permissions))).fetch();
-				const isAdmin = permissionsFiltered.reduce((result, permission) => (
-					result || permission.roles.some(r => roles.indexOf(r) !== -1)),
+				const isAdmin = permissionsFiltered.reduce((result, permission) => {
+					this.setState({ isSetAdminNeedCall: false });
```
diegolmello, line 127: Not a good practice to change the state inside of an iteration.
It's going to set the state for every permissions on the list.
Even if `sCU` prevents the re-render to happen, you're calling unnecessary resources.

diegolmello (commented):
Left a few comments here and added a new task on ClickUp as blocking.
The issue is really on `permissions` fetch.

diegolmello:
Closed in favour of #2914

PR #2869, opened 2021-02-04
[FIX] Thread showing typing indicator from main room

kresnaputra:
## Proposed changes

When another user on the room starts typing, we show `User is typing` on the header.
Currently, we only have plans to add that for threads, so it's not implemented yet.
We've been showing that indicator on threads and that might cause the user to think there's someone else typing on the thread, which is not the case.

## Types of changes

## Checklist

### DIFF
--- a/app/views/RoomView/Header/Header.js
+++ b/app/views/RoomView/Header/Header.js
@@ -119,7 +119,7 @@ HeaderTitle.propTypes = {
 };
 
 const Header = React.memo(({
-	title, subtitle, parentTitle, type, status, usersTyping, width, height, prid, tmid, connecting, goRoomActionsView, roomUserId, theme, isThread
+	title, subtitle, parentTitle, type, status, usersTyping, width, height, prid, tmid, connecting, goRoomActionsView, roomUserId, theme
 }) => {
 	const portrait = height > width;
 	let scale = 1;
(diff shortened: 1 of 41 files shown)


app/views/RoomView/Header/Header.js
```diff
@@ -168,13 +168,14 @@ const Header = React.memo(({
 					theme={theme}
 				/>
 			</View>
-			<SubTitle usersTyping={usersTyping} subtitle={subtitle} theme={theme} renderFunc={renderFunc} />
+			<SubTitle usersTyping={isThread ? 0 : usersTyping} subtitle={subtitle} theme={theme} renderFunc={renderFunc} />
```
diegolmello, line 171: `usersTyping` is an array, so it should be `[]` instead of `0`.

app/views/RoomView/index.js
```diff
@@ -110,6 +110,7 @@ class RoomView extends React.Component {
 		this.rid = props.route.params?.rid;
 		this.t = props.route.params?.t;
 		this.tmid = props.route.params?.tmid;
+		this.isThread = props.route.params?.t === 'thread' || false;
```
diegolmello, line 113: No need to add `isThread`.
If `tmid` (aka thread message id) has content, it's a thread.

diegolmello (changes requested):

diegolmello (approved):
LGTM

PR #2885, opened 2021-02-08
[FIX] Can't copy or edit media's description

kresnaputra:
## Proposed changes

If message has attachment, I copy the text from message.attachments[0].description, otherwise I just copy from message.msg

## How to test or reproduce

- Edit messages text that have image in it:
1. Go to channel or DM.
2. Upload some image and add the caption
3. Edit the message.

- Copy messages text that have image in it:
1. Go to channel or DM.
2. Upload some image and add the caption
3. Copy the message.

## Types of changes

## Checklist

### DIFF
--- a/app/containers/MessageActions/index.js
+++ b/app/containers/MessageActions/index.js
@@ -171,7 +171,7 @@ const MessageActions = React.memo(forwardRef(({
 
 	const handleCopy = async(message) => {
 		logEvent(events.ROOM_MSG_ACTION_COPY);
-		await Clipboard.setString(message.msg);
+		await Clipboard.setString(message.attachments.length > 0 ? message.attachments[0].description : message.msg);
 		EventEmitter.emit(LISTENER, { message: I18n.t('Copied_to_clipboard') });
 	};
 
--- a/app/containers/MessageActions/index.js
+++ b/app/containers/MessageActions/index.js
@@ -171,7 +171,7 @@ const MessageActions = React.memo(forwardRef(({
 
 	const handleCopy = async(message) => {
 		logEvent(events.ROOM_MSG_ACTION_COPY);
-		await Clipboard.setString(message.attachments.length > 0 ? message.attachments[0].description : message.msg);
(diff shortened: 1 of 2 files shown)


app/containers/MessageActions/index.js
```diff
@@ -171,7 +171,7 @@ const MessageActions = React.memo(forwardRef(({
 
 	const handleCopy = async(message) => {
 		logEvent(events.ROOM_MSG_ACTION_COPY);
-		await Clipboard.setString(message.msg);
+		await Clipboard.setString(message.attachments.length > 0 ? message.attachments[0].description : message.msg);
```
diegolmello, line 174: Can you simplify with something like `message?.attachments?.[0]?.description || message.msg`?

app/views/RoomView/index.js
```diff
@@ -531,6 +531,9 @@ class RoomView extends React.Component {
 	}
 
 	onEditInit = (message) => {
+		if (message.attachments.length > 0) {
+			message._raw.msg = message.attachments[0].description;
```
diegolmello, line 535: `message` on this context is an observable from WatermelonDB, so it shouldn't be changed directly (it's error prone).
Instead, you can pick only the properties needed from the object to do the edit.
Also, are you sure this is going to work?

diegolmello (changes requested):

app/views/RoomView/index.js
kresnaputra: I tested it and it is work well, but like you said I need change the code to make sure the db is not change directly

diegolmello (commented):
@kresnaputra Can you add details to test/reproduce this?

kresnaputra:
@diegolmello updated

diegolmello:
@kresnaputra Does it work for other types of attachments, like Video and Audio?

kresnaputra:
@diegolmello yes, I already tried and it is work

diegolmello (approved):

Issue #2917, opened 2021-02-18
[BUG] Android - auth challenge with no cert loaded causes authentication webview to hang


newmanw:
### Description:

If a certificate auth challenge is requested on Android the webview will hang if no certificate is loaded. 

This is a recent breaking change, https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2624.

Our authentication providers will first present a cert auth challenge, if that auth challenge is canceled, the auth provider will present a view allowing users to authenticate via a different means.  Our users do not have client certificates they can load into the RocketChat mobile application.  As such the auth challenge presented by our auth provider is never canceled, and the page load never completes.

### Environment Information:

- Rocket.Chat Server Version: Any
- Rocket.Chat App Version: Since 4.13.0 https://github.com/RocketChat/Rocket.Chat.ReactNative/releases/tag/4.13.0
- Device Name: any
- OS Version: Android, any version

### Steps to reproduce:

1. Do not load a client certificate into Android app.
2. OAuth authentication against webpage that presents cert challenge.
3. Page will not load as auth challenge is not canceled if no client cert is present.

### Expected behavior:

Auth challenge should fallback to default Android behavior and cancel the authentication challenge if now client cert is present.

### Actual behavior:

Auth challenge is not canceled, page hangs.

![oauth-android](https://user-images.githubusercontent.com/3382274/108393959-6a45e080-71d1-11eb-9657-086daf139d43.png)

### Additional context:

This is a breaking change in which our Android users cannot currently authenticate.

PR #2777, opened 2021-01-08
WIP [FIX] Mutual TLS without client cert (#2452)

jangaraj:
## Proposed changes
Proceed request also when no certificate is configured but it is was requested by the server. Server should handle this situation, for example IDP server can return html login/password form instead of mutual TLS authentication. The majority of the codes on Android is doing that - https://www.programcreek.com/java-api-examples/?api=android.webkit.ClientCertRequest

## Issue(s)	
https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/2452

## How to test or reproduce

Configure SAML auth and set Custom Entry Point to `https://client.badssl.com/` - try auth with SAML without any cert:

Before: empty AuthenticationWebView (IDP site is not loaded)

After: IDP page is loaded (there will be expected error page 400 in the `https://client.badssl.com/` case)

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2918, opened 2021-02-18
[FIX] Webview not falling back to default auth challenge when no cert is provided

newmanw:
## Proposed changes

If a certificate auth challenge is requested on Android the webview will hang if no certificate is loaded. To prevent this, fallback to default Android behavior and cancel the challenge with request.cancel().  This is the same behavior on iOS, https://github.com/react-native-webview/react-native-webview/blob/master/apple/RNCWebView.m#L818.

## Issue(s)	

#2917 

## Types of changes

## Checklist

## Further comments

Co-authored-by: Jan Garaj <jan.garaj@gmail.com>

### DIFF
--- a/patches/react-native-webview+10.3.2.patch
+++ b/patches/react-native-webview+10.3.2.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java b/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
-index ab869cf..2aa7a9e 100644
+index ab869cf..b1da7cc 100644
 --- a/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
 +++ b/node_modules/react-native-webview/android/src/main/java/com/reactnativecommunity/webview/RNCWebViewManager.java
 @@ -84,6 +84,12 @@ import java.util.Map;
@@ -44,7 +44,7 @@ index ab869cf..2aa7a9e 100644
    }
  
    @Override
-@@ -742,12 +754,54 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
+@@ -742,12 +754,56 @@ public class RNCWebViewManager extends SimpleViewManager<WebView> {
  
    protected static class RNCWebViewClient extends WebViewClient {
  
@@ -93,6 +93,8 @@ index ab869cf..2aa7a9e 100644
 +          }
 +        };
 +        task.execute();
++      } else {
++        request.cancel();
 +      }
 +    }
 +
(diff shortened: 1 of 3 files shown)


patches/react-native-webview+10.3.2.patch
```diff
@@ -93,6 +93,8 @@ index ab869cf..2aa7a9e 100644
 +          }
 +        };
 +        task.execute();
++      } else {
++        request.cancel();
```
diegolmello, line 97: It's similar to https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/2777

@newmanw @jangaraj which one is correct?

jangaraj: @newmanw for me it doesn't make sense to cancel request `request.cancel();`.

It should proceed `super.onReceivedClientCertRequest(view, request);`. Server will handle that request was made without client certificate (for example in my case my IDP offers standard username/password authentication instead of mutual TLS; in other cases server return). IMHO that should be default option. Real world examples: https://www.programcreek.com/java-api-examples/?api=android.webkit.ClientCertRequest

newmanw: @diegolmello @jacotec I apologize I missed the other PR.

With the current Android implementation either is correct, as android.webkit.WebViewClient.onReceivedClientCertRequest()  will call request.cancel() if its not overridden, which is what used to happen.

request.cancel() only cancels the auth challenge request and is what the client should do in the case where there is not a client certificate to pass to the server.

https://android.googlesource.com/platform/frameworks/base/+/a8352f4/core/java/android/webkit/WebViewClient.java#L313

`
public void onReceivedClientCertRequest(WebView view, ClientCertRequest request) {
request.cancel();
}
`  

I am glad this came up as I went back and forth on the "right" fix.  I am ok with calling super and going with what Android provides as the default.  It will only matter if Android changes its implementation, which is currently request.cancel().

diegolmello: @newmanw can you change your PR to call super then as #2777 does?
I remember @jangaraj was having some issues there and that's why it's still a WIP.

jangaraj: Thank you for the explanation. @diegolmello so it's up to you what do you want to implement. Both options (`request.cancel()`, `super()`) resolve my problem.

newmanw: @jangaraj agreed, both PRs match now and call super class implementation in the case of no client cert.  Wouldn't hurt my feelings if you close this PR, thanks for finding this @jangari.

@diegolmello I appreciate you looking into this.  Currently all our android users are unable to authenticate as our authentication provider displays an alternate way to login if a client cert is not present.  However that page will never load without this fix.  Possible to bump this in priority?

diegolmello: @newmanw Yes, we've been working on fixes only and the plan is to release a patch next week.
I'm merging this now.

@jangaraj I'm adding you as co-author of this PR.

Thanks for your PRs.

diegolmello (approved):

diegolmello (changes requested):
@newmanw It's failing on CI.
image

Can you try running patch-package again?

newmanw:
@diegolmello removed patch file and re-ran  _npx patch-package react-native-webview_, which generated the same patch file.  Possible CI has an old node_modules folder, resulting in a conflict?

diegolmello:
> @diegolmello removed patch file and re-ran _npx patch-package react-native-webview_, which generated the same patch file. Possible CI has an old node_modules folder, resulting in a conflict?

@newmanw Makes sense.
You can try changing dummy stuff on `package.json` and run `yarn` again to change `yarn.lock`.
Like from `"react-native-webview": "10.3.2",` to `"react-native-webview": "^10.3.2",`

diegolmello (approved):

PR #2922, opened 2021-02-19
[CHORE] Simplify server version comparison

gerzonc:
## Proposed changes

Sometimes we have to test server version depending on the feature availability. For that, we use `semver` lib. Server version might have a suffix for development, like `-develop` or `-rc`, which breaks the default of semantic versioning. We have `coerce` to transform it back to `semver ` (e.g. `3.11.0-develop` to `3.11.0`). Otherwise, lower than or greater than won't work. The problem is that we have all of these conditions spread throughout the code.

In order to refactor this in a more simple way I just created two utility functions which work just as `lt` (lower than) and `gte` (greater than) but including `coerce` so we don't have to call it every time we want to compare server versions.

## Types of changes

## Checklist


app/lib/utils.js
```diff
@@ -7,3 +9,7 @@ export const formatAttachmentUrl = (attachmentUrl, userId, token, server) => {
 	}
 	return encodeURI(`${ server }${ attachmentUrl }?rc_uid=${ userId }&rc_token=${ token }`);
 };
+
+export const lt = (currentServerVersion, oldServerVersion) => lowerThan(coerce(currentServerVersion), oldServerVersion);
```
diegolmello, line 13: Currently, we're using the functions from `semver` directly, but that's not meaninful.
Can you rename `lt` to `isServerVersionLowerThan`?
Same for `gte`.

@gerzonc

gerzonc: Alright

app/lib/methods/enterpriseModules.js
```diff
@@ -33,7 +33,7 @@ export function getEnterpriseModules() {
 	return new Promise(async(resolve) => {
 		try {
 			const { version: serverVersion, server: serverId } = reduxStore.getState().server;
-			if (serverVersion && gte(serverVersion, '3.1.0')) {
+			if (serverVersion && isServerVersionGreaterThan(serverVersion, '3.1.0')) {
```
diegolmello, line 36: @gerzonc One more thing here.
We're repeating ourselves on `serverVersion && ` in all calls.
Do you agree that should be inside of the util functions? :)

app/lib/utils.js
```diff
@@ -7,3 +9,7 @@ export const formatAttachmentUrl = (attachmentUrl, userId, token, server) => {
 	}
 	return encodeURI(`${ server }${ attachmentUrl }?rc_uid=${ userId }&rc_token=${ token }`);
 };
+
+export const isServerVersionLowerThan = (currentServerVersion, oldServerVersion) => currentServerVersion && lt(coerce(currentServerVersion), oldServerVersion);
+
+export const isServerVersionGreaterThan = (currentServerVersion, oldServerVersion) => currentServerVersion && gte(coerce(currentServerVersion), oldServerVersion);
```
diegolmello, line 15: I have concerns about this naming still.
`server greater than` is actually `greater than or equal to`.
That might cause some issues in the future.
Do you agree?

What if we do something like this?
```js
export const methods = {
	lowerThan: lt,
	lowerThanOrEqualTo: lte,
	greaterThan: gt,
	greaterThanOrEqualTo: gte
};

export const compareServerVersion = (currentServerVersion, versionToCompare, func) => currentServerVersion && func(coerce(currentServerVersion), versionToCompare);
```

Usage would be like this:
```js
console.log('1.0.0 lower than 1.1.0', compareServerVersion('1.0.0', '1.1.0', methods.lowerThan));
console.log('1.0.0 lower than or equal to 1.1.0', compareServerVersion('1.0.0', '1.1.0', methods.lowerThanOrEqualTo));
console.log('1.0.0 greater than 1.1.0', compareServerVersion('1.0.0', '1.1.0', methods.greaterThan));
console.log('1.0.0 greater than or equal to 1.1.0', compareServerVersion('1.0.0', '1.1.0', methods.greaterThanOrEqualTo));
```

WDYT? @gerzonc

diegolmello:
@gerzonc Lint is failing

diegolmello (approved):
Great job!

PR #2934, opened 2021-02-22
[FIX] Support chats order for older versions of the server

### LINKED ISSUE #2692: [Bug] "Sort by activity" not working on iOS
### Description:

I've configured my RocketChat clients to sort the channel / user list by activity. In the apps on Ubuntu and MacOS, this works pretty fine (such that the channels and users with most recent activity are listed on top), but on iOS, this does not work: some channels with recent activity are shown where they should be shown, but some are listed at positions I do not understand. For example: one of the channels with most recent activity two days ago is listed where I would expect it to be listed in the MacOS app (the channel listed above had its last activity after that channel, the next in list was active a bit earlier).

That same channel is listed way off in the iOS app: below a user I had the last contact on Dec 06 2019.

Additional information:
- I already tried to resolve this error by clearing my app's local cache (on the iPhone), but that does change anything
- This does not happen for all channels, but only some (I see no connection between them)
- This is not resolved by writing something to that channel on my own

### Environment Information:

- Rocket.Chat Server Version: 3.7.1 (running in a Docker container, using MongoDB 3.4.24)
- Rocket.Chat App Version: 4.12.1.317
- Device Name: iPhone 12 Mini
- OS Version: 14.2.1

### LINKED ISSUE #2781: [Bug] "Chat listed ordered by first massage instead of last massage" on android
### Description:

Android application shows the date of the first chat on the chat page instead of the last chat date. This is while the date on the web and Ubuntu show correctly. Also, if we open the chat, the date will be displayed correctly.

#### Additional information:

- I already this problem by using version 4.12.1 android app on my phone
- This happen for all android user use version 4.12.1

### Environment Information:

- Rocket.Chat Server Version: 3.7
- Rocket.Chat App Version: 4.13.1.22186
- Device Name: Android phone
- OS Version: Android 10

gerzonc:
## Proposed changes

On #2763, we started using the new way of sorting front-end is using, but we dropped support for older versions of the server. All we need to do is check for the current `serverVersion`.

## Issue(s)	

Closes #2692
Closes #2781 

## How to test or reproduce

In order to test this you should be logged in from Web client and the app. In all this cases (with a small exception in discussions creation [*] ) chat order should be the same in Web client and the app

1. Log in
1.1 Login from the Web client
1.2 Login from the app

2. Rooms
2.1 Create a room from the Web client
2.2 Send a message from the Web client
2.3 Leave the room
2.4 Create a new room
2.5 Go back to the old room
2.6 Create a thread and send a message

3. Discussions
3.1 Create a room from the Web client
3.2 Create a discussion in that room from the Web client

## Types of changes

## Checklist

## Further comments

- This PR requires #2922 to be merged in order to apply the simplified server comparison
- [*] Instead of moving the room as the first chat it always moves the discussion as the first; not a real issue because it does moves the discussion and the room to be the first chats, but just to take in note.


diegolmello (approved):
LGTM

PR #2972, opened 2021-03-10
[FIX] E2EE password hiding automatically

kresnaputra:
## Proposed changes

We add some UserPreferences's value for compare the Encryption banner to hiding the banner or not.

## Types of changes

## Checklist


e2e/tests/assorted/09-savepasswordbanner.spec.js
```diff
@@ -0,0 +1,65 @@
+const {
```
diegolmello, line 1: Move these tests to `01-e2eencryption.spec` with the rest of E2E Encryption tests

diegolmello (changes requested):
https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/sagas/encryption.js#L60

In the code above we create and store locally the master password already.
What if we check that key instead of creating a new one to track if it was saved?
Then we could just clear it when the user taps save.

diegolmello (approved):

PR #2992, opened 2021-03-17
[IMPROVE] Subscribe to roles

kresnaputra:
## Proposed changes

If a roles is changed by another user or on another client, the roles's local database isn't change right away. So we need to subscribe or stream the roles for update the local database every time roles is change

## Types of changes

## Checklist


app/views/RoomInfoView/index.js
```diff
@@ -133,13 +133,12 @@ class RoomInfoView extends React.Component {
...
+		const { roles } = this.props;
 		try {
-			const rolesCollection = db.get('roles');
-			const role = await rolesCollection.find(id);
+			const role = roles[id];
 			if (role) {
-				return role.description;
+				return role;
 			}
 			return null;
```
diegolmello, line 143: This whole block is unnecessary after using id as description

app/lib/methods/getRoles.js
```diff
@@ -2,9 +2,20 @@ import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
...
+import reduxStore from '../createStore';
 import protectedFunction from './helpers/protectedFunction';
+import { setRoles as setRolessAction } from '../../actions/roles';
 
-export default function() {
+export async function setRoles() {
+	const db = database.active;
+	const rolesCollection = db.collections.get('roles');
+	const allRoles = await rolesCollection.query().fetch();
+	const parsed = allRoles.reduce((acc, item) => ({ ...acc, [item.id]: item.description }), {});
```
diegolmello, line 13: If description is empty, can you add the id as the description?
Role's descriptions aren't required on admin panel, but they might exist anyway.
This is going to prevent roles from not showing.
Something like this:

```suggestion
	const parsed = allRoles.reduce((acc, item) => ({ ...acc, [item.id]: item.description || item.id }), {});
```

app/lib/methods/getRoles.js
```diff
@@ -2,9 +2,20 @@ import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
 
 import database from '../database';
 import log from '../../utils/log';
+import reduxStore from '../createStore';
 import protectedFunction from './helpers/protectedFunction';
+import { setRoles as setRolessAction } from '../../actions/roles';
```
diegolmello, line 7: Typo on your naming
```suggestion
import { setRoles as setRolesAction } from '../../actions/roles';
```

app/lib/rocketchat.js
```diff
@@ -254,6 +254,41 @@ const RocketChat = {
 
 			this.usersListener = this.sdk.onStreamData('users', protectedFunction(ddpMessage => RocketChat._setUser(ddpMessage)));
 
+			this.rolesListener = this.sdk.onStreamData('stream-roles', protectedFunction(async(ddpMessage) => {
```
diegolmello, line 257: Can you move this whole block to `getRoles` file?

app/lib/rocketchat.js
```diff
@@ -254,6 +254,41 @@ const RocketChat = {
...
+					try {
+						const rolesRecord = await rolesCollection.find(_id);
+						await db.action(async() => {
+							await rolesRecord.destroyPermanently();
+						});
+					} catch (err) {
+						console.log(err);
+					}
+				}
+				setRoles();
```
diegolmello, line 289: When `'stream-roles'` is triggered, one role was changed.
We shouldn't overwrite the whole `roles` reducer because of it.
We should have extra redux actions to change a single role on the reducer.

diegolmello (changes requested):

gerzonc (approved):
LGTM

app/views/RoomMembersView/index.js
```diff
@@ -114,6 +114,13 @@ class RoomMembersView extends React.Component {
 		}
 	}
 
+	componentDidUpdate(_, prevState) {
```
diegolmello, line 117: This is getting called because `subscription` is being updated and not because of the user's roles being changed.
In order to do that, we should listen to changes on `login.user.roles`

app/views/RoomMembersView/index.js
```diff
@@ -114,6 +114,13 @@ class RoomMembersView extends React.Component {
...
+		const { room } = this.state;
+		if (prevState.room.roles !== room.roles) {
```
diegolmello, line 119:
```js
[] !== [] // always true
```

app/lib/methods/getRoles.js
```diff
@@ -2,9 +2,58 @@ import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
...
-export default function() {
+export async function setRoles() {
+	const db = database.active;
+	const rolesCollection = db.collections.get('roles');
+	const allRoles = await rolesCollection.query().fetch();
+	const parsed = allRoles.reduce((acc, item) => ({ ...acc, [item.id]: item.description || item.id }), {});
+	reduxStore.dispatch(setRolesAction(parsed));
+}
+
+export async function subscribeRoles(ddpMessage) {
```
diegolmello, line 19: This function is receiving data from DDP, updating local db and dispatching new values to Redux store.
We should rename it to something like `onRolesChanged`

app/lib/methods/getRoles.js
```diff
@@ -2,9 +2,58 @@ import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
...
+			const rolesRecord = await rolesCollection.find(_id);
+			await db.action(async() => {
+				await rolesRecord.update((u) => {
+					u.description = description;
+				});
+			});
+			reduxStore.dispatch(updateRoles(_id, description));
+		} catch (err) {
+			await db.action(async() => {
+				await rolesCollection.create((post) => {
```
diegolmello, line 34: There's no `try/catch` here.
If it fails, it's going to crash the app.

app/lib/methods/getRoles.js
diegolmello, line 19: Can you catch and log bugs using `log()`?

app/reducers/roles.js
```diff
@@ -0,0 +1,28 @@
...
+				...state,
+				[action.payload.id]: action.payload.desc || action.payload.id
+			};
+		case ROLES.ADD:
+			return {
+				...state,
+				...{ [action.payload.id]: action.payload.desc }
+			};
+		case ROLES.REMOVE:
+			delete state[action.payload.id];
```
diegolmello, line 22: You should never mutate the state.
Try something like this:
```
return state.filter(element => element !== action.payload);
```

https://stackoverflow.com/questions/34582678/is-this-the-correct-way-to-delete-an-item-using-redux

app/reducers/roles.js
```diff
@@ -0,0 +1,28 @@
...
+export default function permissions(state = initialState, action) {
+	switch (action.type) {
+		case ROLES.SET:
+			return action.roles;
+		case ROLES.UPDATE:
+			return {
+				...state,
+				[action.payload.id]: action.payload.desc || action.payload.id
+			};
+		case ROLES.ADD:
```
diegolmello, line 16: `ROLES.UPDATE` and `ROLES.ADD` are doing virtually the same thing.
Can we keep only one?

Pay attention to the `desc || id` logic :)

diegolmello (changes requested):

gerzonc (approved):
LGTM

diegolmello (approved):

Issue #3276, opened 2021-07-15
Search on main page is case sensitive for russian text


Goden4you:
### Description:

It seems like search for channels/users use case sensitivity for russian text. English text works fine
For example imagine that you have channel "ПРОВЕРКА". And your search will be like this:
1) input - "Пр"
    expected result - "ПРОВЕРКА"
    real result - nothing
<img width="361" alt="Снимок экрана 2021-07-15 в 14 47 21" src="https://user-images.githubusercontent.com/44268968/125783137-f03faf33-656c-4f7b-b441-2b25a185f895.png">
<img width="359" alt="Снимок экрана 2021-07-15 в 14 47 40" src="https://user-images.githubusercontent.com/44268968/125783141-da533034-c389-4c07-aca5-2fbb44baafb4.png">
<img width="362" alt="Снимок экрана 2021-07-15 в 14 47 51" src="https://user-images.githubusercontent.com/44268968/125783144-961ad86d-6a46-4b2f-9924-8396eb850451.png">

### Environment Information:

- Rocket.Chat Server Version: 3.13.2
- Rocket.Chat App Version: 4.17.0.24389
- Device Name: iPhone XS
- OS Version: iOS 14.4

### Steps to reproduce:

1. Go to main page
2. Click on search icon
3. Try to find channel with case insensitivity by russian words

### Expected behavior:

Search must be case insensitive for all languages

### Actual behavior:

Search sensitive for russian words

diegolmello:
Hey, @Goden4you.
Can you test on web client and shows the results?
Thank you.

Goden4you:
@diegolmello 
<img width="286" alt="Снимок экрана 2021-07-15 в 15 52 05" src="https://user-images.githubusercontent.com/44268968/125791172-13e0a26a-922f-448d-8e99-b6528c77db96.png">

There is no problem on web and desktop. Only mobile

ankar84:
I don't know, but working good for me at 4.13
![IMG_20210720_120834.jpg](https://user-images.githubusercontent.com/4023037/126265329-75735deb-7e7b-4bc5-bdaf-5768c939448b.jpg)

Goden4you:
@ankar84 your code of "localSearch" and "search" in app/lib/rocketchat.js is the same as in this repository? If not, can you provide your code of this func?

ankar84:
[...]

PR #3309, opened 2021-07-29
feat: case insensitive for non-ASCII text on main search

reinaldonetof:
## Proposed changes

The proposed change is to create a new column in the database called "sanitized_fname." This column is for the local database only and will store the "fname || name" variable after slugification.

We will use "fname || name" because fname is the string of the channel, discussion, or user that accepts special characters, but in some cases, such as with the #General channel, there is no fname, only the name. If the UTF-8 slugify option in the Workspace admin is enabled, the name will already come with slugify.

The "limax" library is responsible for slugifying in the backend, but I couldn't make it work in React Native. Therefore, I searched for a library that could slugify characters in Russian, traditional Chinese, Japanese, and also Arabic, and I found the "transliteration" library, as can be seen in the unit tests in utils.test.ts.

Regarding the "localSearchSubscription" function, I added a new comparison related to "sanitized_fname," but I needed to keep the other comparisons. When this new database migration is added, only the "sanitized_fname" column is added, and the slugify is not done for all subscriptions already present in the local db. The slugify of these subscriptions will be done as the user uses the app, for example, when entering a room, receiving a new message, or changing the room name. Therefore, over time, all subscriptions will have the sanitized_fname column written, except for very old rooms that the user has never opened or changed. Another case to have all sanitized_fname is to perform a clear cache, which would also slugify "fname || name" when rooms are created in the local db.

Why all of this?
Well, it all started with a community issue reporting that local search was case-sensitive for Cyrillic characters on iOS, while on Android, this search was case-insensitive. So, I researched if it was possible to change WatermelonDB or FMDB (the objective C library used by WatermelonDB for iOS), but the result was that it would be much more complex. In WatermelonDB documentation, there's a note saying: "For instance, while the comparison is case-insensitive, SQLite cannot by default compare non-ASCII characters case-insensitively (unless you install ICU extension)". However, I could not find a step-by-step guide to install this extension, and as it's only for iOS, we would have to change a lot of native code and keep maintaining it every time we updated the library. So, I abandoned this idea and went for the slugify approach.

Although there is only one open issue from a while ago, this fix is very important. Otherwise, users who have rooms with non-Latin characters will have to search through the list or remember upper and lower case letters.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/3276

## How to test or reproduce

- Create a channel `ПРОВЕРКА`
- On `RoomListView` click on the search icon
- Write `Пр`
- Should return the channel created

## Screenshots
<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</thead>
</tr>
<body>
<tr>
<td>

</td>
<td>

</td>
</tr>
</tbody>
</table>

## Types of changes

## Checklist

## Further comments

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/3276
[TC-394]

[TC-394]: https://rocketchat.atlassian.net/browse/TC-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


diegolmello:
@reinaldonetof Is this still valid?

GleidsonDaniel (approved):

GleidsonDaniel (changes requested):
Two questions:
This slugfied fname merge will only work with new rooms correct? The old ones continued without this slugfied fname.

As you changed the db, can you create a test for this task?
Creating a room with Russian characters and then validating that its search works.

app/lib/methods/helpers/mergeSubscriptionsRooms.ts
```diff
@@ -98,7 +99,7 @@ export const merge = (
 	mergedSubscription.blocker = !!mergedSubscription.blocker;
 	mergedSubscription.blocked = !!mergedSubscription.blocked;
 	mergedSubscription.hideMentionStatus = !!mergedSubscription.hideMentionStatus;
-
+	mergedSubscription.sanitizedFname = slugifyLikeString(mergedSubscription.fname || mergedSubscription.name);
```
diegolmello, line 102: What's going to happen if both fname and name are null? Does `slugifyLikeString` with empty param return an error?

app/lib/methods/search.ts
```diff
@@ -15,10 +15,17 @@ export const localSearchSubscription = async ({ text = '', filterUsers = true, f
...
 	const db = database.active;
 	const likeString = sanitizeLikeString(searchText);
+	const slugifiedString = slugifyLikeString(searchText);
 	let subscriptions = await db
 		.get('subscriptions')
 		.query(
-			Q.or(Q.where('name', Q.like(`%${likeString}%`)), Q.where('fname', Q.like(`%${likeString}%`))),
+			Q.or(
+				Q.where('sanitized_fname', Q.like(`%${slugifiedString}%`)),
+				// TODO: We still need the others comparisons, otherwise we needed to clear the cache
```
diegolmello, line 24: This comment doesn't quite document what's happening nor creates an action for the future.
Can you elaborate?

app/lib/database/utils.ts
```diff
@@ -1,7 +1,15 @@
 import XRegExp from 'xregexp';
+import slug from 'slug';
 
 // Matches letters from any alphabet and numbers
 const likeStringRegex = XRegExp('[^\\p{L}\\p{Nd}]', 'g');
 export const sanitizeLikeString = (str?: string): string | undefined => str?.replace(likeStringRegex, '_');
 
+export const slugifyLikeString = (str?: string) => {
```
diegolmello, line 8: Can you comment this?
I'm watching this PR from distance, but I still don't understand why you need this.
It's even worse that you're adding a new function here, but you're not updating `utils.test.ts`.
Remember to write code for future us.

app/definitions/ISubscription.ts
```diff
@@ -46,6 +46,7 @@ export interface ISubscription {
 	ls: Date;
 	name: string;
 	fname?: string;
+	sanitizedFname: string;
```
diegolmello, line 49: You created a new column on db as optional, but here you're declaring as required.

diegolmello (changes requested):
Besides my comments, please elaborate your PR description to make it clear what you've done on the patch as well the other things.
This PR touches sensitive parts of the app and I'm not quite following the motivation behind it.

diegolmello:
Also the title shouldn't be `russian` only. It's non-latin.

app/lib/methods/helpers/mergeSubscriptionsRooms.ts
reinaldonetof: the function will return an empty string

patches/@nozbe+watermelondb+0.23.0.patch
```diff
@@ -1,3 +1,29 @@
+diff --git a/node_modules/@nozbe/watermelondb/adapters/sqlite/encodeQuery/index.js b/node_modules/@nozbe/watermelondb/adapters/sqlite/encodeQuery/index.js
+index 4d49faf..262edb5 100644
+--- a/node_modules/@nozbe/watermelondb/adapters/sqlite/encodeQuery/index.js
++++ b/node_modules/@nozbe/watermelondb/adapters/sqlite/encodeQuery/index.js
+@@ -60,6 +60,10 @@ var encodeComparison = function (table, comparison) {
+     return right.values ? "between ".concat((0, _encodeValue.default)(right.values[0]), " and ").concat((0, _encodeValue.default)(right.values[1])) : '';
+   }
+ 
++  if (comparison.operator === 'like' && !!comparison.right.value && typeof comparison.right.value === 'string') {
```
diegolmello, line 9: On the PR description you said you dropped the WMDB idea in favor of the slugify one.
Yet there's this change.
What does it do?

e2e/tests/room/01-createroom.spec.ts
```diff
@@ -1,9 +1,21 @@
 import { device, waitFor, element, by, expect } from 'detox';
 
-import { tapBack, navigateToLogin, login, platformTypes, TTextMatcher, tapAndWaitFor } from '../../helpers/app';
+import { tapBack, navigateToLogin, login, platformTypes, TTextMatcher, tapAndWaitFor, sleep } from '../../helpers/app';
 import { createRandomUser } from '../../helpers/data_setup';
 import random from '../../helpers/random';
 
+async function searchRoomNonLatin(room: string, roomTestID: string) {
```
diegolmello, line 7: Why do you need this function?
If `searchRoom` isn't enough, you should not create another one in a separate file.
You should consider changing its behavior.

e2e/tests/room/01-createroom.spec.ts
```diff
@@ -261,6 +276,61 @@ describe('Create room screen', () => {
 					.withTimeout(60000);
 				await expect(element(by.id(`rooms-list-view-item-${room}`))).toExist();
 			});
+
+			it('should create a room with non-latin alphabet and do a case insensitive search for it', async () => {
+				const randomValue = random();
+				const roomName = `ПРОВЕРКА${randomValue}`;
+				const roomNameLower = `проверка${randomValue}`;
```
diegolmello, line 283: `ПРОВЕРКА${randomValue}`.toLowerCase()

e2e/tests/room/01-createroom.spec.ts
```diff
@@ -261,6 +276,61 @@ describe('Create room screen', () => {
...
+				await expect(element(by.id(`room-view-title-${roomName}`))).toExist();
+				await tapBack();
+				await waitFor(element(by.id('rooms-list-view')))
+					.toExist()
+					.withTimeout(2000);
+				await waitFor(element(by.id(`rooms-list-view-item-${roomName}`)))
+					.toExist()
+					.withTimeout(60000);
+				await expect(element(by.id(`rooms-list-view-item-${roomName}`))).toExist();
+				await searchRoomNonLatin(`${roomNameLower}`, `rooms-list-view-item-${roomName}`);
```
diegolmello, line 328: Why do you need to specify a lowercase text?
Isn't dealing with case insensitive the whole goal of this PR?

diegolmello (changes requested):

app/lib/methods/search.ts
```diff
@@ -15,10 +15,24 @@ export const localSearchSubscription = async ({ text = '', filterUsers = true, f
...
+				// For those rooms that were already on local db the columns will begins as null,
+				// and each interact with the room will sanitize the fname or name and add to the sanitized_fname.
+				// The interacts can be:
+				// - open the room
+				// - receive any updates as changing the name, last message
+				// - join or create a new room
+				// Otherwise, if we want to only search for sanitized_fname, we should clear the local DB
+				Q.where('name', Q.like(`%${likeString}%`)),
+				Q.where('fname', Q.like(`%${likeString}%`)),
+				Q.where('name', Q.like(`%${slugifiedString}%`))
```
diegolmello, line 34: I didn't get this one either.

app/lib/database/utils.test.ts
```diff
@@ -39,3 +39,30 @@ describe('sanitizer', () => {
 		expect(utils.sanitizer(content)).toBe(content);
 	});
 });
+
+describe('slugifyLikeString', () => {
+	test('render empty', () => {
+		expect(utils.slugifyLikeString('')).toBe('');
+		expect(utils.slugifyLikeString(undefined)).toBe('');
+	});
+	test('slugify the russian alphabet', () => {
```
diegolmello, line 48: Missing latin test

package.json
```diff
@@ -146,6 +146,7 @@
 		"rn-fetch-blob": "^0.12.0",
 		"rn-root-view": "RocketChat/rn-root-view",
 		"semver": "^7.3.8",
+		"transliteration": "^2.3.5",
```
diegolmello, line 149: I'm pretty sure we added a slug lib a while ago

e2e/tests/room/01-createroom.spec.ts
reinaldonetof: Yeap, the room name is uppercase and we are searching only in lowercase. Also, the testID uses the room name in uppercase. So, the test is: we are searching for a room expecting the result is case insensitive.

e2e/tests/room/01-createroom.spec.ts
reinaldonetof: Why? This is specific for this type of test, I'm testing if the searching is case insensitive for non-ASCII

app/lib/methods/search.ts
reinaldonetof: Because, as the sanitized_fname isn't filled yet, the user could be on a `channel or team`(only those, because the name is slugified by the server), and while the sanitized_fname isn't done, the search could find through the last condition.

package.json
reinaldonetof: This war the PR that introduced another type of "slug" https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4364
That library works well with URI and parser. However, I was not able to use it as the library to slugify the non-ASCII character to ASCII. Also, I tested with other libraries, but the transliteration was the only one that was able to convert Russian, Arabic, Chinese, and Japanese characters to ASCII characters.

GleidsonDaniel (approved):

GleidsonDaniel (approved):
Now it got better, removed the need for two functions.

PR #3313, opened 2021-08-03
[NEW] Support new message parser

gerzonc:
## Proposed changes

Adds support for the new message parser.

## How to test or reproduce

In order to test this feature you need to check that you have the new message parser enabled from the server global preferences and in the app go to Profile > Preferences > Enable Message Parser

## Types of changes

## Checklist

## Further comments

Some edge cases may fail because they're not yet part of the message parser (e.g. formatted links)

### DIFF
--- a/__tests__/__snapshots__/Storyshots.test.js.snap
+++ b/__tests__/__snapshots__/Storyshots.test.js.snap
@@ -12073,7 +12073,6 @@ exports[`Storyshots LoadMore black theme 1`] = `
                     Array [
                       Object {
                         "backgroundColor": "transparent",
-                        "flexShrink": 1,
                         "fontFamily": "System",
                         "fontSize": 16,
                         "fontWeight": "400",
@@ -12093,7 +12092,6 @@ exports[`Storyshots LoadMore black theme 1`] = `
                       Array [
                         Object {
                           "backgroundColor": "transparent",
-                          "flexShrink": 1,
                           "fontFamily": "System",
                           "fontSize": 16,
                           "fontWeight": "400",
@@ -12178,7 +12176,6 @@ exports[`Storyshots LoadMore black theme 1`] = `
                     Array [
                       Object {
                         "backgroundColor": "transparent",
-                        "flexShrink": 1,
                         "fontFamily": "System",
                         "fontSize": 16,
                         "fontWeight": "400",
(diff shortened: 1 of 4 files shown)


app/views/UserPreferencesView/index.js
```diff
@@ -1,48 +1,82 @@
...
+				</List.Section>
+				<List.Section>
+					<List.Separator />
+					<List.Item
+						title='Enable_Message_Parser'
+						testID='preferences-view-enable-message-parser'
+						right={() => renderMessageParserSwitch()}
+					/>
+					<List.Separator />
+				</List.Section>
```
gerzonc, line 72: In order to see it we should validate if we have enabled the new message parser on the server global preferences

app/containers/message/Content.js
```diff
@@ -97,6 +98,9 @@ const Content = React.memo((props) => {
 	if (prevProps.msg !== nextProps.msg) {
 		return false;
 	}
+	if (prevProps.md !== nextProps.md) {
```
diegolmello, line 101: `md` is an object

app/containers/markdown/MessageBody/index.js
```diff
@@ -0,0 +1,69 @@
...
+const isBigEmoji = tokens => tokens.length === 1 && tokens[0].type === 'BIG_EMOJI';
+
+const Body = ({
+	tokens, mentions, channels, navToRoomInfo, style
+}) => {
+	if (isBigEmoji(tokens)) {
+		return <BigEmoji value={tokens[0].value} />;
+	}
+
+	return (
```
diegolmello, line 23: You're not applying styles to text.
New markdown should be rendered with the same styles of the old one.
The only difference is the parser.

diegolmello (changes requested):

app/containers/markdown/MessageBody/Heading.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Text } from 'react-native';
+import PropTypes from 'prop-types';
+import { themes } from '../../../constants/colors';
+import styles from '../styles';
+import { useTheme } from '../../../theme';
+
+const Heading = ({ value, level }) => {
+	const { theme } = useTheme();
+	const textStyle = styles[`heading${ level }`];
```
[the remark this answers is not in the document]
gerzonc: I suppose so, yes. But I don't think that should be possible because you always receive the level when it's a heading, if you don't then the parser is not supposed to identify it as a heading but something else. 🤔

Issue #3088, opened 2021-04-24
Sharing content results in Black screen 


avmentzer:
### Description:

Sharing anything via Rocket.Chat results in a black screen 

### Environment Information:

- Rocket.Chat Server Version: 3.13.2
- Rocket.Chat App Version: 4.16.1
- Device Name:  IPhone XR
- OS Version:  14.4.2

### Steps to reproduce:

Share anything via Rocket.Chat 

### Expected behavior:

No blank screen 

### Actual behavior:

blank screen 
![DA009225-D901-4DAB-94CF-A678ECABC37A](https://user-images.githubusercontent.com/74241215/115965829-3b2e5100-a52b-11eb-967d-8e0767c6fa92.jpeg)
![300DF5B1-55C4-4B47-9ECE-6E4BEDFC6659](https://user-images.githubusercontent.com/74241215/115965830-3cf81480-a52b-11eb-90c8-c13a58a8ae20.png)

sumukhah:
Hello @avmentzer, I tested this with (app: 4.17, server: 3.14), but could not reproduce the issue. Can you share a video record it the error still exist? 
Thank you

avmentzer:
I already provided a Video to official support.
As you can see @ the Screenshot the screen stays blank, if I try to share sth via Rocket.Chat (the red button)

Von meinem iPhone gesendet

cprice-kgi:
I ran into this same issue when testing 3.14. I looked into it and it appears the server setting for mobile screen lock is causing this issue. When it is turned on, I also get the black screen when sharing. I fixed it in my code and will open a PR for the fix.

If you have screen lock enabled on the server side, this could be the issue you are seeing as well.

PR #3320, opened 2021-08-06
[FIX] Black screen during Share

cprice-kgi:
## Proposed changes
Something about RNSplashScreen.hide is causing an issue here. I wrapped the call in a try/catch.

## Issue(s)	
#3088

## How to test or reproduce
Try to share an image either from within the app or from an another application to the app. After investigation, I have found this issue appears only when the server has the mobile screen lock value enabled. If it is disabled, sharing will work fine. If enabled, the call to RNSplashScreen.hide will cause a black screen to appear, and the user is unable to proceed.

## Types of changes

- [ x] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ x] Lint and unit tests pass locally with my changes

## Further comments
N/A


LevyCosta (approved):
lgtm

app/utils/localAuthentication.js
```diff
@@ -103,7 +104,11 @@ export const localAuthenticate = async(server) => {
 	// if screen lock is enabled
 	if (serverRecord?.autoLock) {
 		// Make sure splash screen has been hidden
-		await RNBootSplash.hide();
+		try {
+			await RNBootSplash.hide();
+		} catch (error) {
+			logEvent(events.LC_SERVER_RECORD_BOOT_SPLASH_HIDE_ERROR);
```
diegolmello, line 110: @reinaldonetof This is wrong imo.
If we want to log something, let's log the error.

app/views/AttachmentView.js
```diff
@@ -106,9 +106,7 @@ class AttachmentView extends React.Component {
 	handleSave = async () => {
 		const { attachment } = this.state;
 		const { user, baseUrl } = this.props;
-		const {
-			title_link, image_url, image_type, video_url, video_type
-		} = attachment;
+		const { title_link, image_url, image_type, video_url, video_type } = attachment;
```
diegolmello, line 109: Unrelated, but prettier asked me to change.
It won't hurt :)

diegolmello:
The issue is that we're calling `hide` without having a splash screen on share extension.
That triggers `react-native-bootsplash has not been initialized`, similar to https://github.com/zoontek/react-native-bootsplash/issues/264
Recently they wrapped the `hide` call in a `try/catch` in the example https://github.com/zoontek/react-native-bootsplash/commit/41e32e8965b8808a3d0db5a32a56c4228fdab863#diff-f92b617bbc90bafc01461177b82175f320578d91293fddfbc810bacb339898edR20

diegolmello:
@cprice-kgi Merged. Thanks for your contribution :)

jsalatiel:
Is this the same problem that happens when one tries to share a audio received on whatsapp with someone on rockechat ? After choosing the contact for share, rocketchat gets  stuck on a blank window.  ( on android , running version 4.20)

PR #3324, opened 2021-08-11
[FIX] Server settings API limit increase

cprice-kgi:
## Proposed changes

This change passes a count limit (100) to the settings.public API call in order to ensure ALL server settings are retrieved by the app. I'm guessing the number of these has increased over time (?) and the default limit of 50 does not provide an appropriate amount of records (I believe there are 53 returned on our server instance).

## How to test or reproduce

In testing, we noticed that a few of the server side settings (e.g. "UI_Use_Real_Name") were not being picked up on the app (they were undefined). After much tracking, I found out that these settings were being excluded due to an insufficient default record limit in the API call.

## Types of changes

## Checklist

## Further comments

server-settings-api-limit-increase

### DIFF
--- a/app/lib/methods/getSettings.js
+++ b/app/lib/methods/getSettings.js
@@ -140,7 +140,7 @@ export default async function() {
 		const db = database.active;
 		const settingsParams = Object.keys(settings).filter(key => !loginSettings.includes(key));
 		// RC 0.60.0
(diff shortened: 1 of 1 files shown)


diegolmello:
Thanks for opening, but it was fixed on https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/3277

PR #3351, opened 2021-08-26
[IMPROVE] Fetch members from API endpoint

gerzonc:
## Proposed changes

Currently we've been using a method call for fetching members, on `RoomMembersView`, but we should use REST API like the Web client does.

## How to test or reproduce

In order to test this you should create a group, direct message with multiple users or team, and server version being at least 3.16.0

1. Go to a room
2. Tap on room's title
3. Tap on Members

It should render the members that are online from that room

## Types of changes

## Checklist


app/lib/rocketchat.js
```diff
@@ -1032,7 +1032,27 @@ const RocketChat = {
...
+			return this.sdk.get('im.members', params);
+		}
 
+		if (roomType === 'p') {
+			return this.sdk.get('groups.members', params);
+		}
+
+		if (roomType === 'c') {
+			return this.sdk.get('channels.members', params);
+		}
```
diegolmello, line 1054: You should use `roomTypeToApiType` for this, like https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/rocketchat.js#L1056

app/lib/rocketchat.js
```diff
@@ -1032,7 +1032,27 @@ const RocketChat = {
 		// RC 0.42.0
 		return this.methodCallWrapper('getUsersOfRoom', rid, allUsers, { skip, limit });
 	},
+	getMembers(rid, roomType, type, debouncedText, start = 0, end = 10) {
+		const params = {
+			roomId: rid,
+			offset: start,
+			count: end,
+			...(type !== 'all' && { 'status[]': type }),
+			...(debouncedText && { filter: debouncedText })
```
diegolmello, line 1041: This function doesn't need to know it's a debounced text.

app/lib/rocketchat.js
```diff
@@ -1032,7 +1032,27 @@ const RocketChat = {
 		// RC 0.42.0
 		return this.methodCallWrapper('getUsersOfRoom', rid, allUsers, { skip, limit });
 	},
+	getMembers(rid, roomType, type, debouncedText, start = 0, end = 10) {
```
diegolmello, line 1035: There's no written rule for that, but there's enough params to be an object instead :)
Also, I didn't get why we gave `offset/count` a new naming.

diegolmello (changes requested):

app/lib/rocketchat.js
gerzonc: No reason actually 🤔 Was using the web implementation as reference, I'll change that :)

app/views/RoomMembersView/index.js
```diff
@@ -389,16 +391,31 @@ class RoomMembersView extends React.Component {
...
 		if (isLoading || end) {
 			return;
 		}
 
 		this.setState({ isLoading: true });
 		try {
-			const membersResult = await RocketChat.getRoomMembers(rid, allUsers, members.length, PAGE_SIZE);
-			const newMembers = membersResult.records;
+			if (compareServerVersion(serverVersion, '3.16.0', methods.greaterThanOrEqualTo)) {
+				const membersResult = await RocketChat.getMembers({
```
diegolmello, line 406: @gerzonc how about reusing the same function name and we just check the server version?
We would have `getRoomMembers` for both.

diegolmello (approved):
lgtm

Issue #3002, opened 2021-03-21
Incoming call, alarm clock etc. trashes voice message recording


jacotec:
### Description:

I'm getting many complaints regarding a reproducible issue when voice messages are being recorded.

When someone is recording a voice message, did already 10 minutes, and i.e. some of these things are happening:

- Incoming call
- Alarm clock of the smart phone fires
- A timer runs off
- iPhone popup asks for network provider updates
- ...

The voice message is fully trashed into nowhere. Users are losing the complete message at this time.

In case of an incoming call the app should save the message at this point and give the option to at least send the message up to this point after returning to the app when the disturbing event is over (like in WhatsApp) or don't bother about the incoming call and just continue to record (like in Signal).

### Environment Information:

- Rocket.Chat Server Version: 3.12.1
- Rocket.Chat App Version: 4.15.0
- Device Name: iPhone, iPad
- OS Version: IOS 14.4.1

### Steps to reproduce:

1. Record a voice message
2. Make a phone call to the device which is recording
3. The voice message recording stops and all recording which was done is gone

### Expected behavior:

Either the recording should continue (and the user can decide to send it in case he wants to pick up the call) or the recording is temporary saved and can be sent after the disturbing event like WhatsApp does. However, it should never be just trashed.

### Actual behavior:

All recorded content is immediately stopped, trashed and will be lost.

### Additional context:

After a recording is lost due to this, in addition the recording function is not working after that anymore. The user must force-quit the complete app and restart it to restore voice recording functionality.

jacotec:
@diegolmello I'm on this, solution works in my tests. Will make a few more crosstests and make a PR then.

jacotec:
#3388 fixes this.

PR #3388, opened 2021-09-17
[FIX] Preserve voice message if recording is interrupted on IOS

jacotec:
## Proposed changes
When an event interrupts an ongoing voice recording in IOS, the OS stops the recording. This was not handled properly before: The message was immediately completely lost and it was not possible to make a new voice message before force-quitting and restarting the app (because the Expo recorder was left in loaded state).

This PR stops the recording in this case, indicating it was stopped. The user can send or discard the voice message up to that point at any time afterwards.

I also changed the GUI to make it more clear. See below.

## Issue(s)	
#3002 

## How to test or reproduce
Run the version on an iPhone. Start a voice message recording, while the recording is running, call the iPhone or activate Siri.
IOS stops the recording.
Old behavior: The recorded message was trashed immediately, after that you can't start a new recording (as the Expo recorder was left loaded) - you need to fully restart the app.
My behavior: The UI indicates that the recording is stopped. After the interruption you can still send or trash it at any time. Expo Recorder will not be stuck in loaded state.

## Screenshots

Record in progress: I've replaced the red cross and green check by a red trash icon and a green send icon. IMHO this is much more clear: The green icon in fact immediately **sends** the message, the red icon immediately **trashes** the message (all my users highly appreciate this change as well).
Behind the time display the "record" symbol states that the recording is running.

This is an incoming call. As soon as IOS stops the recording, I'll indicate this with the "pause" symbol behind the time display. The time is stopped and shows the length of the message.

After the call was finished, you can still send or trash the waiting message at any time:

## Types of changes

## Checklist

## Further comments
Would be fine to have this in the next release together with my previous fix merge to have the voice message fixes complete.

### DIFF
--- a/app/containers/MessageBox/RecordAudio.js
+++ b/app/containers/MessageBox/RecordAudio.js
@@ -59,17 +59,19 @@ export default class RecordAudio extends React.PureComponent {
 	constructor(props) {
 		super(props);
 		this.isRecorderBusy = false;
+		this.LastDuration = 0;
 		this.state = {
 			isRecording: false,
+			isRecorderActive: false,
 			recordingDurationMillis: 0
 		};
 	}
 
 	componentDidUpdate() {
 		const { recordingCallback } = this.props;
-		const { isRecording } = this.state;
+		const { isRecorderActive } = this.state;
 
-		recordingCallback(isRecording);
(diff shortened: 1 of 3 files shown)


diegolmello:
To be merged on #3397

PR #3447, opened 2021-10-19
Chore: Migrate MessagesView to Typescript

AlexAlexandre:
## Proposed changes
The proposal of this PR is migrate `MessagesView` to TypeScript.

## Issue(s)	
N/A

## How to test or reproduce
On the app, check if the items below works normally:
* messages list;
* file messages list;
* mentions messages list;
* starred messages list;
* pinned messages list;

## Screenshots
N/A

## Types of changes

## Checklist

## Further comments
N/A


app/views/MessagesView/index.tsx
```diff
@@ -48,7 +77,7 @@ class MessagesView extends React.Component {
 		this.load();
 	}
 
-	shouldComponentUpdate(nextProps, nextState) {
+	shouldComponentUpdate(nextProps: any, nextState: any) {
```
gerzonc, line 80: `shouldComponentUpdate` params actually are supposed to have the same properties as the component, you can use `IMessagesViewProps` as `nextProps` type, for example.

app/views/MessagesView/index.tsx
```diff
@@ -73,17 +102,17 @@ class MessagesView extends React.Component {
...
 			return;
 		}
 		navigation.navigate('RoomInfoView', navParam);
 	};
 
-	jumpToMessage = async ({ item }) => {
+	jumpToMessage = async ({ item }: { item: IMessageItem }) => {
 		const { navigation, isMasterDetail } = this.props;
-		let params = {
+		let params: any = {
```
gerzonc, line 115: Can't we create an `IParams` interface for this? It seems it has values by default so we are able to know which values it receives.

app/views/MessagesView/index.tsx
```diff
@@ -107,9 +136,9 @@ class MessagesView extends React.Component {
 		}
 	};
 
-	defineMessagesViewContent = name => {
+	defineMessagesViewContent = (name?: string) => {
```
gerzonc, line 139: This is not supposed to be optional.

app/views/MessagesView/index.tsx
```diff
@@ -227,7 +257,7 @@ class MessagesView extends React.Component {
 		}
 	};
 
-	getCustomEmoji = name => {
+	getCustomEmoji = (name: any) => {
```
gerzonc, line 260: `name: { [name: string]: string }` ?

app/views/MessagesView/index.tsx
```diff
@@ -257,8 +287,8 @@ class MessagesView extends React.Component {
 		try {
 			const result = await this.content.handleActionPress(message);
 			if (result.success) {
-				this.setState(prevState => ({
-					messages: prevState.messages.filter(item => item._id !== message._id),
+				this.setState((prevState: any) => ({
```
gerzonc, line 290: `IState`

app/views/MessagesView/index.tsx
```diff
@@ -306,12 +336,13 @@ class MessagesView extends React.Component {
...
-const mapStateToProps = state => ({
+const mapStateToProps = (state: any) => ({
 	baseUrl: state.server.server,
 	user: getUserSelector(state),
 	customEmojis: state.customEmojis,
 	useRealName: state.settings.UI_Use_Real_Name,
 	isMasterDetail: state.app.isMasterDetail
 });
 
+// @ts-ignore
```
gerzonc, line 347: Why?

gerzonc (changes requested):

app/views/MessagesView/index.tsx
AlexAlexandre: to resolve that, I needed change the `customEmojis` to receive an index signature... 
the code was like this: `customEmojis: { [key: string]: string };` and the `getCustomEmoji = (name: string) => {`

app/views/MessagesView/index.tsx
AlexAlexandre: When I started the typescript migration, I didn't have all the knowledge about the code base.
So the `app/containers/ActionSheet/Provider.tsx` file only received `Functional Components` but in fact we use it with class components too. 

I was not planning do this change in this PR, to do it separate in the future, but I fixed. 
Thanks.

diegolmello (approved):

PR #3459, opened 2021-10-23
Chore: Migrate ShareListView to Typescript

reinaldonetof:
## How to test or reproduce

- You should go to mobile's gallery and share with Rocket Chat and the list should show properly your rooms

## Types of changes

## Checklist


app/presentation/DirectoryItem/index.tsx
```diff
@@ -21,9 +21,9 @@ interface IDirectoryItem {
 	type: string;
 	onPress(): void;
 	testID: string;
-	style: any;
-	rightLabel: string;
-	rid: string;
+	style?: any;
```
gerzonc, line 24: `style?: ViewStyle`

app/views/ShareListView/Header/Header.android.tsx
```diff
@@ -24,7 +23,13 @@ const styles = StyleSheet.create({
 	}
 });
 
-const Header = React.memo(({ searching, onChangeSearchText, theme }) => {
+interface IHeader {
+	searching: boolean;
```
gerzonc, line 27: This should be optional

app/views/ShareListView/Header/Header.android.tsx
```diff
@@ -24,7 +23,13 @@ const styles = StyleSheet.create({
...
+	onChangeSearchText(text: string): void;
```
gerzonc, line 28: I always have my doubts on this case if we should just rename the function (so we can just extend `TextInputProps`) or use `onChangeSearchText: TextInputProps['onChangeText']` in order to keep the types as the component actually receives (of course, we know it's a function that takes `text: string` as param and returns `void`, but that doesn't mean it could change in the future). We should keep this as future proof as possible, so either you can rename `onChangeSearchText` to `onChangeText` and extend `TextInputProps` or make it `onChangeSearchText: TextInputProps['onChangeText']`

app/views/ShareListView/Header/Header.ios.tsx
```diff
@@ -16,10 +15,18 @@ const styles = StyleSheet.create({
 	}
 });
 
-const Header = React.memo(({ searching, onChangeSearchText, initSearch, cancelSearch, theme }) => {
+interface IHeader {
+	searching: boolean;
```
gerzonc, line 19: Optional as well

app/views/ShareListView/Header/Header.ios.tsx
```diff
@@ -16,10 +15,18 @@ const styles = StyleSheet.create({
...
+	onChangeSearchText(text: string): void;
```
gerzonc, line 20: Same comment as Android

app/views/ShareListView/Header/index.tsx
```diff
@@ -1,10 +1,19 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 
+// @ts-ignore
+// eslint-disable-next-line import/extensions,import/no-unresolved
```
gerzonc, line 4: No need. There are two ways to solve this: You can either omit one platform (it'll still detect the other one, e.g. if you omit iOS it'll keep working fine, since Android is still specified and it will detect if is Android or not) or you can create a type definition file (e.g. `Header.d.ts`) so it can detect it.

app/views/ShareListView/Header/index.tsx
```diff
@@ -1,10 +1,19 @@
...
 
-const ShareListHeader = React.memo(({ searching, initSearch, cancelSearch, search, theme }) => {
-	const onSearchChangeText = text => {
+interface IShareListHeader {
+	searching: boolean;
+	initSearch?: () => void;
+	cancelSearch?: () => void;
+	search(text: string): void;
+	theme: string;
+}
```
gerzonc, line 13: What do you think of creating a separate `interface.ts` file for an interface that you can share across the three files?

app/views/ShareListView/index.tsx
```diff
@@ -25,24 +25,75 @@ import { sanitizeLikeString } from '../../lib/database/utils';
...
+}
+interface IState {
+	searching: boolean;
+	searchText: string;
+	searchResults: IChat[];
+	chats: IChat[];
+	serversCount: number;
+	attachments: IAttachment[];
+	text: string;
+	loading: true;
```
gerzonc, line 66: ?

app/views/ShareListView/index.tsx
```diff
@@ -178,32 +229,33 @@ class ShareListView extends React.Component {
 		});
 	};
 
-	// eslint-disable-next-line react/sort-comp
-	internalSetState = (...args) => {
+	internalSetState = (...args: any[]) => {
```
gerzonc, line 232: It's not an array, it's an object. And since that's the case I think (I'm not sure, that would be something to test), but, what if you just leave it as `any` and then set `this.setState` as `this.setState({ ... args })`?

app/views/ShareListView/index.tsx
```diff
@@ -178,32 +229,33 @@ class ShareListView extends React.Component {
...
 		const { navigation } = this.props;
 		if (navigation.isFocused()) {
 			animateNextTransition();
 		}
+		// @ts-ignore
```
gerzonc, line 237: Follow the previous comment, that way you prevent this.

gerzonc (changes requested):

app/views/ShareListView/Header/index.tsx
reinaldonetof: I could fix this with this comment: https://stackoverflow.com/a/59972215 when he talks the `Update`

app/views/ShareListView/index.tsx
reinaldonetof: I just changed this any[] to object[], but the TS continues to complain about the array. When we migrate the roomview, that has the same function, we refactor here too!

app/views/ShareListView/Header/Header.android.tsx
reinaldonetof: I fixed this with https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/3459/files#diff-57cd459b4f487e2435bae7298f83fba2a0ca18d7c60c1d965bbffd27cf9ac304R3

app/views/ShareListView/Header/index.tsx
gerzonc: ```You can either omit one platform (it'll still detect the other one, e.g. if you omit iOS it'll keep working fine, since Android is still specified and it will detect if is Android or not)```

That was what I was referring to :)

gerzonc (approved):
LGTM

PR #3470, opened 2021-10-29
[NEW] Preview or download attachments

AlexAlexandre:
## Proposed changes
Currently, when we try open a document (pdf / word / unsupported video files), we open the browser and expose the user token on the url. 
The propose of this PR, is open a native app to show this documents or download the file to the device.

## Issue(s)	
N/A

## How to test or reproduce
* on `mobile` server, look for `TestAuthenticateFiles` room, then test all files;
* you can test on `open` server as well, but the `mobile` server already has all type files. 

#### Regression test
* open a video and image, and try download it.

## Screenshots

<table>
    <thead>
    <tr>
        <th>Before</th>
    </tr>
    </thead>
    <tbody>
    <tr>
        <td>
            
            
        </td>
    </tr>
    </tbody>
</table>

### After

https://user-images.githubusercontent.com/14894380/140212287-5abf10c0-f048-4d2d-b5d2-c40dd9c576ce.mov

https://user-images.githubusercontent.com/14894380/141383909-9529ccf5-5037-44f7-a2cf-b2d824d8f427.mov

https://user-images.githubusercontent.com/14894380/141383960-7d69009e-4dab-460a-8750-2d22730f9d5e.mov

## Types of changes

## Checklist

## Further comments
N/A


app/utils/fileDownload/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	title_link: string;
+	type: string;
+	description: string;
+}
+
+export const getExtensionType = (text: string): string | undefined => text.split('.').pop();
+
+export const getLocalFilePathFromFile = (localPath: string, attachment: IAttachment): string => {
+	const fileName = attachment.title.split('.')[0];
+	return `${localPath}${fileName}.${getExtensionType(attachment.title_link)}`;
```
diegolmello, line 20: I didn't get why we need to do this.
Wouldn't this work `${localPath}${attachment.title}` without getting the extension type?

app/utils/fileDownload/index.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const fileDownloadAndPreview = async (url: string, attachment: IAttachment): Promise<void> => {
+	const path = getLocalFilePathFromFile(DOCUMENTS_PATH, attachment);
+	const file = await RNFetchBlob.config({
+		timeout: 10000,
+		indicator: true,
+		path
+	}).fetch('GET', url);
+
+	if (file) {
```
diegolmello, line 49: What if we don't have a file?
Shouldn't we try catch `RNFetchBlob.config` and return a proper error?

diegolmello (changes requested):
Left a few comments related to Android mostly.
I couldn't open all the files on RoomActionsView > Files.

app/utils/fileDownload/index.ts
AlexAlexandre: Great question! 
I created this function, because I preferred get the extension from the `attachment.title_link`, because  in my thought Is not very safe get this information from the name of the file. The best idea shoud be get a new field `extensions` from the API. 

But, after your comment I stopped to think about, and this function:

```const fileName = attachment.title.split('.')[0];```

was not safe either, because if the file has a `.` in the file name, will crash our app. 

So I just changed the code as you suggested, thanks

File: app/utils/fileDownload/index.ts [hunk 3]
```diff
@@ -0,0 +1,59 @@
...
+		FileViewer.open(file.data, {
+			showOpenWithDialog: true,
+			showAppsSuggestions: true
+		})
+			.then(res => res)
+			.catch(() => {
+				EventEmitter.emit(LISTENER, { message: I18n.t('Downloaded_file') });
```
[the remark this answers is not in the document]
AlexAlexandre: Yep, on my Android tests, when the user don't have a default app installed we redirect to the play store, I forgot to upload the Android prints screens, but I did it right now. 

I fixed the catch to download the files when the app can't show the preview, it's works now!

app/utils/fileDownload/index.ts
AlexAlexandre: Done!

app/containers/message/Video.tsx
```diff
@@ -39,15 +46,34 @@ interface IMessageVideo {
...
+		const onPress = async () => {
 			if (isTypeSupported(file.video_type)) {
 				return showAttachment(file);
 			}
-			const uri = formatAttachmentUrl(file.video_url, user.id, user.token, baseUrl);
-			openLink(uri, theme);
+
+			if (!isIOS) {
+				const uri = formatAttachmentUrl(file.video_url, user.id, user.token, baseUrl);
+				await downloadVideo(uri);
```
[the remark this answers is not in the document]
AlexAlexandre: In this case I didn't want to redirect the user, because is not all android has a default app to reproduce .WEB videos, so as we cant insurence that, I preferred just download the file to downloads folder.

diegolmello (approved):

PR #3499, opened 2021-11-12
Chore: Migrate LivechatEditView to Typescript

reinaldonetof:
## Proposed changes

Migrate LivechatEditView to Typescript, create 2 definitions used in LiveChatEditView

## How to test or reproduce

- You should be an omnichannel agent
- Join a chat from Omnichannel's queue, then click on the header and click on the name
- You should be moved to another screen, then click on the `edit` button, on the header
- And the title's header should be `Edit`. So this is the screen.

## Types of changes

## Checklist


gerzonc (approved):
LGTM

app/definition/ITagsOmnichannel.ts
```diff
@@ -0,0 +1,5 @@
+export interface ITagsOmnichannel {
```
diegolmello, line 1: These files are creating a new folder
image

app/definition/IVisitor.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface IVisitor {
+	_id?: string;
+	token: string;
+	username: string;
+	updatedAt?: Date;
+	name: string;
+	department?: string;
+	phone?: Array<IVisitorPhone>;
+	visitorEmails?: Array<IVisitorEmail>;
```
diegolmello, line 17: It's not wrong, but just for the sake of consistency, we should use the same you've used on the other file: `string[]`
```suggestion
	phone?: IVisitorPhone[];
	visitorEmails?: IVisitorEmail[];
```

app/views/LivechatEditView.tsx
```diff
@@ -39,20 +41,65 @@ const styles = StyleSheet.create({
...
+}
+
+interface ICustomFields {
+	visitor?: { [key: string]: string };
+	livechat?: { [key: string]: string };
+}
+interface ILivechatEditViewProps {
+	user: any;
+	navigation: StackNavigationProp<any, 'LivechatEditView'>;
+	route: RouteProp<any, 'LivechatEditView'>;
```
diegolmello, line 80: Missing proper stack instead of `any`

diegolmello (dismissed):

app/containers/UIKit/MultiSelect/index.tsx
```diff
@@ -22,8 +22,8 @@ interface IMultiSelect {
 	context?: number;
 	loading?: boolean;
 	multiselect?: boolean;
-	onSearch: Function;
-	onClose: Function;
+	onSearch?: Function;
+	onClose?: Function;
```
gerzonc, line 26: Can we send how the function is composed instead of setting `Function`? I think `Function` is too ambiguous and open to anything

app/definitions/IRoom.ts
```diff
@@ -24,4 +24,9 @@ export interface IRoom extends IRocketChatRecord {
 	autoTranslate?: boolean;
 	observe?: Function;
 	usedCannedResponse: string;
+	livechatData?: {
+		[key: string]: any;
```
gerzonc, line 28: Why `any`?

app/views/LivechatEditView.tsx
```diff
@@ -39,20 +42,65 @@ const styles = StyleSheet.create({
...
+	topic: RNTextInput | null;
+	[index: string]: RNTextInput | null;
+}
+
+interface ICustomFields {
+	visitor?: { [key: string]: string };
+	livechat?: { [key: string]: string };
+}
+interface ILivechatEditViewProps {
+	user: any;
```
gerzonc, line 79: Just adding a comment here in case that someone else sees it, but the reason why this is `any` is because it comes from `mapStateToProps`

app/views/LivechatEditView.tsx
```diff
@@ -39,20 +42,65 @@ const styles = StyleSheet.create({
...
+	tag: string[];
+}
+
+type TParams = IVisitor & IInputs;
+
+interface IInputsRefs {
+	name: RNTextInput | null;
+	phone: RNTextInput | null;
+	topic: RNTextInput | null;
+	[index: string]: RNTextInput | null;
```
gerzonc, line 71:
```js
interface IInputsRefs {
	[index: string]: RNTextInput | null;
	name: RNTextInput | null;
	phone: RNTextInput | null;
	topic: RNTextInput | null;
}
```

For some reason I prefer `index` to be the first one always, but that's on you :)

gerzonc (changes requested):

app/containers/TextInput.tsx
```diff
@@ -59,7 +59,7 @@ interface IRCTextInputProps extends TextInputProps {
 	loading?: boolean;
 	containerStyle?: StyleProp<ViewStyle>;
 	inputStyle?: StyleProp<TextStyle>;
-	inputRef?: React.Ref<unknown>;
+	inputRef?: React.Ref<RNTextInput>;
```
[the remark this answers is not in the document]
reinaldonetof: Actually, the approach to ForwardRef was weird. Now I adjusted.

https://www.carlrippon.com/react-forwardref-typescript/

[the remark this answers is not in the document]
reinaldonetof: @AlexAlexandre can you take a look at this later? Thanks!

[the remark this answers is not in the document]
GleidsonDaniel: @reinaldonetof Can this file sometimes not receive the ref?
If yes the type is correct.
However, if he always receives the ref, he can put React.RefObject which is nice.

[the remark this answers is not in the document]
GleidsonDaniel: Actually looking here your logic is correct, ignore my previous comment hahaha

GleidsonDaniel (approved):
LGTM

The solution of the refs seems to me satisfactory.

app/definitions/ISubscription.ts
```diff
@@ -79,6 +79,7 @@ export interface ISubscription {
 	avatarETag?: string;
 	teamId?: string;
 	teamMain?: boolean;
+	sms?: boolean;
```
diegolmello, line 82: Isn't sms a dynamic column on Omnichannel?
If so, we shouldn't add it here.
Same for the code, as we shouldn't count on having it.

diegolmello (changes requested):

app/views/LivechatEditView.tsx
```diff
@@ -39,36 +43,87 @@ const styles = StyleSheet.create({
...
-	const [availableUserTags, setAvailableUserTags] = useState([]);
+const LivechatEditView = ({
+	user,
+	navigation,
+	route,
+	theme,
+	editOmnichannelContact,
+	editLivechatRoomCustomfields
+}: ILivechatEditViewProps) => {
+	const [customFields, setCustomFields] = useState({} as ICustomFields);
```
GleidsonDaniel, line 105: Why are you using it this way and not the default way?
Like that:
`const [customFields, setCustomFields] = useState<ICustomFields>({});`

GleidsonDaniel (changes requested):

PR #3537, opened 2021-12-02
[FIX] Certificate stops working after app update on iOS

diegolmello:
## Proposed changes

When the app is updated, NSDocumentDirectory changes as well.
Since we were persisting the whole path on certificate is set, that makes certificate to be not found.

A valid option for the fix could be to stop persisting the path, but we need the full path here https://github.com/RocketChat/react-native/blob/0.64.2/Libraries/Network/RCTHTTPRequestHandler.mm#L78
Instead of that, we're checking and updating the path every time we reconnect to the server.

### Apply the certificate
- Make sure you downloaded another build (not this one), because we want to try the app update
- On the new server screen, `Apply your certificate`, select the cert and enter the cert password (credentials on Clickup task)
- Enter the server
- Enter username/password
- It should connect properly

### Reproduce the issue
- Follow `Apply the certificate` test
- Update the app to another build (not this one)
- App should not be able to connect to the server, because the cert isn't found

### Reproduce the fix
- Follow `Apply the certificate` test
- Update the app to this build
- Everything should work fine

## Types of changes

## Checklist


app/utils/sslPinning.js
```diff
@@ -49,16 +56,19 @@ const RCSSLPinning = Platform.select({
...
+					certificate = await persistCertificate(name, certificate.password);
+				}
 				await UserPreferences.setMapAsync(extractHostname(server), certificate);
 			}
 		}
 	},
 	android: {
 		pickCertificate: () => SSLPinning?.pickCertificate(),
-		setCertificate: alias => SSLPinning?.setCertificate(alias)
+		setCertificate: name => SSLPinning?.setCertificate(name)
```
diegolmello, line 71: Changed to `name` here just to standardize the param as on iOS

AlexAlexandre (approved):
LGTM

PR #3553, opened 2021-12-13
Chore: Migrate CannedResponsesListView to Typescript

reinaldonetof:
## Proposed changes

Migrate CannedResponsesListView and CannedResponseDetail to TS

## How to test or reproduce

- You need to be a Livechat agent
- Then log in, take a channel, click on the title header and click on Canned Responses button
- You should be moved to Canned Response List View, and when click in any Canned response, you will be redirect to Canned Response Detail
- Both views were migrated.

## Types of changes

## Checklist


AlexAlexandre (approved):
LGTM

diegolmello (dismissed):
Lint is failing.
We need to evaluate the calls to goRoom.
We're sending `usedCannedResponse` twice on params.

app/views/CannedResponsesListView/Dropdown/index.tsx
```diff
@@ -50,16 +49,15 @@ class Dropdown extends React.Component {
...
+		const { theme, currentDepartment, onDepartmentSelected, departments } = this.props;
+		const heightDestination = 0;
 		const translateY = this.animatedValue.interpolate({
 			inputRange: [0, 1],
 			outputRange: [-300, heightDestination] // approximated height of the component when closed/open
 		});
 		const backdropOpacity = this.animatedValue.interpolate({
 			inputRange: [0, 1],
-			outputRange: [0, themes[theme].backdropOpacity]
+			outputRange: [0, themes[theme!].backdropOpacity]
```
GleidsonDaniel, line 60: Here if you don't have the theme as a prop and it comes undefined it will give an error, because the "non-null assertion" doesn't work that way, it will be undefined there the same way.
As it can come undefined the correct thing is to add a fall back if it is undefined.

GleidsonDaniel (changes requested):

app/views/CannedResponsesListView/Dropdown/index.tsx
GleidsonDaniel: Add a check or confirm that the theme will always come out correct.

reinaldonetof: Unfortunately, this non-null assertion is happening because the HOC `withTheme`, that must have the theme but isn't passing to the component as a prop outside the interface. So because of this, we need to insert a theme in the interface, but as optional just to make TS happy when calling the component.

GleidsonDaniel: ok, good point, thanks for explaining.

GleidsonDaniel (approved):

PR #3558, opened 2021-12-16
Chore: Migrate Markdown to Typescript

reinaldonetof:
## Proposed changes

Migrate Markdown to TS

## How to test or reproduce

Actually, this PR add types to Markdown, that is the component that you see the messages, the E2E banner, Room info, replies, quotes. Almost all components that you see text, could be a Markdown component!

So, I suggest testing all types of messages: Links, image previews, replies, emojis, threads and the E2E banner

## Types of changes

## Checklist


app/containers/markdown/Preview.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+interface IPreview {
+	msg: string;
+	numberOfLines: number;
+	theme: string;
+	testID?: string;
+	style?: StyleProp<TextStyle>[];
+}
+
+const Preview = ({ msg, numberOfLines, testID, theme, style = [] }: IPreview): JSX.Element => {
```
diegolmello, line 18: Can we use useTheme instead?

diegolmello, line 18: Can we make numberOfLines default 1?

app/containers/markdown/Preview.tsx
```diff
@@ -0,0 +1,44 @@
...
+	m = m.replace(/^\[([\s]*)\]\(([^)]*)\)\s/, '').trim();
+	m = shortnameToUnicode(m);
+	// Removes sequential empty spaces
+	m = m.replace(/\s+/g, ' ');
+	m = removeMarkdown(m);
+	m = m.replace(/\n+/g, ' ');
+	return (
+		<Text
+			accessibilityLabel={m}
+			style={[styles.text, { color: themes[theme].bodyText }, ...style]}
```
diegolmello, line 36: Can we remove the style prop from the component?

app/containers/markdown/Preview.tsx
```diff
@@ -0,0 +1,44 @@
...
+			numberOfLines={numberOfLines}
+			testID={testID}>
+			{m}
+		</Text>
+	);
+};
+
+export default Preview;
```
diegolmello, line 44: Can we rename it to MarkdownPreview instead?

app/containers/markdown/formatText.ts
```diff
@@ -0,0 +1,6 @@
+// Support <http://link|Text>
+export const formatText = (text: string) =>
```
diegolmello, line 2: Can we write tests?

app/containers/MessageBox/ReplyPreview.tsx
```diff
@@ -3,7 +3,7 @@ import { StyleSheet, Text, View } from 'react-native';
 import moment from 'moment';
 import { connect } from 'react-redux';
 
-import Markdown from '../markdown';
+import Preview from '../markdown/Preview';
```
diegolmello, line 6: Can we export Preview from markdown index?

app/containers/UIKit/index.tsx
```diff
@@ -49,10 +50,10 @@ class MessageParser extends UiKitParserMessage {
 		}
 
 		const isContext = context === BLOCK_CONTEXT.CONTEXT;
-		return (
-			// @ts-ignore
-			<Markdown msg={text} theme={theme} style={[isContext && { color: themes[theme].auxiliaryText }]} preview={isContext} />
-		);
+		if (isContext) {
+			return <Preview msg={text} theme={theme} style={[isContext && { color: themes[theme].auxiliaryText }]} numberOfLines={0} />;
```
diegolmello, line 54: numberOfLines 0?
You can use UIKit stories to test this.

app/containers/markdown/index.tsx
```diff
@@ -33,40 +32,32 @@ type UserMention = Pick<IUser, '_id' | 'username' | 'name'>;
...
 	}[];
-	enableMessageParser: boolean;
-	navToRoomInfo: Function;
-	preview: boolean;
-	theme: string;
-	testID: string;
-	style: any;
-	onLinkPress: Function;
+	enableMessageParser?: boolean;
+	navToRoomInfo?: Function;
```
diegolmello, line 51: Can we try to define this and onLinkPress?

app/containers/markdown/index.tsx
```diff
@@ -153,7 +144,7 @@ class Markdown extends PureComponent<IMarkdownProps, any> {
 			renderParagraphsInLists: true
 		});
 
-	get isNewMarkdown(): boolean {
+	get isNewMarkdown(): boolean | undefined {
```
diegolmello, line 147: Can we make sure to return boolean?
Maybe making sure `enableMessageParser` is boolean?

app/containers/markdown/index.tsx
```diff
@@ -256,9 +247,9 @@ class Markdown extends PureComponent<IMarkdownProps, any> {
 			<MarkdownAtMention
 				mentions={mentions}
 				mention={mentionName}
-				useRealName={useRealName}
-				username={username}
-				navToRoomInfo={navToRoomInfo}
+				useRealName={!!useRealName}
```
diegolmello, line 250: We could remove the `!!` making sure default `useRealName` is false.

app/containers/markdown/index.tsx
```diff
@@ -33,40 +32,32 @@ type UserMention = Pick<IUser, '_id' | 'username' | 'name'>;
...
+	mentions?: UserMention[];
+	getCustomEmoji?: Function;
+	baseUrl?: string;
+	username?: string;
+	tmid?: string;
+	isEdited?: boolean;
+	numberOfLines?: number;
+	customEmojis?: boolean;
+	useRealName?: boolean;
+	channels?: {
```
diegolmello, line 46: Can we create a definition similar to UserMention and apply it on the other places?

app/containers/markdown/new/index.tsx
```diff
@@ -49,7 +49,7 @@ const Body = ({
 			baseUrl,
 			onLinkPress
 		}}>
-		{tokens.map(block => {
+		{tokens?.map(block => {
```
diegolmello, line 52: Imo we should return null instead (before of the context) in case of `!tokens`?

app/containers/message/Content.tsx
```diff
@@ -61,8 +63,7 @@ const Content = React.memo(
 					enableMessageParser={user.enableMessageParserEarlyAdoption}
 					username={user.username}
 					isEdited={props.isEdited}
-					numberOfLines={isPreview ? 1 : 0}
-					preview={isPreview}
+					numberOfLines={0}
```
diegolmello, line 66: I think we can remove this

app/containers/markdown/index.tsx
```diff
@@ -33,40 +32,32 @@ type UserMention = Pick<IUser, '_id' | 'username' | 'name'>;
...
-	tmid: string;
-	isEdited: boolean;
-	numberOfLines: number;
-	customEmojis: boolean;
-	useRealName: boolean;
-	channels: {
+	theme: string;
+	md?: MarkdownAST;
+	mentions?: UserMention[];
+	getCustomEmoji?: Function;
```
diegolmello, line 38: I think it should be required.

app/views/RoomView/Banner.js
```diff
@@ -22,7 +23,7 @@ const Banner = React.memo(
 						style={[styles.bannerContainer, { backgroundColor: themes[theme].bannerBackground }]}
 						testID='room-view-banner'
 						onPress={toggleModal}>
-						<Markdown msg={text} theme={theme} numberOfLines={1} style={[styles.bannerText]} preview />
+						<Preview msg={text} theme={theme} numberOfLines={1} style={[styles.bannerText]} />
```
diegolmello, line 26:
```suggestion
						<Preview msg={text} theme={theme} numberOfLines={1} style={styles.bannerText} />
```

app/views/SearchMessagesView/index.tsx
```diff
@@ -308,7 +308,6 @@ class SearchMessagesView extends React.Component<ISearchMessagesViewProps, ISear
 						testID='search-message-view-input'
 						theme={theme}
 					/>
-					{/* @ts-ignore */}
 					<Markdown msg={I18n.t('You_can_search_using_RegExp_eg')} username='' baseUrl='' theme={theme} />
```
diegolmello, line 311: We should remove those extra props

storybook/stories/Markdown.js
```diff
@@ -4,6 +4,7 @@ import { ScrollView, StyleSheet, View } from 'react-native';
 import { storiesOf } from '@storybook/react-native';
 
 import Markdown from '../../app/containers/markdown';
+import Preview from '../../app/containers/markdown/Preview';
```
diegolmello, line 7: Can we remove those extra props?
IPreview should make sure we only send the right ones from now on.

storybook/stories/Markdown.js
```diff
@@ -4,6 +4,7 @@ import { ScrollView, StyleSheet, View } from 'react-native';
 import { storiesOf } from '@storybook/react-native';
```
diegolmello, line 4: Can you try to migrate to ts?
I'm not sure it's going to work though, because we might not have the prop lint.

app/views/RoomView/Banner.js
reinaldonetof: I understand why you are asking that, but actually here: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/3558/files#diff-acf5acde58285628166dc989f5f1b1f4b7a4a928f0a46d2b0405045eb01aaf26R38

We are using the spread operator for every style that is coming, so I think we can continue with this array, just to maintain the pattern as other components, and to be isn't necessary to check if it is an array, WDYT?

app/containers/markdown/interfaces.ts
```diff
@@ -0,0 +1,13 @@
...
+	name?: string;
+}
+
+export interface UserChannel {
+	[index: number]: string | number;
+	name: string;
+	_id: string;
+}
+
+export type TOnLinkPress = (link: string) => void;
```
GleidsonDaniel, line 13: Usually a file that concentrates the types we call "types" (including vscode understands this alias) and not interfaces.
Why not call this file types.ts?

app/containers/message/Reply.tsx
```diff
@@ -115,22 +116,22 @@ interface IMessageTitle {
 
 interface IMessageDescription {
 	attachment: Partial<IMessageReplyAttachment>;
-	getCustomEmoji: Function;
+	getCustomEmoji: TGetCustomEmoji;
```
GleidsonDaniel, line 119: I know you only touched this part, but do you realize that there are two interfaces with exactly the same properties? Besides the fact that they are very similar?

Maybe I could clean up this code, or just check a TODO.

GleidsonDaniel (commented):
LGTM
There are only those two comments that I left there, but they are not impeding.

app/containers/message/Reply.tsx
reinaldonetof: You are correct! However, I didn't "refactor" this, because I understood that they are 2 different props for 2 different components

So, WDYT? No problem for me, up to you

app/containers/markdown/interfaces.ts
reinaldonetof: We started to do the files that handle interfaces or types like this. But, WDYT?

GleidsonDaniel: I think it's better to use the pattern that the community uses hahah.
But in this case, leave it that way, in the next few I think it's cool to use this pattern :)

app/containers/message/Reply.tsx
GleidsonDaniel: Okay, as it's not necessarily your implementation, we leave it that way, but I think that at most it would create an interface that extends the other, since the data is virtually equal.

app/containers/markdown/Preview.tsx
```diff
@@ -0,0 +1,46 @@
...
+		return null;
+	}
+
+	const { theme } = useTheme();
+
+	let m = formatText(msg);
+
+	// Ex: '[ ](https://open.rocket.chat/group/test?msg=abcdef)  Test'
+	// Return: 'Test'
+	m = m.replace(/^\[([\s]*)\]\(([^)]*)\)\s/, '').trim();
```
diegolmello, line 29: We reusing this regex on markdown/index.
We should create a separate function.
Very similar to the work we've done to `formatText`.

diegolmello (changes requested):

GleidsonDaniel (approved):

app/containers/markdown/Hashtag.tsx
```diff
@@ -23,7 +20,7 @@ const Hashtag = React.memo(({ hashtag, channels, navToRoomInfo, style = [] }: IH
 		const index = channels?.findIndex(channel => channel.name === hashtag);
 		const navParam = {
 			t: 'c',
-			rid: channels[index]._id
+			rid: channels?.[index!]._id
```
diegolmello, line 23: It might crash our app.

app/containers/markdown/interfaces.ts
```diff
@@ -0,0 +1,13 @@
+export interface UserMention {
```
diegolmello, line 1: IUserMention and IUserChannel maybe?

app/containers/markdown/new/Inline.tsx
```diff
@@ -41,16 +41,16 @@ const Inline = ({ value }: IParagraphProps): JSX.Element => {
 						return (
 							<AtMention
 								mention={block.value.value}
-								useRealName={useRealName}
-								username={username}
-								navToRoomInfo={navToRoomInfo}
+								useRealName={useRealName!}
+								username={username!}
+								navToRoomInfo={navToRoomInfo!}
```
diegolmello, line 46: Unsafe

app/containers/markdown/new/Emoji.tsx
```diff
@@ -21,7 +21,7 @@ const Emoji = ({ value, isBigEmoji }: IEmojiProps): JSX.Element => {
 	const emoji = getCustomEmoji?.(value.value);
 
 	if (emoji) {
-		return <CustomEmoji baseUrl={baseUrl} style={[isBigEmoji ? styles.customEmojiBig : styles.customEmoji]} emoji={emoji} />;
+		return <CustomEmoji baseUrl={baseUrl!} style={[isBigEmoji ? styles.customEmojiBig : styles.customEmoji]} emoji={emoji} />;
```
diegolmello, line 24: non-null assertion

PR #3580, opened 2021-12-30
Chore: Migrate Database to Typescript

reinaldonetof:
## How to test or reproduce

Test cleaning the cache, new messages that are coming when the app is in foreground and background, edit users profiles and room info, create new threads, talk in other threads, use the search for threads (`ThreadMessagesView`) and clean again the cache. The data should be the same on the web. 

## Types of changes

## Checklist


app/containers/message/Reply.tsx
```diff
@@ -198,7 +182,7 @@ const Fields = React.memo(
 						<Text style={[styles.fieldTitle, { color: themes[theme].bodyText }]}>{field.title}</Text>
 						{/* @ts-ignore*/}
 						<Markdown
-							msg={field.value}
+							msg={field.value!}
```
GleidsonDaniel, line 185: Same thing I commented on in the previous review.
Sure the value can be undefined?
However, you can see that in this case they did a treatment within the markdown in case the message was null.

app/views/NewMessageView.tsx
```diff
@@ -153,7 +153,9 @@ class NewMessageView extends React.Component<INewMessageViewProps, INewMessageVi
 	};
 
 	search = async (text: string) => {
-		const result = await RocketChat.search({ text, filterRooms: false });
+		// TODO: When migrate rocketchat.js pass the param ISearch to there and the return should be
+		// ISearch | TSubscriptionModel, this because we do a local search too
+		const result = (await RocketChat.search({ text, filterRooms: false })) as unknown as ISearch[];
```
GleidsonDaniel, line 158: A doubt, I saw in other files but I can't let it go.
why use `as unknown as ISearch`?
Is it some detail of the DB model?

diegolmello (changes requested):
Since we're only overriding/declaring `db.get`, we need to migrate all calls using the old `db.collections.get` to `db.get`.

GleidsonDaniel:
@reinaldonetof Can you resolve conflicts here?
Solving them I think we can now pass to the QA.

GleidsonDaniel (approved):
LGTM

GleidsonDaniel:
> Since we're only overriding/declaring `db.get`, we need to migrate all calls using the old `db.collections.get` to `db.get`.

@diegolmello This is a case where I think I need you to double check to see if your request has been answered.

app/views/NewMessageView.tsx
```diff
@@ -295,10 +294,14 @@ class NewMessageView extends React.Component<INewMessageViewProps, INewMessageVi
...
+
+		const itemSearch = item as ISearch;
+		const itemModel = item as TSubscriptionModel;
+
 		return (
 			<UserItem
-				name={item.search ? item.name : item.fname}
-				username={item.search ? item.username : item.name}
+				name={itemSearch.search ? itemSearch.name : itemModel.fname || ''}
+				username={itemSearch.search ? itemSearch.username : itemModel.name}
```
diegolmello, line 304: It feels awkward to me.
I tried playing with type guards, but I couldn't do something better without creating a new util function.

@GleidsonDaniel do you have something?

diegolmello (changes requested):
Almost there!
Left a few comments :)

app/containers/MessageErrorActions.tsx
```diff
@@ -36,7 +36,7 @@ const MessageErrorActions = forwardRef(({ tmid }: any, ref): any => {
 				try {
 					// Find the thread header and update it
 					const msg = await msgCollection.find(tmid);
-					if (msg.tcount <= 1) {
+					if (msg?.tcount && msg.tcount <= 1) {
```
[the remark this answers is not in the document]
reinaldonetof: The TS complain when I do this

app/views/NewMessageView.tsx
reinaldonetof: This I did together with @GleidsonDaniel . Was the way that we found

app/containers/EmojiPicker/index.tsx
```diff
@@ -132,8 +133,8 @@ class EmojiPicker extends Component<IEmojiPickerProps, IEmojiPickerState> {
 	updateFrequentlyUsed = async () => {
 		const db = database.active;
 		const frequentlyUsedRecords = await db.get('frequently_used_emojis').query().fetch();
-		let frequentlyUsed: any = orderBy(frequentlyUsedRecords, ['count'], ['desc']);
-		frequentlyUsed = frequentlyUsed.map((item: IEmoji) => {
+		let frequentlyUsed: any[] = orderBy(frequentlyUsedRecords, ['count'], ['desc']);
```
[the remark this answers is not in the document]
reinaldonetof: I refactor this, instead use as `let` and switch between 3 types, I created another const

diegolmello (approved):

PR #3586, opened 2022-01-05
[FIX] RoomInfoView displaying different info depending on the origin

reinaldonetof:
## Proposed changes

Navigating to RoomInfoView from Avatar and from RoomActionsView displays different information.

## How to test or reproduce

- Make sure you're logged in user with admin (or correct permission)
- Go to a room, and click on the avatar of another user that talked before
- You will be moved directly to RoomInfoView and you can see the user's roles
- Then, create a DM to the same user that you clicked before,
- Click on Header, then when you are inside the Actions, click on the user's name
- You will see the RoomInfoView and should appear the roles too

### Before

https://user-images.githubusercontent.com/47038980/148281211-f9de2bc0-b289-486c-b3d0-b9a4aa994602.mov

### After

https://user-images.githubusercontent.com/47038980/148281336-8b4e6bb1-89d9-49a5-b2b7-ec197926cf6a.mov

## Types of changes

## Checklist


app/views/RoomInfoView/index.js
```diff
@@ -189,19 +197,25 @@ class RoomInfoView extends React.Component {
...
 					this.setState({ roomUser: user });
 				}
 			} catch {
 				// do nothing
 			}
+		} else {
+			try {
+				const { roles } = roomUser;
+				if (roles && roles.length) {
+					roomUser.parsedRoles = await this.parseRoles(roles);
```
GleidsonDaniel, line 212: As I understand it, are you here manipulating the state directly?
I don't think this is a good practice, it can generate several side-effects.
The right thing would be to create a variable receiving the parseRoles and then unstructure it to receive the value.

GleidsonDaniel: > if (roles && roles.length) {
>   const parsedRoles = await this.parseRoles(roles);
>   this.setState({ roomUser: { ...roomUser, parsedRoles } });
> } else {
>   this.setState({ roomUser });
> }

GleidsonDaniel (changes requested):

app/views/RoomInfoView/index.js
GleidsonDaniel: I committed the changes

GleidsonDaniel (approved):

AlexAlexandre (approved):
LGTM

PR #3621, opened 2022-01-17
[FIX] Add search and fix pagination for omnichannels departments

reinaldonetof:
## Proposed changes

When an agent tries to forward an ongoing chat to another department, the list shows only 50 departments. So we are proposing to the agent should be able to either search on the list or see the full list

## How to test or reproduce

- To do a better test for this task, use the mobile.rocket.chat
- Activate as an online agent
- To create a new chat you can go to https://mobile.rocket.chat/livechat
- Then, take the chat from queue and go to actions room and click to forward the chat

- You should be able to see the pagination for departments
- Add to this, you can search by typing the department's name

### Before

https://user-images.githubusercontent.com/47038980/150029564-712332b1-e426-481b-aa08-789e824b4c78.mov

### After

https://user-images.githubusercontent.com/47038980/150029612-d5920d82-95db-4250-9115-f56bca8ea037.mov

## Types of changes

## Checklist


app/views/PickerView.tsx
```diff
@@ -37,6 +37,9 @@ interface IItem {
 interface IPickerViewState {
 	data: IOptionsField[];
 	value: string;
+	offset: number;
+	total: number;
+	text: string;
```
diegolmello, line 42: On PickerView's context, `text` is the search text, right?
Maybe we can rename it to `searchText` so it's more clear?

diegolmello (changes requested):
I was not able to forward a chat following the test plan.
Is there a configuration needed on mobile server?

app/views/ForwardLivechatView.tsx
```diff
@@ -148,6 +160,11 @@ const ForwardLivechatView = ({ forwardRoom, navigation, route, theme }: IForward
 			value: room?.departmentId,
 			data: departments,
 			onChangeValue: setDepartment,
+			onChangeText: getDepartments,
```
diegolmello, line 163: We did have search on `PickerView`, but this is the first time we're using it.
So we should fix this.
Search header shouldn't be sticky to header (rendered on top of the list while scrolling).


It should be part of the scroll and render only on top.

app/stacks/types.ts
```diff
@@ -98,7 +98,14 @@ export type ChatsStackParamList = {
 		title: string;
 		data: IOptionsField[];
 		value?: any; // TODO: Change
-		onChangeText?: ((text: string) => IOptionsField[]) | ((term?: string) => Promise<any>);
+		onChangeText?: (text?: string) => Promise<{ data: IOptionsField[] } | undefined>;
+		onEndReached?: (
+			text: string,
+			offset?: number
+		) => Promise<{ data: IOptionsField[]; total: number; offset: number } | undefined>;
```
diegolmello, line 105: Since we're adding this, we should fix this return of `IOptionsField`.
That's coming from `NotificationPreferencesView`.
PickerView is supposed to be generic.

app/views/ForwardLivechatView.tsx
```diff
@@ -54,22 +54,33 @@ interface IForwardLivechatViewProps {
...
+			const result = await RocketChat.getDepartments({ count: COUNT_DEPARTMENT, text, offset });
 			if (result.success) {
-				setDepartments(
-					result.departments.map((department: ILivechatDepartment) => ({ label: department.name, value: department._id }))
-				);
+				const parsedDepartments = result.departments.map((department: ILivechatDepartment) => ({
+					label: department.name,
+					value: department._id
+				}));
+				if (!text && !offset) {
```
reinaldonetof, line 78: I insert this to save the departments only once when the text is empty and offset is 0, instead of all time that this function is called.

app/views/ForwardLivechatView.tsx
```diff
@@ -86,13 +97,14 @@ const ForwardLivechatView = ({ forwardRoom, navigation, route, theme }: IForward
 			});
 			if (result.success) {
 				const parsedUsers = result.items.map((user: IUser) => ({ label: user.username, value: user._id }));
-				setUsers(parsedUsers);
-				return parsedUsers;
+				if (!term) {
```
reinaldonetof, line 100: Same here!

app/views/PickerView.tsx
```diff
@@ -81,16 +89,31 @@ class PickerView extends React.PureComponent<IPickerViewProps, IPickerViewState>
...
 		},
 		300,
 		true
 	);
 
+	onEndReached = async () => {
+		const { route } = this.props;
+		const { data, offset, total, searchText } = this.state;
+		const onEndReached = route.params?.onEndReached;
+		const newOffset = offset + this.count;
```
diegolmello, line 108: Isn't `offset + this.count` the same as `data.length`?
I think it's confusing to receive `count` as a prop on `PickerView` and limiting as a generic component.
It shouldn't have to know `count` is the value we're sending as param for a http request.

diegolmello (changes requested):

app/stacks/types.ts
```diff
@@ -97,8 +97,10 @@ export type ChatsStackParamList = {
 	PickerView: {
 		title: string;
 		data: IOptionsField[];
-		value?: any; // TODO: Change
-		onChangeText?: ((text: string) => IOptionsField[]) | ((term?: string) => Promise<any>);
+		value?: string;
+		onChangeText?: (text?: string) => Promise<any>;
```
reinaldonetof, line 101: This is Promise<any> to keep the PickerView too much generic as possible

reinaldonetof: I changed the name of the param from `onChangeText` to `onSearch`

AlexAlexandre (approved):

PR #3636, opened 2022-01-20
Chore/migration ts redux room

GleidsonDaniel:
## Types of changes

## Checklist


app/views/TeamChannelsView.tsx
```diff
@@ -435,7 +436,8 @@ class TeamChannelsView extends React.Component<ITeamChannelsViewProps, ITeamChan
 				{
 					text: I18n.t('Yes_action_it', { action: I18n.t('delete') }),
 					style: 'destructive',
-					onPress: () => deleteRoom(item._id, item.t)
+					// VERIFY ON PR
+					onPress: () => dispatch(deleteRoom(item._id, item))
```
diegolmello, line 440: We need to address this and open a new task with proper info.
Or is it easy enough to fix on this PR?

app/actions/room.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+type TSubscribeRoom = IBaseReturn;
+type TUnsubscribeRoom = IBaseReturn;
+type TCloseRoom = IBaseReturn;
+
+type TRoom = Record<string, any>;
+
+interface ILeaveRoom extends Action {
+	roomType: string;
```
diegolmello, line 26: I think it could be a type of `'channel' | 'team'`

diegolmello (changes requested):

app/views/TeamChannelsView.tsx
GleidsonDaniel: Actually, I think it doesn't matter, but as I said it was already broken in the previous version, because everywhere they instantiate the return object and not the string that was passed.

app/actions/room.ts
GleidsonDaniel: in fact I saw that there are several of these types, but there is no Enum, I created one.
Check if that's the case.

AlexAlexandre (approved):
LGTM

PR #3637, opened 2022-01-20
Chore: Migrate lib/utils to TypeScript

gerzonc:
## Proposed changes

Migrate `lib/utils` to TypeScript and remove `methods` usage across the app so we can use the union type 
```ts
TMethods = 'lowerThan' | 'lowerThanOrEqualTo' | 'greaterThan' | 'greaterThanOrEqualTo';
```

Now we don't need to import `methods` in order to use the right function from `semver`.

## How to test or reproduce

We use this function to differentiate which functionality runs in the current and older versions of the app. Most uses are internal (Like API endpoints changes or how we handle something before/after a specific server version), but best way to test this is using the Teams feature, in my opinion.

In order to test this you should have a server with version previous to 3.13.0 (Like [3.12.5](https://github.com/RocketChat/Rocket.Chat/releases/tag/3.12.5), for example) and a current one (like `open`), since Teams feature didn't exist before  3.13.0. 

`3.12.5`: Set up the server and check if Teams feature doesn't exist (In order to test from physical device, use your machine IP as URL following the port the server its using, something like this: `http://192.168.100.0:3000`)

`open.rocket.chat`: Teams feature should be there.
   

## Types of changes

## Checklist


diegolmello (changes requested):

app/lib/methods/getUsersPresence.js
```diff
@@ -11,7 +11,7 @@ export function subscribeUsersPresence() {
 	const serverVersion = reduxStore.getState().server.version;
 
 	// if server is lower than 1.1.0
-	if (compareServerVersion(serverVersion, '1.1.0', methods.lowerThan)) {
+	if (compareServerVersion(serverVersion, 'lowerThan', '1.1.0')) {
```
GleidsonDaniel, line 14: If you are not exporting the lib method, the right thing would be to use the value informed by semver and not use a "fixed" value.
This can cause future problems if change something in semver.

GleidsonDaniel (changes requested):

GleidsonDaniel (approved):
LGTM

PR #3641, opened 2022-01-21
[FIX] Display emojis when used just after the text without any space

try-catch-stack:
## Proposed changes

The emojis used just after the text without any space are not rendered as emojis.

## How to test or reproduce

Send a message with some text and use an emoji just after the text like `Text:smiley:`  You will see that the emoji will not be rendered .

## Types of changes

Before:
emoji-before

After:
emoji-after

## Checklist


try-catch-stack:
@diegolmello @GleidsonDaniel I cannot figure out why that one jest test is failing . I tried `yarn test -u` which updated the storyshots file but still something's not right. Any suggestions?

PR #3658, opened 2022-01-26
[FIX] Regression: Prevent duplicated .jpg on file upload

reinaldonetof:
## Proposed changes

Prevent app from uploading files with ".jpg.jpg" on the filename on "forceJpgExtension" method

## How to test or reproduce

Upload a jpg image from using "Choose from library"

### Before
Screen Shot 2022-01-26 at 13 47 59

### After
Screen Shot 2022-01-26 at 13 48 19

## Types of changes

## Checklist

### DIFF
--- a/app/containers/MessageBox/index.tsx
+++ b/app/containers/MessageBox/index.tsx
@@ -131,7 +131,8 @@ interface IMessageBoxState {
 }
 
 const forceJpgExtension = (attachment: ImageOrVideo) => {
-	if (isIOS && attachment.mime === 'image/jpeg' && attachment.filename) {
+	const regexJPG = new RegExp(/.jpg$/i); // Test if the attachment.filename is ending with '.jpg', ignoring case
+	if (isIOS && attachment.mime === 'image/jpeg' && attachment.filename && !attachment.filename.match(regexJPG)) {
 		const regex = new RegExp(/.heic$/i);
 		if (attachment.filename.match(regex)) {
 			attachment.filename = attachment.filename.replace(regex, '.jpg');


GleidsonDaniel:
Given that we've already reached the point of adding more of this check, wouldn't it be simpler to just override all sorts of extensions?

Something like:
```
const forceJpgExtension = (attachment: ImageOrVideo) => {
	if (isIOS && attachment.mime === 'image/jpeg' && attachment.filename) {
		attachment.filename = attachment.filename.replace(/\.[^/.]+$/, 'jpg');
	}
	return attachment;
};
```

Playground:
https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhjAvDA5AExAIwHQCsAOA5igFCiSyZKoY4ASAogJIDCp50Mw16W2+YYmXCc0PWnnwBTIcMggANlOwKQhABRxsAJyn4FcYFPUB6ADrYA2gD0T2ALoBqACQmANKknEAlN7kRFZVUNHF19Q2NzK1sHF3dPAh8-DkCVNXVgHT0DI1MLGzsnVw8ULxRff1Tg9TQs8NyogtjihKJyvxISIA

@diegolmello @reinaldonetof

GleidsonDaniel:
Would add a check if it is png.

reinaldonetof:
> Would add a check if it is png.

Actually, I think isn't necessary because the type of `.png` is `image/png`. And we can keep the same type. Just the `.heic` that needs to be replaced, because just iOS 11+ and macOS that could see the image.

GleidsonDaniel:
> @reinaldonetof @GleidsonDaniel I think it's converting any image to jpg, right? If so, I'd do as Daniel says and also test `'image/*'` instead of `'image/jpeg'`. It might not be necessary because it's a return from the lib, but it won't hurt (:
> 
> If that's correct, we don't need anything for `.png` or other types.

That's what I was commenting on.
The issue with png is why it maintains transparency, jpg does not.

But in my opinion, overriding the all extension is better than adding a new check.
And if it's like reinaldo said there of png coming as another type, just apply the code I commented on.

app/containers/MessageBox/index.tsx
```diff
@@ -132,12 +132,8 @@ interface IMessageBoxState {
 
 const forceJpgExtension = (attachment: ImageOrVideo) => {
 	if (isIOS && attachment.mime === 'image/jpeg' && attachment.filename) {
-		const regex = new RegExp(/.heic$/i);
-		if (attachment.filename.match(regex)) {
-			attachment.filename = attachment.filename.replace(regex, '.jpg');
-		} else {
-			attachment.filename += '.jpg';
-		}
+		const regex = new RegExp(/\.[^/.]+$/); // Check from last '.' of the string
```
diegolmello, line 135: You're making a change for "all types", but you're requesting tests for jpg only.

Plus: Maybe you could take this opportunity to create a separate file for this method and write tests.

diegolmello (changes requested):

GleidsonDaniel:
I think the right one would be a method for `forceJpgExtension` and not a method for `replaceImageFilename`.
Why in this case you tested this function and on the screen it only called it instead of implementing a method on the screen and importing another method within it.

app/containers/MessageBox/forceJpgExtension.test.ts
```diff
@@ -0,0 +1,121 @@
+import { forceJpgExtension } from './forceJpgExtension';
+
+interface Image {
```
GleidsonDaniel, line 3: No need to create the interface, just import.
`import { Image } from 'react-native-image-crop-picker';`

app/containers/MessageBox/forceJpgExtension.ts
```diff
@@ -0,0 +1,12 @@
+import { ImageOrVideo } from 'react-native-image-crop-picker';
+
+import { isIOS } from '../../utils/deviceInfo';
+
+const regex = new RegExp(/\.[^/.]+$/); // Check from last '.' of the string
+
+export const forceJpgExtension = (attachment: ImageOrVideo) => {
```
GleidsonDaniel, line 7: It's just a detail, but put the function return, as it will return the parameter informed, just use the same for the return.
`export const forceJpgExtension = (attachment: ImageOrVideo): ImageOrVideo => {`

GleidsonDaniel (changes requested):

GleidsonDaniel (approved):
🚀

app/containers/MessageBox/forceJpgExtension.test.ts
```diff
@@ -0,0 +1,107 @@
...
+	modificationDate: '1643984790',
+	localIdentifier: 'device/L0/001',
+	size: 16623,
+	sourceURL: '',
+	mime: 'image/jpeg',
+	cropRect: null,
+	creationDate: '1641490665'
+};
+
+describe('forceJpgExtension for iOS', () => {
```
diegolmello, line 21: Why do we need this platform separation?

app/containers/MessageBox/forceJpgExtension.ts
```diff
@@ -0,0 +1,12 @@
+import { ImageOrVideo } from 'react-native-image-crop-picker';
+
+import { isIOS } from '../../utils/deviceInfo';
+
+const regex = new RegExp(/\.[^/.]+$/); // Check from last '.' of the string
+
+export const forceJpgExtension = (attachment: ImageOrVideo): ImageOrVideo => {
+	if (isIOS && attachment.mime === 'image/jpeg' && attachment.filename) {
+		attachment.filename = attachment.filename.replace(regex, '.jpg'); // Replace files that ends with .jpg | .heic | .jpeg to .jpg;
```
diegolmello, line 9: It's not `.jpg | .heic | .jpeg` only

app/containers/MessageBox/forceJpgExtension.test.ts
reinaldonetof: Because the return of the function depends on which device the user is using

reinaldonetof: WDYT?

app/containers/MessageBox/forceJpgExtension.ts
reinaldonetof: You are right, so I changed the comment to // Replace files extension that mime type is 'image/jpeg' to .jpg;

WDYT?

PR #3667, opened 2022-01-28
[FIX] Differ to Last Session Authenticated

reinaldonetof:
## Proposed changes

When you set the PIN code to enter the application, the blocking occurs after the time set in the settings after the activity is closed. System time is used as a starting point. It is possible to bypass PIN by setting the system time back to a value when the application has not yet been blocked.

To fix this problem, now we are using the route `{server}/_timesync` provided by the server. However, if the device is offline/out of the internet or the server fall, will request the pin, even it is between the free PIN time.

#### Step to step
- Enable Screen Lock and set lock time (e.g. 1 min)
- Close chat activity, mark the current time (e.g. 00:02), wait for blocking.
- Open the app and make sure it is blocked, close the app.
- Change the system time to a value when the lock has not yet been triggered (e.g. 00:01 or 00:02 )
- Start the app, it should be blocked.

#### Additional tests with different connections
- Test with poor connection, with lack of connection
- Try to do a poor connection, which will make the request be more than 2 seconds. And even you are between the permission time, will request the code.

- Test with the `mobile.rocket.chat` server too! Because the server is forcing lock screen on devices

## Types of changes

## Checklist


app/lib/rocketchat.js
```diff
@@ -169,6 +169,17 @@ const RocketChat = {
 			message: I18n.t('Not_RC_Server', { contact: I18n.t('Contact_your_server_admin') })
 		};
 	},
+	async getServerTimeSync(server) {
+		try {
+			const response = await RNFetchBlob.fetch('GET', `${server}/_timesync`);
```
reinaldonetof, line 174: I use the same way to fetch to server as: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/3667/files#diff-88ed18506a065f3ea03357ad43d30e19ac28e6725abb7ba2477eb9c535fa781bR131

Is there another best way to do this?

diegolmello: This way is fine

diegolmello, line 174: This request takes 1600ms on https://mobile.rocket.chat, so we should be careful on the impact.
If screen lock is enabled, we do it right on app open.
Pay attention on the white screen after the splash screen is hidden.

https://user-images.githubusercontent.com/804994/153067384-32401176-9f43-43ee-8ec0-500453ea54bf.mp4

https://user-images.githubusercontent.com/804994/153068220-e3d8a199-887d-473a-a964-60ce661041f8.mp4

A good solution for this is to move `RNBootSplash.hide()` to after the call.

app/utils/localAuthentication.ts
```diff
@@ -22,6 +22,8 @@ import EventEmitter from './events';
 import { isIOS } from './deviceInfo';
 
 export const saveLastLocalAuthenticationSession = async (server: string, serverRecord?: TServerModel): Promise<void> => {
+	const timesync: number = (await RocketChat.getServerTimeSync(server)) ?? 0;
```
diegolmello, line 25: It's awkward to call this request again here.
`localAuthenticate()` already calls it.

diegolmello (changes requested):
Did you try logging in?
It's requesting FaceID twice.

app/lib/rocketchat/rocketchat.js
```diff
@@ -175,6 +175,17 @@ const RocketChat = {
 			message: I18n.t('Not_RC_Server', { contact: I18n.t('Contact_your_server_admin') })
 		};
 	},
+	async getServerTimeSync(server) {
+		try {
+			const response = await RNFetchBlob.fetch('GET', `${server}/_timesync`);
```
diegolmello, line 180: I just realized we should do a race here.
This request might take some time to happen in some cases (poor connection fast enough to try to make the request, but not slow enough to be offline).
In such cases, it'd be better to just try for some time and show the lock screen than hanging for 10s.

I think a race of 2 seconds should be ok.
DM me if you're in doubt.

diegolmello (changes requested):

app/lib/rocketchat/rocketchat.js
```diff
@@ -164,6 +164,20 @@ const RocketChat = {
 			message: I18n.t('Not_RC_Server', { contact: I18n.t('Contact_your_server_admin') })
 		};
 	},
+	async getServerTimeSync(server) {
```
diegolmello, line 167: We've been working on dehydrating this file.
We shouldn't add more code here.

app/utils/localAuthentication.ts
```diff
@@ -116,10 +124,10 @@ export const localAuthenticate = async (server: string): Promise<void> => {
 		// `checkHasPasscode` results newPasscode = true if a passcode has been set
 		if (!result?.newPasscode) {
 			// diff to last authenticated session
-			const diffToLastSession = moment().diff(serverRecord?.lastLocalAuthenticatedSession, 'seconds');
+			const diffToLastSession = moment(timesync).diff(serverRecord?.lastLocalAuthenticatedSession, 'seconds');
 
-			// if last authenticated session is older than configured auto lock time, authentication is required
-			if (diffToLastSession >= serverRecord.autoLockTime!) {
+			// if can't receive the timesync value from the server or last authenticated session is older than configured auto lock time, authentication is required
```
diegolmello, line 129:
```suggestion
			// it it was not possible to get `timesync` from server or the last authenticated session is older than the configured auto lock time, authentication is required
```

app/lib/rocketchat/rocketchat.js
diegolmello: Move it to `services` folder

diegolmello (approved):

PR #3685, opened 2022-02-03
[FIX] Messages read after a lack of connection

reinaldonetof:
## Proposed changes

When a user loses connection while a chat room is open, new messages are not "read" properly once the connection is restored.

## How to test or reproduce

- open a Room with two different users
- Cut the connection of one of them
- Send messages with the other
- Turn the connection on again.

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/subscriptions/room.js
+++ b/app/lib/methods/subscriptions/room.js
@@ -32,9 +32,10 @@ export default class RoomSubscription {
 			await this.unsubscribe();
 		}
 		this.promises = RocketChat.subscribeRoom(this.rid);
+		this.isDisconnected = false;
 
 		this.connectedListener = RocketChat.onStreamData('connected', this.handleConnection);
-		this.disconnectedListener = RocketChat.onStreamData('close', this.handleConnection);
+		this.disconnectedListener = RocketChat.onStreamData('close', this.handleDisconnection);
 		this.notifyRoomListener = RocketChat.onStreamData('stream-notify-room', this.handleNotifyRoomReceived);
 		this.messageReceivedListener = RocketChat.onStreamData('stream-room-messages', this.handleMessageReceived);
 		if (!this.isAlive) {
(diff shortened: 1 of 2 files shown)


app/lib/methods/subscriptions/room.js
```diff
@@ -80,6 +80,7 @@ export default class RoomSubscription {
 	handleConnection = () => {
 		reduxStore.dispatch(clearUserTyping());
 		RocketChat.loadMissedMessages({ rid: this.rid }).catch(e => console.log(e));
+		RocketChat.subscriptionRead(this.rid);
```
diegolmello, line 83: - This is a call to a REST API, so it might throw an error. You're not handling it, so it might crash the app
- We should only call it after `loadMissedMessages` is resolved, otherwise we would be marking as read messages the user never downloaded.

app/lib/rocketchat.js
```diff
@@ -1057,6 +1057,9 @@ const RocketChat = {
 		}
 		return this.post('subscriptions.read', { rid: roomId });
 	},
+	subscriptionRead(roomId) {
```
diegolmello, line 1060: We don't need this. We should call `RoomServices.readMessages`

diegolmello (changes requested):

app/lib/methods/loadMissedMessages.js
```diff
@@ -31,6 +31,8 @@ export default function loadMissedMessages(args) {
 		try {
 			const data = await load.call(this, { rid: args.rid, lastOpen: args.lastOpen });
 
+			await this.sdk.post('subscriptions.read', { rid: args.rid });
```
diegolmello, line 34: This is wrong. You should not add another method here.
The goal for this file is to `load missed messages`.
Adding another method here is going to either add coupling or hide its purpose.

app/lib/methods/subscriptions/room.js
```diff
@@ -78,8 +78,11 @@ export default class RoomSubscription {
 	};
 
 	handleConnection = () => {
+		const _lastOpen = this.lastOpen;
 		reduxStore.dispatch(clearUserTyping());
 		RocketChat.loadMissedMessages({ rid: this.rid }).catch(e => console.log(e));
```
diegolmello, line 83: You can wrap this in a `try/catch`, call it with `async/await` instead of `then/catch` and call `read` after it's resolved.

app/lib/methods/subscriptions/room.js
```diff
@@ -78,8 +78,11 @@ export default class RoomSubscription {
 	};
 
 	handleConnection = () => {
+		const _lastOpen = this.lastOpen;
```
diegolmello, line 81: Can you explain this usage of lastOpen?

diegolmello (changes requested):

PR #3686, opened 2022-02-03
[FIX] Remove deprecated database methods and other database operations

diegolmello:
## Types of changes

## Checklist


app/lib/encryption/encryption.js
```diff
@@ -240,15 +240,15 @@ class Encryption {
 						msg,
 						tmsg
 					});
-					if (message._hasPendingUpdate) {
-						console.log(message);
-						return;
+					try {
```
diegolmello, line 243: Instead of relying on `_hasPendingUpdate` to prevent the error, let's wrap the whole call on `try/catch` and handle the error.

app/lib/methods/getRoles.js
```diff
@@ -20,41 +21,38 @@ export async function onRolesChanged(ddpMessage) {
 		const db = database.active;
 		const rolesCollection = db.get('roles');
 		try {
-			const rolesRecord = await rolesCollection.find(_id);
-			try {
```
diegolmello, line 24: If something bad happens inside the block, we would try to re-create the role again, which would raise a SQLite PK error.

app/lib/methods/updateMessages.js
```diff
@@ -58,6 +58,25 @@ export default function updateMessages({ rid, update = [], remove = [], loaderIt
 			// filter loaders to delete
 			let loadersToDelete = allMessagesRecords.filter(i1 => update.find(i2 => i1.id === generateLoadMoreId(i2._id)));
 
+			// Delete
```
diegolmello, line 61: Moved delete to top, because I found a couple of issues on Bugsnag complaining about trying to delete a record with pending updates.
So let's schedule to delete first and handle the issue on update block.
Of course it doesn't solve the issue, but the issue itself is a matter of timing and might be fixed only with JSI (and its instant changes).

GleidsonDaniel (approved):

PR #3687, opened 2022-02-03
[IMPROVE] Add support for ephemeral messages inside threads

diegolmello:
## Proposed changes

- Added support for sending ephemeral messages inside threads when they have `tmid`
- Added all extra info from the message payload, like user object

## How to test or reproduce

- Send `/sendlater help` on a thread from the app
- It should receive and render the ephemeral message inside the thread (not on main room)

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/subscriptions/rooms.js
+++ b/app/lib/methods/subscriptions/rooms.js
@@ -1,5 +1,6 @@
 import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
 import { InteractionManager } from 'react-native';
+import EJSON from 'ejson';
 
 import database from '../../database';
 import { merge } from '../helpers/mergeSubscriptionsRooms';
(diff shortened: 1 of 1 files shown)


app/lib/methods/subscriptions/rooms.js
```diff
@@ -309,32 +311,18 @@ export default function subscribeRooms() {
...
-							Object.assign(m, message);
-						})
-					);
-				});
+				const [args] = ddpMessage.fields.args;
+				const _id = random(17);
+				const message = {
+					u: {
+						_id,
+						username: 'rocket.cat',
```
GleidsonDaniel, line 320: A doubt, this name should be fixed?

GleidsonDaniel (changes requested):
Another thing is that testing here, sometimes the "cat" appeared and other times only a message box, is this the expected operation? Since we are fixing the name to rocket.cat.

Screen Shot 2022-02-04 at 11 35 32

diegolmello:
> Another thing is that testing here, sometimes the "cat" appeared and other times only a message box, is this the expected operation? Since we are fixing the name to rocket.cat.
> 
> Screen Shot 2022-02-04 at 11 35 32

@GleidsonDaniel I was not able to reproduce that 😕

GleidsonDaniel (approved):

PR #3695, opened 2022-02-08
Chore: Migrate methods/getRoomInfo to TS

gerzonc:
## Proposed changes

Migrate `methods/getRoomInfo` to TS

## How to test or reproduce

Check room information screen is working properly

## Types of changes

## Checklist

### DIFF
--- a/app/definitions/ICustomEmoji.ts
+++ b/app/definitions/ICustomEmoji.ts
@@ -1,8 +1,9 @@
 import Model from '@nozbe/watermelondb/Model';
 
 export interface ICustomEmoji {
+	_id: string;
 	name?: string;
-	aliases?: string;
+	aliases?: string[];
 	extension: string;
 	_updatedAt: Date;
 }
--- a/app/definitions/rest/v1/emojiCustom.ts
+++ b/app/definitions/rest/v1/emojiCustom.ts
@@ -9,7 +9,7 @@ export type EmojiCustomEndpoints = {
 		} & PaginatedResult;
 	};
 	'emoji-custom.list': {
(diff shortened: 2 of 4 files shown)


app/lib/methods/getRoomInfo.ts
```diff
@@ -1,7 +1,8 @@
+import { IRoom } from '../../definitions';
 import { getSubscriptionByRoomId } from '../database/services/Subscription';
 import RocketChat from '../rocketchat';
 
-const getRoomInfo = async rid => {
+const getRoomInfo = async (rid: string): Promise<Partial<IRoom> | null> => {
```
AlexAlexandre, line 5: I don't know if use a `Partial` here is the best approach, take a look at the lint erros.

AlexAlexandre (changes requested):

app/lib/methods/getRoomInfo.ts
gerzonc: It's a room, just that `getRoomInfo` returns less information than what the `IRoom` interface does, that's why lint complains, we defined `this.room` with an object and some properties from `IRoom`, but on the context we're using it is actually a room with less attributes, that's why I opted for `Partial<IRoom>`.

AlexAlexandre: So you can use the `Pick<Type, keys>`. 
Using the keys required at RoomInfo
E.g: Pick<IRoom, 'id' | 'name'>

AlexAlexandre (approved):
LGTM

diegolmello (approved):

PR #3696, opened 2022-02-08
[FIX] Mention from suggestions concatenates to query string on autocomplete

### LINKED ISSUE #3342: [bug] Autocomplete don't clear search query
### Description:

Few versions before a very good feature was introduced - autocomplete as it works on web client. That autocomplete function search not only name of user, but username too (Real Name on Russian in our case).
But now one pitty bug appeared in thst feature - if you start mention with `@` sign and fill few symbols of username and then tap on suggested user - that few symbols remain in message box.

### Environment Information:

- Rocket.Chat Server Version: 3.16.3
- Rocket.Chat App Version: 4.19.0
- Device Name: Xiaomi Mo 9 SE
- OS Version: Android 11

### Steps to reproduce:

1. Start typing `@карла` to mention `Карлан Антон` (that my name) 
2. See suggestions
3. Tap correct user from suggestion
4. Correct username concatenated to search query

### Expected behavior:

Suggested username should remain without query string

try-catch-stack:
## Proposed changes

https://user-images.githubusercontent.com/77742477/153057487-1dbb2bb9-b105-4e44-a77b-734beb40f7b5.mp4

## Issue(s)	

Fixes #3342 

## Types of changes

## Checklist


app/containers/MessageBox/index.tsx
```diff
@@ -496,7 +496,11 @@ class MessageBox extends Component<IMessageBoxProps, IMessageBoxState> {
 		const msg = this.text;
 		const { start, end } = this.selection;
 		const cursor = Math.max(start, end);
-		const regexp = /([a-z0-9._-]+)$/im;
+		let regexp = /([a-z0-9._-]+)$/im;
+		if (trackingType === MENTIONS_TRACKING_TYPE_USERS) {
+			// Selects query text after '@' for all languages while mentioning a user
+			regexp = /([^@p{L}]+)$/im;
```
diegolmello, line 502: You should reuse this regexp on your tests instead of repeating yourself.
You can create a small function inside `Messagebox` with it and a separate file to describe those tests.

diegolmello (changes requested):

app/containers/MessageBox/index.tsx
try-catch-stack: Made some changes @diegolmello . Created a function to get Regexp and generate its tests. Also added the tests for other mentions like emoji,room and commands.
Please review .

ankar84:
Great fix, @try-catch-stack !

diegolmello (changes requested):
Not working here.

https://user-images.githubusercontent.com/804994/154136555-cef303e1-e05f-4c21-b6d4-03115d490e36.mp4

diegolmello:
Maybe iOS only

try-catch-stack:
Possible! 
But I can't digest the fact that it is just a regexp doing the job and how can it work differently on IOS and android. Also tested the regexp on this website https://www.regextester.com/93547 and it works fine. Do you have any idea what can be causing the issue?

try-catch-stack:
Sure! That would be great. Unfortunately I won't be able to test for IOS as of now :(

app/containers/MessageBox/mentionSuggestionRegexp.test.js
```diff
@@ -0,0 +1,66 @@
+import getRegexp from './getRegexp';
```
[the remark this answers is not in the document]
try-catch-stack: How about getMentionRegexp?

GleidsonDaniel:
Some points
As we use the cursor and we are getting the last incidence of @, it is not necessary to use `/im`
`i` ignore case/lower case - As we are taking the last incident of @ and ignoring the rest after, it is not necessary.
`m` multiline - As we are taking the last incident of @ and ignoring the rest after, it is not necessary.

Another thing would be that we are using two different regex, one for mentions and one for emoji and channels.
So they should be different regex, not just one with different returns, given the different responsibilities.

However the solution only includes the mentions problem, if any channel or emoji contains Cyrillic characters or other types the problem will persist, and that was the reason I tried other things after this [solution](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/3777/commits/e2b11e301e034598c9e2199723da6b4b51b83af4).

try-catch-stack:
Got it @GleidsonDaniel !
Can you please confirm for once if this regex is working fine on iOS and I'll make the required changed.

GleidsonDaniel:
> Got it @GleidsonDaniel ! Can you please confirm for once if this regex is working fine on iOS and I'll make the required changed.
> 
> Tried another regex that would work for all the mention types (User ,Emoji ,Room/Channel ,Command) . Check the following test for it. [Regex test](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAthA5gVQgUwE4Fk1igS3BgF4YByAGRAzThnwAcIBXOgExABtqYJ9YAhnDRQANDFCQ0wKCOYYYAto3wRg+MIhhpO-GAAFAXCCAGEEACIAA4AnACYAzAEZ7ANnuWrcMgChJ0eEgBKICBwOHiEYCTkVDR0jCzsXDx8gsJiEuDoMnIKSipqGlo6egDEJhY2Ds6u7l4+sAiIAKJwIABW+KEERKSU1LT0TKwwHNwKyYqp4j7SslDyisoMquqa2rqwAFxlVnaOLm6WHt4Z9UgAwsFwAmBsneGRvTED8cOJY3pCIlMZM9kLeStCusYAB6bYVPbVQ5eY6QWA0RBoAAekRBAG0AHr6DbFEEAXQAVAASEEAbk8sIgXDQADpuIgABQNVCYO7gGk0BicATANAMhHI8RkMgAShFoJBACJov04kMRkkPpN0lIsnMcotlgU1np9JLKdS6SBGQ1AsE2WAOWguTy+QKkULReKQVKZbFBglRrwlV8VZlZvNckt8qsirBivqfIb6UykM02h1cF1LZzubz+WhEQ7yE6JdK+u6Xgr3ilfdM1YHNSGgXoNpGMtHjbHEBc4FcbharTb0-bHWK827nvK3t7S2lywGNQDtWHQfqgA)

Yes, worked.

diegolmello (approved):

GleidsonDaniel:
🎉

PR #3701, opened 2022-02-09
Chore: Migrate methods/loadMessagesForRoom to Typescript

AlexAlexandre:
## Proposed changes
Migrate methods/loadMessagesForRoom to typescript

## Issue(s)	
N/A

## How to test or reproduce
 * Open a chat,  messages should load correctly.
 * Clear the app cache, open a chat, messages should load correctly.

## Screenshots
N/A

## Types of changes

- [X] Documentation update (if none of the other choices apply)

## Checklist

## Further comments
N/A


app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -5,29 +5,37 @@ import log from '../../utils/log';
...
+	const data: any = await sdk.get(`${apiType}.history`, params);
 	if (!data || data.status === 'error') {
 		return [];
 	}
 	return data.messages;
 }
 
-export default function loadMessagesForRoom(args) {
+export default function loadMessagesForRoom(
+	this: IRocketChat,
```
GleidsonDaniel, line 36: Here you don't need `this`, since you don't use `this` anymore in the other function.
You can call `load` directly without being `.call`.

GleidsonDaniel (changes requested):

app/lib/methods/loadMessagesForRoom.ts
AlexAlexandre: cool! thanks!

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -5,32 +5,41 @@ import log from '../../utils/log';
...
 import updateMessages from './updateMessages';
+import { IMessage, IRoom, SubscriptionType, TMessageModel } from '../../definitions';
+import sdk from '../rocketchat/services/sdk';
+import RocketChat from '../rocketchat';
 
 const COUNT = 50;
 
-async function load({ rid: roomId, latest, t }) {
-	let params = { roomId, count: COUNT };
+async function load({ rid: roomId, latest, t }: Pick<IRoom, 'rid' | 'latest' | 't'>) {
```
diegolmello, line 14: We should type this as basic types for simplicity.
We need to work on those broader types.

For example, you said IRoom, but shouldn't it be ISubscription?
SubscriptionType is also broken, so use string for now.

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -5,32 +5,41 @@ import log from '../../utils/log';
...
+	let params = { roomId, count: COUNT } as { roomId: string; count: number; latest: string };
```
diegolmello, line 15: Latest can be undefined.
Otherwise we wouldn't need the `if latest`

diegolmello (changes requested):

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -5,32 +5,41 @@ import log from '../../utils/log';
...
 	if (latest) {
 		params = { ...params, latest: new Date(latest).toISOString() };
 	}
 
-	const apiType = this.roomTypeToApiType(t);
+	const apiType = RocketChat.roomTypeToApiType(t);
```
diegolmello, line 20: Use the method directly here.

PR #3712, opened 2022-02-10
Chore: Migrate sendMessage to TS

gerzonc:
## Proposed changes

Migrate `sendMessage` to TypeScript
## How to test or reproduce

Ensure sending messages and resending messages is working properly (channels and threads)
## Types of changes

## Checklist


app/lib/methods/sendMessage.ts
```diff
@@ -37,33 +48,33 @@ const changeMessageStatus = async (id, tmid, status, message) => {
...
+		await db.write(async () => {
 			await db.batch(...successBatch);
 		});
 	} catch (error) {
 		// Do nothing
 	}
 };
 
-export async function sendMessageCall(message) {
+export async function sendMessageCall(this: IRocketChat, message: TMessageModel) {
```
gerzonc, line 59: It is complaining about the `this` because of the properties we're adding, not sure why since they all live on the same context; we will check on this later

app/lib/methods/sendMessage.ts
```diff
@@ -72,41 +75,42 @@ export async function resendMessage(message, tmid) {
...
-		await sendMessageCall.call(this, m);
+
+		await sendMessageCall(m);
 	} catch (e) {
 		log(e);
 	}
 }
 
-export default async function (rid, msg, tmid, user, tshow) {
+export default async function (rid: string, msg: string, tmid: string, user: IUser, tshow?: boolean) {
```
diegolmello, line 93: `IUser` is wrong here.
We're actually expecting id and token only.

app/lib/methods/sendMessage.ts
```diff
@@ -72,41 +75,42 @@ export async function resendMessage(message, tmid) {
...
 	try {
 		const db = database.active;
 		const subsCollection = db.get('subscriptions');
 		const msgCollection = db.get('messages');
 		const threadCollection = db.get('threads');
 		const threadMessagesCollection = db.get('thread_messages');
 		const messageId = random(17);
-		const batch = [];
+		const batch: (TMessageModel | TThreadMessageModel | TThreadModel | TSubscriptionModel)[] = [];
```
diegolmello, line 101:
```suggestion
		const batch: Model[] = [];
```

diegolmello (changes requested):

app/lib/methods/sendMessage.ts
gerzonc: We use `username` and `name` as well 👀 
```ts
tm.u = {
    _id: user.id || '1',
    username: user.username,
    name: user.name
};
```

PR #3722, opened 2022-02-11
[FIX] Rooms are losing the time after search

try-catch-stack:
## Proposed changes

Favorite toggle and alter toggle don't work as expected in search View . Also the date of last activity/message is not visible on the rooms.
Returning back to the main page brings the bug to those rooms on the main page (Dates won't show up, favorite and alert toggle not work as expected to) .

https://user-images.githubusercontent.com/77742477/153672954-5e0ecc90-0e51-4ffd-b60b-8b9c7e7c0244.mp4

https://user-images.githubusercontent.com/77742477/153672963-8ff43d39-3f2a-4040-8335-c34948144120.mp4

## Types of changes

## Checklist


app/lib/rocketchat/rocketchat.js
```diff
@@ -703,6 +703,10 @@ const RocketChat = {
 
 		data = data.slice(0, 7);
 
+		if (searchView) {
+			return data;
+		}
+
 		data = data.map(sub => ({
```
diegolmello, line 710: We probably want to add extra info here.

app/lib/rocketchat/rocketchat.js
```diff
@@ -683,7 +683,7 @@ const RocketChat = {
 	readMessages,
 	resendMessage,
 
-	async localSearch({ text, filterUsers = true, filterRooms = true }) {
+	async localSearch({ text, filterUsers = true, filterRooms = true, searchView = false }) {
```
diegolmello, line 686: We shouldn't have to add this extra param.

diegolmello (changes requested):

app/lib/rocketchat/rocketchat.js
try-catch-stack: Extra info as in? I mean the data required to display a room card is return here white the data object returned below lack some of the fields. Do we need some sort of comment or something here?

app/lib/rocketchat/rocketchat.js
try-catch-stack: Added this just to check if we want the data for the search view . Do we have any other workaround?

app/lib/rocketchat/rocketchat.js
try-catch-stack: Or should we return the existing data object with the extra field in all cases regardless of where is the function called?

try-catch-stack: It would work just fine if we return the whole object without the searchView as well. I just thought that there is some reason behind returning some fields only and not the whole object (Maybe some performance benefit or something) .

diegolmello:
Can you resolve the conflicts?

try-catch-stack:
Resolved!
But not sure if it is okay to use `TSubscriptionModel` as return type. Also what do we want to do with the object returned finally?

diegolmello:
> Resolved! But not sure if it is okay to use `TSubscriptionModel` as return type. Also what do we want to do with the object returned finally?

If you want to make it easier, you can rollback to the way you did before and only add `f` to the `map`.
Then you'll only have to change the search definition.

try-catch-stack:
> > Resolved! But not sure if it is okay to use `TSubscriptionModel` as return type. Also what do we want to do with the object returned finally?
> 
> If you want to make it easier, you can rollback to the way you did before and only add `f` to the `map`. Then you'll only have to change the search definition.

Sorry , not sure which way you are talking about (Returning different object for this case or adding more fields to it?) . I mean what is this `f` for?

diegolmello:
> I mean what is this f for?

`favorite`

try-catch-stack:
Yes got it but don't we need more fields like the time of last message etc. in the object? Or am I missing your point?

diegolmello:
> Yes got it but don't we need more fields like the time of last message etc. as well in the object? Or am I missing your point?

Not really. [Here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/e792f2a49b5956f34ae8af81a1b59e37fa86e131/app/views/RoomsListView/index.tsx#L485) we're composing a query that sorts by `time` (by default). If something changes the order (like a new message), the [`subscribe`](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/e792f2a49b5956f34ae8af81a1b59e37fa86e131/app/views/RoomsListView/index.tsx#L499) is going to receive the new data and update the FlatList.

But let's say the user has `unread on top` set. If they read a channel, we should trigger an update as well. That's what [`observeWithColumns`](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/e792f2a49b5956f34ae8af81a1b59e37fa86e131/app/views/RoomsListView/index.tsx#L488) does.

Re-renders that don't change the sorting are made on the [RoomItem](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/e792f2a49b5956f34ae8af81a1b59e37fa86e131/app/presentation/RoomItem/index.tsx#L116).

Using this technique, we prevent tons of re-renders :)

---- 

That's why adding `f` to `observeWithColumns` could be enough.

diegolmello:
Got it. I think it could be good enough. I'm going to review properly soon.

try-catch-stack:
Sure (:

reinaldonetof (approved):
LGTM

PR #3731, opened 2022-02-14
Chore: Migrate selector/login to TS

gerzonc:
## Proposed changes

Migrate `selector/login` to TypeScript

## How to test or reproduce

1. Check login services are being displayed properly
2. Check login form shows properly
3. Check login button shows properly

## Types of changes

## Checklist

### DIFF
--- a/app/selectors/login.ts
+++ b/app/selectors/login.ts
@@ -1,15 +1,17 @@
 import { createSelector } from 'reselect';
 import isEmpty from 'lodash/isEmpty';
 
-const getUser = state => {
+import { IApplicationState } from '../definitions';
+
+const getUser = (state: IApplicationState) => {
 	if (!isEmpty(state.share?.user)) {
 		return state.share.user;
 	}
 	return state.login?.user;
 };
-const getLoginServices = state => state.login.services || {};
-const getShowFormLoginSetting = state => state.settings.Accounts_ShowFormLogin || false;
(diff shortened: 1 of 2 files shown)


app/selectors/login.ts
```diff
@@ -0,0 +1,22 @@
...
+import { IApplicationState } from '../definitions';
+
+const getUser = (state: IApplicationState) => {
+	if (!isEmpty(state.share?.user)) {
+		return state.share.user;
+	}
+	return state.login?.user;
+};
+const getLoginServices = (state: IApplicationState) => state.login.services || {};
+const getShowFormLoginSetting = (state: IApplicationState) => state.settings.Accounts_ShowFormLogin || false;
```
GleidsonDaniel, line 13: If you are sure of the return of this property, the right thing would be to explain its value.
As it is, it can be `number`, `string` or `boolean`, but the right thing would be to define the value, even because you already said that it can be `Accounts_ShowFormLogin` or `false`.
Something like that:
`const getShowFormLoginSetting = (state: IApplicationState) => state.settings.Accounts_ShowFormLogin as boolean || false;`

GleidsonDaniel: The same for the other properties.

GleidsonDaniel (changes requested):

app/selectors/login.ts
gerzonc: Makes sense to me

diegolmello (approved):

PR #3743, opened 2022-02-15
Chore: Migrate normalizeMessage to TS

gerzonc:
## Proposed changes

Migrate `normalizeMessage` to TypeScript
## How to test or reproduce

Send a message and add reactions or mark it as unread
## Types of changes

## Checklist


app/lib/methods/helpers/buildMessage.ts
```diff
@@ -2,7 +2,7 @@ import { IMessage } from '../../../definitions';
 import messagesStatus from '../../../constants/messagesStatus';
 import normalizeMessage from './normalizeMessage';
 
-export default (message: IMessage): IMessage => {
+export default (message: IMessage): IMessage | null => {
```
GleidsonDaniel, line 5: Why could the return be null when we always pass the same object?

app/lib/methods/helpers/mergeSubscriptionsRooms.ts
```diff
@@ -18,7 +25,7 @@ export const merge = (
 	if (room) {
 		room = EJSON.fromJSONValue(room) as ISubscription;
 		if (room._updatedAt) {
-			subscription.lastMessage = normalizeMessage(room.lastMessage);
+			subscription.lastMessage = normalizeMessage(room.lastMessage) as IMessage;
```
GleidsonDaniel, line 28: This here is related to putting null in the return of that function.

app/lib/methods/helpers/normalizeMessage.ts
```diff
@@ -11,17 +14,17 @@ function normalizeAttachments(msg) {
...
 		}
-		att = normalizeAttachments(att);
+		att = normalizeAttachments(att as IMessage & IAttachment);
 		return att;
 	});
 	return msg;
 }
 
-export default msg => {
+export default (msg: any): IMessage | null => {
```
GleidsonDaniel, line 23: `any` ?

GleidsonDaniel (changes requested):
Update to dev that some changes there in dev affect your code.

app/lib/methods/helpers/buildMessage.ts
gerzonc: Because of `normalizeMessage`, which returns null if message is falsy.
```ts
export default msg => {
	if (!msg) {
		return null;
	}
```

app/lib/methods/helpers/mergeSubscriptionsRooms.ts
gerzonc: Reason previously explained

app/lib/methods/helpers/normalizeMessage.ts
gerzonc: Diego specified me to do this because this isn't the same as an `IMessage`, it just gets converted into one.

app/lib/methods/helpers/mergeSubscriptionsRooms.ts
diegolmello: But if it's null, we shouldn't cast to IMessage.
`subscription.lastMessage` can be null.
For example, when you create a room, it has no messages.

app/lib/methods/helpers/normalizeMessage.ts
```diff
@@ -11,17 +14,17 @@ function normalizeAttachments(msg) {
 		if (att.ts) {
 			att.ts = moment(att.ts).toDate();
 		}
-		att = normalizeAttachments(att);
+		att = normalizeAttachments(att as IMessage & IAttachment);
```
diegolmello, line 17:
```suggestion
		att = normalizeAttachments(att as TMsg);
```

diegolmello (changes requested):

app/lib/methods/helpers/normalizeMessage.ts
```diff
@@ -11,17 +14,17 @@ function normalizeAttachments(msg) {
...
 		}
-		att = normalizeAttachments(att);
+		att = normalizeAttachments(att as TMsg);
 		return att;
 	});
 	return msg;
 }
 
-export default msg => {
+export default (msg: any): IMessage | IThreadResult | null => {
```
GleidsonDaniel, line 23: In the `buildMessage` method the `msg` can be `Partial<IMessage> | IThreadResult`, in `mergeSubscriptionsRooms` the `msg` can be `ILastMessage | null | undefined`.
Why are we using any here, when we know where this function is used and what types are passed?

GleidsonDaniel (changes requested):

app/lib/methods/helpers/normalizeMessage.ts
gerzonc: As previously mentioned, Diego specified me to do this because this isn't the same as an `IMessage`, it just gets converted into one; if any of those other functions are setting a return type related to this function, then we might be doing something wrong. To be honest, I'm quite unsure.

GleidsonDaniel (approved):

PR #3751, opened 2022-02-16
Chore: Migrate search and localSearch methods to Typescript

GleidsonDaniel:
## Proposed changes

Migrate search and localSearch methods to Typescript
t
## How to test or reproduce

It is necessary to test all search fields, such as new conversation and room search.

## Types of changes

## Checklist


app/lib/rocketchat/methods/search.ts
```diff
@@ -0,0 +1,98 @@
...
+			if (filterUsers) {
+				users
+					.filter((item1, index) => users.findIndex(item2 => item2._id === item1._id) === index) // Remove duplicated data from response
+					.filter(user => !data.some(sub => user.username === sub.name)) // Make sure to remove users already on local database
+					.forEach(user => {
+						data.push({
+							...user,
+							rid: user.username,
+							name: user.username,
+							t: SubscriptionType.R_THREAD,
```
diegolmello, line 39: This is actually `'d'`, so you should use SubscriptionType.DIRECT.

We're just mapping the users returned from the search and adding the type manually to make things easier.

app/lib/rocketchat/methods/search.ts
```diff
@@ -0,0 +1,98 @@
...
+
+import { sanitizeLikeString } from '../../database/utils';
+import database from '../../database/index';
+import { spotlight } from '../services/restApi';
+import isGroupChat from './isGroupChat';
+import { ISearch, ISearchLocal, SubscriptionType } from '../../../definitions';
+
+let debounce: null | ((reason: string) => void) = null;
+
+export const search = async ({ text = '', filterUsers = true, filterRooms = true }): Promise<(ISearch | ISearchLocal)[]> => {
```
diegolmello, line 11: We should declare `localSearch` before `search` 🙈

app/views/NewMessageView.tsx
```diff
@@ -313,7 +301,7 @@ class NewMessageView extends React.Component<INewMessageViewProps, INewMessageVi
 			<FlatList
 				data={search.length > 0 ? search : chats}
 				extraData={this.state}
-				keyExtractor={item => item._id}
+				keyExtractor={item => item._id || item.rid}
```
diegolmello, line 304: Now with the code more readable, I think we can remove that `_id`, can't we?

diegolmello (changes requested):

app/views/NewMessageView.tsx
GleidsonDaniel: Worse than not, when the object comes directly from the api it doesn't have the rid.

diegolmello (approved):

PR #3778, opened 2022-02-25
Chore: Migrate RoomInfoView to Typescript

reinaldonetof:
## Proposed changes

Migrate RoomInfoView to Typescript

## How to test or reproduce

- Access RoomInfoView from RoomView clicking on: profile pic, name
- Inside the RoomView click on the header, then click on the card that is showing the avatar of the channel or user, you should be redirected to RoomInfoView

## Types of changes

## Checklist


app/definitions/rest/v1/omnichannel.ts
```diff
@@ -20,11 +20,7 @@ export type OmnichannelEndpoints = {
 	};
 	'livechat/visitors.info': {
 		GET: (params: { visitorId: string }) => {
-			visitor: {
-				visitorEmails: Array<{
-					address: string;
-				}>;
-			};
+			visitor: ILivechatVisitor;
```
reinaldonetof, line 23: We need to evaluate this later. Because the only param `visitorEmails` returned is not the truth!

app/views/RoomInfoView/index.tsx
```diff
@@ -60,28 +75,57 @@ const getRoomTitle = (room, type, name, username, statusText, theme) =>
...
+
+export interface ILivechatVisitorModified extends ILivechatVisitor {
+	os?: string;
+	browser?: string;
+}
+
+interface IRoomInfoViewState {
+	room: ISubscription;
+	// TODO: Could be IUserParsed or ILivechatVisitorModified
+	roomUser: any;
```
reinaldonetof, line 106: I wasn't able to do here the type.

app/containers/Avatar/interfaces.ts
```diff
@@ -9,7 +11,7 @@ export interface IAvatar {
 	size?: number;
 	borderRadius?: number;
 	type?: string;
-	children?: JSX.Element;
+	children?: React.ReactNode;
```
diegolmello, line 14: We should avoid React.ReactNode.
I don't think there's a use case for it on React Native at all.

It let's you do this:
```tsx
<Avatar>
	anything
</Avatar>
```
That wouldn't be an issue on web, but it'd crash in RN.
For that to work on RN, we'd need to wrap the content with a Text element.
```tsx
<Avatar>
	<Text>anything</Text>
</Avatar>
```

```suggestion
	children?: React.ReactElement | null;
```

Good read: https://changelog.com/posts/the-react-reactnode-type-is-a-black-hole

app/definitions/rest/v1/omnichannel.ts
diegolmello: It's probably going to be fixed on our requests effort.

app/views/RoomInfoView/Livechat.tsx
```diff
@@ -50,37 +51,38 @@ const Livechat = ({ room, roomUser, theme }) => {
...
-			<Item label={I18n.t('Facebook')} content={room.facebook?.page.name} theme={theme} />
-			<Item label={I18n.t('SMS')} content={room.sms && 'SMS Enabled'} theme={theme} />
-			<Item label={I18n.t('Topic')} content={room.topic} theme={theme} />
-			<Item label={I18n.t('Tags')} content={room.tags?.join(', ')} theme={theme} />
-			<Item label={I18n.t('Department')} content={department.name} theme={theme} />
-			<CustomFields customFields={room.livechatData} theme={theme} />
+			<Item label={I18n.t('Agent')} content={room.servedBy?.username} />
+			{/* TODO: Check if this param is outdated */}
+			{/* @ts-ignore */}
+			<Item label={I18n.t('Facebook')} content={room.facebook?.page.name} />
```
diegolmello, line 76: I couldn't find this (or sms) on main repo definitions.
Is this a custom column or is it for Omnichannel only?
If you don't know, can you get this info from Omnichannel team?

diegolmello (changes requested):

app/containers/Avatar/interfaces.ts
reinaldonetof: Nice, awesome point!

diegolmello (approved):

PR #3959, opened 2022-03-24
[IMPROVE] Support Omnichannel icons

reinaldonetof:
## How to test or reproduce

- First, with a Whatsapp account, send a message to: +55 51 93469973
- Then, with a livechat agent logged in `multiverse.rocket.chat`, take the chat take will come to Sales Department
- The icon needs to be the Whatsapp

#### Icon if the internet fall
- If you had taken it a chat with a custom icon and your internet fall, will show the omnichannel's custom icon

#### Futher infos
- You can test with `candidate.rocket.chat` too,
- To create a new chat you need to go to https://www.facebook.com/104516168873143 and send a message through this page

#### Before
Screen Shot 2022-03-28 at 11 13 02

#### After
Screen Shot 2022-03-28 at 10 45 42

## Types of changes

## Checklist


app/containers/RoomTypeIcon/OmnichannelRoomIcon.tsx
```diff
@@ -0,0 +1,50 @@
...
+	size: number;
+	type: string;
+	style?: StyleProp<ViewStyle>;
+	status?: string;
+	sourceType?: IOmnichannelSourceConnected;
+}
+
+export const OmnichannelRoomIcon = ({ size, style, sourceType, status }: IOmnichannelRoomIconProps) => {
+	if (sourceType?.type === OmnichannelSourceType.APP && sourceType.id && sourceType.sidebarIcon && sourceType.connected) {
+		const baseUrl = useSelector((state: IApplicationState) => state.server?.server);
```
diegolmello, line 29: You should never call hooks inside conditions.
Please read this: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

My suggestion is to create another component for the SVG.

app/containers/RoomTypeIcon/index.tsx
```diff
@@ -14,21 +15,22 @@ const styles = StyleSheet.create({
...
 	style?: ViewStyle;
+	sourceType?: IOmnichannelSourceConnected;
 }
 
-const RoomTypeIcon = React.memo(({ type, isGroupChat, status, style, theme, teamMain, size = 16 }: IRoomTypeIcon) => {
+const RoomTypeIcon = React.memo(({ type, isGroupChat, status, style, teamMain, size = 16, sourceType }: IRoomTypeIcon) => {
 	if (!type) {
 		return null;
 	}
+	const { theme } = useTheme();
```
diegolmello, line 31: Same here.
Hooks always at the top level.

app/containers/RoomHeader/index.tsx
```diff
@@ -102,6 +120,8 @@ class RoomHeaderContainer extends Component<IRoomHeaderContainerProps, any> {
 			subtitle = subtitleProp;
 		}
 
+		const sourceT = sourceType ? { ...sourceType, connected } : undefined;
```
diegolmello, line 123: This is awkward.
We already receive the data properly.
Why should we mutate it to add `connected`?
Why not just use `useSelector` to fetch `connected` inside the component?

app/definitions/IRoom.ts
```diff
@@ -65,6 +65,27 @@ export enum OmnichannelSourceType {
...
+	// The sidebar icon
+	sidebarIcon?: string;
+	// The default sidebar icon
+	defaultIcon?: string;
+	_updatedAt?: Date;
+	queuedAt?: Date;
+}
+
+export interface IOmnichannelSourceConnected extends IOmnichannelSource {
+	connected: boolean;
```
diegolmello, line 87: This is only needed because we're mutating the data.

diegolmello (changes requested):
Besides those comments, have you reviewed those snapshots?
I'm missing new stories for these new states.

app/containers/RoomTypeIcon/index.tsx
```diff
@@ -38,6 +40,10 @@ const RoomTypeIcon = React.memo(({ type, isGroupChat, status, style, theme, team
 		return <Status style={[iconStyle, { color: STATUS_COLORS[status] }]} size={size} status={status} />;
 	}
 
+	if (type === 'l') {
```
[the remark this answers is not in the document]
reinaldonetof: Just to be more clear that when is 'l' we are returning a component, instead of a text.

app/containers/RoomHeader/index.tsx
GleidsonDaniel, line 123: Why are you passing `connect` as a prop here, if in the `OmnichannelRoomIcon` file you are getting this prop with useSelector?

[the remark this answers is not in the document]
GleidsonDaniel: As it is passing via prop drilling, it can cause a lot of re-rendering in the dependency tree to pass it like this.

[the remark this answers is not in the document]
GleidsonDaniel: As it is passing via prop drilling, it can cause a lot of re-rendering in the dependency tree to pass it like this.

GleidsonDaniel (changes requested):

app/containers/RoomHeader/index.tsx
reinaldonetof: You are right! I forgot to remove the prop in this file

GleidsonDaniel (approved):

PR #3966, opened 2022-03-25
[IMPROVE] Enable multiline for Android tablets

gerzonc:
## Proposed changes

Changes how keyboard behaves on Android tablets in order to enable multiline message box; now it behaves just like the phone version
## How to test or reproduce

Test keyboard usage with or without physical keyboard, just to ensure it is working properly.
## Screenshots
<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
      </thead>
    </tr>
    <tbody>
      <tr>
        <td>
        image
        </td>
        <td>
        image
        </td>
      </tr>
    </tbody>
</table >

## Types of changes

## Checklist


diegolmello (changes requested):
> Changes how keyboard behaves on Android tablets 

What are the changes?
Can you add videos to show the difference in behavior?

app/containers/MessageBox/index.tsx
```diff
@@ -1040,9 +1040,7 @@ class MessageBox extends Component<IMessageBoxProps, IMessageBoxState> {
 		const isAndroidTablet: Partial<IThemedTextInput> =
 			isTablet && isAndroid
 				? {
-						multiline: false,
-						onSubmitEditing: this.submit,
-						returnKeyType: 'send'
+						onSubmitEditing: this.submit
```
diegolmello, line 1043: If it behaves like the smartphone version, why do we need `isAndroidTablet`?

diegolmello (changes requested):

app/containers/MessageBox/index.tsx
gerzonc: Lmao true

GleidsonDaniel:
This is a change that changes the behavior of message sending.
Since now it not only "enables" the multiline, it also disables sending with enter, because the enter now jumps the text line as it is done in the mobile app.

I don't see any problems, but was that the purpose of the task from the beginning? I'm talking about changing the behavior of the app.

@diegolmello

gerzonc:
> This is a change that changes the behavior of message sending. Since now it not only "enables" the multiline, it also disables sending with enter, because the enter now jumps the text line as it is done in the mobile app.
> 
> I don't see any problems, but was that the purpose of the task from the beginning? I'm talking about changing the behavior of the app.
> 
> @diegolmello

Yes and no. Originally we were looking for a way to make `MessageBox` to work with `multiline`, but this is not possible without changing how the keyboard behaves on Android 😭 It's been an open issue for a couple of years now on the react-native repo and sadly, this is the best solution. In any case, I think not only we should do it but as well as to evaluate the user experience so is not a complication for users later on.

diegolmello:
> > This is a change that changes the behavior of message sending. Since now it not only "enables" the multiline, it also disables sending with enter, because the enter now jumps the text line as it is done in the mobile app.
> > I don't see any problems, but was that the purpose of the task from the beginning? I'm talking about changing the behavior of the app.
> > @diegolmello
> 
> Yes and no. Originally we were looking for a way to make `MessageBox` to work with `multiline`, but this is not possible without changing how the keyboard behaves on Android 😭 It's been an open issue for a couple of years now on the react-native repo and sadly, this is the best solution. In any case, I think not only we should do it but as well as to evaluate the user experience so is not a complication for users later on.

It's still single line because we found the issue you're talking about back then when we implemented it.
If the issue is still there, nothing changed.
I don't think we should merge this.

gerzonc:
> > > This is a change that changes the behavior of message sending. Since now it not only "enables" the multiline, it also disables sending with enter, because the enter now jumps the text line as it is done in the mobile app.
> > > I don't see any problems, but was that the purpose of the task from the beginning? I'm talking about changing the behavior of the app.
> > > @diegolmello
> > 
> > 
> > Yes and no. Originally we were looking for a way to make `MessageBox` to work with `multiline`, but this is not possible without changing how the keyboard behaves on Android 😭 It's been an open issue for a couple of years now on the react-native repo and sadly, this is the best solution. In any case, I think not only we should do it but as well as to evaluate the user experience so is not a complication for users later on.
> 
> It's still single line because we found the issue you're talking about back then when we implemented it. If the issue is still there, nothing changed. I don't think we should merge this.

I'll close this then and we'll retry on the future.

PR #4200, opened 2022-05-16
Chore: Hooks - migrate RoomHeader to hook

reinaldonetof:
## How to test or reproduce

- Open any room
- The room header should work properly

#### Test the status of another user
- Login with user A through the mobile app
- Then, on another device, login with user B
- Start a direct channel between both
- Then when user A changes his status from online, busy, away, offline, etc.
- The status in user B's RoomHeader about user A should be rendered properly

## Types of changes

## Checklist


app/containers/RoomHeader/index.tsx
```diff
@@ -1,124 +1,79 @@
...
+			if (connecting) {
+				setSubtitle(I18n.t('Connecting'));
+			} else if (!connected) {
+				setSubtitle(I18n.t('Waiting_for_network'));
+			} else {
+				setSubtitle(subtitleProp);
+			}
+		}, [connecting, connected]);
+
+		useEffect(() => {
```
diegolmello, line 65: You don't have to create a state to render the value you want.
You already receive those values from `props`, the same way we did on `mapStateToProps`.
If any of those `useSelector` changes, the re-render is going to be triggered and your value would work just fine.
Adding this extra state the way you did is just re-rendering unnecessarily.

app/containers/RoomHeader/index.tsx
```diff
@@ -1,124 +1,79 @@
...
 
-		let subtitle;
-		if (connecting) {
-			subtitle = I18n.t('Connecting');
-		} else if (!connected) {
-			subtitle = I18n.t('Waiting_for_network');
-		} else {
-			subtitle = subtitleProp;
-		}
+		useEffect(() => {
```
diegolmello, line 55: Same here, you can just use the value from the selector.

app/containers/RoomHeader/index.tsx
```diff
@@ -139,29 +94,28 @@ class RoomHeaderContainer extends Component<IRoomHeaderContainerProps, any> {
...
-	let status = 'offline';
-	const { roomUserId, type, visitor = {}, tmid } = ownProps;
-
-	if (state.meteor.connected) {
-		if ((type === 'd' || (tmid && roomUserId)) && state.activeUsers[roomUserId]) {
-			({ status, statusText } = state.activeUsers[roomUserId]);
-		} else if (type === 'l' && visitor?.status) {
-			({ status } = visitor);
+	},
+	(prevProps, nextProps) => {
```
diegolmello, line 98: Do we still need this?

app/views/RoomView/index.tsx
```diff
@@ -522,13 +522,12 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
...
 			headerTitle: () => (
 				<RoomHeader
-					rid={rid}
 					prid={prid}
 					tmid={tmid}
 					title={title}
 					teamMain={teamMain}
 					parentTitle={parentTitle}
-					subtitle={subtitle}
+					subtitle={subtitle || undefined}
```
diegolmello, line 530: Undefined is awkward.
Why is this change needed?

app/containers/RoomHeader/index.tsx
```diff
@@ -1,124 +1,79 @@
...
-import { withDimensions } from '../../dimensions';
+import { IApplicationState, TUserStatus, IOmnichannelSource, IVisitor } from '../../definitions';
+import { useDimensions } from '../../dimensions';
 import I18n from '../../i18n';
 import RoomHeader from './RoomHeader';
 
 interface IRoomHeaderContainerProps {
-	title: string;
-	subtitle: string;
+	title?: string;
```
diegolmello, line 11: A room without title?
Can you explain?

diegolmello (changes requested):

app/containers/RoomHeader/index.tsx
reinaldonetof: Yeap. You send a message, then reply in a thread. After, you delete the message sent in room view, the thread continues to exist and if you go to threadMessagesView you can see the thread without message and when you click to open it, the header will be without title.

This error occurred a quite time ago, when we was fixing something in Thread Messages View.

app/containers/RoomHeader/index.tsx
reinaldonetof: Yes and no. Let me explain,

For a normal user/channel/room the RoomHeader will be rendered only once when open the room. But, for omnichannel's chat the roomheader re-render 3 times when open the room! WDYT? Keep the check difference or just remove them?

PR #4207, opened 2022-05-18
[FIX] Showing blank screen on share extension if the user is not logged in

reinaldonetof:
## Proposed changes

Redirect users to WithoutServersView in case there was no logged in server.

## How to test or reproduce

- Make sure the app is not logged in any server (if you open from app icon, shows onboarding screen)
- Open Photos or Gallery and try to share a photo using [Rocket.Chat](http://rocket.chat/)

## Types of changes

## Checklist

### DIFF
--- a/app/containers/ActionSheet/ActionSheet.tsx
+++ b/app/containers/ActionSheet/ActionSheet.tsx
@@ -63,7 +63,6 @@ const ActionSheet = React.memo(
 
 		const hide = () => {
 			bottomSheetRef.current?.close();
-			toggleVisible();
 		};
 
 		const show = (options: TActionSheetOptions) => {


app/share.tsx
```diff
@@ -104,6 +104,9 @@ class Root extends React.Component<{}, IState> {
 			scale,
 			fontScale
 		};
+	}
+
+	UNSAFE_componentWillMount() {
```
diegolmello, line 109: I'm missing a proper explanation on why this is needed on the description.
Why don't we need this on `/app/index`? https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/f5625cd5f3fa16e2ff06f45728174ecbe6d91fd1/app/index.tsx#L84

diegolmello (changes requested):

app/share.tsx
diegolmello: If your whole state explanation is correct, why does it break it if `currentServer === null` only?
It works properly inside `if (currentServer) {}`.

We used to call `const currentServer = UserPreferences.getString(CURRENT_SERVER);` as asynchronously before #3634 and it might have broken it.
We should try to understand what really happened before making this change you're proposing.
Also, `UNSAFE_componentWillMount` shouldn't be used.

app/share.tsx
```diff
@@ -117,10 +123,12 @@ class Root extends React.Component<{}, IState> {
 
 		if (currentServer) {
 			await localAuthenticate(currentServer);
-			this.setState({ root: 'inside' });
+			// @ts-ignore
+			this.mounted ? this.setState({ root: 'inside' }) : (this.state.root = 'inside');
```
diegolmello, line 127: You don't have to do this here, as `await localAuthenticate(currentServer)` already makes sure a render occurred.

app/share.tsx
```diff
@@ -117,10 +123,12 @@ class Root extends React.Component<{}, IState> {
...
 			await shareExtensionInit(currentServer);
 		} else {
-			this.setState({ root: 'outside' });
+			// @ts-ignore
+			this.mounted ? this.setState({ root: 'outside' }) : (this.state.root = 'outside');
```
diegolmello, line 131: Can we not use inline if?
It's not as clear as a normal if and you're actually hiding what's getting ignored on that ts-ignore.

diegolmello (changes requested):

diegolmello (approved):

PR #4216, opened 2022-05-25
Chore: Migrate LoginServices to Hooks

reinaldonetof:
## How to test or reproduce

- Join into `open.rocket.chat`
- Go to Login screen and test the oauth's buttons, `more options` and login with email too

## Types of changes

## Checklist


app/containers/LoginServices.tsx
```diff
@@ -92,154 +94,145 @@ interface IOauthProvider {
...
+	const { length } = Object.values(services);
+
+	const heightButtons = useSharedValue(SERVICES_COLLAPSED_HEIGHT);
+
+	const animatedStyle = useAnimatedStyle(() => ({
+		overflow: 'hidden',
+		height: withTiming(heightButtons.value, { duration: 300, easing: Easing.inOut(Easing.quad) })
+	}));
+
+	const onPressFacebook = () => {
```
diegolmello, line 123: On our refinement, we talked about extracting all these service logic functions away from the component.

app/containers/LoginServices.tsx
```diff
@@ -271,77 +264,51 @@ class LoginServices extends React.PureComponent<ILoginServicesProps, ILoginServi
...
+			twitter: onPressTwitter,
+			wordpress: onPressWordpress
 		};
 		return oauthProviders[name];
 	};
 
-	renderServicesSeparator = () => {
-		const { collapsed } = this.state;
-		const { services, separator, theme } = this.props;
+	const renderServicesSeparator = () => {
```
diegolmello, line 285: You don't want to use `renderSomething`.
That's ok for class components, but not for function components, as it's going to get recreated on every render.
The ideal fix is to extract the logic to its own component `<ServicesSeparator />` and send the necessary props.

A _not-so-valid_ fix would be to use `useCallback`, but it's hard to maintain and it's kind of anti-pattern.
There are use cases for this, but 99% of times a separate component is the right way.

app/containers/LoginServices.tsx
```diff
@@ -271,77 +264,51 @@ class LoginServices extends React.PureComponent<ILoginServicesProps, ILoginServi
...
+						}}
 						style={styles.options}
 						color={themes[theme].actionTintColor}
 					/>
 					<OrSeparator theme={theme} />
 				</>
 			);
 		}
-		if (length > 0 && separator) {
+		if (length > 0 && Accounts_ShowFormLogin) {
```
diegolmello, line 305: I didn't get this change.

diegolmello (changes requested):

app/containers/LoginServices/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+		<>
+			{Object.values(services).map((service: IItemService) => (
+				<ServicesList
+					CAS_enabled={CAS_enabled}
+					CAS_login_url={CAS_login_url}
+					Gitlab_URL={Gitlab_URL}
+					server={server}
+					service={service}
+				/>
+			))}
```
diegolmello, line 73: This is wrong.
`ServicesList` should be the whole map and render a `Service` component (called `ServiceList` now).

```tsx
// index.tsx
<ServicesList services={services} />

// ServicesList.tsx
{Object.values(services).map((service: IItemService) => (
	<Service
		CAS_enabled={CAS_enabled}
		CAS_login_url={CAS_login_url}
		Gitlab_URL={Gitlab_URL}
		server={server}
		service={service}
	/>
))}
```

app/containers/LoginServices/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+			<ServicesSeparator
+				services={services}
+				separator={separator}
+				collapsed={collapsed}
+				onPressButtonSeparator={onPressButtonSeparator}
```
diegolmello, line 78: No need to name it using the context.

```suggestion
				onPress={onPressButtonSeparator}
```

app/containers/LoginServices/interfaces.ts
```diff
@@ -0,0 +1,37 @@
...
+	url: string;
+	ssoToken?: string;
+	authType?: string;
+}
+
+export interface IItemService {
+	_id: string;
+	name: string;
+	service: string;
+	authType: string;
```
diegolmello, line 11: You know the available types.
We could benefit from having them defined here.

```suggestion
	authType: 'facebook' | 'github' | ...;
```

app/containers/LoginServices/serviceLogin.ts
```diff
@@ -0,0 +1,160 @@
...
+import { Base64 } from 'js-base64';
+
+import { events, logEvent } from '../../utils/log';
+import { Services } from '../../lib/services';
+import Navigation from '../../lib/navigation/appNavigation';
+import { IItemService, IOpenOAuth, IServiceLogin } from './interfaces';
+import random from '../../utils/random';
+
+const LOGIN_STYPE_POPUP = 'popup';
+const LOGIN_STYPE_REDIRECT = 'redirect';
```
diegolmello, line 12: Unify into TLoginStyle

app/containers/LoginServices/serviceLogin.ts
```diff
@@ -0,0 +1,160 @@
...
+		await Services.loginOAuthOrSso({ fullName, email, identityToken });
+	} catch {
+		logEvent(events.ENTER_WITH_APPLE_F);
+	}
+};
+
+const getOAuthState = (loginStyle = LOGIN_STYPE_POPUP) => {
+	const credentialToken = random(43);
+	let obj: {
+		loginStyle: string;
```
diegolmello, line 144: Apply TLoginStyle

app/containers/LoginServices/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+					collapsed={collapsed}
+					onPressButtonSeparator={onPressButtonSeparator}
+				/>
+			</>
+		);
+	}
+	return (
+		<>
+			{Object.values(services).map((service: IItemService) => (
+				<ServicesList
```
diegolmello, line 66: You might need to add `key` to prevent re-renders

diegolmello (changes requested):

app/containers/LoginServices/Service.tsx
```diff
@@ -0,0 +1,101 @@
...
+		CAS_login_url: string;
+		storiesTestOnPress?: () => void;
+	}) => {
+		const { colors } = useTheme();
+
+		const { name } = service;
+		let modifiedName = name === 'meteor-developer' ? 'meteor' : name;
+		const icon = `${modifiedName}-monochromatic` as TIconsName;
+		const isSaml = service.service === 'saml';
+		let onPress: any = () => {};
```
GleidsonDaniel, line 33: Will this value be changed?
Why is it not recommended to use let inside components, as you have no reference to which value that is, which causes some re-rendering.
Same buttonText

GleidsonDaniel (changes requested):

GleidsonDaniel (approved):

PR #4195, opened 2022-05-12
Redesign searchbox

AlexAlexandre:
## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4256, opened 2022-05-30
Chore: Migration to Hooks - FormTextInput

AlexAlexandre:
## How to test or reproduce
As we changed the FormTextInput, we should test all inputs related: 
* New server view;
* Login;
* Register;
* Forgot password; 
* Change Status;
* Share view;
* Search Message view;
* Room view -> join code;
* Room View -> Edit info;
* Profile;
* E2E screens... (Enter your password)
* Create Discussions;
* Two Factor

## Types of changes

## Checklist


app/containers/TextInput/FormTextInput.tsx
```diff
@@ -58,112 +57,91 @@ export interface IRCTextInputProps extends TextInputProps {
...
+		inputRef,
+		iconLeft,
+		iconRight,
+		left,
+		testID,
+		secureTextEntry,
+		placeholder,
+		...inputProps
+	}: IRCTextInputProps) => {
+		const { colors, theme } = useTheme();
```
diegolmello, line 81: Stop using theme here.
Use it on TextInput/index directly.

app/containers/TextInput/FormTextInput.tsx
```diff
@@ -58,112 +57,91 @@ export interface IRCTextInputProps extends TextInputProps {
...
-export default class FormTextInput extends React.PureComponent<IRCTextInputProps, IRCTextInputState> {
-	static defaultProps = {
-		error: {},
-		theme: 'light'
-	};
-
-	state = {
-		showPassword: false
-	};
+const FormTextInput = React.memo(
```
diegolmello, line 65: This memo isn't doing anything.
TextInput doesn't handle its own state, so it receives props all the time, which is going to trigger re-renders anyway.
Even if it wasn't the case, it receives non-primitive values which also would trigger re-renders.

app/containers/TextInput/FormTextInput.tsx
```diff
@@ -58,112 +57,91 @@ export interface IRCTextInputProps extends TextInputProps {
...
 
-	get iconRight() {
-		const { iconRight, onIconRightPress, theme } = this.props;
-		return iconRight ? (
-			<Touchable onPress={onIconRightPress} style={[styles.iconContainer, styles.iconRight]}>
-				<CustomIcon name={iconRight} size={20} color={themes[theme].bodyText} />
-			</Touchable>
-		) : null;
-	}
+		const showIconRight = () =>
```
diegolmello, line 95: You should prevent the usage of this kind of function creation inside function components.
You're only recreating something that React can't really understand.
On this case, it's actually easy to just abstract and reuse it inside the return.

app/containers/TextInput/FormTextInput.tsx
```diff
@@ -58,112 +57,91 @@ export interface IRCTextInputProps extends TextInputProps {
...
-
-	get loading() {
-		const { theme } = this.props;
-		return <ActivityIndicator style={[styles.iconContainer, styles.iconRight]} color={themes[theme].bodyText} />;
-	}
 
-	tooglePassword = () => {
-		this.setState(prevState => ({ showPassword: !prevState.showPassword }));
-	};
+		const showLoading = () => (
```
diegolmello, line 117: Similar to the icons. There's no reason to return a function here.
```jsx
{loading ? <ActivityIndicator ... /> : null}
```

diegolmello (changes requested):
Left a few comments, but you should actually wait for https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4195 so you prevent rework.

app/containers/TextInput/TextInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+			ref={ref}
+			style={[{ color: themes[theme].titleText }, style, styles.input]}
+			placeholderTextColor={themes[theme].auxiliaryText}
+			keyboardAppearance={theme === 'light' ? 'light' : 'dark'}
+			{...props}
+		/>
+	);
+});
+
+// export default ThemedTextInput;
```
GleidsonDaniel, line 31: 👀

GleidsonDaniel (approved):

PR #4264, opened 2022-06-02
Chore: Migration to Hooks - Markdown

AlexAlexandre:
## How to test or reproduce
We are changing our Markdown component, so we should test everything related it. 
E.G:
* formatted texts:
  * fonts with bold style: `*bold test*` should be **bold test**;
  * fonts with italic style: `_italic test_` should be _italic test_
  * fonts with strike style: `~strike test~` should be ~strike test~
* all list options:
  * bulet lists
  * ordered lists
* Different text style;
* Preview messages;
* Mentions like: `@all, @here, @some-user and @some-user-that-does-not-exist` 
* Hashtags like: `#channel-that-exist and #channel-that-does-not-exist`
* Emojis like: custom emojis, unicode, short names
* Block quote message;
* Links: verify if is correctly formatted, if shows a preview;
* Images;
* Headers like: `# header, ## Header ### Header. #### Header. ##### Header and ###### Header`
* Collapsible messages should still works fine
* messages with code format:
  *  \` code inline \` should be `code inline` . 
  * \``` code multiline \``` should be 
```
here we have more then one line 
here we have more then one line 
```

* Tables: this text 
'First Header | Second Header
------------ | -------------
Content from cell 1 | Content from cell 2
Content in the first column | Content in the second column'

should be formatted like: 

| First Header | Second Header | 
------------ | -------------
| Content from cell 1 | Content from cell 2 |
| Content in the first column | Content in the second column |

Remember to disable the message parse to get the table correctly formatted.

Also test on Ipads and tablets, please.

## Types of changes

## Checklist


app/containers/markdown/index.tsx
```diff
@@ -86,271 +84,244 @@ const emojiCount = (str: string) => {
...
-					},
-					...style
-				]}>
-				{literal}
-			</Text>
-		);
-	};
+	const formattedMessage = formatHyperlink(formatText(msg));
+	const ast = mergeTextNodes(parser.parse(formattedMessage));
+	isMessageContainsOnlyEmoji.current = isOnlyEmoji(formattedMessage) && emojiCount(formattedMessage) <= 3;
```
reinaldonetof, line 165: I think you can move these lines 163 -165 to before the line 325. Because if isNewMarkdown is true, you don't need to waste time parsing the message with the lib. WDYT?

reinaldonetof (changes requested):

GleidsonDaniel (approved):

PR #4310, opened 2022-06-20
Chore: Migrate RoomItem to Hooks

GleidsonDaniel:
## Proposed changes

Migrate RoomItem to hooks

## How to test or reproduce

Check if the status is changing correctly.
Check the gestures, for right or left.

## Types of changes

## Checklist


app/containers/RoomItem/index.tsx
```diff
@@ -23,149 +26,106 @@ const attrs = [
...
-			status,
-			showLastMessage,
-			username,
-			useRealName,
-			swipeEnabled,
-			autoJoin,
-			showAvatar,
-			displayMode
-		} = this.props;
+let roomSubscription: Subscription | null;
```
diegolmello, line 29: If I got it right, this should be a ref inside `RoomItemContainer`.
The way you're proposing is a global variable and it's going to be overwritten by other subscriptions rendered on the list.

app/containers/RoomItem/index.tsx
```diff
@@ -23,149 +26,106 @@ const attrs = [
...
+let accessibilityLabel = '';
```
diegolmello, line 30: Similar to `roomSubscription`, but there's actually no benefit from having it out of the component.

app/containers/RoomItem/index.tsx
```diff
@@ -23,149 +26,106 @@ const attrs = [
...
-			accessibilityLabel += `, ${I18n.t('you_were_mentioned')}`;
-		}
-
-		if (date) {
-			accessibilityLabel += `, ${I18n.t('last_message')} ${date}`;
-		}
+		const connected = useAppSelector(state => state.meteor.connected);
+		const userStatus = useAppSelector(state => state.activeUsers[id || '']?.status);
+		const [_, forceUpdate] = useReducer(x => x + 1, 0);
+		const { theme } = useTheme();
```
diegolmello, line 63: Should we stop using prop drilling `theme` and use `colors` inside the other component instead?

app/containers/RoomItem/index.tsx
```diff
@@ -23,149 +26,106 @@ const attrs = [
...
+
+		useEffect(() => {
+			const init = () => {
+				if (item?.observe) {
+					const observable = item.observe();
+					roomSubscription = observable?.subscribe?.(() => {
+						if (_) forceUpdate();
+					});
+				}
+				return () => roomSubscription?.unsubscribe && roomSubscription.unsubscribe();
```
diegolmello, line 73: Shorter without hurting readability 🙈 

```suggestion
				return () => roomSubscription?.unsubscribe?.();
```

app/containers/RoomItem/index.tsx
```diff
@@ -23,149 +26,106 @@ const attrs = [
...
+		}, []);
+
+		useEffect(() => {
+			const isDirect = !!(item.t === 'd' && id && !isGroupChat(item));
+			if (connected && isDirect) {
+				getUserPresence(id);
+			}
+		}, [connected]);
+
+		useEffect(() => {
```
diegolmello, line 85: I don't think this effect is actually needed (not really impacting performance).
It could be inside return and than you could just remove the `let accessibilityLabel`, creating here instead

app/containers/RoomItem/interfaces.ts
```diff
@@ -64,48 +64,43 @@ export interface ITypeIconProps {
...
-	width: number;
-	status: TUserStatus;
-	toggleFav(): void;
-	toggleRead(): void;
-	hideChannel(): void;
+interface IRoomItemTouchables {
+	toggleFav?: (rid: string, favorite: boolean) => Promise<void>;
+	toggleRead?: (rid: string, tIsRead: boolean) => Promise<void>;
+	hideChannel?: (rid: string, type: SubscriptionType) => Promise<void>;
+	onPress: (item?: any) => void;
```
diegolmello, line 71: `item` is still too unclear, right?
I know it could be `TSubscriptionModel`, but this component is also used on some other places :(

GleidsonDaniel: This one has a problem with types in "general" it's not just because it's used in other places, also because some types aren't so clear.

app/containers/RoomItem/index.tsx
```diff
@@ -23,149 +25,102 @@ const attrs = [
...
+			const init = () => {
+				if (item?.observe) {
+					const observable = item.observe();
+					roomSubscription.current = observable?.subscribe?.(() => {
+						if (_) forceUpdate();
+					});
+				}
+				return () => roomSubscription.current?.unsubscribe();
```
davidlougheed, line 70: this would never be called, no?

diegolmello: Nice catch. Thanks!

PR #4327, opened 2022-06-28
Chore: Migrate NotificationPreferencesView to hooks

GleidsonDaniel:
## Proposed changes

Migrate NotificationPreferencesView to hooks and improves layout and user feedback 
- It is no longer necessary to change screens to select some options, they are via actionSheet.
- Now when an error occurs when changing the option, it reverts to the old option.

## How to test or reproduce

- Try changing some option and see if it will actually replicate on the web or in tests.
- Force an error (too many requests for example) and see the error message and the go back option.

## Types of changes

## Checklist


app/containers/ActionSheet/Item.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-nested-ternary */
```
diegolmello, line 1: I don't think this is needed.

app/containers/ActionSheet/styles.ts
```diff
@@ -63,5 +62,6 @@ export default StyleSheet.create({
 	},
 	rightContainer: {
 		paddingLeft: 12
-	}
+	},
+	none: { width: 20 }
```
diegolmello, line 66: You're not using it

app/views/NotificationPreferencesView/index.tsx
```diff
@@ -1,282 +1,234 @@
...
-					this.state.room = changes;
-				}
-			});
-		}
-	}
-
-	componentDidMount() {
-		this.mounted = true;
-	}
+const RenderListPicker = React.memo(
```
diegolmello, line 29: `room` is an object and `onChangeValue` is a function, so React.memo isn't really doing anything.

app/views/NotificationPreferencesView/index.tsx
```diff
@@ -1,282 +1,234 @@
...
-			});
+			// @ts-ignore
+			logEvent(events[`NP_${key.toUpperCase()}`]);
+			await Services.saveNotificationSettings(rid, params);
 		} catch (e) {
 			// @ts-ignore
 			logEvent(events[`NP_${key.toUpperCase()}_F`]);
 			log(e);
+			onError();
+			showErrorAlertWithEMessage(e);
```
diegolmello, line 113: This is confusing.
When is it calling `onError`?
Will it do both `onError` and `showErrorAlertWithEMessage`?

app/views/NotificationPreferencesView/index.tsx
```diff
@@ -1,282 +1,234 @@
...
+const RenderSwitch = ({ preference, room, onChangeValue }: IBaseParams) => {
+	const [switchValue, setSwitchValue] = useState(!room[preference]);
+	return (
+		<Switch
+			value={switchValue}
+			testID={preference}
+			trackColor={SWITCH_TRACK_COLOR}
+			onValueChange={value => {
+				onChangeValue(preference, { [preference]: switchValue ? '1' : '0' }, () => setSwitchValue(switchValue));
+				setSwitchValue(value);
```
diegolmello, line 83: Should we call `setSwitchValue` twice?

app/views/NotificationPreferencesView/index.tsx
```diff
@@ -1,282 +1,234 @@
...
-interface INotificationPreferencesViewProps {
-	navigation: StackNavigationProp<ChatsStackParamList, 'NotificationPrefView'>;
-	route: RouteProp<ChatsStackParamList, 'NotificationPrefView'>;
-	theme: TSupportedThemes;
-	serverVersion: string | null;
-}
-
-interface INotificationPreferencesViewState {
+interface IBaseParams {
+	preference: string;
```
diegolmello, line 24: I think we could actually type this.
This preference is an abstraction of both `OPTIONS` or `IRoomNotifications`.

For the first one, you could
- use `keyof typeof OPTIONS`
- remove [this nonsense](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/60f781d20e24ca92f9cd2c5dd9ef41de16fd0767/app/views/NotificationPreferencesView/options.ts#L7)
- remove the assertion `as IOptionsField` 
- Everything `RenderListPicker` is going to be inferred

Apply something like that to `IRoomNotifications` and we're good.

diegolmello (dismissed):

app/views/NotificationPreferencesView/index.tsx
```diff
@@ -1,282 +1,236 @@
...
+
+const RenderSwitch = ({ preference, room, onChangeValue }: IBaseParams) => {
+	const [switchValue, setSwitchValue] = useState(!room[preference]);
+	return (
+		<Switch
+			value={switchValue}
+			testID={preference as string}
+			trackColor={SWITCH_TRACK_COLOR}
+			onValueChange={value => {
+				onChangeValue(preference, { [preference]: switchValue ? '1' : '0' }, () => setSwitchValue(switchValue));
```
reinaldonetof, line 84: WDYT to use named params?

app/views/NotificationPreferencesView/index.tsx
```diff
@@ -1,282 +1,236 @@
...
+			title: I18n.t('Notification_Preferences')
+		});
+	}, []);
 
-	saveNotificationSettings = async (key: string, value: string | boolean, params: IRoomNotifications) => {
-		// @ts-ignore
-		logEvent(events[`NP_${key.toUpperCase()}`]);
-		const { room } = this.state;
-		const db = database.active;
+	const rid = route.params?.rid ?? '';
```
reinaldonetof, line 102: Is this really needed? Is there any possibility the rid be undefined?

reinaldonetof (changes requested):

app/views/NotificationPreferencesView/index.tsx
GleidsonDaniel: named params?

reinaldonetof (approved):

PR #4328, opened 2022-06-28
[IMPROVE] Redesign emoji picker

try-catch-stack:
## Proposed changes

* Migrated `EmojiPicker` to hooks.
* Revamped `EmojiPicker` component. Added functionality to `search` and `delete` emojis.

## How to test or reproduce

**Note:** We have a default emojis list on some places and they are composed of `['clap', 'thumbsup', 'heart_eyes', 'grinning', 'thinking', 'smiley']`
image
Once the user starts selecting emojis, the app starts to replace this default list with this `frequently used` emojis.

### Add emoji to message composer from emoji keyboard
- Tap on emoji button on left side of message composer to open emoji picker
  - select an emoji from custom emojis tab (rocket icon)
  - select an emoji from frequently used tab (clock icon) (it's based on emojis you previously selected)
  - select a normal emoji from any other tab

### Add emoji to message composer from emoji keyboard's search (clear cache before running this test)
- With emoji keyboard open, tap on search on bottom left
   - search should display default emojis at first. Select one of them and they should appear on the composer.
   - Close the search, open it again and the emoji you selected should appear first on the list (now as a frequently used)
   - Search for a custom emoji and select it
   - Search for a normal emoji and select it
   - Close the search, open it again and the list should be based on the emojis you selected

### Add emoji to message composer from autocomplete
- Focus on message composer and type `:` to open emoji autocomplete
- Select a normal and a custom one

### React a message from message actions (clear cache before this one)
- Long press on a message and it should show default emojis on the header of the bottom sheet
- Select one of them and the message should be reacted with it
- Open it again and the emoji should now appear first on the list

### React a message from reaction picker
- Long press on a message and tap on the `"smile emoji +"`  on the top right of the bottom sheet. It should open the reaction picker. Then select an emoji to react.
image
- On the right side of the reaction, the same icon above should appear to add another reaction. Tap on it to open the reaction picker.
- Same drill as the other tests. Try frequently used, custom emojis and the normal ones.
- Also try the search.

### Remove reaction
- Remove a reaction of a custom emoji
- Remove a reaction of a normal emoji

### Others
- Run these tests on tablets
- Check for any visual regression on avatars and markdowns

## Screenshots

|Before|After|
|-------|------|
|emoji-picker-before1|new emoji-picker-after1|
|emoji-picker-before2|emoji-picker-after2|
|reaction-picker-before|reaction-picker-after1|

* ## Searching emojis:
reaction-picker-after2 reaction-picker-search

https://user-images.githubusercontent.com/77742477/179019484-7d9cec0e-30ba-494b-9e94-78c846931d97.mp4

## Types of changes

## Checklist

## Further comments

[NATIVE-109]

[NATIVE-109]: https://rocketchat.atlassian.net/browse/NATIVE-109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -26,48 +26,39 @@ const renderEmoji = (emoji: IEmoji, size: number, baseUrl: string) => {
...
+		return null;
 	}
-}
+
+	return (
+		<FlatList
+			// rerender FlatList in case of width changes
+			key={`emoji-category-${width}`}
+			// @ts-ignore
+			keyExtractor={item => (item && item.isCustom && item.content) || item}
```
GleidsonDaniel, line 52:
```suggestion
			keyExtractor={item => item && item.isCustom ? item.content : item.name}
```

app/containers/EmojiPicker/TabBar.tsx
```diff
@@ -1,55 +1,36 @@
...
-		}
-		return false;
-	}
+	return (
+		<View style={styles.tabsContainer}>
+			{tabs?.map((tab, i) => (
+				<TouchableOpacity
+					activeOpacity={0.7}
+					key={tab}
+					onPress={() => {
```
GleidsonDaniel, line 17:
```suggestion
					onPress={() => goToPage?.(i)}
```

app/containers/EmojiPicker/EmojiCategory.tsx
try-catch-stack: This won't work for now as the non-custom emojis are plain strings.

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -26,48 +28,44 @@ const renderEmoji = (emoji: IEmoji, size: number, baseUrl: string) => {
 	);
 };
 
-class EmojiCategory extends React.Component<IEmojiCategory> {
-	renderItem(emoji: IEmoji) {
-		const { baseUrl, onEmojiSelected } = this.props;
-		return (
+const EmojiCategory = React.memo(
```
diegolmello, line 31: Same as the comment I left on the other PR.
`React.memo` is possibly making the performance worse here.
`onEmojiSelected` might change over re-renders and it's an inline function, which is going to passed as a forever changing prop.
If that's true, the component is always re-rendering anyway :)

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -26,48 +28,44 @@ const renderEmoji = (emoji: IEmoji, size: number, baseUrl: string) => {
...
+	({ baseUrl, onEmojiSelected, emojis, width, tabsCount, isBottomSheet, ...props }: IEmojiCategory) => {
+		const emojiSize = width ? Math.min(width / tabsCount, MAX_EMOJI_SIZE) : MAX_EMOJI_SIZE;
+		const numColumns = Math.trunc(width ? width / emojiSize : tabsCount);
+
+		const FlatList = isBottomSheet ? BottomSheetFlatList : GHFlatList;
+
+		const renderItem = (emoji: IEmoji) => (
 			<TouchableOpacity
```
diegolmello, line 39: Can you get rid of this `TouchableOpacity`?

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -26,48 +28,44 @@ const renderEmoji = (emoji: IEmoji, size: number, baseUrl: string) => {
...
 				activeOpacity={0.7}
 				// @ts-ignore
 				key={emoji && emoji.isCustom ? emoji.content : emoji}
 				onPress={() => onEmojiSelected(emoji)}
 				testID={`reaction-picker-${emoji && emoji.isCustom ? emoji.content : emoji}`}>
-				{renderEmoji(emoji, EMOJI_SIZE, baseUrl)}
+				{renderEmoji(emoji, emojiSize, baseUrl)}
```
diegolmello, line 45: We should turn this into a component.
The way you're using (`renderEmoji()`), is a function and not a component, so React is not aware of it, it doesn't know what to expect as the result and so it needs to re-evaluate it every time.
As a component (`<EmojiSomething />`), React will be able to reconcile it, making use of the props appropriately.

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -26,48 +28,44 @@ const renderEmoji = (emoji: IEmoji, size: number, baseUrl: string) => {
...
 			<FlatList
-				contentContainerStyle={{ marginHorizontal }}
 				// rerender FlatList in case of width changes
 				key={`emoji-category-${width}`}
 				// @ts-ignore
 				keyExtractor={item => (item && item.isCustom && item.content) || item}
 				data={emojis}
-				extraData={this.props}
-				renderItem={({ item }) => this.renderItem(item)}
+				extraData={{ baseUrl, onEmojiSelected, width, ...props }}
```
diegolmello, line 60: Can you explain?

app/containers/EmojiPicker/Footer.tsx
```diff
@@ -0,0 +1,27 @@
...
+import styles from './styles';
+import { IFooterProps } from './interfaces';
+
+const BUTTON_HIT_SLOP = { top: 15, right: 15, bottom: 15, left: 15 };
+
+const Footer = React.memo(({ onSearchPressed, onBackspacePressed }: IFooterProps) => {
+	const { colors } = useTheme();
+	return (
+		<View style={[styles.footerContainer, { backgroundColor: colors.bannerBackground }]}>
+			<BorderlessButton activeOpacity={0.7} onPress={onSearchPressed} style={styles.footerButtonsContainer}>
```
diegolmello, line 16: These opacities are probably not working as expected on darker themes.

app/containers/EmojiPicker/Footer.tsx
```diff
@@ -0,0 +1,27 @@
...
+				<CustomIcon color={colors.auxiliaryTintColor} size={25} name='search' />
```
diegolmello, line 17: It's 24 on Figma

app/containers/EmojiPicker/TabBar.tsx
```diff
@@ -1,55 +1,28 @@
...
+import { ITabBarProps } from './interfaces';
 
-interface ITabBarProps {
-	goToPage?: (page: number) => void;
-	activeTab?: number;
-	tabs?: string[];
-	tabEmojiStyle: StyleProp<TextStyle>;
-	theme: TSupportedThemes;
-}
+const TabBar = React.memo(({ activeTab, tabs, goToPage, tabEmojiStyle }: ITabBarProps) => {
```
diegolmello, line 8: Same as the other memos

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,204 +1,187 @@
...
+		searching = false,
+		searchedEmojis = []
+	}: IEmojiPickerProps) => {
+		const [frequentlyUsed, setFrequentlyUsed] = useState<IEmoji[]>([]);
+		const [show, setShow] = useState(false);
+		const [width, setWidth] = useState(null);
+		const { colors } = useTheme();
+
+		const allCustomEmojis: ICustomEmojis = useAppSelector(state => state.customEmojis);
+		const customEmojis = Object.keys(allCustomEmojis)
```
diegolmello, line 37: While as a Class Component, `constructor()` is triggered only once before the component actually mounting, as a Function Component everything is triggered on render all the time.
We should evaluate this and maybe turn it either into a custom hook or make use of `useMemo`

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,204 +1,187 @@
...
-				this._addFrequentlyUsed({ content, isCustom: false });
-				const shortname = `:${emoji}:`;
-				onEmojiSelected(shortnameToUnicode(shortname), shortname);
+		useEffect(() => {
+			const init = async () => {
+				await updateFrequentlyUsed();
+				setShow(true);
+			};
+			init();
+		}, []);
```
diegolmello, line 51: Can you turn this into a custom hook?
You would extract most of this logic away from the component.

app/containers/EmojiPicker/styles.ts
```diff
@@ -54,5 +54,18 @@ export default StyleSheet.create({
 	},
 	customCategoryEmoji: {
 		margin: 8
+	},
+	footerContainer: {
+		height: 45,
```
diegolmello, line 59: It's 44 on Figma

app/containers/MessageActions/index.tsx
```diff
@@ -264,7 +264,8 @@ const MessageActions = React.memo(
 					// TODO: evaluate unification with IEmoji
 					onReactionPress(shortname as any, message.id);
 				} else {
-					reactionInit(message);
+					// Wait for the Action Sheet to close before opening reaction picker
+					setTimeout(() => reactionInit(message), 500);
```
diegolmello, line 268: Don't we have a smarter way of doing it?

app/views/RoomView/index.tsx
```diff
@@ -821,12 +824,32 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 		this.setState({ selectedMessage: undefined, replying: false, replyWithMention: false });
 	};
 
+	showReactionPicker = () => {
+		const { showActionSheet, width, height } = this.props;
+		const { reacting, selectedMessage } = this.state;
+		showActionSheet &&
```
diegolmello, line 830: Change it to a normal if, please.

diegolmello (changes requested):

diegolmello (commented):
- [ ] Review your React.memo usage
- [ ] Make sure Detox tests are passing on Android

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -24,50 +33,46 @@ const renderEmoji = (emoji: IEmoji, size: number, baseUrl: string) => {
...
+		return null;
 	}
-}
+
+	return (
+		<FlatList
+			// rerender FlatList in case of width changes
+			key={`emoji-category-${width}`}
+			// @ts-ignore
+			keyExtractor={item => (item && item.isCustom && item.content) || item}
```
GleidsonDaniel, line 66:
```suggestion
			keyExtractor={item => item.name}
```
This here doesn't make sense, given that supposedly there will always be an emoji here and this emoji always having a name is easier like this:

app/containers/EmojiPicker/categories.ts
```diff
@@ -1,43 +1,43 @@
 const list = ['frequentlyUsed', 'custom', 'people', 'nature', 'food', 'activity', 'travel', 'objects', 'symbols', 'flags'];
```
GleidsonDaniel, line 1:
```suggestion
const list = ['clock', 'rocket', 'emoji' ...];
```
Isn't it easier to correct the names here and make a map from this list?

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,106 +1,92 @@
...
+		[allCustomEmojis]
+	);
 
-	onEmojiSelected = (emoji: IEmoji) => {
+	const handleEmojiSelect = (emoji: IEmoji) => {
 		try {
-			const { onEmojiSelected } = this.props;
 			if (emoji.isCustom) {
-				this._addFrequentlyUsed({
+				_addFrequentlyUsed({
```
GleidsonDaniel, line 72: This method is being called before being initialized.
Not really a problem, but the default is set first then use.

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,106 +1,92 @@
...
 					content: emoji.content,
 					extension: emoji.extension,
 					isCustom: true
 				});
-				onEmojiSelected(`:${emoji.content}:`);
+				onItemClicked(EventTypes.EMOJI_PRESSED, `:${emoji.content}:`);
 			} else {
 				const content = emoji;
-				this._addFrequentlyUsed({ content, isCustom: false });
+				_addFrequentlyUsed({ content, isCustom: false });
```
GleidsonDaniel, line 80:
```suggestion
				_addFrequentlyUsed({ content: emoji, isCustom: false });
```
remove the upper line

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+		return <CustomEmoji style={{ height: EMOJI_SIZE, width: EMOJI_SIZE, margin: 4 }} emoji={emoji} baseUrl={baseUrl} />;
+	}
+	return (
+		<Text style={[styles.searchedEmoji, { fontSize: EMOJI_SIZE, color: colors.backdropColor }]}>
+			{shortnameToUnicode(`:${emoji}:`)}
+		</Text>
+	);
+});
+
+const EmojiSearchbar = React.forwardRef<TextInput, IEmojiSearchbarProps>(
```
GleidsonDaniel, line 42:
```suggestion
const EmojiSearchBar = React.forwardRef<TextInput, IEmojiSearchBarProps>(
```

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,106 +1,92 @@
...
-			return true;
-		}
-		return false;
-	}
+import { useTheme } from '../../theme';
+import { IEmoji, ICustomEmojis, TFrequentlyUsedEmojiModel } from '../../definitions';
+import { useAppSelector } from '../../lib/hooks';
+import { IEmojiPickerProps, EventTypes } from './interfaces';
+
+export const useFrequentlyUsedEmoji = () => {
```
GleidsonDaniel, line 22: This function is being used in more than one place, it is better to hook it in a separate file.

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+	({ openEmoji, onChangeText, emojis, onEmojiSelected, baseUrl }, ref) => {
+		const { colors } = useTheme();
+		const [searchText, setSearchText] = useState<string>('');
+		const [frequentlyUsedEmojis, setFrequentlyUsed] = useState<(string | ICustomEmoji)[]>();
+		const { frequentlyUsed, loaded } = useFrequentlyUsedEmoji();
+
+		useEffect(() => {
```
GleidsonDaniel, line 49: Wouldn't it be better to put this logic inside the emoji hook?

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,120 @@
...
+				setFrequentlyUsed(frequentlyUsedWithDefaultEmojis);
+			}
+		}, [loaded]);
+
+		const handleTextChange = (text: string) => {
+			setSearchText(text);
+			onChangeText(text);
+		};
+
+		const renderItem = (emoji: IEmoji) => (
```
GleidsonDaniel, line 66: This component was better to use memo or create it out of this component, because every re-render it will be created again.

app/views/RoomView/ReactionPicker.tsx
```diff
@@ -1,85 +1,99 @@
...
 }
 
-class ReactionPicker extends React.Component<IReactionPickerProps> {
-	shouldComponentUpdate(nextProps: IReactionPickerProps) {
-		const { show, width, height } = this.props;
-		return nextProps.show !== show || width !== nextProps.width || height !== nextProps.height;
-	}
+const MAX_EMOJIS_TO_DISPLAY = 20;
+
+const ReactionPicker = React.memo(({ onEmojiSelected, message, reactionClose }: IReactionPickerProps) => {
```
GleidsonDaniel, line 29: This react.memo will not work properly, even more so since message is a giant object.

GleidsonDaniel (changes requested):
errors:
When you send an emoji the keyboard does not close automatically.
When clicking directly on the emoji without clicking on the keyboard, when I switch to keyboard it doesn't work, the keyboard is black.

app/containers/MessageBox/EmojiSearchbar.tsx
try-catch-stack: I thought of doing so but this logic doesn't have any relation with frequently used emojis, also we don't want this behavior at other place so maybe it would not make sense to merge the login IMO.

app/containers/EmojiPicker/EmojiCategory.tsx
try-catch-stack: Actually item here is not an object all the time. For non-custom emojis, item is a string and hence this logic.

app/containers/EmojiPicker/categories.ts
try-catch-stack: Using category name ('custom', 'people' ...) instead of icon name ('rocker', 'emoji' ...) makes more sense IMO. Also we'll have to change the keys in `./emoji.ts` to these icon names.

try-catch-stack:
> errors: When you send an emoji the keyboard does not close automatically. When clicking directly on the emoji without clicking on the keyboard, when I switch to keyboard it doesn't work, the keyboard is black.

I believe, this issue should now be fixed.

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,106 +1,50 @@
...
+	searching = false,
+	searchedEmojis = []
+}: IEmojiPickerProps): React.ReactElement | null => {
+	const [width, setWidth] = useState(null);
+	const { colors } = useTheme();
+	const { frequentlyUsed, loaded } = useFrequentlyUsedEmoji();
+
+	const baseUrl = useAppSelector(state => state.server?.server);
+	const allCustomEmojis: ICustomEmojis = useAppSelector(state => state.customEmojis);
+	const customEmojis = useMemo(
```
diegolmello, line 35: It won't work, because `allCustomEmojis` is the dependency, but it's a non-primitive value.
It can't be compared.

This might fix what you expect
```tsx
const allCustomEmojis: ICustomEmojis = useAppSelector(state => state.customEmojis, shallowEqual);
```

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,106 +1,50 @@
...
-		} catch (e) {
-			log(e);
-		}
-	};
+				})),
+		[allCustomEmojis]
+	);
 
-	_addFrequentlyUsed = protectedFunction(async (emoji: IEmoji) => {
+	const _addFrequentlyUsed = protectedFunction(async (emoji: IEmoji) => {
```
diegolmello, line 47:
```suggestion
	const addFrequentlyUsed = async (emoji: IEmoji) => {
```

Can you remove the `_` and the `protectedFunction`?

app/containers/EmojiPicker/index.tsx
```diff
@@ -160,47 +107,59 @@ class EmojiPicker extends Component<IEmojiPickerProps, IEmojiPickerState> {
 		}
 		return (
 			<EmojiCategory
-				emojis={emojis}
-				onEmojiSelected={(emoji: IEmoji) => this.onEmojiSelected(emoji)}
+				emojis={emojis as IEmoji[]}
```
diegolmello, line 110: Can you type this properly?

app/containers/MessageBox/EmojiKeyboard.tsx
```diff
@@ -6,21 +6,23 @@ import { Provider } from 'react-redux';
...
 
-const EmojiKeyboard = ({ theme }: { theme: TSupportedThemes }) => {
-	const onEmojiSelected = (emoji: string) => {
-		KeyboardRegistry.onItemSelected('EmojiKeyboard', { emoji });
+const EmojiKeyboard = () => {
+	const onItemClicked = (eventType: EventTypes, emoji: string | undefined) => {
+		KeyboardRegistry.onItemSelected('EmojiKeyboard', { eventType, emoji });
 	};
 
+	const {colors} = useTheme()
```
diegolmello, line 17: Hooks always on top https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

Also, you're not using Prettier yet.
Please do it.

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { CustomIcon } from '../CustomIcon';
+import { IEmoji } from '../../definitions';
+import shortnameToUnicode from '../../lib/methods/helpers/shortnameToUnicode';
+import CustomEmoji from '../EmojiPicker/CustomEmoji';
+import styles from './styles';
+import useFrequentlyUsedEmoji from '../EmojiPicker/frequentlyUsedEmojis';
+
+const BUTTON_HIT_SLOP = { top: 15, right: 15, bottom: 15, left: 15 };
+const EMOJI_SIZE = 30;
+const DEFAULT_EMOJIS = ['clap', '+1', 'heart_eyes', 'grinning', 'thinking_face', 'smiley'];
```
diegolmello, line 16: Can you reuse this from MessageActions?
It could be a constant outside of both components.

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+import I18n from '../../i18n';
+import { CustomIcon } from '../CustomIcon';
+import { IEmoji } from '../../definitions';
+import shortnameToUnicode from '../../lib/methods/helpers/shortnameToUnicode';
+import CustomEmoji from '../EmojiPicker/CustomEmoji';
+import styles from './styles';
+import useFrequentlyUsedEmoji from '../EmojiPicker/frequentlyUsedEmojis';
+
+const BUTTON_HIT_SLOP = { top: 15, right: 15, bottom: 15, left: 15 };
+const EMOJI_SIZE = 30;
```
diegolmello, line 15: 40

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+import { useTheme } from '../../theme';
+import I18n from '../../i18n';
+import { CustomIcon } from '../CustomIcon';
+import { IEmoji } from '../../definitions';
+import shortnameToUnicode from '../../lib/methods/helpers/shortnameToUnicode';
+import CustomEmoji from '../EmojiPicker/CustomEmoji';
+import styles from './styles';
+import useFrequentlyUsedEmoji from '../EmojiPicker/frequentlyUsedEmojis';
+
+const BUTTON_HIT_SLOP = { top: 15, right: 15, bottom: 15, left: 15 };
```
diegolmello, line 14: 4

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,122 @@
...
+				setFrequentlyUsed(frequentlyUsedWithDefaultEmojis);
+			}
+		}, [loaded]);
+
+		const handleTextChange = (text: string) => {
+			setSearchText(text);
+			onChangeText(text);
+		};
+
+		const ListItem = React.memo(({ emoji }: { emoji: IEmoji }) => (
```
diegolmello, line 66: Avoid nested components

app/containers/MessageBox/EmojiSearchbar.tsx
diegolmello: You're on the correct path, but still not quite right.
Your custom hooks is already returning the data you need.
You don't need an extra `useEffect` with `loaded` as dependency.

Just use `loaded` and `frequentlyUsed` to render what you need.

app/containers/MessageBox/index.tsx
```diff
@@ -555,18 +577,43 @@ class MessageBox extends Component<IMessageBoxProps, IMessageBoxState> {
...
+			case EventTypes.EMOJI_PRESSED:
+				newText = `${text.substr(0, cursor)}${emoji}${text.substr(cursor)}`;
+				newCursor = cursor + emoji.length;
+				this.setInput(newText, { start: newCursor, end: newCursor });
+				this.setShowSend(true);
+				break;
+			case EventTypes.SEARCH_PRESSED:
+				this.setState({ showEmojiKeyboard: false, showEmojiSearchbar: true });
+				setTimeout(() => {
+					this.emojiSearchbarRef.current?.focus();
```
diegolmello, line 611: https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs

diegolmello (changes requested):
Missing detox tests

app/containers/MessageBox/EmojiSearchbar.tsx
diegolmello: Are you going to fix this?

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,128 @@
...
+		}, [loaded]);
+
+		const handleTextChange = (text: string) => {
+			setSearchText(text);
+			onChangeText(text);
+		};
+
+		return (
+			<View style={{ borderTopWidth: 1, borderTopColor: colors.borderColor, backgroundColor: colors.backgroundColor }}>
+				<FlatList
```
diegolmello, line 80: Using emojis from this list isn't updating frequently used

app/views/RoomView/ReactionPicker.tsx
```diff
@@ -1,85 +1,98 @@
...
 }
 
-class ReactionPicker extends React.Component<IReactionPickerProps> {
-	shouldComponentUpdate(nextProps: IReactionPickerProps) {
-		const { show, width, height } = this.props;
-		return nextProps.show !== show || width !== nextProps.width || height !== nextProps.height;
-	}
+const MAX_EMOJIS_TO_DISPLAY = 20;
+
+const ReactionPicker = ({ onEmojiSelected, message, reactionClose }: IReactionPickerProps): React.ReactElement => {
```
diegolmello, line 28: image

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}, [loaded]);
+
+		const handleTextChange = (text: string) => {
+			setSearchText(text);
+			onChangeText(text);
+		};
+
+		return (
+			<View style={{ borderTopWidth: 1, borderTopColor: colors.borderColor, backgroundColor: colors.backgroundColor }}>
```
diegolmello, line 79: Remove all static inline styles like this one

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -1,75 +1,74 @@
...
-	render() {
-		const { emojis, width } = this.props;
+});
 
-		if (!width) {
-			return null;
-		}
+const EmojiCategory = ({ baseUrl, onEmojiSelected, emojis, width, tabsCount }: IEmojiCategory): React.ReactElement | null => {
+	const emojiSize = width ? Math.min(width / tabsCount, MAX_EMOJI_SIZE) : MAX_EMOJI_SIZE;
+	const numColumns = Math.trunc(width ? width / emojiSize : tabsCount);
```
diegolmello, line 35: Emojis are very small this way.
We should aim for sizes closer to 44x44.
Usually, that is going to render 7~8 emojis per row.
Just pay attention to how landscape and tablets behave.

diegolmello (changes requested):

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -1,75 +1,74 @@
...
-		);
+	if (!width) {
+		return null;
 	}
-}
+
+	return (
+		<FlatList
+			// rerender FlatList in case of width changes
+			key={`emoji-category-${width}`}
```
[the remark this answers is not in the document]
try-catch-stack: Tried this but it was still complaining about `cannot change the number of columns on the fly` so I kept it this way.

app/views/RoomView/ReactionPicker.tsx
```diff
@@ -1,85 +1,98 @@
...
+	const [searchedEmojis, setSearchedEmojis] = React.useState<(string | IEmoji)[]>([]);
+	const [searching, setSearching] = React.useState<boolean>(false);
+
+	const handleTextChange = (text: string) => {
+		setSearching(text !== '');
+		setSearchText(text);
+		searchEmojis(text);
+	};
+
+	const searchEmojis = debounce(async (keyword: string) => {
```
diegolmello, line 40: You should never use debounce like this on a function component.
It won't work, because they're going to get recreated on every render.

app/views/RoomView/styles.ts
```diff
@@ -64,5 +62,14 @@ export default StyleSheet.create({
 	previewMode: {
 		fontSize: 16,
 		...sharedStyles.textMedium
+	},
+	searchbarContainer: {
```
diegolmello, line 66: Not multiple of 4

app/containers/EmojiPicker/index.tsx
try-catch-stack: I did add the `shallowEqual` as you suggested. Did I miss anything?

app/containers/MessageBox/EmojiSearchbar.tsx
try-catch-stack: Fixed!

app/containers/MessageBox/EmojiSearchbar.tsx
try-catch-stack: 40 seemed too large on android. I attached a screenshot but probably missed adding the comment. 
Screenshot 2022-09-01 at 10 46 27 PM

I'll do it if you think it looks fine.

app/containers/MessageBox/EmojiSearchbar.tsx
```diff
@@ -0,0 +1,143 @@
...
+					keyboardShouldPersistTaps='always'
+				/>
+				<View style={styles.emojiSearchbarContainer}>
+					<Pressable
+						style={({ pressed }: { pressed: boolean }) => [styles.openEmojiKeyboard, { opacity: pressed ? 0.7 : 1 }]}
+						onPress={openEmoji}
+						hitSlop={BUTTON_HIT_SLOP}
+						testID='openback-emoji-keyboard'
+					>
+						<CustomIcon name='chevron-left' size={30} color={colors.collapsibleChevron} />
```
diegolmello, line 117: 24

app/containers/MessageBox/EmojiSearchbar.tsx
diegolmello: It's 36 on Figma

try-catch-stack: 32 looks good to me.

app/containers/EmojiPicker/index.tsx
```diff
@@ -1,155 +1,68 @@
...
+		[allCustomEmojis]
+	);
 
-	onEmojiSelected = (emoji: IEmoji) => {
+	const handleEmojiSelect = (emoji: IEmoji | string) => {
 		try {
-			const { onEmojiSelected } = this.props;
-			if (emoji.isCustom) {
-				this._addFrequentlyUsed({
+			if (typeof emoji === 'string') {
```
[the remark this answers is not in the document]
try-catch-stack: Now that `IEmoji` is either a string or an object, this logic should make more sense.

__tests__/containers/Avatar/__snapshots__/Avatar.stories.storyshot
```diff
@@ -18,7 +18,7 @@ exports[`Storyshots Avatar Custom Style 1`] = `"{\\"type\\":\\"View\\",\\"props\
...
+exports[`Storyshots Avatar Emoji 1`] = `"{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":[{\\"width\\":56,\\"height\\":56,\\"borderRadius\\":4},null],\\"testID\\":\\"avatar\\"},\\"children\\":[{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":[{\\"overflow\\":\\"hidden\\"},[{\\"width\\":30,\\"height\\":30},{\\"width\\":56,\\"height\\":56,\\"borderRadius\\":4}]]},\\"children\\":[{\\"type\\":\\"FastImageView\\",\\"props\\":{\\"style\\":{\\"position\\":\\"absolute\\",\\"left\\":0,\\"right\\":0,\\"top\\":0,\\"bottom\\":0},\\"source\\":{\\"uri\\":\\"/emoji-custom/troll.jpg\\",\\"priority\\":\\"high\\"},\\"resizeMode\\":\\"contain\\"},\\"children\\":null}]}]}"`;
```
diegolmello, line 21: I didn't get why server URL isn't on the snapshots anymore, but it's working on Storybook.

__tests__/containers/RoomItem/__snapshots__/RoomItem.stories.storyshot
```diff
@@ -1,25 +1,25 @@
...
... unread priority 3\\"]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":{\\"alignItems\\":\\"flex-end\\"}},\\"children\\":[{\\"type\\":\\"Text\\",\\"props\\":{\\"style\\":[{\\"fontSize\\":13,\\"marginLeft\\":4,\\"textAlign\\":\\"left\\",\\"backgroundColor\\":\\"transparent\\",\\"fontFamily\\":\\"Inter\\",\\"fontWeight\\":\\"400\\"},{\\"color\\":\\"#9ca2a8\\"},[{\\"textAlign\\":\\"left\\",\\"backgroundColor\\":\\"transparent\\",\\"fontFamily\\":\\"Inter\\",\\"fontWeight\\":\\"600\\"},{\\"color\\":\\"#1d74f5\\"}]],\\"ellipsizeMode\\":\\"tail\\",\\"numberOfLines\\":1},\\"children\\":[\\"10:00\\"]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":[{\\"height\\":21,\\"paddingVertical\\":3,\\"paddingHorizontal\\":5,\\"borderRadius\\":10.5,\\"alignItems\\":\\"center\\",\\"justifyContent\\":\\"center\\",\\"marginLeft\\":10},{\\"backgroundColor\\":\\"#1d74f5\\",\\"minWidth\\":21},null]},\\"children\\":[{\\"type\\":\\"Text\\",\\"props\\":{\\"style\\":[{\\"fontSize\\":13,\\"textAlign\\":\\"left\\",\\"backgroundColor\\":\\"transparent\\",\\"fontFamily\\":\\"Inter\\",\\"fontWeight\\":\\"600\\"},null,{\\"color\\":\\"#ffffff\\"}],\\"numberOfLines\\":1},\\"children\\":[\\"1\\"]}]}]}]}]}]}]}]}]}]"`;
```
diegolmello, line 3: Somehow FastImage started to appear on these snapshots

__tests__/containers/Chip/__snapshots__/Chip.stories.storyshot
```diff
@@ -1,11 +1,11 @@
...
...\\",\\"props\\":{\\"style\\":[{\\"overflow\\":\\"hidden\\"},{\\"width\\":28,\\"height\\":28,\\"borderRadius\\":4}]},\\"children\\":[{\\"type\\":\\"FastImageView\\",\\"props\\":{\\"style\\":{\\"position\\":\\"absolute\\",\\"left\\":0,\\"right\\":0,\\"top\\":0,\\"bottom\\":0},\\"source\\":{\\"uri\\":\\"https://open.rocket.chat/avatar/rocket.cat?format=png&size=56\\",\\"headers\\":{\\"User-Agent\\":\\"RC Mobile; ios unknown; vunknown (unknown)\\"},\\"priority\\":\\"high\\"},\\"resizeMode\\":\\"cover\\"},\\"children\\":null}]}]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":{\\"marginRight\\":8,\\"maxWidth\\":120}},\\"children\\":[{\\"type\\":\\"Text\\",\\"props\\":{\\"style\\":[{\\"fontSize\\":16,\\"textAlign\\":\\"left\\",\\"backgroundColor\\":\\"transparent\\",\\"fontFamily\\":\\"Inter\\",\\"fontWeight\\":\\"500\\"},{\\"color\\":\\"#2f343d\\"}],\\"numberOfLines\\":1},\\"children\\":[\\"Rocket.Cat\\"]}]},{\\"type\\":\\"Text\\",\\"props\\":{\\"selectable\\":false,\\"allowFontScaling\\":false,\\"style\\":[{\\"fontSize\\":16,\\"color\\":\\"#6C727A\\"},null,{\\"fontFamily\\":\\"custom\\",\\"fontWeight\\":\\"normal\\",\\"fontStyle\\":\\"normal\\"},{}]},\\"children\\":[\\"\\"]}]}]}]}"`;
```
diegolmello, line 3: In most of these snapshots, the change is FastImage wasn't rendered properly inside of Storyshots

__tests__/containers/ReactionsList/__snapshots__/ReactionsList.stories.storyshot
```diff
@@ -1,5 +1,5 @@
...
...\"children\\":null}]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":{\\"width\\":750}},\\"children\\":[{\\"type\\":\\"View\\",\\"props\\":{\\"tabLabel\\":{\\"emoji\\":\\":react_rocket:\\",\\"_id\\":\\"react_rocket\\",\\"usernames\\":[\\"rocket.cat\\",\\"diego.mello\\"],\\"names\\":[\\"Rocket Cat\\",\\"Diego Mello\\"]}},\\"children\\":null}]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":{\\"width\\":750}},\\"children\\":[{\\"type\\":\\"View\\",\\"props\\":{\\"tabLabel\\":{\\"emoji\\":\\":nyan_rocket:\\",\\"_id\\":\\"nyan_rocket\\",\\"usernames\\":[\\"rocket.cat\\"],\\"names\\":[\\"Rocket Cat\\"]}},\\"children\\":null}]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":{\\"width\\":750}},\\"children\\":[{\\"type\\":\\"View\\",\\"props\\":{\\"tabLabel\\":{\\"emoji\\":\\":grinning:\\",\\"_id\\":\\"grinning\\",\\"usernames\\":[\\"diego.mello\\"],\\"names\\":[\\"Diego Mello\\"]}},\\"children\\":null}]},{\\"type\\":\\"View\\",\\"props\\":{\\"style\\":{\\"width\\":750}},\\"children\\":[{\\"type\\":\\"View\\",\\"props\\":{\\"tabLabel\\":{\\"emoji\\":\\":tada:\\",\\"_id\\":\\"tada\\",\\"usernames\\":[\\"diego.mello\\"],\\"names\\":[\\"Diego Mello\\"]}},\\"children\\":null}]}]}]}]}]}]}"`;
```
diegolmello, line 3: On others, the FastImage is rendered but without server

app/containers/EmojiPicker/styles.ts
```diff
@@ -2,20 +2,23 @@ import { StyleSheet } from 'react-native';
...
-		paddingTop: 5
+		flexDirection: 'row'
 	},
 	tab: {
 		flex: 1,
 		alignItems: 'center',
 		justifyContent: 'center',
-		paddingBottom: 10
+		paddingVertical: 10,
+		width: 44
```
diegolmello, line 21: EMOJI_BUTTON_SIZE

app/containers/EmojiPicker/styles.ts
```diff
@@ -49,10 +51,34 @@ export default StyleSheet.create({
...
+		width: EMOJI_SIZE
+	},
+	emojiButton: {
+		alignItems: 'center',
+		justifyContent: 'center',
+		height: EMOJI_BUTTON_SIZE,
+		width: EMOJI_BUTTON_SIZE
+	},
+	footerContainer: {
+		height: 44,
```
diegolmello, line 70: EMOJI_BUTTON_SIZE

app/containers/EmojiPicker/styles.ts
```diff
@@ -49,10 +51,34 @@ export default StyleSheet.create({
...
+		paddingHorizontal: 12,
+		flexDirection: 'row',
+		justifyContent: 'space-between',
+		alignItems: 'center',
+		borderTopWidth: 1
+	},
+	footerButtonsContainer: {
+		height: 44,
+		width: 44,
```
diegolmello, line 79: EMOJI_BUTTON_SIZE

app/i18n/locales/en.json
```diff
@@ -478,6 +478,7 @@
   "Search_Messages": "Search Messages",
   "Search": "Search",
   "Search_by": "Search by",
+  "Search_emoji": "Search emoji",
```
diegolmello, line 481: Missing ptbr

app/views/RoomView/index.tsx
```diff
@@ -1026,7 +1044,7 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 		});
 	};
 
-	getCustomEmoji = (name: string): Pick<IEmoji, 'name' | 'extension'> | null => {
+	getCustomEmoji = (name: string): Pick<ICustomEmoji, 'name' | 'extension'> | null => {
```
diegolmello, line 1047: TGetCustomEmoji

diegolmello (approved):

app/containers/markdown/new/Emoji.tsx
```diff
@@ -16,7 +16,7 @@ interface IEmojiProps {
 
 const Emoji = ({ block, isBigEmoji }: IEmojiProps) => {
 	const { theme } = useTheme();
```
GleidsonDaniel, line 18: Since we're working on the file, I think it's worth removing the theme in the places it's being used.

GleidsonDaniel (changes requested):

GleidsonDaniel:
Screen Shot 2022-10-21 at 10 34 31
When I didn't send any emoji the most used emojis tab is showing empty (It shouldn't appear I think)

GleidsonDaniel (approved):

PR #4291, opened 2022-06-08
[FIX] Share Extension shows a server that has no user logged in

reinaldonetof:
## How to test or reproduce

1 - Start the process of logging in to a server by choosing one, but do not log in to any user account.
       a - If this step is not done correctly, at step 4, the user will probably see the screen stating there is no server logged in.
2 - Minimize app
3 - Open Share Extension (by sharing something from the device's gallery)
4 - Try choosing a server to share the file to.

#### Before
https://user-images.githubusercontent.com/47038980/172525539-58c63246-0fb5-488f-a674-0edfaf84ad0e.mov

#### After
https://user-images.githubusercontent.com/47038980/172525273-174297ca-1746-41a1-861d-7d150e46faa3.mov

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4336, opened 2022-07-04
[FIX] Share Extension shows a server that has no user logged in

GleidsonDaniel:
## Proposed changes

Changes the logic to only set the chosen server if the login is successful.
Since it is not necessary to set the server and I have not even completed the login.

Note: If it is necessary to keep the server even not logging into it, the right thing would be to create a key, something like a temp key, different from the main key that indicates the logged server.

Closes #4291
I left the old PR just for history and in case you want to compare.

## How to test or reproduce

1 - Start the process of logging in to a server by choosing one, but do not log in to any user account.
a - If this step is not done correctly, at step 4, the user will probably see the screen stating there is no server logged in.
2 - Minimize app
3 - Open Share Extension (by sharing something from the device's gallery)
4 - Try choosing a server to share the file to.

## Types of changes

## Checklist

### DIFF
--- a/app/sagas/login.js
+++ b/app/sagas/login.js
@@ -20,7 +20,7 @@ import { inquiryRequest, inquiryReset } from '../ee/omnichannel/actions/inquiry'
 import { isOmnichannelStatusAvailable } from '../ee/omnichannel/lib';
 import { RootEnum } from '../definitions';
 import sdk from '../lib/services/sdk';
-import { TOKEN_KEY } from '../lib/constants';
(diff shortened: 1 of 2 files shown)


diegolmello (changes requested):
I think the solution is a lot clever than the other one, but it's not free of bugs :)

- Open the app
- Login on `open`
- Close the app and reopen
- You're now logged out
- Only if you tap `join community`, it's going to trigger `handleSelectServer` with a user.

Maybe we should add a logic like that to `loginSuccess`?

PR #4342, opened 2022-07-05
[NEW] Create useEndpointData for call endpoints using hooks

GleidsonDaniel:
## Proposed changes

Create useEndpointData for call endpoints using hooks

## Types of changes

## Checklist


app/lib/hooks/useEndpointData.ts
```diff
@@ -0,0 +1,59 @@
...
+				} else {
+					// handle error
+				}
+			})
+			.catch((e: TError) => {
+				// handle error
+				setLoading(false);
+				showErrorAlert(e.error);
+			});
+	}, [params]);
```
diegolmello, line 48: If `params` is an object, it'll trigger an infinite loop.

app/lib/hooks/useEndpointData.ts
```diff
@@ -0,0 +1,66 @@
...
+				if (e.success) {
+					setResult(e);
+				} else {
+					// handle error
+				}
+			})
+			.catch((e: TError) => {
+				// handle error
+				setLoading(false);
+				showErrorAlert(e.error);
```
diegolmello, line 53: I think we should return the error as data to the component calling it, so we call handle it there.
Calling the alert here could result in unexpected behaviors, like multiple components failing and spamming alerts at the same time.

app/lib/hooks/useEndpointData.ts
```diff
@@ -0,0 +1,66 @@
...
+		if (!endpoint) return;
+		setLoading(true);
+		sdk
+			.get(endpoint, params)
+			.then(e => {
+				setLoading(false);
+				if (e.success) {
+					setResult(e);
+				} else {
+					// handle error
```
diegolmello, line 47: This sucks, but we should handle it similarly to `catch`, otherwise it's going to just stop loading without any further info.

app/lib/hooks/useEndpointData.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+		);
+	}
+	return (
+		<View>
+			<Text testID='load complete'>load complete</Text>
+		</View>
+	);
+}
+
+describe('useFetch', () => {
```
diegolmello, line 46: Given the changes requested about error, we should test them here as well

diegolmello (changes requested):

diegolmello (approved):

PR #4364, opened 2022-07-14
[FIX] Add support to non ascii url characters on ios devices

GleidsonDaniel:
## Proposed changes

Add support to non ascii url characters on ios devices

## How to test or reproduce

Try opening a url that doesn't contain ACSII characters like: 
```https://чат24.рф```
## Screenshots
Before:

https://user-images.githubusercontent.com/18687097/179036118-cc9d9a56-e4d7-41f8-80af-21fd4724575c.mp4

After:

https://user-images.githubusercontent.com/18687097/179036126-037e3565-b430-4616-bd69-080fb45f4f80.mp4

## Types of changes

## Checklist

## Further comments

Note: On android this is done automatically.

### CODE REVIEW

File: app/lib/methods/urlParse.ts [hunk 1]
```diff
@@ -0,0 +1,3 @@
+import * as uri from 'uri-js';
+
+export const urlParse = (url: string): string => uri.serialize(uri.parse(url));
```
diegolmello, line 3: I'm not sure about this.
We already have `parseUrls` https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/lib/methods/helpers/parseUrls.ts
Having `urlParse` is going to create confusion for sure.

Also I'm not sure how useful it is to export this code
```jsx
uri.serialize(uri.parse(url))
```

instead of the actual reusable thing

```jsx
let newUrl = url.replace(/\/+$/, '').replace(/\\/g, '/');
const ascii = /^[ -~\t\n\r]+$/;
if (isIOS && !ascii.test(newUrl)) {
	newUrl = urlParse(newUrl);
}
```

Maybe we could increase the scope and unify some URL logic, as we also have `parseQuery`?
Ping me in chat :)

package.json
```diff
@@ -132,6 +132,7 @@
 		"rn-root-view": "1.0.3",
 		"semver": "7.3.5",
 		"ua-parser-js": "^1.0.2",
+		"uri-js": "^4.4.1",
```
diegolmello, line 135: We should consider replacing this lib with the other ones, if this one is better.

diegolmello (changes requested):

app/lib/methods/serializeAsciiUrl.test.ts
```diff
@@ -0,0 +1,30 @@
+import { serializeAsciiUrl } from '.';
+
+const ASCIIUrl = 'https://чат24.рф';
+const NonASCIIUrl = 'open.rocket.chat';
+const ASCIIUrlSerialized = 'https://xn--24-6kc6exa.xn--p1ai/';
```
reinaldonetof, line 5: I'm wondering about the last slash. I guess that we need to implement a regex to remove the last slash, because we are expecting that the url doesn't have that slash.

reinaldonetof (changes requested):

app/lib/methods/serializeAsciiUrl.test.ts
GleidsonDaniel: Why should we remove it?

PR #4381, opened 2022-07-21
[IMPROVE] Redesign flow of create room

reinaldonetof:
## Proposed changes

We're proposing lots of UX improvements on this task.
We're going to make use of the new Searchbox, the new forms pattern and improve the UX overall.

Screen Shot 2022-07-21 at 15 14 42

#### Before

https://user-images.githubusercontent.com/47038980/180284356-2c46517f-9666-4bbb-a611-9ef8011537e6.mov

#### After

https://user-images.githubusercontent.com/47038980/180282740-57f948ee-5fec-4a89-9321-95280a0866af.mov

## Types of changes

## Checklist


app/views/CreateChannelView.tsx
```diff
@@ -32,38 +31,41 @@ const styles = StyleSheet.create({
...
 	},
-	input: {
-		height: 54,
-		paddingHorizontal: 18,
-		fontSize: 17,
-		...sharedStyles.textRegular
-	},
 	switchContainer: {
-		height: 54,
+		minHeight: 54,
```
diegolmello, line 35: Why min? Is it expected to grow?

app/views/CreateChannelView.tsx
```diff
@@ -94,11 +96,13 @@ interface ICreateChannelViewProps extends IBaseScreen<ChatsStackParamList, 'Crea
...
+	useRealName: boolean;
 }
 
 interface ISwitch extends SwitchProps {
 	id: string;
 	label: string;
+	hint: string;
 }
 
 class CreateChannelView extends React.Component<ICreateChannelViewProps, ICreateChannelViewState> {
```
diegolmello, line 108: It's a redesign. It's expected to migrate all these screens of the flow to Hooks.

app/views/NewMessageView.tsx
```diff
@@ -71,6 +77,7 @@ interface INewMessageViewProps extends IBaseScreen<NewMessageStackParamList, 'Ne
 	createPublicChannelPermission?: string[];
 	createPrivateChannelPermission?: string[];
 	createDiscussionPermission?: string[];
+	useRealName: boolean;
 }
 
 class NewMessageView extends React.Component<INewMessageViewProps, INewMessageViewState> {
```
diegolmello, line 83: Hooks

diegolmello (changes requested):

app/views/CreateChannelView.tsx
reinaldonetof: Screen Shot 2022-08-02 at 19 10 35


In this case, I added minHeight and MaxHeight because of the different descriptions

app/containers/Chip/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+	text: string;
+	item: ISelectedUser;
+	onPress?: Function;
+	testID?: string;
+	iconName?: TIconsName;
+	style?: StyleProp<ViewStyle>;
+	textStyle?: StyleProp<TextStyle>;
+}
+
+const Chip = ({ item, avatar, text, onPress = () => {}, testID, iconName, style, textStyle }: IChip) => {
```
diegolmello, line 49: What are the odds of sending a different icon for a Chip other than `close`?
It makes `iconName` verbose.
So let's remove it and make `close` as the default icon.

app/containers/Chip/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+				color: colors.bannerBackground
+			}}>
+			<View style={styles.container}>
+				{avatar ? <Avatar text={avatar} size={28} style={styles.avatar} /> : null}
+				<View style={styles.textContainer}>
+					<Text style={[styles.name, { color: colors.bodyText }, textStyle]} numberOfLines={1}>
+						{text}
+					</Text>
+				</View>
+				{iconName ? <CustomIcon name={iconName} size={16} color={colors.auxiliaryTintColor} /> : null}
```
diegolmello, line 73: On this implementation, we want Chips to be pressable only when the user wants to remove them.
As we removed `iconName` from props, we can use `onPress` as the prop responsible to render the icon.

```suggestion
				{onPress ? <CustomIcon name={iconName} size={16} color={colors.auxiliaryTintColor} /> : null}
```

app/containers/Chip/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+		<Pressable
+			testID={testID}
+			style={({ pressed }) => [
+				styles.pressable,
+				{
+					backgroundColor: pressed ? colors.bannerBackground : colors.auxiliaryBackground
+				},
+				style
+			]}
+			onPress={() => onPress(item)}
```
diegolmello, line 62: It should be optional, so we might not have it.
```suggestion
			onPress={() => onPress?.()}
```

`Pressable` shouldn't have active state if they don't have `onPress`.
They should behave just like a `View`.

Make sure to test the existence of the `onPress` on unit tests.

app/containers/Chip/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import sharedStyles from '../../views/Styles';
+import Avatar from '../Avatar';
+
+const styles = StyleSheet.create({
+	pressable: {
+		marginVertical: 16,
+		paddingHorizontal: 8,
+		marginRight: 8,
+		borderRadius: 2,
+		height: 40,
```
diegolmello, line 16: What happens if the user has a larger text setting on their device?
Your accessibility breaks.
image

We should try not to set a fixed height for components like this and use padding instead.

app/containers/UserItem.tsx
```diff
@@ -64,14 +61,11 @@ const UserItem = ({ name, username, onPress, testID, onLongPress, style, icon, t
...
-				<Text style={[styles.name, { color: themes[theme].titleText }]} numberOfLines={1}>
+				<Text style={[styles.name, { color: themes[theme].bodyText }]} numberOfLines={1}>
 					{name}
 				</Text>
-				<Text style={[styles.username, { color: themes[theme].auxiliaryText }]} numberOfLines={1}>
-					@{username}
-				</Text>
 			</View>
-			{icon ? <CustomIcon name={icon} size={22} color={themes[theme].actionTintColor} style={styles.icon} /> : null}
+			{icon ? <CustomIcon name={icon} size={22} color={iconColor || themes[theme].actionTintColor} style={styles.icon} /> : null}
```
diegolmello, line 68: This is possibly wrong and should have a unified color instead.
Let's discuss internally if we need this differentiation.

app/containers/Chip/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+				style
+			]}
+			onPress={() => onPress(item)}
+			android_ripple={{
+				color: colors.bannerBackground
+			}}>
+			<View style={styles.container}>
+				{avatar ? <Avatar text={avatar} size={28} style={styles.avatar} /> : null}
+				<View style={styles.textContainer}>
+					<Text style={[styles.name, { color: colors.bodyText }, textStyle]} numberOfLines={1}>
```
diegolmello, line 69: Why do we need `textStyle`?

app/views/CreateChannelView.tsx
diegolmello: I see.
I added a comment on this review about this usage.
On your screenshot, you can see that, although working as you expect, the layout has no padding for `broadcast`.
Instead of heights, you should be using padding.

Also broadcast description is pt-BR, while the rest is en.

app/views/CreateChannelView.tsx
```diff
@@ -73,357 +74,295 @@ interface IOtherUser {
...
-			type: true,
-			readOnly: false,
-			encrypted: false,
-			broadcast: false,
-			isTeam,
-			permissions: []
-		};
-		this.setHeader();
-	}
+const RenderItem = memo(
```
diegolmello, line 83: I explain the issue with `RenderItem` on another PR.

diegolmello, line 83: `memo` can't memoize objects.
Even though you have a shallow equal there.
Just don't do it.

```ts
interface IOtherUser {
	_id: string;
	name: string;
	fname: string;
}
```

diegolmello, line 83: Another file

app/views/CreateChannelView.tsx
```diff
@@ -73,357 +74,295 @@ interface IOtherUser {
...
+			/>
+		);
+	},
+	shallowEqual
+);
 
-	setHeader = () => {
-		const { navigation } = this.props;
-		const { isTeam } = this.state;
+const RenderSwitch = memo(({ id, value, label, hint, onValueChange, disabled = false }: ISwitch) => {
```
diegolmello, line 102: No memo.
No Render.
Another file.

app/views/CreateChannelView.tsx
```diff
@@ -73,357 +74,295 @@ interface IOtherUser {
...
-					value={value}
-					onValueChange={onValueChange}
-					testID={`create-channel-${id}`}
-					trackColor={SWITCH_TRACK_COLOR}
-					disabled={disabled}
-				/>
-			</View>
+			<RenderSwitch
+				id={'type'}
+				value={permissions[1] ? type : false}
```
diegolmello, line 207: We agreed not to use `permissions[N]` when we implemented `usePermissions`.
Instead you should make it clear on your call.
```tsx
const [nameYourPermission, nameIt] = usePermissions(['create-c', 'create-p']);
```

app/views/CreateChannelView.tsx
```diff
@@ -73,357 +74,295 @@ interface IOtherUser {
...
+		(value: boolean) => {
+			logEvent(events.CR_TOGGLE_TYPE);
+			// If we set the channel as public, encrypted status should be false
+			setType(value);
+			setEncrypted(value && encrypted);
+		},
+		[encrypted]
+	);
+
+	const renderType = () => {
```
diegolmello, line 187: This whole thing should be a different file as a `<Type />` component.
Same for the other functions.

diegolmello (changes requested):
Apply the comments to the other files as well, please.

app/views/CreateChannelView.tsx
reinaldonetof: The translation was fixed :) About the padding, you are right! I will change it

app/containers/Chip/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+			testID={testID}
+			style={({ pressed }) => [
+				styles.pressable,
+				{
+					backgroundColor: pressed ? colors.bannerBackground : colors.auxiliaryBackground
+				},
+				style
+			]}
+			disabled={!onPress}
+			onPress={() => onPress?.()}
```
GleidsonDaniel, line 58:
```suggestion
			onPress={onPress}
```

app/views/CreateChannelView/RoomSettings/SwitchItemEncrypted.tsx
```diff
@@ -0,0 +1,46 @@
...
+		}
+		if (!isTeam && !type) {
+			hint = 'Channel_hint_encrypted_not_available';
+		}
+
+		return (
+			<SwitchItem
+				id={'encrypted'}
+				value={encrypted}
+				label={'Encrypted'}
```
GleidsonDaniel, line 38: This here cannot be translated into other languages or is it really fixed?

app/views/CreateChannelView/RoomSettings/SwitchItemReadOnly.tsx
```diff
@@ -0,0 +1,41 @@
+import React, { memo } from 'react';
+import { shallowEqual } from 'react-redux';
+
+import { SwitchItem } from './SwitchItem';
+
+export const SwitchItemReadOnly = memo(
```
GleidsonDaniel, line 6: I wonder the real efficiency of it, first because it has a function as a prop and another because this component needs to be re-rendered and doesn't have much logic to be memorized.

GleidsonDaniel (changes requested):

app/views/NewMessageView/ButtonCreate.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { themes } from '../../lib/constants';
+import { CustomIcon, TIconsName } from '../../containers/CustomIcon';
+import { useTheme } from '../../theme';
+import Touch from '../../containers/Touch';
+import sharedStyles from '../Styles';
+import { PADDING_HORIZONTAL } from '../../containers/List/constants';
+
+const styles = StyleSheet.create({
+	button: {
+		height: 46,
```
diegolmello, line 13: Can you use paddingVertical instead so it's accessible? (won't cut on larger font sizes)

app/views/NewMessageView/ButtonCreate.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+interface IButton {
+	onPress: () => void;
+	testID: string;
+	title: string;
+	icon: TIconsName;
+	first?: boolean;
+}
+
+const ButtonCreate = ({ onPress, testID, title, icon, first }: IButton) => {
```
diegolmello, line 40: Can't you use `List.Item` here?

app/views/NewMessageView/index.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+	return (
+		<SafeAreaView testID='new-message-view'>
+			<StatusBar />
+			<FlatList
+				data={search.length > 0 ? search : chats}
+				keyExtractor={item => item._id || item.rid}
+				ListHeaderComponent={<HeaderNewMessage maxUsers={maxUsers} onChangeText={handleSearch} />}
+				renderItem={({ item, index }) => (
+					<UserItem goRoom={goRoom} chats={chats} index={index} item={item} search={search} useRealName={useRealName} />
```
diegolmello, line 94: Can you change UserItem to match Figma with 16 font size?
It's going to change other places, but it's quite noticeable now with Inter font

app/views/NewMessageView/UserItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+			return false;
+		}
+		if (prevProps.chats.length !== nextProps.chats.length) {
+			return false;
+		}
+		if (prevProps.search.length !== nextProps.search.length) {
+			return false;
+		}
+		return true;
+	}
```
diegolmello, line 64: This is probably not needed.
There's no logic to add reactivity here depending on a user name change.

app/views/NewMessageView/UserItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+	item: TItem;
+	index: number;
+	search: TItem[];
+	chats: TSubscriptionModel[];
+	useRealName: boolean;
+	goRoom: (item: TGoRoomItem) => void;
+}
+
+const UserItem = memo(
+	({ item, index, chats, search, useRealName, goRoom }: IUserItem) => {
```
diegolmello, line 22: chats and search are wrong here.
You're adding complexity to the component and coupling it to the parent view when you could just send the data you actually want as props.
It could be just a single `listLength` prop.

Theory aside, I couldn't find why this is still needed.
Couldn't it just be a `ListFooterComponent`?
When we're doing refactors like this, we need to actually evaluate the code and understand what needs to be improved.

app/views/SelectedUsersView/Header.tsx
```diff
@@ -0,0 +1,58 @@
...
+import SearchBox from '../../containers/SearchBox';
+import I18n from '../../i18n';
+import { ISelectedUser } from '../../reducers/selectedUsers';
+import { useTheme } from '../../theme';
+import sharedStyles from '../Styles';
+import { useAppSelector } from '../../lib/hooks';
+import SelectedChipItem from './SelectedChipItem';
+
+const ITEM_WIDTH = 250;
+const getItemLayout = (_: any, index: number) => ({ length: ITEM_WIDTH, offset: ITEM_WIDTH * index, index });
```
diegolmello, line 14: This is wrong.
The new Chip component has a dynamic size, so we can't calculate the positions.
Just remove the whole `getItemLayout` logic.

app/views/SelectedUsersView/SelectedChipItem.tsx
```diff
@@ -0,0 +1,35 @@
...
+			return false;
+		}
+		if (prevProps.item?.fname !== nextProps.item?.fname) {
+			return false;
+		}
+		if (prevProps.useRealName !== nextProps.useRealName) {
+			return false;
+		}
+		return true;
+	}
```
diegolmello, line 32: We need to avoid this.
This is not reactive.
Why do we need to keep track of user name change?
If you just send primitive values as props, it's going to work out of box.

app/views/SelectedUsersView/UserItem.tsx
```diff
@@ -0,0 +1,69 @@
...
+	index: number;
+	name: string;
+	username: string;
+	searchLength: number;
+	chatsLength: number;
+	isChecked: boolean;
+	onPressItem: (item: ISelectedUser) => void;
+}
+
+const UserItem = memo(
```
diegolmello, line 19: Apply the same principles I've been commenting.

app/views/SelectedUsersView/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+		state => ({
+			users: state.selectedUsers.users,
+			loading: state.selectedUsers.loading,
+			useRealName: state.settings.UI_Use_Real_Name as boolean,
+			user: getUserSelector(state)
+		}),
+		shallowEqual
+	);
+
+	sendLoadingEvent({ visible: loading });
```
diegolmello, line 56: Why are you sending this on every render? 😨

app/views/SelectedUsersView/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+	const isChecked = useCallback((username: string) => users.findIndex(el => el.name === username) !== -1, [users]);
```
diegolmello, line 58: This is being recreated on every render, because users is a non-primitive value.
Why do you need the useCallback?
You're wasting time trying to improve the performance of something without knowing if it's needed.

app/views/CreateChannelView/UserItem.tsx
```diff
@@ -0,0 +1,22 @@
...
+export const UserItem = memo(
+	({ item, removeUser, useRealName }: { item: IOtherUser; useRealName: boolean; removeUser: (item: IOtherUser) => void }) => {
+		const name = useRealName && item.fname ? item.fname : item.name;
+		const username = item.name;
+
+		return (
+			<Chip text={name} avatar={username} onPress={() => removeUser(item)} testID={`create-channel-view-item-${item.name}`} />
+		);
+	},
+	shallowEqual
```
diegolmello, line 21: Avoid this

app/views/CreateChannelView/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+	const { isFetching, useRealName, users } = useAppSelector(
+		state => ({
+			isFetching: state.createChannel.isFetching,
+			users: state.selectedUsers.users,
+			useRealName: state.settings.UI_Use_Real_Name as boolean
+		}),
+		shallowEqual
+	);
+
+	sendLoadingEvent({ visible: isFetching });
```
diegolmello, line 84: Why do we need to do this on every render?

diegolmello (changes requested):
Please review as a whole.

app/views/NewMessageView/UserItem.tsx
reinaldonetof: As we talked about yesterday, I removed this component and implemented it directly in the render item.

app/views/CreateChannelView/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+const styles = StyleSheet.create({
+	container: {
+		flex: 1
+	},
+	list: {
+		width: '100%'
+	},
+	invitedHeader: {
+		marginTop: 18,
```
diegolmello, line 34: Not a multiple of 4 and not the same spacing as seen on Figma.
The total spacing you're applying is 26 and should be 24

app/views/CreateChannelView/UserItem.tsx
diegolmello: 🤔

app/views/SelectedUsersView/Header.tsx
```diff
@@ -0,0 +1,61 @@
...
+	}));
+
+	const onContentSizeChange = () => flatlist?.current?.scrollToEnd({ animated: true });
+
+	return (
+		<View style={{ backgroundColor: themes[theme].backgroundColor }}>
+			<SearchBox onChangeText={(text: string) => onChangeText(text)} testID='select-users-view-search' />
+			{users.length === 0 ? null : (
+				<View>
+					<Text style={{ ...sharedStyles.textRegular, color: themes[theme].auxiliaryTintColor, marginLeft: 16 }}>
```
diegolmello, line 35: Missing fontSize and inline style

app/views/SelectedUsersView/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+						<HeaderButton.Item
+							title={users.length > 0 ? buttonTextHeader : I18n.t('Skip')}
+							onPress={nextActionHeader}
+							testID='selected-users-view-submit'
+						/>
+					</HeaderButton.Container>
+				)
+		};
+		navigation.setOptions(options);
+	}, [users.length, maxUsers]);
```
diegolmello, line 83: missing navigation

app/views/SelectedUsersView/Header.tsx
```diff
@@ -0,0 +1,61 @@
...
+						renderItem={({ item }) => {
+							const name = useRealName && item.fname ? item.fname : item.name;
+							const username = item.search ? (item.username as string) : item.name;
+
+							return (
+								<Chip text={name} avatar={username} onPress={() => onPressItem(item)} testID={`selected-user-${item.name}`} />
+							);
+						}}
+						keyboardShouldPersistTaps='always'
+						contentContainerStyle={{ paddingLeft: 16 }}
```
diegolmello, line 52: paddingHorizontal. When you reach the end of the list, it's expected to have a padding at the end.

app/views/SelectedUsersView/index.tsx
```diff
@@ -0,0 +1,188 @@
...
+				renderItem={({ item, index }) => {
+					const name = useRealName && item.fname ? item.fname : item.name;
+					const username = item.search ? (item.username as string) : item.name;
+					let style = { borderColor: colors.separatorColor };
+					if (search.length > 0 && index === search.length - 1) {
+						style = { ...style, ...sharedStyles.separatorBottom };
+					}
+					if (search.length === 0 && index === chats.length - 1) {
+						style = { ...style, ...sharedStyles.separatorBottom };
+					}
```
diegolmello, line 166: We talked about this.
Can't you use ListFooterComponent with a small condition?
It seems unnecessary to test this on every item.

diegolmello (changes requested):

app/containers/Chip/Chip.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+	<>
+		<View style={styles.container}>
+			<Chip avatar='rocket.cat' text={'Rocket.Cat'} />
+			<Chip avatar='rocket.cat' text={'Short'} />
+			<Chip text='Without Avatar' />
+			<Chip avatar='rocket.cat' text='Without Icon' />
+			<Chip text='Without Avatar and Icon' />
+		</View>
+	</>
+));
```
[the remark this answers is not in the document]
reinaldonetof: Yeap

diegolmello (approved):

PR #4399, opened 2022-07-29
[NEW] Unify members section

GleidsonDaniel:
## Proposed changes

Unifies the session of adding or inviting users to the members screen.

## How to test or reproduce

- Try to add or remove some user through the new menu.

Check if the settings are working correctly:
- Possibility to add or not a user.
- Possibility of inviting or not a user.
## Types of changes

## Checklist


app/containers/CheckIcon/index.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { RadioButton } from 'react-native-ui-lib';
+
+import { useTheme } from '../../theme';
+
+export const CheckRadioButton = ({
```
diegolmello, line 6: I'm not sure about this function and file name.
Checkbox and Radio button are different things https://material.io/components/radio-buttons#usage

I think it's worth keeping it as RadioButton, because of size and colors.

app/containers/MembersSection/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+	StackNavigationProp<MasterDetailInsideStackParamList>
+>;
+
+interface IMembersSection {
+	rid: TSubscriptionModel['rid'];
+	t: TSubscriptionModel['t'];
+	joined: boolean;
+}
+
+export default function MembersSection({ rid, t, joined }: IMembersSection): React.ReactElement {
```
diegolmello, line 28: I can't visualize where we would be using this component outside of RoomMembersView.
Can we colocate it closer to the component where it's going to be used?

app/lib/hooks/useUserPermissions.ts
```diff
@@ -0,0 +1,51 @@
...
+import { useAppSelector } from './useAppSelector';
+
+interface IMembersSection {
+	rid: TSubscriptionModel['rid'];
+	t: TSubscriptionModel['t'];
+	joined: boolean;
+}
+// Promise.allSettled() is not working for react native, in the future use this instead of various useEffects
+// https://github.com/facebook/react-native/issues/30236#issuecomment-939286987
+export const useUserPermissions = ({ rid, t, joined }: IMembersSection): { canAddUser: boolean; canInviteUser: boolean } => {
```
diegolmello, line 14: I don't think we need this custom hook.
We have `usePermissions` for this use case.

Also, this custom hooks wouldn't be reactive.

app/views/RoomMembersView/index.tsx
```diff
@@ -72,6 +74,7 @@ interface IRoomMembersViewState {
 
 class RoomMembersView extends React.Component<IRoomMembersViewProps, IRoomMembersViewState> {
```
diegolmello, line 75: It's a redesign, so it's expected to migrate it to Hooks.

diegolmello (changes requested):

app/views/RoomMembersView/index.tsx
```diff
@@ -420,251 +506,91 @@ class RoomMembersView extends React.Component<IRoomMembersViewProps, IRoomMember
...
+			? state.members.filter(
+					m =>
+						m.username.toLowerCase().match(state.filter.toLowerCase()) || m.name?.toLowerCase().match(state.filter.toLowerCase())
+			  )
+			: null;
+
+	return (
+		<SafeAreaView testID='room-members-view'>
+			<StatusBar />
+			<FlatList
```
diegolmello, line 577: Not on Figma, but we should add `No members found`

app/views/RoomMembersView/index.tsx
```diff
@@ -420,251 +506,91 @@ class RoomMembersView extends React.Component<IRoomMembersViewProps, IRoomMember
...
-		const { theme } = this.props;
-
-		return (
-			<UserItem
-				name={item.name as string}
-				username={item.username}
-				onPress={() => this.onPressUser(item)}
-				testID={`room-members-view-item-${item.username}`}
-				theme={theme}
+	const Member = React.memo(({ member }: { member: TUserModel }) => (
```
diegolmello, line 550: We shouldn't add Nested Components.
Also, memo is `receiving` an observable obj from WatermelonDB and it's not memoizing the component.

diegolmello (changes requested):

e2e/tests/room/03-roomactions.spec.js
```diff
@@ -442,13 +446,27 @@ describe('Room actions screen', () => {
...
+						.withTimeout(2000);
+					await element(by.id('room-members-view-toggle-status-all')).tap();
 					await waitFor(element(by.id(`room-members-view-item-${user.username}`)))
 						.toExist()
 						.withTimeout(60000);
+					await tapBack();
 				});
 
 				it('should filter user', async () => {
+					await element(by.id('room-actions-members')).tap();
```
diegolmello, line 464: Always `waitFor(...).toExist().withTimeout(2000)` before tapping stuff.
It makes tests more reliable.
The way it's right now, the test is a bit flaky, because it relies only on the `tapBack` from the previous test.
The `tapBack` triggers the pop animation from react-navigation and only then we can actually `tap` the item.

app/views/RoomMembersView/index.tsx
```diff
@@ -420,251 +516,86 @@ class RoomMembersView extends React.Component<IRoomMembersViewProps, IRoomMember
...
+					</View>
+				)}
+				style={[styles.list]}
+				keyExtractor={item => item._id}
+				ItemSeparatorComponent={List.Separator}
+				ListHeaderComponent={renderHeader(state.room.rid, state.room.t)}
+				ListFooterComponent={() => (state.isLoading ? <ActivityIndicator /> : null)}
+				onEndReachedThreshold={0.1}
+				onEndReached={() => fetchMembers(state.allUsers)}
+				ListEmptyComponent={() => (
```
diegolmello, line 592: It's flashing `no members found` when we open the screen.

diegolmello (changes requested):

PR #4316, opened 2022-06-24
Chore: Upgrade React Native to 0.68.2

GleidsonDaniel:
## Proposed changes

Upgrades React Native to 0.68.2, which is a huge jump from 0.64.2.

## Issue(s)	

Depends on https://github.com/RocketChat/rn-root-view/pull/3
Depends on https://github.com/RocketChat/react-native/tree/0.68.2

## How to test or reproduce

- We need to test it deeper than a beta, because things might break
  - For example, attachment zoom was failing and I had to fix

### This PR left side vs Production app right side

Running on Samsung S10e.
I didn't find a visible difference in performance on iOS.

https://user-images.githubusercontent.com/804994/182207449-559cc15e-8699-4f8a-a88a-037a3943e278.mp4

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4401, opened 2022-07-30
Chore: Migrate App/index.tsx to hooks

reinaldonetof:
## Proposed changes

Migrate the App/index.tsx to hooks.

This is the second file, after the index.js. Here we start with a lot of providers, the redux and some modals.

## How to test or reproduce

- Test the themes, change them to another, and back
- Test the deeplink
- Test when dimension changes
- Test the tablet
- Test the theme in share extension

## Types of changes

## Checklist

## Further comments

Jira: [NATIVE-45]

[NATIVE-45]: https://rocketchat.atlassian.net/browse/NATIVE-45?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


guilhermedfs (commented):
LGTM!

liversonp:
Hello. I was reviewing about hooks in react native and I found this way to hook consts using usereducer, I think this can help in clean code
 - https://pt-br.reactjs.org/docs/hooks-reference.html#usereducer

app/index.tsx
```diff
@@ -81,57 +72,54 @@ const parseDeepLinking = (url: string) => {
...
 			});
 		}, 5000);
-		Dimensions.addEventListener('change', this.onDimensionsChange);
-	}
+		Dimensions.addEventListener('change', onDimensionsChange);
 
-	componentWillUnmount() {
-		clearTimeout(this.listenerTimeout);
-		Dimensions.removeEventListener('change', this.onDimensionsChange);
+		return () => {
```
diegolmello, line 110: I think you should run prettier to fix this spacing.

app/index.tsx
```diff
@@ -153,54 +141,54 @@ export default class Root extends React.Component<{}, IState> {
...
+		subscribeTheme(themePreferences, setThemeFunction);
 	};
 
-	setDimensions = ({ width, height, scale, fontScale }: IDimensions) => {
-		this.setState({ width, height, scale, fontScale });
+	const setDimensions = ({ width, height, scale, fontScale }: IDimensions) => {
+		setWidth(width);
+		setHeight(height);
+		setScale(scale);
+		setFontScale(fontScale);
```
diegolmello, line 181: We'll need to test this on #4316, which uses React 18, because if automatic batching isn't working on React + RN, this is going to re-render this component 4 times.
https://reactjs.org/blog/2022/03/29/react-v18.html#new-feature-automatic-batching

app/index.tsx
```diff
@@ -81,57 +72,54 @@ const parseDeepLinking = (url: string) => {
...
-			this.initTablet();
+			initTablet();
 		}
 		setNativeTheme(theme);
-	}
+	}, []);
 
-	componentDidMount() {
-		this.listenerTimeout = setTimeout(() => {
+	useEffect(() => {
```
diegolmello, line 99: Why do you need 2 useEffects with the same dependencies?
Are they going to run in parallel or something? 🤔

app/index.tsx
```diff
@@ -81,57 +72,54 @@ const parseDeepLinking = (url: string) => {
 	return null;
 };
 
-export default class Root extends React.Component<{}, IState> {
-	private listenerTimeout!: any;
+const Root = () => {
+	const { width: widthWindow, height: heightWindow, scale: scaleWindow, fontScale: fontScaleWindow } = Dimensions.get('window');
```
diegolmello, line 76: You're re-running this `Dimensions.get` on every render, which used to be called only on `constructor`.
What you can do is actually call it outside the component.

app/index.tsx
```diff
@@ -81,57 +72,54 @@ const parseDeepLinking = (url: string) => {
...
+	const [theme, setTheme] = useState(getTheme(initialTheme()));
+	const [themePreferences, setThemePreferences] = useState(initialTheme());
+	const [width, setWidth] = useState(widthWindow);
+	const [height, setHeight] = useState(heightWindow);
+	const [scale, setScale] = useState(scaleWindow);
+	const [fontScale, setFontScale] = useState(fontScaleWindow);
 
-	private onKeyCommands: any;
+	const listenerTimeout = useRef<any>();
+	const onKeyCommands = useRef<any>();
```
diegolmello, line 85: types?

app/index.tsx
```diff
@@ -81,57 +72,54 @@ const parseDeepLinking = (url: string) => {
...
+			Dimensions.removeEventListener('change', onDimensionsChange);
 
-		unsubscribeTheme();
+			unsubscribeTheme();
 
-		if (this.onKeyCommands && this.onKeyCommands.remove) {
-			this.onKeyCommands.remove();
-		}
-	}
+			if (onKeyCommands.current && onKeyCommands.current.remove) {
```
diegolmello, line 116:
```suggestion
			if (onKeyCommands?.current?.remove) {
```

app/index.tsx
```diff
@@ -81,57 +72,54 @@ const parseDeepLinking = (url: string) => {
...
+				onKeyCommands.current.remove();
```
diegolmello, line 117:
```suggestion
				onKeyCommands.current.remove?.();
```

app/share.tsx
```diff
@@ -141,7 +141,7 @@ class Root extends React.Component<{}, IState> {
 	setTheme = (newTheme = {}) => {
 		// change theme state
 		this.setState(
-			prevState => newThemeState(prevState, newTheme as IThemePreference),
+			prevState => newThemeState(prevState.themePreferences, newTheme as IThemePreference),
```
diegolmello, line 144: Why?

app/index.tsx
```diff
@@ -153,54 +141,54 @@ export default class Root extends React.Component<{}, IState> {
...
-	setTheme = (newTheme = {}) => {
-		// change theme state
-		this.setState(
-			prevState => newThemeState(prevState, newTheme as IThemePreference),
-			() => {
-				const { themePreferences } = this.state;
-				// subscribe to Appearance changes
-				subscribeTheme(themePreferences, this.setTheme);
-			}
+	const setThemeFunction = (newThemeObject = {}) => {
```
diegolmello, line 167: This new naming is awkward.
`setThemeFunction`? `newThemeObject`?
Why?

app/index.tsx
```diff
@@ -153,54 +141,54 @@ export default class Root extends React.Component<{}, IState> {
...
-		this.onKeyCommands = KeyCommandsEmitter.addListener('onKeyCommand', (command: ICommand) => {
+	const initTablet = () => {
+		setMasterDetail(width);
+		onKeyCommands.current = KeyCommandsEmitter.addListener('onKeyCommand', (command: ICommand) => {
 			EventEmitter.emit(KEY_COMMAND, { event: command });
 		});
 	};
 
-	initCrashReport = () => {
+	const initCrashReport = () => {
```
diegolmello, line 191: As it's the root function component and we know its behavior, I think you should wrap all these functions in `useCallback` to prevent these recreations.

app/index.tsx
```diff
@@ -153,54 +141,54 @@ export default class Root extends React.Component<{}, IState> {
...
-		const isMasterDetail = this.getMasterDetail(width);
+	const setMasterDetail = (width: number) => {
+		const isMasterDetail = getMasterDetail(width);
 		store.dispatch(setMasterDetailAction(isMasterDetail));
 	};
 
 	// Dimensions update fires twice
-	onDimensionsChange = debounce(({ window: { width, height, scale, fontScale } }: { window: IDimensions }) => {
-		this.setDimensions({
+	const onDimensionsChange = debounce(({ window: { width, height, scale, fontScale } }: { window: IDimensions }) => {
```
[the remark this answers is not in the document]
reinaldonetof: It's working properly here! Because the properties are coming from the listener.

Another issue was happening when changes some property from the state and the function inside the debounce needs them.

app/index.tsx
reinaldonetof: Here, the React also batch the setState, instead of re-render 4 times. However, one interesting thing that I noticed just after you commented here is that it re-renders 2 times when changing the orientation, because the orientation active the debounce, and debounce is an asynchronous function when we call setState from synchronous function and the asynchronous function will re-render multiple times. Because of that, I needed to use the [unstable_batchedUpdates](https://stackoverflow.com/a/48610973) from react-native, but I need to test when change to React 18

app/share.tsx
reinaldonetof: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4401/files#diff-b788f0c6b6e92a2a43dedd61dd4e596b4b0f981e30bcfd88e2f5d30b2de3a7dbR39

Because I changed the function newThemeState to receive only the property required, not the state completely.

app/index.tsx
```diff
@@ -80,57 +71,53 @@ const parseDeepLinking = (url: string) => {
...
-		}
-	}
+			if (onKeyCommands?.current?.remove) {
+				onKeyCommands.current.remove?.();
+			}
+		};
+	}, []);
 
-	init = async () => {
+	const init = useCallback(async () => {
```
diegolmello, line 120: Can you remove the non-null assertion?

app/index.tsx
```diff
@@ -150,103 +137,107 @@ export default class Root extends React.Component<{}, IState> {
...
+	const initTablet = useCallback(() => {
+		setMasterDetail(widthWindow);
+		onKeyCommands.current = KeyCommandsEmitter.addListener('onKeyCommand', (command: ICommand) => {
 			EventEmitter.emit(KEY_COMMAND, { event: command });
 		});
-	};
+	}, []);
 
-	initCrashReport = () => {
+	const initCrashReport = useCallback(() => {
```
diegolmello, line 197: It doesn't need to be a function inside the component.
It should be a separate file.

app/index.tsx
```diff
@@ -150,103 +137,107 @@ export default class Root extends React.Component<{}, IState> {
...
+		setScale(scale);
+		setFontScale(fontScale);
 	};
 
-	initTablet = () => {
-		const { width } = this.state;
-		this.setMasterDetail(width);
-		this.onKeyCommands = KeyCommandsEmitter.addListener('onKeyCommand', (command: ICommand) => {
+	const initTablet = useCallback(() => {
+		setMasterDetail(widthWindow);
```
diegolmello, line 191: Same as initCrashReport

app/index.tsx
```diff
@@ -80,57 +71,53 @@ const parseDeepLinking = (url: string) => {
...
-		Dimensions.addEventListener('change', this.onDimensionsChange);
-	}
+		Dimensions.addEventListener('change', onDimensionsChange);
 
-	componentWillUnmount() {
-		clearTimeout(this.listenerTimeout);
-		Dimensions.removeEventListener('change', this.onDimensionsChange);
+		return () => {
+			clearTimeout(listenerTimeout.current as ReturnType<typeof setTimeout>);
+			Dimensions.removeEventListener('change', onDimensionsChange);
```
diegolmello, line 110: `removeEventListener` is deprecated.
Can you migrate to the new one?

app/index.tsx
```diff
@@ -150,103 +137,107 @@ export default class Root extends React.Component<{}, IState> {
...
+				setMasterDetail(width);
+			});
+		}),
+		[]
+	);
+
+	const setNewTheme = (newThemePreference = {}) => {
+		const { theme: newTheme, themePreferences: newThemePreferences } = newThemeState(
+			themePreferences,
+			newThemePreference as IThemePreference
```
diegolmello, line 176: You're kind of just making TS happy here.
You should avoid `as` as much as you can.

```tsx
const setNewTheme = (newThemePreference: IThemePreference) => {
	const { theme: newTheme, themePreferences: newThemePreferences } = newThemeState(
		themePreferences,
		newThemePreference
	);
	setThemePreferences(newThemePreferences);
	setTheme(newTheme);
	subscribeTheme(themePreferences, setNewTheme);
};

...

subscribeTheme = (themePreferences: IThemePreference, setTheme: Function): void
```

Please evaluate the need of `{}` there.
I don't think it's needed.

app/lib/methods/helpers/theme.ts
```diff
@@ -36,10 +36,10 @@ export const getTheme = (themePreferences: IThemePreference): TSupportedThemes =
 	return theme === 'dark' ? darkLevel : 'light';
 };
 
-export const newThemeState = (prevState: { themePreferences: IThemePreference }, newTheme: IThemePreference) => {
+export const newThemeState = (prevthemePreferences: IThemePreference, newTheme: IThemePreference) => {
```
diegolmello, line 39:
```suggestion
export const newThemeState = (prevThemePreferences: IThemePreference, newTheme: IThemePreference) => {
```

👀

diegolmello (changes requested):

app/index.tsx
```diff
@@ -150,103 +137,107 @@ export default class Root extends React.Component<{}, IState> {
...
-		// change theme state
-		this.setState(
-			prevState => newThemeState(prevState, newTheme as IThemePreference),
-			() => {
-				const { themePreferences } = this.state;
-				// subscribe to Appearance changes
-				subscribeTheme(themePreferences, this.setTheme);
-			}
+	const onDimensionsChange = useCallback(
+		debounce(({ window: { width, height, scale, fontScale } }: { window: IDimensions }) => {
```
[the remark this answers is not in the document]
reinaldonetof: Actually, the re-render will be triggered because the window is an object, isn't it?

app/index.tsx
reinaldonetof: you are right!

app/lib/methods/helpers/theme.ts
reinaldonetof: Ok, fixed

GleidsonDaniel (approved):
Add a comment to your task saying that it has to be well tested.
All app works.
Open and close, push etc.

Issue #4360, opened 2022-07-13
[single-server] When I follow notification - app stucks on loading messages


nekkitl:
Hi!

### Description:

In our neighbourhoods we chat a lot and we found issue. When we get a lot massages and want to read them - we got limitless loading messages frame…
We try to debug via logs/wireshark and find in code why its not working well and why app stuck on this.

### Environment Information:

- Rocket.Chat Server Version: 4.6 - Docker
- Rocket.Chat App Version: 4.26 - Single-Server
- Device Name: iPhone 12 Pro Max / Google Pixel 6
- OS Version: iOS 15.5 (iOS 16beta3) / Android 12

### Steps to reproduce:

1. Somebody send u messages (we spam them to faster reproduce)
2. Get all notifies on device, then
3. follow notifications and get this issue (look screenshot)

### Actual behavior:

Stucks on this, but if we continue write something - we see behind this loader.
(we changed loader picture for fun)
![screenshot](https://user-images.githubusercontent.com/107915590/178701214-7fd1a632-fdc4-4a66-9d50-e525d7309d7e.png)

diegolmello:
Yep, so the issue is where I sent you.
When you tap the push notification, the app redirects you to the room and tries to scroll to the message you've tapped.
Since you're receiving too many messages while it tries to scroll, the app gets lost.
The logic should have a timeout in case it doesn't find the message.

PR #4410, opened 2022-08-04
[FIX] Jump to message stuck on loading animation

diegolmello:
## Proposed changes

- [x] Fixed a few race conditions causing the loading to get stuck
- [x] Changed the loading to component to be global
- [x] Added a cancel option to the loading, so depending on the process the user can stop
- [x] Updated E2E tests
- [x] Unit tests for Loading component
- [x] Apply global Loading to other screens

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4360

## How to test or reproduce

- Tests should be passing
- Share a file from share extension

#### Following actions have E2E tests
- Add existing channel to a team
- Create a channel
- Create a discussion
- Edit a room
- Jump to a message on the same room
- Jump to a message inside a thread
- Jump to a message from another room
- Change user status

## Types of changes

## Checklist


app/views/RoomInfoEditView/index.tsx
```diff
@@ -339,7 +337,7 @@ class RoomInfoEditView extends React.Component<IRoomInfoEditViewProps, IRoomInfo
 			log(e);
 		}
 
-		await this.setState({ saving: false });
+		EventEmitter.emit(LOADING_EVENT, { visible: false });
```
GleidsonDaniel, line 340: Why, instead of repeating this logic of importing so many separate things every time, not create a generic handler into loading container? Something like that?

`export const sendLoadingEvent = (value: boolean): void => EventEmitter.emit(LOADING_EVENT, { visible: value });`

In the long run it would be less expensive, since it is two less imports and would keep the logic in a single function and in a single file.

GleidsonDaniel (approved):

PR #4414, opened 2022-08-08
Chore: Improve code and migrate some props to hooks on ThemeView

GleidsonDaniel:
## Proposed changes

Improve code and migrate some props to hooks on ThemeView

## How to test or reproduce

Test the screen ThemeView

## Types of changes

## Checklist


app/views/ThemeView.tsx
```diff
@@ -92,45 +90,48 @@ const ThemeView = ({ navigation }: IThemeViewProps): React.ReactElement => {
...
-					title={label}
-					onPress={() => onClick(item)}
-					testID={`theme-view-${value}`}
-					right={() => (isSelected(item) ? renderIcon() : null)}
-				/>
-				<List.Separator />
-			</>
-		);
-	};
+	const RenderItem = React.memo(({ theme }: { theme: ITheme }) => (
```
reinaldonetof, line 104: This `React.memo` doesn't work, because the functions `onClick` and `isSelected` are recreated all re-render. In this case, we should use the `useCallback`.

What do you think about removing the RenderItem from inside the `father` and create a new component to `RenderItem`.

reinaldonetof (changes requested):

app/views/ThemeView.tsx
GleidsonDaniel: I prefer useCallback

reinaldonetof: Separate RenderItem to another file `Item`

reinaldonetof (changes requested):

app/views/ThemeView.tsx
```diff
@@ -58,79 +57,112 @@ interface ITheme {
...
-			logEvent(events.THEME_SET_THEME_GROUP, { theme_group: value });
-			changes = { currentTheme: value as TThemeMode };
-		}
-		if (group === DARK_GROUP && darkLevel !== value) {
-			logEvent(events.THEME_SET_DARK_LEVEL, { dark_level: value });
-			changes = { darkLevel: value as TDarkLevel };
-		}
-		_setTheme(changes);
-	};
+	const onClick = useCallback(
```
diegolmello, line 109: Not really doing anything apparently (or almost nothing).
The only state we have is `themePreferences` and you had to add it as dependencies.
It might be less memory consuming to just recreate it on every render.

app/views/ThemeView.tsx
```diff
@@ -58,79 +57,112 @@ interface ITheme {
 	group: string;
 }
 
-type IThemeViewProps = IBaseScreen<SettingsStackParamList, 'ThemeView'>;
+const Item = ({
```
diegolmello, line 60: We discussed this on Today's daily.
One component per file, just to make it easier to read and files with fewer lines :)

diegolmello (changes requested):

app/views/ThemeView.tsx
reinaldonetof: I asked to @GleidsonDaniel and he told me that you both agreed to keep the item in the same file. So, considering that, I can approve this PR.

reinaldonetof (approved):

PR #4417, opened 2022-08-08
Chore: Hooks App/share

reinaldonetof:
## Proposed changes

Migrate the app/share.tsx from Class to Hooks. And cleaned some functions that aren't in use.

## How to test or reproduce

- Open the gallery or folder from your smartphone 
- Share something by Rocket.Chat
- Should open the Share Extension
- Try logged in some server and without logged

## Types of changes

## Checklist


app/share.tsx
```diff
@@ -90,115 +71,68 @@ export const App = ({ root }: any) => (
...
-		this.init();
-	}
+const { width, height, scale, fontScale } = Dimensions.get('screen');
 
-	componentDidMount() {
-		this.mounted = true;
-	}
+const Root = () => {
+	const [root, setRoot] = useState('');
+	const theme = getTheme(initialTheme());
```
diegolmello, line 78: While this is ok on React.Component's `constructor`, it's not on a function component.
It's going to re-run these functions on every render.
You should do it outside the component, as you did for Dimensions.

If you compare this logic with `App/Index`, you'll notice you did this:
```jsx
const [theme, setTheme] = useState(getTheme(initialTheme()));
```

This is totally fine, since you're running the functions only once for the default value of `useState()`.
Here, you can do it outside the component, since we don't really care if the theme changes inside the share extension

diegolmello (changes requested):

app/share.tsx
reinaldonetof: Sure, You are right!

app/share.tsx
```diff
@@ -89,115 +70,68 @@ export const App = ({ root }: any) => (
...
-			root: '',
-			width,
-			height,
-			scale,
-			fontScale
-		};
-		this.init();
-	}
+const { width, height, scale, fontScale } = Dimensions.get('screen');
+const theme = getTheme(initialTheme());
```
GleidsonDaniel, line 74: Why are these properties that are only used inside Root outside of it?
As it is here, they are allocated in memory while the app is running, and not just at screen execution time.

GleidsonDaniel (changes requested):

app/share.tsx
reinaldonetof: @GleidsonDaniel this was discussed with @diegolmello 

https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4417#discussion_r942735088

GleidsonDaniel (approved):

PR #4421, opened 2022-08-08
[FIX] Can't connect to RC through Gitlab Oauth

### LINKED ISSUE #4354: Can't connect to RC through Gitlab Oauth
### Description:

Before upgrading to RocketChat v4.8.1 (previously in 4.4.X) we couldn't connect through Gitlab Oauth (gitlab self hosted) because of an issue (https://github.com/RocketChat/Rocket.Chat/issues/23613) that has been fixed on RC 4.5.2.
Now that we use RocketChar 4.8.1, we can connect through Gitlab Oauth on the desktop app but it is still impossible with iOS and Android app since it automatically redirect to gitlab.com instead of our own gitlab URL

### Environment Information:

- Rocket.Chat Server Version: 4.8.1
- Rocket.Chat App Version: 4.28.0.33046 (Android) /  4.28.0.33048 (iOS)
- Device Name: OnePlus 9 Pro / Iphone 11
- OS Version: Android 11 Oxygen OS build 11.2.8.8.LE15BA

### Steps to reproduce:

- Try to connect to a rocketchat instance that has a Gitlab Oauth with self hosted Gitlab
- Connect with Gitlab 
- Get stuck in the wrong gitlab url page

### Expected behavior:

Should be redirected to the login page of my self hosted gitlab instance.

### Actual behavior:

- Get stuck on this page to gitlab.com, instead of my self hosted gitlab URL.
292381442_604801684290098_7750065793581607085_n

PetMoln:
## Proposed changes

Fix missing initalValue in Array.reduce() which is causing this and potentially some other similar config parsing bugs.

## Issue(s)	

Closes #4354

## How to test or reproduce

To reproduce the original problem you need a server with custom GitLab URL configured. I sent an example of such a server  to @diegolmello in private on the Open RC server on Jul 19th, I'm happy to share the URL in private if needed.

## Types of changes

## Checklist

## Further comments

Processing of the first item of the settings array was missed because of missing initial value in Array.reduce().
Because of that, the first element of loginSettings in getSettings.ts (which happens to be API_Gitlab_URL) was not parsed, making the gitlab URL fall back to the default value (gitlab.com).

According to the [documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) if the `initialValue` is not provided, then `array[0]` is used as the `initialValue`. In case of this code `ret` is used as the `previousValue` argument, hence the first item of the array is returned without processing, and only the subsequent values are processed by the `Array.reduce()` function.


GleidsonDaniel (approved):
🎉

NoxInmortus:
When will this patch be released ?

diegolmello:
@NoxInmortus Beta is planned to land next Monday, Aug 22. Public release Aug 29.

PR #4435, opened 2022-08-15
[IMPROVE] Remove show message in main thread preference

reinaldonetof:
## Proposed changes

This PR https://github.com/RocketChat/Rocket.Chat/pull/26002 introduced a breaking change in Rocket Chat server 5.0.0. You can see the behavioral differences in the PR

Now there are three behavior options
Gifs copied from the PR
`Default`: it unchecks after sending the first message

`Always`: stay checked for all messages, but the user can unchecked.

`Never`: stay unchecked for all messages, but the user can check it.

The tests are running correctly for servers < 5.0.0. If your server is >= 5.0.0, please change the configuration to "default"

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4453

## Types of changes

## Checklist


GleidsonDaniel (approved):

diegolmello (changes requested):
- [ ] You're checking for the value on the wrong place
  - It's a user preference, not a server setting
  - You can see it on My Account > Preferences > Messages > Also send thread message to channel behavior
  - I showed you this on our refinement call
- [ ] Can you add the new preference to `UserPreferencesView`?
- [ ] The `default` option is not working as expected
  - `default` is `true` for the message that creates the thread, not for every first message you send since you opened a thread
  - also, it's not checking tshow for the message that creates the thread (see video below)
- [ ] Pay attention to the behavior of the `always/never`
  - It should consider the preference only on the first time
  - If the user changes the value, it shouldn't clear it after a message is sent
  - example: if the user has `always` and unchecks it, it should remain unchecked for all messages until they leave the room

https://user-images.githubusercontent.com/804994/185681609-5842c949-cfa3-4b36-96eb-85ffa3f7ad11.mp4

app/containers/List/ListPicker.tsx
```diff
@@ -0,0 +1,57 @@
...
+interface IListPicker {
+	title: string;
+	testID: string;
+	value?: string;
+	options: TOptions;
+	preference: string;
+	onChangeValue: (pref: string, param: { [key: string]: string }, onError: () => void) => void;
+}
+
+const ListPicker = ({ preference, options, value, title, testID, onChangeValue }: IListPicker) => {
```
diegolmello, line 23: Missing storybook

diegolmello, line 23: This is a generic component.
What's `preference` on this context?
It should be some other name.

app/containers/List/ListPicker.tsx
```diff
@@ -0,0 +1,57 @@
...
+		right: option?.value === i.value ? () => <CustomIcon name={'check'} size={20} color={colors.tintActive} /> : undefined
+	}));
+
+	return (
+		<List.Item
+			title={title}
+			testID={testID}
+			onPress={() => showActionSheet({ options: actionSheetOptions })}
+			right={() => (
+				<Text style={[{ ...sharedStyles.textRegular, fontSize: 16 }, { color: colors.actionTintColor }]}>
```
diegolmello, line 47: `{ ...sharedStyles.textRegular, fontSize: 16 }` should have its own style rather than inline

app/containers/List/ListPicker.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+	const actionSheetOptions: TActionSheetOptionsItem[] = options[preference].map(i => ({
+		title: I18n.t(i.label, { defaultValue: i.label, second: i.second }),
+		onPress: () => {
+			hideActionSheet();
+			onChangeValue(preference, { [preference]: i.value.toString() }, () => setOption(option));
+			setOption(i);
+		},
+		right: option?.value === i.value ? () => <CustomIcon name={'check'} size={20} color={colors.tintActive} /> : undefined
+	}));
```
diegolmello, line 39: Besides this whole logic being hard to be reused because it makes use of  a `preference` context, you're re-executing the action sheet logic even when it's not needed (action sheet is closed).
Instead of that, you can create a `getOptions` function and call it when it's actually required. Example: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/17be449d4e81649ba2025a230b7a523e248e41be/app/containers/MessageActions/index.tsx#L459

```tsx
onPress={() => showActionSheet({ options: getOptions() })}
```

app/containers/List/ListPicker.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Text } from 'react-native';
+
+import { TActionSheetOptionsItem, useActionSheet } from '../ActionSheet';
+import { CustomIcon } from '../CustomIcon';
+import * as List from './index';
+import I18n from '../../i18n';
+import { useTheme } from '../../theme';
+import sharedStyles from '../../views/Styles';
+
+type TOption = { label: string; value: string; second?: string };
```
diegolmello, line 11: What's `second`?
Can you try a more semantic naming?

Edit: After reading through your PR, is something I already commented.
This component is supposed to be generic instead of couple to a logic.
Why would a `list picker` understand what's a `second` of a notification option?

app/i18n/locales/en.json
```diff
@@ -842,5 +842,7 @@
   "error-init-video-conf": "Error starting video call",
   "totp-invalid": "Code or password invalid",
   "Close_Chat": "Close Chat",
-  "Select_tags": "Select tags"
+  "Select_tags": "Select tags",
+  "Also_send_thread_message_to_channel_behavior": "Also send thread message to channel",
+  "Accounts_Default_User_Preferences_alsoSendThreadToChannel_Description": "Allow users to select the Also send to channel behavior"
```
diegolmello, line 847: Missing ptbr

app/views/RoomView/index.tsx
```diff
@@ -260,6 +260,7 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 			fname,
 			prid
 		};
+		console.log('🚀 ~ file: index.tsx ~ line 257 ~ RoomView ~ constructor ~ room', room);
```
diegolmello, line 263: Nope

e2e/helpers/app.js
```diff
@@ -222,6 +223,15 @@ async function prepareAndroid() {
 	await runCommand('adb shell settings put global animator_duration_scale 0.0');
 }
 
+async function expectToBeVisible(id) {
```
diegolmello, line 226: Don't do this.
This is error prone, since you avoid the synchronization of Detox.
Also, why would do this anyway?
The test should assert a specific scenario.
Why would it change depending on the value?

app/views/UserPreferencesView/index.tsx
```diff
@@ -69,9 +83,30 @@ const UserPreferencesView = ({ navigation }: IUserPreferencesViewProps): JSX.Ele
 						<List.Item
 							title='Enable_Message_Parser'
 							testID='preferences-view-enable-message-parser'
-							right={() => renderMessageParserSwitch(enableMessageParserEarlyAdoption)}
+							right={() => (
```
diegolmello, line 86: Is there something wrong here?
It's not related to the goal of this PR.

app/definitions/ISubscription.ts
```diff
@@ -35,7 +35,19 @@ export enum ERoomTypes {
 
 type RelationModified<T extends Model> = { fetch(): Promise<T[]> } & Relation<T>;
 
-export interface ISubscription {
+export interface ISubscriptionNotificationOptionsDraft {
```
diegolmello, line 38: Draft? Why draft?

diegolmello (changes requested):

app/i18n/locales/en.json
diegolmello: Still missing ptbr

app/lib/services/connect.ts
```diff
@@ -287,8 +289,9 @@ async function login(credentials: ICredentials, isFromWebView = false): Promise<
 			roles: result.me.roles,
 			avatarETag: result.me.avatarETag,
 			isFromWebView,
-			showMessageInMainThread: result.me.settings?.preferences?.showMessageInMainThread ?? true,
-			enableMessageParserEarlyAdoption
+			showMessageInMainThread,
+			enableMessageParserEarlyAdoption,
+			alsoSendThreadToChannel: result.me.settings?.preferences?.alsoSendThreadToChannel
```
diegolmello, line 294: This field is not reactive.
If you change it on other client, it's not reflecting the change on mobile.

diegolmello (changes requested):
Besides my comments, there's a visual bug the we should fix on this PR.
If the thread is already created and we long press the message, we're showing the `Reply in thread` as an action.
We could apply these rules of this PR to show tshow accordingly, but I think the best user experience would be to redirect the user to the thread.

- create a thread
- on the main room, long press any message from the thread (make sure to fix for both)
- tap reply in thread
- create a check to identify if the message is from a thread
- if it is, redirect to the thread
- if not, keep the current behavior to start a new thread

app/lib/services/connect.ts
reinaldonetof: Fixed

e2e/helpers/app.js
reinaldonetof: Deleted!

app/i18n/locales/en.json
reinaldonetof: Added

PR #4447, opened 2022-08-18
Chore: Refactor CustomEmoji to Typescript

liversonp:
Co-authored-by: guilhermedfs <guilhermedfs11@gmail.com>

## Proposed changes

Change emoji file from Javascript to Typescript

## Types of changes

## Checklist

### DIFF
--- a/app/lib/database/model/CustomEmoji.ts
+++ b/app/lib/database/model/CustomEmoji.ts
@@ -8,11 +8,11 @@ export const CUSTOM_EMOJIS_TABLE = 'custom_emojis';
 export default class CustomEmoji extends Model {
 	static table = CUSTOM_EMOJIS_TABLE;
 
-	@field('name') name;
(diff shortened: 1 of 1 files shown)


diegolmello (commented):
Hey, thanks for your contribution!
What's the benefit of migrating models to TS?
I don't think we would actually use these types on schemes or on queries.

What we use really is the `CUSTOM_EMOJIS_TABLE`, which is going to be part of operations and it was introduced here https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/3580

For the sake of migrating all files to Typescript, I'd say it's a good thing to migrate all models, but then it would require a different lint.
Otherwise it would be too cumbersome to type record on several files.

In other words, we could migrate all models, but without having to type `@field('name') name;` to `string | undefined`.
I'm ok with a simple ts-ignore for the whole file.

PR #4470, opened 2022-08-25
Chore: Create useDebounce hook

GleidsonDaniel:
## Proposed changes

Create useDebounce hook
Apply useDebounce hook where it is necessary

## How to test or reproduce

- Test the DiscussionsView and CannedResponsesListView search filter.

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/helpers/debounce.ts
+++ b/app/lib/methods/helpers/debounce.ts
@@ -24,6 +24,6 @@ export function debounce(func: Function, wait?: number, immediate?: boolean) {
 	return _debounce;
 }
 
-export function useDebounce(func: Function, wait?: number, immediate?: boolean): () => void {
+export function useDebounce(func: Function, wait?: number, immediate?: boolean): (...args: any[]) => void {
 	return useCallback(debounce(func, wait, immediate), []);
 }
--- a/app/lib/methods/helpers/debounce.ts
+++ b/app/lib/methods/helpers/debounce.ts
@@ -1,3 +1,5 @@
+import { useCallback } from 'react';
+
 export function debounce(func: Function, wait?: number, immediate?: boolean) {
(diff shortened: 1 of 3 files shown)


app/lib/methods/helpers/debounce.ts
```diff
@@ -21,3 +23,7 @@ export function debounce(func: Function, wait?: number, immediate?: boolean) {
 	_debounce.stop = () => clearTimeout(timeout!);
 	return _debounce;
 }
+
+export function useDebounce(func: Function, wait?: number, immediate?: boolean): (...args: any[]) => void {
```
diegolmello, line 27: We discussed this internally, but we should consider using a lib instead of creating our own.

My main concern is related to the useCallback usage and here's why https://github.com/xnimorz/use-debounce/blob/8020962aed8038c065f9468fd7cf1d9a1751ce62/src/useDebouncedCallback.ts#L150

diegolmello (changes requested):

app/lib/methods/helpers/debounce.ts
```diff
@@ -21,3 +23,7 @@ export function debounce(func: Function, wait?: number, immediate?: boolean) {
 	_debounce.stop = () => clearTimeout(timeout!);
 	return _debounce;
 }
+
+export function useDebounce(func: (...args: any) => any, wait?: number): (...args: any[]) => void {
```
reinaldonetof, line 27: IMHO, I think we can export `useDebounce` from libs/hooks, WDYT?

GleidsonDaniel: The question here is more to maintain debounce things on the same local.
If we change the useDebounce to hooks, we will have two files doing the same thing.

reinaldonetof (approved):

PR #4487, opened 2022-09-02
[FIX] Secured push notifications not fetching content on Android

diegolmello:
## Proposed changes

Documentation: https://docs.rocket.chat/guides/mobile-guides/push-notifications/push-notification-security

Every time we receive a push notification with this feature, we need to fetch the content before displaying the push notification.
We have a retry logic to keep trying to fetch the content in case of a failure with a limit of 5 tries.
After that, we display a generic push notification "You have a new message".

A part of the logic was created as static, so it would keep the value instantiated forever.
After the limit was reached, it would display the generic message "forever" (or until the app was force closed).

On this PR, I removed statics and started instantiating a new class for every push, so each push is going to have it's own retry context.

**Note:** I'm quite aware of a similar issue on iOS, but I was not able to pinpoint any issues on the logic. I kept this PR open for while in order to try another approach, but I couldn't.
I'm moving forward with it and, once we release it in production, we keep an eye on the issues and understand if the iOS issues are gone or not. It could be an issue on backend rate limiter, for example.

### How to make it fail before this PR
- Send lots of messages in order to receive so many push notifications that rate limiter is reached and starts to fail the requests
- Once reached, the static value mentioned above is going to be at max and any new push notifications is going to be displayed as the generic message

### The fix
- Try the same steps as "How to make it fail before this PR"
- Rate limiter is still going to be reached and it's still going to display the generic message
- After the 1 min of the rate limiter, new push notifications are going to start fetching the content again, like this

## Types of changes

## Checklist

## Further comments

[TC-74]

[TC-74]: https://rocketchat.atlassian.net/browse/TC-74?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


diegolmello:
> I just have a doubt if this is really the problem, or could be another problem. This task solves a problem, but is it the same one reported? Here's the question...

Do you mean because it's not an iOS fix?
Perhaps we should keep the task open and create a new one for this?

GleidsonDaniel:
> > I just have a doubt if this is really the problem, or could be another problem. This task solves a problem, but is it the same one reported? Here's the question...
> 
> Do you mean because it's not an iOS fix? Perhaps we should keep the task open and create a new one for this?

Not necessarily, yes because the is not fully replicable and what we replicate may not be the problem.
But I think we can close the old one, if someone complains, ask him to go through the step by step of how to reproduce.

Even closing this will force people to do this :)

PR #4533, opened 2022-09-16
[FIX] Header cutting Next text on SelectListView

GleidsonDaniel:
## Proposed changes

This PR solve the problem of the text cutting, but the problem is a deep more complex.
Today we use the react navigation to create our header and that implies that we use their spacing rules and etc, and this create some problems, example:
Spacing is not controllable, on this print you see:
Screen Shot 2022-09-16 at 10 19 50
The space enter title and buttons is not controllable, its fixed, and because of this the text is cutting, they don't have a correct wrap rule, if you add a padding or margin the text will break, because the size will not increase.

## Types of changes

## Checklist

## Further comments

Jira: [TC-54]

[TC-54]: https://rocketchat.atlassian.net/browse/TC-54?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/containers/HeaderButton/HeaderButtonItem.tsx
+++ b/app/containers/HeaderButton/HeaderButtonItem.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Platform, StyleSheet, Text } from 'react-native';
+import { Platform, StyleSheet, Text, ViewStyle } from 'react-native';
 import { PlatformPressable } from '@react-navigation/elements';
 
 import { CustomIcon, ICustomIcon, TIconsName } from '../CustomIcon';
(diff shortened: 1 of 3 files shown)


reinaldonetof (changes requested):
Can you try to do like the RoomView? Because if we increase the font size in system settings, will hide the "next"

Screen Shot 2022-09-16 at 12 38 46

Example of RoomVIew: 
Screen Shot 2022-09-16 at 12 39 11


WDYT?

PR #4537, opened 2022-09-16
Chore: Hooks app/views/LanguageView

reinaldonetof:
## Proposed changes

Migration to Hooks app/views/LanguageView

## Types of changes

## Checklist


GleidsonDaniel (changes requested):
Dude the migration is correct, however I think that some things could be improved in the code taking advantage of this migration.
For example, you only use two user properties and still get the entire reducer object.
Another thing is when selecting a new language, if I select my current language it shouldn't change :)

I think it would be worthwhile to "correct" these two points by taking advantage of this PR.

PR #4582, opened 2022-09-29
[NEW] Reply in direct

GleidsonDaniel:
## Proposed changes

Adds the option to reply to a message sent in a group or team via direct

## How to test or reproduce

- In a group or team, select a message and click "Reply in Direct Message".

## Types of changes

## Checklist

## Further comments

Jira: [TC-121]

[TC-121]: https://rocketchat.atlassian.net/browse/TC-121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reinaldonetof (approved):

app/containers/MessageActions/index.tsx
```diff
@@ -237,6 +239,23 @@ const MessageActions = React.memo(
 				replyInit(message, false);
 			};
 
+			const handleReplyInDM = async (message: TAnyMessageModel) => {
+				if (message.u.username) {
```
diegolmello, line 243: We should be a little bit more careful about this here.
I'm not sure what could happen if there's no `username` if the message is from some sort of bot or idk.

app/containers/MessageActions/index.tsx
```diff
@@ -237,6 +239,23 @@ const MessageActions = React.memo(
...
+					const result = await Services.createDirectMessage(message.u.username);
+					if (result.success) {
```
diegolmello, line 245: Did we test what happens if the DM already exists?

app/containers/MessageActions/index.tsx
```diff
@@ -237,6 +239,23 @@ const MessageActions = React.memo(
...
+						const { room } = result;
+						const params = {
+							rid: room.rid,
+							name: getRoomTitle(room),
+							t: room.t,
+							roomUserId: getUidDirectMessage(room),
+							replyInDM: message
+						};
+						Navigation.replace('RoomView', params);
```
diegolmello, line 254: We should try to use `goRoom`

app/containers/MessageActions/index.tsx
reinaldonetof: Yeap! Works properly, the user is redirected to the correct DM.

app/containers/MessageActions/index.tsx
reinaldonetof: The `goRoom` uses the `Navigation.navigate`, this navigation's type doesn't work when it goes to the "same" view. Also, this handleInDM introduces that behavior that we commented on some time ago, about the navigation between chats.

app/containers/MessageActions/index.tsx
reinaldonetof: Sure, I added the `optional chain operator`

app/containers/MessageActions/index.tsx
diegolmello: `goRoom` accepts a `navigationMethod` param.
However remember we're going to deprecate nested rooms soon and we'll have to analyze this case again to use `goRoom` anyway.

Issue #4178, opened 2022-05-08
Animated GIFs from Photos on iOS upload as JPGs and are static.


fc757:
We've been having an issue with animated gifs and Rocket.Chat. If you upload a photo in the iOS Rocket.Chat application from Photos, it uploads as a JPG and is not animated.

If you save the GIF from Photos to your iCloud drive and upload it from Files it is uploaded as a GIF and it animates properly.

If you use Safari on iOS and upload from Photos, it animates as expected and is uploaded a GIF. The aspect ratio for images is also a lot better in Safari as opposed to the App. See below:

![safari](https://user-images.githubusercontent.com/4654695/167313324-f41b5f55-1068-45a5-8758-c16942474517.jpg)
![ios](https://user-images.githubusercontent.com/4654695/167313326-7ee4a32d-6683-4c41-a689-771c26ea2ea4.jpg)

Please let me know if there is any information I can provide to help troubleshoot and hopefully resolve the issue.

stefanandres:
I've noticed the same. This actually was working before the code rebuild for the ".heic pictures are not being shown on non-iOS devices" feature happened.

PR #4668, opened 2022-11-09
[RKT-17] Animated GIFs from Photos on iOS upload as JPGs and are static. #4178

gitstart:
## Proposed changes

1- GIF images was forced changed to JPG, set bool to `forceJpg: false`, also library 'react-native-image-crop-picker' was saving .gif files as .jpg in tmp directory so instead of using tmp directory url for uploading .gif now using original sourcePath.
2- List was not displaying .gif and audios uploaded from iOS properly it was showing these as Reply so check the issue and some how iOS uploaded audio and gif files do not having `audio_url` and `image_url` so check for `title_url` and fixed issue.
## Issue(s)	

[#4178](https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4178)

## How to test or reproduce

1- Open App on iOS device, click on + sign inside a Room and send a .gif image.
2- Image will appear static instead of animated.

## Types of changes

## Checklist

## Further comments

Bug video: https://www.loom.com/share/4998f29fbde74e388e480501bc9fa8d4
After Fix Video: https://www.loom.com/share/fbfe787430ba4f91afe8992098544253


app/containers/MessageBox/index.tsx
```diff
@@ -76,7 +76,7 @@ const libraryPickerConfig: Options = {
 	multiple: true,
 	compressVideoPreset: 'Passthrough',
 	mediaType: 'any',
-	forceJpg: true
+	forceJpg: false
```
diegolmello, line 79: This way you would fix the issue you're trying to fix, but stop converting images to jpg.
It would be a regression.

diegolmello (changes requested):

app/containers/MessageBox/index.tsx
gitstart: `forceJpg` flag set to `false` to stop converting images to `.jpg`, previously it was set to `true` that's why `.gif` images was converted to `.jpg` and was displayed as static images.

diegolmello: It doesn't change the fact that we would still have issues with `.HEIC` rendering on the web if we don't convert to `.jpg`.
That's why I said it's a regression.

gitstart: image

HEIC image displayed on web correctly, is there any other issue, please let me know. Please check attach image
Also if we just change extension of HEIC image would it work as expected on web?
@diegolmello

diegolmello: Ok. Thanks, we're discussing this internally.
We have a few issues rendering HEIC depending on the OS/browser and the client how uploaded it.

PR #4669, opened 2022-11-09
[FEAT] Encode audio

GleidsonDaniel:
## Proposed changes

Changes the sending of default audios to send the audios encoded as "default" mp3.
Before, the audios were sent as m4a which has a better audio quality, but it doesn't have universal support (This was even the reason for the audios to have a problem on iOS in the past).
It uses FFmpeg as a base (open source) to encode the audios to mp3.

## How to test or reproduce

Send an audio on a normal channel and try to listen to it on a device of another brand or browser.
Same test as above, but sending in a channel that redirects messages to whatsapp.

## Types of changes

## Checklist

## Further comments

This solution adds a little more processing to the sending of audios, it's not that much, but there is an increase.
Testing on a 2 min audio there was a delay of 1.5 sec more than when sent without the encoder.

This solution is not the "best" solution, because if another service needed to receive an audio in a different encode, we should send this encode, adding another layer, which in reality should be done by the backend, but it works!

I tested with other libraries too, none of them do the encode as simple mp3.
https://github.com/goodatlas/react-native-audio-record/blob/master/index.d.ts
https://github.com/hyochan/react-native-audio-recorder-player/blob/main/index.ts#L49

Jira: [NATIVE-114]

[NATIVE-114]: https://rocketchat.atlassian.net/browse/NATIVE-114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -11,7 +11,7 @@ buildscript {
 
     ext {
         buildToolsVersion = "31.0.0"
-        minSdkVersion = 23
+        minSdkVersion = 24
         compileSdkVersion = 31
         targetSdkVersion = 31
         if (System.properties['os.arch'] == "aarch64") {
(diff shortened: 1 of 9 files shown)


GleidsonDaniel:
Waiting for the backend staff to see other possible solutions.

PR #4708, opened 2022-11-25
[IMPROVE] Disable Jitsi native module on Android

GleidsonDaniel:
## Proposed changes

Instead of running the Jitsi module via the native module, let's run it via the webview.

It was necessary to make some changes to the code since we "disabled" jitsi in the android app.
- The permissions were those that had in jitsi were "embedded" in the app, need to add explicitly.
- As the module was only deactivated on android, it was necessary to create a separate "module" for the android implementation, since when assembling the bundle it looks at all the imports.
- Added the intent lib, basically an alias for native android functions. Open intent, check intent, etc.

## How to test or reproduce

- Join and leave a call using an android device. (Try several times and in different rooms.)
- Test the integration with the Jitsi app (by installing the jitsi app on the device and seeing if the call is redirected to it)
- Test the call features, turn camera and microphone on or off, audio and image output. 
- Test on a server running a version lower than version 5.0

## Types of changes

## Checklist

## Further comments

Jira: [VC-37]

[VC-37]: https://rocketchat.atlassian.net/browse/VC-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/JitsiMeetView.android.tsx
```diff
@@ -0,0 +1,144 @@
...
+			isAppInstalled(JITSI_INTENT).then(function (isInstalled) {
+				if (!isInstalled) {
+					PermissionsAndroid.requestMultiple([
+						PermissionsAndroid.PERMISSIONS.CAMERA,
+						PermissionsAndroid.PERMISSIONS.RECORD_AUDIO
+					]);
+					return;
+				}
+				navigation.pop();
+				openAppWithData(JITSI_INTENT, route.params?.url);
```
diegolmello, line 75: Isn't possible to use this? https://reactnative.dev/docs/linking#sendintent-android

diegolmello (changes requested):

diegolmello (approved):

android/app/src/main/AndroidManifest.xml
```diff
@@ -69,5 +76,10 @@
             </intent-filter>
         </activity>
     </application>
-
+    <queries>
```
diegolmello, line 79: Remove `android:host` pointing to Jitsi above on this file to test if it fixes the crash

diegolmello (changes requested):

ankar84:
I tested:
- Call in Jitsi Meet app
- Deleted Jitsi Meet app and test WebView
- Few calls in a row in different chats
- ScreenSharing from PC
- Rotate smartphone during call
- mute/unmute, turn and off camera

And all tests was successful!
So, that improvement is looks good to me!
One thing - on call start a white screen appears on smartphone for 5-10 seconds (like loading of call). But maybe it's related to our Jitsi deployment.

diegolmello (approved):

PR #4723, opened 2022-12-01
fix: link preview without embed image

### LINKED ISSUE #4482: Link preview and Autolinker not working in mobile client.
### Description:

The behavior of embedded link previews is not the same between the regular browser client and the official mobile (Android/iPhone) client. What I’m seeing is that when I paste a directly-linked image (ending in .jpg or similar), the browser client will display a preview for it properly. In the mobile client, there will be no preview shown.

The Autolinker function is similarly inconsistent, and changing the Regex field doesn’t seem to have any effect. A good example is direct links for https://i.redd.it/ images. These are hyperlinked and previewed properly on the desktop client, no issues there. However, on the mobile client these links are NOT hyperlinked and no preview is shown. I’ve tried editing the RegEx field in Settings to account for i.redd.it links, nothing I do seems to make a difference. I’ve tried restarting the RocketChat instance after changing it, no difference.

SOME links do get a preview properly in the mobile client, but not direct image links. If I link http://dilbert.com/, for example, that gets a preview in the mobile client just fine. YouTube videos get an embed preview as well.

I'm mainly testing on my Android phone, but I've done a few side tests with an iPhone and run into the same problem. I have also tested with the browser client through Chrome on the Android phone, and it behaves as I would expect.

### Environment Information:

- Rocket.Chat Server Version: 4.8.3
- Rocket.Chat App Version: 4.29
- Device Name: Pixel 6 Pro
- OS Version: Android 13

### Steps to reproduce:

1. Go to a channel and post a direct url to an image. In my example, https://i.redd.it/t2mul61342l91.jpg
2. In the browser client, observe that the url is hyperlinked and a link preview is shown.
3. In the mobile client, observe that the url is bare, no hyperlink, no link preview shown.

### Expected behavior:

I should expect link preview and autolinker behavior to be consistent between the browser and mobile clients.

reinaldonetof:
## Proposed changes

The behavior of embedded link previews is not the same between the regular browser client and the official mobile (Android/iPhone) client. What I’m seeing is that when I paste a directly-linked image (ending in .jpg or similar), the browser client will display a preview for it properly. In the mobile client, there will be no preview shown.

The Autolinker function is similarly inconsistent, and changing the Regex field doesn’t seem to have any effect. A good example is direct links for https://i.redd.it/ images. These are hyperlinked and previewed properly on the desktop client, no issues there. However, on the mobile client these links are NOT hyperlinked and no preview is shown. I’ve tried editing the RegEx field in Settings to account for [i.redd.it](http://i.redd.it/) links, nothing I do seems to make a difference. I’ve tried restarting the RocketChat instance after changing it, no difference.

SOME links do get a preview properly in the mobile client, but not direct image links. If I link [Welcome to Dilbert](http://dilbert.com/), for example, that gets a preview in the mobile client just fine. YouTube videos get an embed preview as well.

I'm mainly testing on my Android phone, but I've done a few side tests with an iPhone and run into the same problem. I have also tested with the browser client through Chrome on the Android phone, and it behaves as I would expect.

## Issue(s)	

Closes #4482 

## How to test or reproduce

- Go to a channel and post a direct url to an image. In my example, https://i.redd.it/t2mul61342l91.jpg
- In the browser client, observe that the url is hyperlinked and a link preview is shown.
- In the mobile client, observe that the url is bare, no hyperlink, no link preview shown.

## Screenshots

| Before | After |
| --- | --- |
| image | image |

## Types of changes

## Checklist

## Further comments

[TC-165]

[TC-165]: https://rocketchat.atlassian.net/browse/TC-165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/message/Urls.tsx
```diff
@@ -123,8 +136,8 @@ const Url = React.memo(
 				background={Touchable.Ripple(themes[theme].bannerBackground)}
 			>
 				<>
-					<UrlImage image={url.image} />
-					<UrlContent title={url.title} description={url.description} theme={theme} />
+					<UrlImage image={imageUrl} />
```
diegolmello, line 139: It's not applying bottom rounded borders

diegolmello (changes requested):
I'm also missing tests

GleidsonDaniel (changes requested):
Won't create test for the file?

> app/lib/methods/helpers/image.ts

GleidsonDaniel (approved):

GleidsonDaniel (approved):

app/containers/message/Urls.tsx
```diff
@@ -108,8 +108,16 @@ const Url = React.memo(
...
+					if (isIOS) {
+						const result = await isImageURL(url.url);
+						setIsImageUrlFromParamUrl(result);
+					} else {
+						Image.getSize(
+							url.url,
+							() => setIsImageUrlFromParamUrl(true),
+							() => setIsImageUrlFromParamUrl(false)
+						);
+					}
```
reinaldonetof, line 120: The function isImageURL is returning some difference between platforms, as example the URL https://play.google.com/apps/test/RQVpXLytHNc/ahAETVdb4gnoB4cOV8_fm8D0Vkswqyp2fluIarrP5O3OpKyh-oAUEiuNV6zfOvGHuI3crF0LIvu5Bh26-Boz0eF0KP :

Android <image src="https://user-images.githubusercontent.com/47038980/214647799-e258bede-2c06-4a0c-8f2f-6e54f0d87824.png" width="200" /> iOS <image src="https://user-images.githubusercontent.com/47038980/214647823-7164b144-c1ea-4764-a498-9d77835ea0f9.png" width="200" />

It should not show any line or preview

Using the Image.getSize for Android, I was able to reproduce the same behavior for Android and iOS.

Also, both platforms aren't able to render any image with .svg format
example: https://global-uploads.webflow.com/611a19b9853b7414a0f6b3f6/6141e37299de7d330c3699c2_graduation-cap%201.svg

app/lib/methods/helpers/image.ts
```diff
@@ -0,0 +1,10 @@
+import { Image } from 'react-native';
+
+export const isImageURL = async (url: string) => {
+	try {
+		const result = await Image.prefetch(url);
```
diegolmello, line 5: We can't do this, because prefetch downloads the image locally to be used later.
You're just using the return to understand if it's an image or not.
That's wrong, because we would be always persisting images on `RN's Image` lib to be rendered on `fast-image`. We would download the same image twice.
Also, since this persist happens on `RN's Image`, `clear cache` won't be able to clean it.

app/containers/message/Urls.tsx
```diff
@@ -96,6 +103,26 @@ const UrlContent = React.memo(
...
+		const [isImageUrlFromParamUrl, setIsImageUrlFromParamUrl] = useState(false);
+
+		useEffect(() => {
+			if (!url.image && url.url) {
+				const testImageUrl = async () => {
+					if (isIOS) {
+						const result = await isImageURL(url.url);
+						setIsImageUrlFromParamUrl(result);
+					} else {
+						Image.getSize(
```
diegolmello, line 115: Same thing. It's going to download twice, wrong place, etc.

app/containers/message/Urls.tsx
```diff
@@ -52,15 +53,21 @@ const styles = StyleSheet.create({
 });
 
 const UrlImage = React.memo(
-	({ image }: { image: string }) => {
+	({ image, hasContent }: { image: string; hasContent: boolean }) => {
```
diegolmello, line 56: This is awkward.
We already have a param `image` we check to understand if there's content or not.
You're adding yet another one to render a border radius?
It looks very suspicious...

You could merge the components.

app/containers/message/Urls.tsx
diegolmello: SVG's can be rendered with expo-image apparently https://docs.expo.dev/versions/unversioned/sdk/image
Different scope and we should give it a try on another task

diegolmello (changes requested):
An idea: all your solutions are based on downloading the image to display it afterwards.
What if you actually try to render it and check `onLoadEnd`?
You could start the render without a size and resize it later.

app/containers/message/Urls.tsx
```diff
@@ -48,21 +48,51 @@ const styles = StyleSheet.create({
...
+	imageWithoutContent: {
+		borderRadius: 4
+	},
+	activityPosition: {
+		zIndex: 2,
+		position: 'absolute'
+	},
+	loadingState: {
+		height: 0,
+		borderWidth: 0
```
reinaldonetof, line 61: I refactor the whole PR to add the possibility of showing nothing during loading, instead appears something loading and then disappears because of an error from  onError function

https://user-images.githubusercontent.com/47038980/219213894-de4ef541-52bc-4b0e-8300-a00f76871387.mov

diegolmello: @reinaldonetof I don't understand what I'm supposed to be looking for on this 3s video.

reinaldonetof: before this refactoring of the component, during the loading state, the `Touchable` (https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4723/files/249a056b7dd0a3b991be58e87b7ed14eb08d7952#diff-b1ec2afd1adf1d4430ea9e5090e7107f68f598bc493ca059ba5b38dade0c97fbR146) was rendering a blank background and was affecting the UX. Now I hide this Touchable while it's loading

app/containers/message/Urls.tsx
```diff
@@ -48,21 +48,51 @@ const styles = StyleSheet.create({
...
+					source={{ uri: image }}
+					style={[styles.image, style]}
+					resizeMode={FastImage.resizeMode.cover}
+					onError={() => setImageLoadedState('error')}
+					onLoad={() => setImageLoadedState('done')}
+				/>
+			</>
+		);
 	},
-	(prevProps, nextProps) => prevProps.image === nextProps.image
```
diegolmello, line 65: This test was unnecessary before this PR, since `image` is primitive and the only prop, so it would work without it.
If your PR is merged, there's no reason to use memo anymore, so you should remove it.

app/containers/message/Urls.tsx
```diff
@@ -48,21 +48,51 @@ const styles = StyleSheet.create({
...
 		}
 
 		image = image.includes('http') ? image : `${baseUrl}/${image}?rc_uid=${user.id}&rc_token=${user.token}`;
-		return <FastImage source={{ uri: image }} style={styles.image} resizeMode={FastImage.resizeMode.cover} />;
+
+		return (
+			<>
+				<FastImage
+					source={{ uri: image }}
+					style={[styles.image, style]}
```
diegolmello, line 87: Both components are on the same file.
Why do you need to send style as a prop?
This is what's breaking memo.

app/containers/message/Urls.tsx
```diff
@@ -48,21 +48,51 @@ const styles = StyleSheet.create({
...
 		borderTopRightRadius: 4
+	},
+	imageWithoutContent: {
+		borderRadius: 4
+	},
+	activityPosition: {
+		zIndex: 2,
+		position: 'absolute'
+	},
+	loadingState: {
```
diegolmello, line 59: Why `loadingState` and not `loading`?

diegolmello (changes requested):
Since you're going to work on this, can you record a better video to show the new behavior?

app/containers/message/Urls.tsx
reinaldonetof: Instead of pass the style or pass hasContent and imageLoadedState to the UrlImage, I merged the UrlImage inside the URL, as you talked here: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4723#discussion_r1107538174

app/containers/message/Urls.tsx
```diff
@@ -118,13 +123,22 @@ const Url = React.memo(
...
+					{image ? (
+						<FastImage
+							source={{ uri: image }}
+							style={[styles.image, !hasContent && styles.imageWithoutContent, imageLoadedState === 'loading' && styles.loading]}
+							resizeMode={FastImage.resizeMode.cover}
+							onError={() => setImageLoadedState('error')}
+							onLoad={() => setImageLoadedState('done')}
+						/>
+					) : null}
+					{hasContent ? <UrlContent title={url.title} description={url.description} theme={theme} /> : null}
```
GleidsonDaniel, line 141:
```suggestion
					{hasContent ? <UrlContent title={url.title} description={url.description} /> : null}
```
remove theme

GleidsonDaniel: Updates the PR and send it to QA.

GleidsonDaniel (changes requested):

GleidsonDaniel (approved):

Issue #4733, opened 2022-12-06
Room header doesn't show selected icon


phriedrich:
### Description:

Room header doesn't show selected icon but it's working correctly in the sidebar.

### Environment Information:

- Rocket.Chat Server Version: 5.4.0
- Rocket.Chat App Version: 4.33.0.36480
- Device Name: iPad Mini 6, iPhone SE2
- OS Version: iPadOS 16.1.1, iOS 16.1.2

### Steps to reproduce:

1. Open Room with set icon

### Expected behavior:

Room should show the selected icon in the header next to the name like it does with users.

### Actual behavior:

Icon is not shown instead the generic colored tile with a letter is used.

<img width="643" alt="Missing-Avatar" src="https://user-images.githubusercontent.com/470153/205914442-9675a4d6-7870-4b3b-8f92-d6dcff23cf93.png">

diegolmello:
Possible regression after #4702 @reinaldonetof

Issue #2655, opened 2020-11-19
iOS: Updated new avatar not refreshed in hamburger menu when user relogin


ramdeveloperbh:
### Environment Information:

- Rocket.Chat Server Version: 3.7.1
- Rocket.Chat App Version: 4.12.1
- Device Name: iPhone 11
- OS Version: 13.6.1

### Steps to reproduce:

1. Login the mobile app.
2. Choose the profile option from hamburger menu.
3. Upload/Update new avatar in Profile screen.
4. Save Changes.
5. Update avatar will refresh in hamburger menu, Profile screen and Direct/Group chat conversations.
6. Choose the logout option from hamburger menu.
7. Re-login the mobile app.
8. Check avatar in hamburger menu, Profile screen and Direct/Group chat conversations.

### Expected behavior:

Updated new avatar should refresh in hamburger menu, Profile screen and Direct/Group chat conversations.

### Actual behavior:

Updated new avatar not refreshed in hamburger menu.

<img width="1655" alt="Screen Shot 2020-11-19 at 1 17 31 PM" src="https://user-images.githubusercontent.com/74710223/99653233-d34e7880-2a69-11eb-956e-780eb88b7490.png">

MrWillTurner:
+1

nmplcpimenta:
I'd like to take on this issue; if anyone has anymore information, I'd be thankful =)

PR #4746, opened 2022-12-14
feat: Change Avatar View

reinaldonetof:
## Proposed changes

This PR introduces a new screen `ChangeAvatarView`. This view could be navigated from `ProfileView`, `RoomInfoView`, and `RoomInfoEditView`.

When we are talking about the navigation from `ProfileView` we aim to change the user's avatar. During the tests, I noticed a known issue [NATIVE-200] `User's avatar is not being updated throughout the app` and decided to fix it inside this PR too. Because if the user wasn't able to see the new avatar on mobile after he changed it, we couldn't validate properly the improvement.

Another behavior implemented here was the capability to change the room's avatar by the user that has that permission. So, to edit it, we can navigate from `RoomInfoView` and `RoomInfoEditView`. Also, was noticed another issue that if the user changes the room's avatar to an uploaded image, then later he back and "deletes" the uploaded image, in other places that appear this room's avatar, will keep the same uploaded image.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/2655
Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4733

## Screenshots

https://user-images.githubusercontent.com/47038980/207864622-a6c9c9d6-dca0-417c-8764-f143ab294ece.mov

## Types of changes

## Checklist

## Further comments

[TC-25]
[TC-45]
[NATIVE-200]

[TC-25]: https://rocketchat.atlassian.net/browse/TC-25?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[TC-45]: https://rocketchat.atlassian.net/browse/TC-45?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NATIVE-200]: https://rocketchat.atlassian.net/browse/NATIVE-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/Avatar/index.tsx
```diff
@@ -21,19 +25,24 @@ const AvatarContainer = ({
 	onPress,
 	getCustomEmoji,
 	isStatic,
-	rid
-}: IAvatar): React.ReactElement => {
+	rid,
+	handleEdit,
+	isUserProfile
+}: IAvatar & { isUserProfile?: boolean; handleEdit?: () => void }): React.ReactElement => {
```
diegolmello, line 31: It looks like you should update IAvatar

app/containers/Avatar/index.tsx
```diff
@@ -74,7 +83,7 @@ const AvatarContainer = ({
 	};
 
 	useEffect(() => {
-		if (!avatarETag) {
+		if (!avatarETag || avatarETag !== previousAvatarEtag.current) {
```
diegolmello, line 86: Can you explain the lifecycle of these effects?
I didn't get how we have a query subscribe, but yet we need two effects to trigger it again.

app/containers/Avatar/index.tsx
```diff
@@ -84,28 +93,44 @@ const AvatarContainer = ({
...
+				getCustomEmoji={getCustomEmoji}
+				isStatic={isStatic}
+				rid={rid}
+				blockUnauthenticatedAccess={blockUnauthenticatedAccess}
+				externalProviderUrl={externalProviderUrl}
+				avatarETag={isUserProfile ? avatarETagUser : avatarETag}
+				serverVersion={serverVersion}
+			/>
+			{handleEdit ? (
+				<Button
```
diegolmello, line 124: This button is very different from Figma

app/containers/Avatar/index.tsx
```diff
@@ -21,19 +25,24 @@ const AvatarContainer = ({
...
 
 	const server = useSelector((state: IApplicationState) => state.share.server.server || state.server.server);
 	const serverVersion = useSelector((state: IApplicationState) => state.share.server.version || state.server.version);
-	const { id, token } = useSelector(
+	const { id, token, avatarETagUser } = useSelector(
 		(state: IApplicationState) => ({
 			id: getUserSelector(state).id,
-			token: getUserSelector(state).token
+			token: getUserSelector(state).token,
+			avatarETagUser: getUserSelector(state).avatarETag
```
diegolmello, line 45: This feels wrong.
`Avatar` is a generic component to render user images.

`id` and `token` are necessary to make requests.
The introduction of `ETag` for the logged in user is wrong, because this way we're going to end up having `avatarETag` on the state coming from a query and `avatarETagUser` coming from Redux (and why doesn't it need to be a ref?).
And then you need to `avatarETag={isUserProfile ? avatarETagUser : avatarETag}`.

It would be a lot more elegant to separate this into a custom hook.

app/containers/Avatar/index.tsx
diegolmello, line 31: Don't you know from params and redux when the avatar is from the logged in user?

diegolmello: It also creates an insane prop drilling

app/stacks/InsideStack.tsx
```diff
@@ -96,6 +97,7 @@ const ChatsStackNavigator = () => {
 			<ChatsStack.Screen name='SelectListView' component={SelectListView} options={SelectListView.navigationOptions} />
 			<ChatsStack.Screen name='RoomInfoView' component={RoomInfoView} options={RoomInfoView.navigationOptions} />
 			<ChatsStack.Screen name='RoomInfoEditView' component={RoomInfoEditView} options={RoomInfoEditView.navigationOptions} />
+			<ProfileStack.Screen name='ChangeAvatarView' component={ChangeAvatarView} />
```
diegolmello, line 100: Copy/Paste issues...

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+}: {
+	item?: IAvatar;
+	testID?: string;
+	onPress: (value?: IAvatar) => void;
+	text?: string;
+}) => {
+	const { colors } = useTheme();
+
+	return (
+		<TouchableOpacity
```
diegolmello, line 25: `Touch`

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,81 @@
...
+import styles from './styles';
+import { useTheme } from '../../theme';
+
+const Item = ({
+	item,
+	onPress,
+	text,
+	testID
+}: {
+	item?: IAvatar;
```
diegolmello, line 17: Are you sure something that comes straight from the rest api is going to have the exact same interface of IAvatar?

app/i18n/locales/en.json
```diff
@@ -876,5 +876,13 @@
   "Call": "Call",
   "Reply_in_direct_message": "Reply in Direct Message",
   "room_archived": "archived room",
-  "room_unarchived": "unarchived room"
+  "room_unarchived": "unarchived room",
```
diegolmello, line 879: Missing ptbr

app/views/ChangeAvatarView/AvatarUrl.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import I18n from '../../i18n';
+import { FormTextInput } from '../../containers/TextInput';
+import { useDebounce, isImage, isValidURLRequest, isValidURL } from '../../lib/methods/helpers';
+
+const AvatarUrl = ({ submit }: { submit: (value: string) => void }) => {
+	const handleChangeText = useDebounce(async (value: string) => {
+		if (isImage(value) && isValidURL(value)) {
+			const result = await isValidURLRequest(value);
```
diegolmello, line 10: Why do you need this?

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+			}
+			if (textAvatar) {
+				return handleError(e, 'resetAvatar', 'changing_avatar');
+			}
+			return handleError(e, 'setAvatarFromService', 'changing_avatar');
+		}
+	};
+
+	const resetAvatar = () => {
+		setAvatar(undefined);
```
diegolmello, line 106: Bad practice.
You're mixing null and undefined usage.
Stick to null here by initializing your useState with null.

app/views/NewMessageView/index.tsx
```diff
@@ -93,6 +93,7 @@ const NewMessageView = () => {
 
 					return (
 						<UserItem
+							id={item._id}
```
diegolmello, line 96: `_id`? Where does this come from?
It looks like it should be normalized to use `id` as everywhere else.

diegolmello (changes requested):

app/views/ChangeAvatarView/AvatarUrl.tsx
reinaldonetof: I'm testing if it's a valid URL, before changing the textURL and in sequence forcing the FastImage to render the URL

app/containers/Avatar/index.tsx
reinaldonetof: Actually, I created a interface IAvatarContainer

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+const Item = ({ item, onPress, text, testID }: { item?: IAvatar; testID?: string; onPress: Function; text?: string }) => {
+	const { colors } = useTheme();
+
+	return (
+		<Touch
+			key={item?.service}
+			testID={testID}
+			onPress={() => onPress(item)}
+			style={[styles.avatarButton, { backgroundColor: colors.borderColor }]}
```
diegolmello, line 20: `avatarButton` for an `Item` component feels not so good.
If you create a single component per file and have one StyleSheet for each file, you'll it's going to be easier to create meaningful style names.

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,72 @@
...
+import Touch from '../../containers/Touch';
+
+const Item = ({ item, onPress, text, testID }: { item?: IAvatar; testID?: string; onPress: Function; text?: string }) => {
+	const { colors } = useTheme();
+
+	return (
+		<Touch
+			key={item?.service}
+			testID={testID}
+			onPress={() => onPress(item)}
```
diegolmello, line 19: `Avatar` has its own `onPress`. Do we need this extra one?

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,72 @@
...
+		</Touch>
+	);
+};
+const AvatarSuggestion = ({
+	onPress,
+	user,
+	resetAvatar
+}: {
+	onPress: (value: IAvatar | null) => void;
+	user?: IUser;
```
diegolmello, line 32: Only primitives as props as much as possible.

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,72 @@
...
+				service,
+				contentType
+			};
+		});
+		setAvatarSuggestions(suggestions);
+	};
+
+	useEffect(() => {
+		getAvatarSuggestion();
+	}, []);
```
diegolmello, line 55: Use `useEndpointData` instead

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+	return (
+		<View style={{ flex: 1 }}>
+			<Text style={[styles.itemLabel, { color: colors.titleText }]}>{I18n.t('Images_uploaded')}</Text>
+			<View style={styles.containerAvatarSuggestion}>
+				{user?.username && resetAvatar ? (
+					<Item text={`@${user.username}`} testID={`reset-avatar-suggestion`} onPress={resetAvatar} />
+				) : null}
+				{avatarSuggestions.slice(0, 7).map(item => (
```
diegolmello, line 64: Why are you limiting to 7?

app/views/ChangeAvatarView/AvatarSuggestion.tsx
```diff
@@ -0,0 +1,72 @@
...
+		});
+		setAvatarSuggestions(suggestions);
+	};
+
+	useEffect(() => {
+		getAvatarSuggestion();
+	}, []);
+
+	return (
+		<View style={{ flex: 1 }}>
```
diegolmello, line 58: Can we create a style?

app/views/ChangeAvatarView/AvatarUrl.tsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+import I18n from '../../i18n';
+import { FormTextInput } from '../../containers/TextInput';
+import { useDebounce, isImage, isValidURLRequest, isValidURL } from '../../lib/methods/helpers';
+
+const AvatarUrl = ({ submit }: { submit: (value: string) => void }) => {
```
diegolmello, line 7: Can we type `submit` properly?
```
{fromUser ? <AvatarUrl submit={value => setAvatar({ url: value, data: value, service: 'url' })} /> : null}
```

app/views/ChangeAvatarView/AvatarUrl.tsx
```diff
@@ -0,0 +1,28 @@
...
+	const handleChangeText = useDebounce(async (value: string) => {
```
diegolmello, line 8: There's no reason to debounce this.
You're doing it for URL and not for a search, so what's most likely to happen is to fail 99% of the time.
You added a `submit` prop for an `onChangeText` event.
It's awkward, because it changes the semantic.

What you should do is to validate on a form submit.

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+import log from '../../lib/methods/helpers/log';
+
+const ChangeAvatarView = () => {
+	const [avatar, setAvatarState] = useState<IAvatar | null>(null);
+
+	const [textAvatar, setTextAvatar] = useState('');
+	const [saving, setSaving] = useState(false);
+	const { colors } = useTheme();
+	const { user, serverVersion } = useAppSelector(state => ({
+		user: getUserSelector(state),
```
diegolmello, line 36: Non primitive. What do you actually want from user here?

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+		isMasterDetail: state.app.isMasterDetail,
```
diegolmello, line 37: It looks like you're not even using this one.

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+				// Change User's Avatar
+				await Services.setAvatarFromService(avatar);
+			} else if (textAvatar) {
+				// Change User's Avatar
+				await Services.resetAvatar(user.id);
+			}
+			setSaving(false);
+			avatarUrl.current = '';
+			return navigation.goBack();
+		} catch (e) {
```
diegolmello, line 92: This feels very hard to read/maintain.
What if you keep this submit as the main method that calls the appropriate endpoint?
Inside each method you can have its own `try/catch` throwing the message you want.

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,195 @@
...
+	const [textAvatar, setTextAvatar] = useState('');
+	const [saving, setSaving] = useState(false);
+	const { colors } = useTheme();
+	const { user, serverVersion } = useAppSelector(state => ({
+		user: getUserSelector(state),
+		isMasterDetail: state.app.isMasterDetail,
+		serverVersion: state.server.version
+	}));
+
+	const avatarUrl = useRef<string | undefined>('');
```
diegolmello, line 41: Why `undefined` if it's default `''`?
I didn't get why do you need this ref.
Can you explain?

app/views/ChangeAvatarView/styles.ts
```diff
@@ -0,0 +1,36 @@
+import { StyleSheet } from 'react-native';
+
+import sharedStyles from '../Styles';
+
+export default StyleSheet.create({
+	disabled: {
```
diegolmello, line 6: Where are you using this?

app/views/ProfileView/index.tsx
```diff
@@ -334,39 +271,13 @@ class ProfileView extends React.Component<IProfileViewProps, IProfileViewState>
...
+			handleError(e, 'resetAvatar', 'changing_avatar');
 		}
 	};
 
-	pickImageWithURL = (avatarUrl: string) => {
-		logEvent(events.PROFILE_PICK_AVATAR_WITH_URL);
-		this.setAvatar({ url: avatarUrl, data: avatarUrl, service: 'url' });
+	handleEditAvatar = () => {
+		const { navigation } = this.props;
+		navigation.navigate('ChangeAvatarView', { fromUser: true });
```
diegolmello, line 280: We talked about this `fromUser` prop.
Perhaps you could make it more dynamic while increasing semantic?

Think like this:
`ChangeAvatarView` is a screen that behaves differently based on the context, because it can be called from `ProfileView`, `RoomInfoView` and `RoomInfoEditView`.
You created `fromUser` to make it behave differently for `ProfileView`.
What if you create TS types to specify the context?

```
type TChangeAvatarViewContext = 'profile' | 'room';
```

Add it to the screen's props and then TS is going to autocomplete it for you.

diegolmello (changes requested):

app/lib/services/restApi.ts
```diff
@@ -553,7 +553,7 @@ export const saveRoomSettings = (
 	rid: string,
 	params: {
 		roomName?: string;
-		roomAvatar?: string;
+		roomAvatar?: string | null;
```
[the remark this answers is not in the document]
reinaldonetof: Yeap, I need to send null to "delete" the previous avatar for rooms

app/views/ChangeAvatarView/AvatarSuggestion.tsx
reinaldonetof: getAvatarSuggestion uses methodCallWrapper

app/views/ChangeAvatarView/AvatarSuggestion.tsx
reinaldonetof: Screen Shot 2023-01-10 at 00 20 41

app/lib/methods/helpers/url.ts
```diff
@@ -13,3 +15,12 @@ export const isValidURL = (url: string): boolean => {
 
 // Use checkUseSsl: false only if server url starts with http://
 export const isSsl = (url: string): boolean => !/http:\/\//.test(url);
+
+export const isValidURLRequest = async (url: string) => {
```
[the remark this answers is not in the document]
reinaldonetof: I removed this fetch. If the avatar URL doesn't exist, the server returns an error. So, we don't need to check this from mobile

app/lib/methods/helpers/info.ts
```diff
@@ -2,6 +2,16 @@ import { Alert } from 'react-native';
 
 import I18n from '../../../i18n';
 
+export const handleError = (e: any, _func: string, action: string) => {
```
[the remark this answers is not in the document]
reinaldonetof: I created both tasks inside the Native Client board

app/lib/methods/helpers/index.ts
```diff
@@ -14,3 +14,4 @@ export * from './server';
 export * from './url';
 export * from './isValidEmail';
 export * from './random';
+export * from './image';
```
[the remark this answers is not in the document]
reinaldonetof: I created it in helpers/index because I'm considering this PR too https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4723

app/views/ChangeAvatarView/AvatarUrl.tsx
reinaldonetof: I think it's properly typed, isn't it?

app/views/ChangeAvatarView/index.tsx
reinaldonetof: I typed for string now! About the ref, I need it to understand when something changed from before to show properly the discard alert  or how to show correctly the image when you are trying to reset the room avatar https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/3db96db70c25b4defb04291602edc3f4337c2afe/app/views/ChangeAvatarView/index.tsx#L176

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+	const { userId, username, serverVersion } = useAppSelector(
+		state => ({
+			userId: getUserSelector(state).id,
+			username: getUserSelector(state).username,
+			serverVersion: state.server.version
+		}),
+		shallowEqual
+	);
+
+	const avatarUrl = useRef<string>('');
```
diegolmello, line 47: `isFormDirty`

diegolmello (changes requested):

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+	};
+
+	const resetAvatar = () => {
+		setAvatar(null);
+		setTextAvatar(`@${username}`);
+		avatarUrl.current = `@${username}`;
+	};
+
+	const resetRoomAvatar = () => {
+		setAvatar({ data: null });
```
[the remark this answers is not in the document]
diegolmello: `useReducer` maybe?

app/views/ProfileView/index.tsx
```diff
@@ -338,35 +275,19 @@ class ProfileView extends React.Component<IProfileViewProps, IProfileViewState>
 		}
 	};
 
-	pickImage = async () => {
-		const { Accounts_AllowUserAvatarChange } = this.props;
-
-		if (!Accounts_AllowUserAvatarChange) {
-			return;
+	handleError = (e: any, _func: string, action: string) => {
```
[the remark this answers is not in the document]
diegolmello: Or maybe you could do it. It won't hurt.

app/containers/Avatar/AvatarWithEdit.tsx
```diff
@@ -0,0 +1,84 @@
...
+			/>
+			{handleEdit && serverVersion && compareServerVersion(serverVersion, 'greaterThanOrEqualTo', '3.6.0') ? (
+				<Button
+					title={I18n.t('Edit')}
+					type='secondary'
+					backgroundColor={colors.editAndUploadButtonAvatar}
+					onPress={handleEdit}
+					testID='avatar-edit-button'
+					style={styles.editAvatarButton}
+					styleText={[styles.textButton]}
```
diegolmello, line 75: Unnecessary array

app/containers/Avatar/AvatarWithEdit.tsx
```diff
@@ -0,0 +1,84 @@
...
+	editAvatarButton: {
+		marginTop: 8,
+		paddingVertical: 8,
+		paddingHorizontal: 12,
+		marginBottom: 0,
+		height: undefined
+	},
+	textButton: {
+		fontSize: 12,
+		fontWeight: '600'
```
diegolmello, line 23: We don't use weight directly, since it works different on iOS and Android.
You might be looking for `textSemibold`

app/lib/methods/helpers/image.ts
```diff
@@ -0,0 +1,10 @@
+import { Image } from 'react-native';
+
+export const isImageURL = async (url: string) => {
```
diegolmello, line 3: I'm reviewing from GH only, so Idk if TS is smart enough to infer the return.
In case it doesn't, you should do it yourself, like this.

```suggestion
export const isImageURL = async (url: string): boolean => {
```

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+		dispatch(action);
+	};
+
+	const submit = async () => {
+		try {
+			setSaving(true);
+			if (context === 'room' && room?.rid) {
+				// Change Rooms Avatar
+				await changeRoomsAvatar(room.rid, state?.data);
+			} else if (state?.url) {
```
diegolmello, line 114: We had a long discussion about unifying this, because this is not running only for url. Yet here it is still.
What changed?

app/views/ChangeAvatarView/submitHelpers.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const resetUserAvatar = async (userId: string) => {
+	try {
+		await Services.resetAvatar(userId);
+	} catch (e) {
+		return handleError(e, 'changing_avatar');
+	}
+};
+
+const handleError = (e: any, action: string) => {
```
diegolmello, line 31: This can be considered a helper. Every other function is a service.

app/views/ChangeAvatarView/index.tsx
```diff
@@ -0,0 +1,237 @@
...
+			dispatchAvatar({
+				type: AvatarStateActions.CHANGE_AVATAR,
+				payload: { url: response.path, data: `data:image/jpeg;base64,${response.data}`, service: 'upload' }
+			});
+		} catch (error) {
+			log(error);
+		}
+	};
+
+	const deletingRoomAvatar = context === 'room' && state.data === null ? {} : { rid: room?.rid };
```
diegolmello, line 152: We talked about using Avatar component rather than container to remove this logic. What changed?

I'm missing a clear reasoning from you about what was discussed on your PRs.
If it can't be done, please say so as a note for the reviewer.
Otherwise it looks like I'm just wasting time repeating same reviews over and over.

app/views/ProfileView/index.tsx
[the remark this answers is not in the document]
diegolmello: Are you not doing this?

diegolmello (changes requested):

app/views/ProfileView/index.tsx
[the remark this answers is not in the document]
reinaldonetof: You sure, I can fix it!

app/containers/Avatar/AvatarWithEdit.tsx
reinaldonetof: Sure, you are right! I needed to fix the type in button interface

app/views/ChangeAvatarView/index.tsx
reinaldonetof: But the state changed from `avatar` and `textAvatar` to one state with all the props needed. We need this `state?.url` just to confirm that exist the URL inside the state. WDYT?

app/views/ChangeAvatarView/index.tsx
reinaldonetof: Yeap, sorry, that was missing explain why I didn't use the AvatarPresentational. So I didn't use it because I needed to get another param inside this View, for example, the `server`, this param is passed from  AvatarContainer passes to Avatar

reinaldonetof: I changed and pushed a new commit with this change too

diegolmello (approved):

diegolmello (approved):
After approved from QA, don't forget to merge fork changes, apply correct branch and update pods before merging this PR.

metro.config.js
```diff
@@ -8,6 +8,10 @@
 // eslint-disable-next-line import/no-extraneous-dependencies
 const blocklist = require('metro-config/src/defaults/exclusionList');
 
+const defaultSourceExts = require('metro-config/src/defaults/defaults').sourceExts;
+
+console.log('🚀 ~ file: metro.config.js:27 ~ process.env.MOCK_E2E_TEST:', process.env.MOCK_E2E_TEST);
```
reinaldonetof, line 13: Remove it later, I'm testing what value is set

package.json
```diff
@@ -26,7 +26,8 @@
 		"e2e:ios-build-debug": "yarn detox build -c ios.sim.debug",
 		"e2e:ios-test-debug": "yarn detox test -c ios.sim.debug",
 		"e2e:ios-build": "yarn detox build -c ios.sim.release",
-		"e2e:ios-test": "yarn detox test -c ios.sim.release"
+		"e2e:ios-test": "yarn detox test -c ios.sim.release",
+		"e2e:start": "RUNNING_E2E_TESTS=true npx react-native start"
```
GleidsonDaniel, line 30:
```suggestion
		"e2e:start": "RUNNING_E2E_TESTS=true npx react-native start --reset-cache"
```

GleidsonDaniel (approved):

PR #4788, opened 2022-12-30
[FIX] Modal freezing app after return from the background

reinaldonetof:
## Proposed changes

This issue was found the first time when the user opened the announcement and the announcement's text is a link. Then the user pressed the link and should open the web browser outside of the app. This behavior was forcing the app to go to the background, then when the user resume the app again and tried to close the modal, the app was freezing. This is a known behavior when using a modal as you can see here: https://github.com/facebook/react-native/issues/32504

However, to do a workaround in this case, we aren't re-render the RoomView at the exact time that the app goes to the background, although we are forced to close the modal when resuming the app again and calling the observable again to the list's query.

## How to test or reproduce

- Create a channel
- Add an announcement with only a link: e.g: [GitHub: Let’s build from here](https://github.com/)
- Click on the announcement 
- Click on the link
- Go back to the app

#### Before

https://user-images.githubusercontent.com/47038980/210114139-0409fdb7-104b-4378-84db-7624019dd932.mov

https://user-images.githubusercontent.com/47038980/210114187-9c48be61-5b87-40fc-920f-9d528b4a566c.mov

#### After
https://user-images.githubusercontent.com/47038980/210114070-379784b0-a9ff-4ec3-8de2-440f1f2f3b92.mov

## Types of changes

## Checklist

## Further comments

[TC-340]

### DIFF
--- a/app/views/RoomView/index.tsx
+++ b/app/views/RoomView/index.tsx
@@ -235,6 +235,7 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 		cancel: () => void;
 	};
 	private sub?: RoomClass;
+	private listObservableQuery = false;
 
 	constructor(props: IRoomViewProps) {
 		super(props);
@@ -359,7 +360,8 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 			if (appState === 'background' && showAnnouncementModal) {
 				// Prevent app freezing while announcement's modal is open and then the user click on a link from announcement
 				// or minimize the app then resume it
-				this.setState({ showAnnouncementModal: false }, () => this.keepListObservableWorking());
+				this.listObservableQuery = true;
+				this.setState({ showAnnouncementModal: false });
 				return false;
 			}
 			return true;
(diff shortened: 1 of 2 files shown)


GleidsonDaniel (changes requested):
Wouldn't it be simpler to just disappear the modal before opening the link?

reinaldonetof:
> Wouldn't it be simpler to just disappear the modal before opening the link?

yeap, but you need to keep in mind that if you minimize the app, then reopen it again, will freeze the app too. So, the flow to freeze the app isn't only when clicking on the link.

app/views/RoomView/index.tsx
```diff
@@ -347,12 +351,19 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
...
 	shouldComponentUpdate(nextProps: IRoomViewProps, nextState: IRoomViewState) {
 		const { state } = this;
-		const { roomUpdate, member, isOnHold } = state;
+		const { roomUpdate, member, isOnHold, showAnnouncementModal } = state;
 		const { appState, theme, insets, route } = this.props;
 		if (theme !== nextProps.theme) {
 			return true;
 		}
 		if (appState !== nextProps.appState) {
+			if (appState === 'background' && showAnnouncementModal) {
```
diegolmello, line 360: We shouldn't do this.
It's going to be a mess to remove it later.

What if you just hide the modal before going background?

diegolmello (changes requested):

reinaldonetof:
The pr https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4795 is fixing this issue

Issue #1408, opened 2019-11-21
Unable to send messages after hitting share on user profile


wreiske:
- Your Rocket.Chat Experimental app version: 1.24.0.12769
- Your Rocket.Chat server version: 2.3.0-develop
- Device (or Simulator) you're running with: Samsung Galaxy Tab S5e 

![20191120_231444](https://user-images.githubusercontent.com/6295044/69304026-1d014a80-0bed-11ea-977b-3b4b31d3475d.jpg)
![20191120_231222](https://user-images.githubusercontent.com/6295044/69304037-25f21c00-0bed-11ea-8189-7794943a80f8.jpg)
![20191120_231144](https://user-images.githubusercontent.com/6295044/69304050-30141a80-0bed-11ea-9fb5-ad7a7fe8d17c.jpg)
![20191120_231044](https://user-images.githubusercontent.com/6295044/69304061-399d8280-0bed-11ea-9dff-30f4ba884c1c.jpg)
![20191120_231021](https://user-images.githubusercontent.com/6295044/69304069-428e5400-0bed-11ea-9bb3-5a37a989b0ce.jpg)
![Screenshot_20191120-231548](https://user-images.githubusercontent.com/6295044/69304085-546ff700-0bed-11ea-997b-01e2e206f2d4.jpg)

PR #4795, opened 2023-01-04
[FIX] Modal freezing app after return from the background

reinaldonetof:
## Proposed changes

This issue was found the first time when the user opened the announcement and the announcement's text is a link. Then the user pressed the link and should open the web browser outside of the app. This behavior was forcing the app to go to the background, then when the user resume the app again and tried to close the modal, the app was freezing.

During some tests, I noticed that is only happening on iOS and we still need to keep the `list.query` to resolve this bug that still occurs on Android https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1410

## How to test or reproduce

You need to test properly for iOS and Android

1 - For both, open the announcement inside any Room
2 - Then, minimize the app or if the announcement is a link, click on it
3 - After the app goes to background, resume it again
4 - You should be able to dismiss the announcement modal
5 - Then, after dismissing the announcement, do a long press in any message, the actions will open
6 - Click to Share, and use the Rocket.Chat to share it, share the message to any room
7 - Then, back again to the app, you should be in the same room as before and be able to send and receive messages, and the message list must be refreshing correctly

The steps from 5 - 7 you can compare them with this Issue https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/1408

We are testing this, to avoid any regression.

## Types of changes

## Checklist

## Further comments

[TC-340]

[TC-340]: https://rocketchat.atlassian.net/browse/TC-340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/RoomView/index.tsx
+++ b/app/views/RoomView/index.tsx
@@ -388,7 +388,7 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 
 		if (appState === 'foreground' && appState !== prevProps.appState && this.rid) {
 			// Fire List.query() just to keep observables working
-			if (this.list && this.list.current) {
+			if (this.list && this.list.current && !isIOS) {
 				this.list.current?.query();
 			}
 		}


diegolmello (approved):
I'm only missing a clear test plan so we don't have a regression to https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/1410

PR #4801, opened 2023-01-11
[FIX] Messages not loading on some edge cases

diegolmello:
## Proposed changes

Fixes a use case where the app would fail to persist data on local db without adding a new loader item ("load more").
It could be reproduced by scrolling through messages history while receiving multiple messages on the same room.
That would make `LoadMore` component to mount -> unmount -> mount, which would call `useEffect` twice and create a race condition during load.

I fixed by adding the load logic inside a saga channel (queue), since we don't have a better option to deal with message loading atm.
`LoadMore` dispatches the redux action to the saga queue, which calls them in sequence.
If the same loader was called more than once, it's ignored.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4793

## How to test or reproduce

- Go to a room with enough messages to trigger pagination
- Receive messages while scrolling through history
- Eventually the app would reach a point where it can't load more messages, but it should

It's easily reproducible by running these two scripts at the same time.

```sh
#!/bin/bash
for i in {1..1500}
do
  curl --location --request POST 'http://localhost:3000/api/v1/chat.postMessage' \
		--header 'content-type: application/json' \
		--header 'x-auth-token: <<TOKEN>>' \
		--header 'x-user-id: <<USERID>>' \
		--data-raw '{
				"roomId": "<<RID>>",
				"msg": '$i'
		}'
	sleep 0.3
done
```

```sh
#!/bin/bash
while true
do
  adb shell input swipe 800 350 800 1000 50
	sleep 0.5
done
```

## Types of changes

## Checklist

## Further comments

[NATIVE-223]

[NATIVE-223]: https://rocketchat.atlassian.net/browse/NATIVE-223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/actions/actionsTypes.ts
+++ b/app/actions/actionsTypes.ts
@@ -30,7 +30,7 @@ export const ROOM = createRequestTypes('ROOM', [
 	'FORWARD',
 	'USER_TYPING',
 	'HISTORY_REQUEST',
-	'HISTORY_FINISHED' // TODO: I'm not sure about this naming, but I don't see value on handling failure atm
+	'HISTORY_FINISHED'
 ]);
 export const INQUIRY = createRequestTypes('INQUIRY', [
 	...defaultTypes,
--- a/app/actions/actionsTypes.ts
+++ b/app/actions/actionsTypes.ts
@@ -28,7 +28,10 @@ export const ROOM = createRequestTypes('ROOM', [
 	'DELETE',
 	'REMOVED',
 	'FORWARD',
-	'USER_TYPING'
+	'USER_TYPING',
+	'HISTORY_REQUEST',
+	'HISTORY_SUCCESS',
+	'HISTORY_FAILURE'
 ]);
 export const INQUIRY = createRequestTypes('INQUIRY', [
 	...defaultTypes,
--- a/app/actions/actionsTypes.ts
+++ b/app/actions/actionsTypes.ts
@@ -30,8 +30,7 @@ export const ROOM = createRequestTypes('ROOM', [
 	'FORWARD',
 	'USER_TYPING',
 	'HISTORY_REQUEST',
-	'HISTORY_SUCCESS',
-	'HISTORY_FAILURE'
+	'HISTORY_FINISHED' // TODO: I'm not sure about this naming, but I don't see value on handling failure atm
 ]);
 export const INQUIRY = createRequestTypes('INQUIRY', [
 	...defaultTypes,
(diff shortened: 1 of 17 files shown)


app/actions/actionsTypes.ts
```diff
@@ -28,7 +28,9 @@ export const ROOM = createRequestTypes('ROOM', [
 	'DELETE',
 	'REMOVED',
 	'FORWARD',
-	'USER_TYPING'
+	'USER_TYPING',
+	'HISTORY_REQUEST',
+	'HISTORY_FINISHED'
```
diegolmello, line 33: This naming is anti-pattern, but I don't see any value adding success/failure atm

app/views/RoomView/LoadMore/index.tsx
```diff
@@ -20,53 +23,50 @@ const styles = StyleSheet.create({
...
 		}
-	}, []);
 
-	let text = 'Load_More';
-	if (type === MessageTypeLoad.NEXT_CHUNK) {
-		text = 'Load_Newer';
+		return (
+			<Touch onPress={handleLoad} style={styles.button} enabled={!loading}>
+				{loading ? (
+					<ActivityIndicator color={themes[theme].auxiliaryText} />
```
GleidsonDaniel, line 63:
```suggestion
					<ActivityIndicator color={colors.auxiliaryText} />
```

app/views/RoomView/services/getMessages.ts
```diff
@@ -1,23 +1,10 @@
...
-	latest?: Date;
-	lastOpen?: Date;
-	loaderItem?: any; // TODO: type this
-}): Promise<void> => {
+const getMessages = ({ rid, t, lastOpen }: { rid: string; t?: string; lastOpen?: Date }): Promise<void> => {
 	if (lastOpen) {
 		return loadMissedMessages({ rid, lastOpen });
 	}
-	return loadMessagesForRoom({ rid, t: t as any, latest, loaderItem });
+	return loadMessagesForRoom({ rid, t: t as any });
```
GleidsonDaniel, line 7:
```suggestion
	return loadMessagesForRoom({ rid, t: t as RoomType });
```

app/views/RoomView/index.tsx
```diff
@@ -1337,7 +1327,9 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 		if (item.t && MESSAGE_TYPE_ANY_LOAD.includes(item.t as MessageTypeLoad)) {
 			content = (
 				<LoadMore
-					load={() => this.onLoadMoreMessages(item)}
+					rid={room.rid}
+					t={room.t as any}
```
GleidsonDaniel, line 1331:
```suggestion
					t={room.t as RoomType}
```

GleidsonDaniel (approved):

Issue #4806, opened 2023-01-12
Separators don't render correctly if font size a11y is decreased


diegolmello:
#### Overview

It happens mostly on Android.

Check how some separators aren’t getting rendered below.
![image](https://user-images.githubusercontent.com/804994/212089241-d51b303d-4b2b-4ce9-a38d-f0a31fdd4f5c.png)
![image](https://user-images.githubusercontent.com/804994/212089255-3f5ddd04-523d-42b5-a907-5e0a0e8e9e82.png)

That happens because StyleSheet.hairlineWidth calculates its value based on PixelRatio.

#### To be done

- Stop using StyleSheet.hairlineWidth and apply 1 instead
- Evaluate all places impacted by this and apply the new value

romiljain5:
hi @diegolmello,
I have raised the MR for this issue please check at https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4849

Gokulsankar-21:
@diegolmello i would like to work on this issue. Could you please assign me to resolve

lakshaylalia:
@diegolmello Is this issue still open?

PR #4849, opened 2023-01-23
#4806-change-value-for-seperator

romiljain5:
- change value to 1 instead of using pixelRatio

## Types of changes

## Checklist


diegolmello:
Lint is failing

romiljain5:
it is failing for return types, which is not caused by my changes I guess, @diegolmello

diegolmello:
image

GleidsonDaniel:
This solution is actually creating a problem.
Because before we had a parameterizable fixed value, it was just changing a variable that would change everywhere, now you are putting a fixed value everywhere.
If in a while we had to change to 1.5 for example, we would have to change several files, instead of changing only one.

diegolmello:
It doesn't fix the problem.
I've opened an issue on React Native repo https://github.com/facebook/react-native/issues/36408

PR #4366, opened 2022-07-14
[IMPROVE] Add emoji picker to iOS

reinaldonetof:
## Proposed changes

Use the same keyboard layout for both Andorid and iOS. Also, add the emoji picker on keyboard.

This task is dependent on the merge of the PR https://github.com/RocketChat/react-native-ui-lib/pull/6

## How to test or reproduce

- Open any room
- Click on the "icon" button to open the emoji picker

#### After

Screen Shot 2022-07-14 at 16 08 12

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4864, opened 2023-01-25
[FIX] "Allow Reaction" does not work properly on "Read Only" rooms

reinaldonetof:
## Proposed changes

When the user is in a Read Only room when he does a long press, the message action doesn't appear and he isn't able to reply directly to the person who sent a message before or react to it with an emoji, or copy the permalink. This is a regression from https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4366

## How to test or reproduce

In any channel, with a user that has the appropriate permissions

- Make it a “Read Only” room
- And activate the “Allow Rection” option
- Send a few messages and react to a couple.
- With another user, without “workaround” permissions” (like being an all-powerful admin or owner)
- Join the room and try reacting to a message

#### Before

https://user-images.githubusercontent.com/47038980/214669583-96a82c78-ab4c-4c82-8883-056feaf427bf.mov

#### After
https://user-images.githubusercontent.com/47038980/214668099-e98bc579-cdd6-41c3-a0a6-e92218fdb2d4.mov

## Types of changes

## Checklist

## Further comments

[TC-326]

[TC-326]: https://rocketchat.atlassian.net/browse/TC-326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


GleidsonDaniel (approved):

app/views/RoomView/index.tsx
```diff
@@ -785,13 +785,22 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
...
 
+	handleCloseEmoji = (action?: Function, params?: any) => {
+		if (this.messagebox?.current) {
+			return this.messagebox?.current.closeEmojiAndAction(action, params);
+		}
+		if (action) {
+			return action(params);
+		}
+	};
+
```
reinaldonetof, line 796: I needed to do this differentiation because there is the possibility that the message box isn't rendered, for example, when the room is `read only`.

So, in that case, besides the possibility to reactions on long press, the user must be able to click on the `emoji +` to add new emoji or long press on the reactions.

PR #4883, opened 2023-02-02
[Feat] Brings the operation of the video call closer to the web

GleidsonDaniel:
## Types of changes

## Checklist


app/lib/hooks/useVideoConf.tsx
```diff
@@ -0,0 +1,98 @@
...
+export const useVideoConf = (rid: string): { showInitCallActionSheet: () => Promise<void>; showCallOption: boolean } => {
+	const [showCallOption, setShowCallOption] = useState(false);
+
+	const serverVersion = useAppSelector(state => state.server.version);
+	const jitsiEnabled = useAppSelector(state => state.settings.Jitsi_Enabled);
+	const jitsiEnableTeams = useAppSelector(state => state.settings.Jitsi_Enable_Teams);
+	const jitsiEnableChannels = useAppSelector(state => state.settings.Jitsi_Enable_Channels);
+	const user = useAppSelector(state => getUserSelector(state));
+
+	const greaterThanFive = compareServerVersion(serverVersion, 'greaterThanOrEqualTo', '5.0.0');
```
diegolmello, line 36: Optional, but the naming doesn't tell the whole story imo.
`greater than five` isn't tied to server so it might cause confusion for someone who never touched the code. Like, five what? It's also incorrect, since the logic is actually `greaterThanOrEqualTo` instead of `greaterThan`.
Thinking of `isServer5OrNewer` or something like that.

We could also come up with a different solution for the future, to make these kinds of tests more clear.

app/views/RoomInfoView/components/UserInfoIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { CustomIcon, TIconsName } from '../../../containers/CustomIcon';
+import styles from '../styles';
+import { useTheme } from '../../../theme';
+import { useVideoConf } from '../../../lib/hooks/useVideoConf';
+import i18n from '../../../i18n';
+import { useAppSelector } from '../../../lib/hooks';
+import { compareServerVersion } from '../../../lib/methods/helpers';
+
+// TODO: change other icons on future
```
diegolmello, line 13: Can you create tasks to do it and remove the comment?
I just read a comment like this asking to remove some logic after we update to RN 0.65.
We're on 0.68.6 and the comment lost meaning 😂

app/views/RoomInfoView/components/UserInfoIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+export function UserInfoIcon({
```
diegolmello, line 14: I find this `*Icon` naming awkward as well.
Ok, it has an icon, but it really is a button, because the goal for the component is to be pressed. Not to merely display something.

It's also odd, because this can happen:
```tsx
<UserInfoIcon showIcon={false} />
```

An icon that has no icon.

app/views/RoomInfoView/components/UserInfoIcon.tsx
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function CallIcon({ rid, isDirect }: { rid: string; isDirect: boolean }): React.ReactElement | null {
+	const { showCallOption, showInitCallActionSheet } = useVideoConf(rid);
+	const serverVersion = useAppSelector(state => state.server.version);
+	const greaterThanFive = compareServerVersion(serverVersion, 'greaterThanOrEqualTo', '5.0.0');
+
+	const showIcon = greaterThanFive ? showCallOption : showCallOption && isDirect;
+
+	return <UserInfoIcon onPress={showInitCallActionSheet} iconName='phone' label={i18n.t('Call')} showIcon={showIcon} />;
```
diegolmello, line 47: Do we need to reuse or could we do something like this?

```tsx
const { showCallOption, showInitCallActionSheet } = useVideoConf(rid);
const serverVersion = useAppSelector(state => state.server.version);
const color = danger ? colors.dangerColor : colors.actionTintColor;
const greaterThanFive = compareServerVersion(serverVersion, 'greaterThanOrEqualTo', '5.0.0');

if (greaterThanFive ? showCallOption : showCallOption && isDirect) {
	return (
		<BorderlessButton>
			...
		</BorderlessButton>
	);
}

return null;
```

Looks a lot easier to read, but there's also that `danger` prop that is never used.
I don't know if you're seeking to use it in the future.
But then you should move it to a separate file, so I'm not sure.

app/views/RoomView/RightButtons.tsx
```diff
@@ -374,6 +370,9 @@ class RightButtonsContainer extends Component<IRightButtonsProps, IRigthButtonsS
 						badge={() => <HeaderButton.BadgeUnread tunread={tunread} tunreadUser={tunreadUser} tunreadGroup={tunreadGroup} />}
 					/>
 				) : null}
+				{/* {isTeamRoom({ teamId, joined }) ? (
```
diegolmello, line 373: I think we can remove this.

app/views/RoomView/components/HeaderCallRightIcon.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+import * as HeaderButton from '../../../containers/HeaderButton';
+import { useVideoConf } from '../../../lib/hooks/useVideoConf';
+
+export default function HeaderCallIcon({ rid }: { rid: string }): React.ReactElement | null {
```
diegolmello, line 6: I see you use a separate file to use hooks, but can you call it `HeaderCallButton`?
You'll be using it on `RightButtons` component.
Just to use the same naming.

app/views/RoomView/index.tsx
```diff
@@ -621,18 +621,20 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 			),
 			headerRight: () => (
 				<RightButtons
-					rid={rid}
+					rid={rid!}
```
diegolmello, line 624: No bangs.
This might break the app on tablets, for example, since it renders RoomView without `rid`.

app/views/RoomView/index.tsx
```diff
@@ -621,18 +621,20 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
...
 					status={room.status}
 					omnichannelPermissions={omnichannelPermissions}
-					t={this.t || t}
+					t={(this.t || t) as SubscriptionType}
 					encrypted={encrypted}
 					navigation={navigation}
 					toggleFollowThread={this.toggleFollowThread}
 					showActionSheet={this.showActionSheet}
 					departmentId={departmentId}
+					name={avatar!}
```
diegolmello, line 636: same

app/containers/UIKit/VideoConferenceBlock/components/StartACallActionSheet.tsx
```diff
@@ -43,25 +42,27 @@ export default function CallAgainActionSheet({ rid }: { rid: string }): React.Re
...
 					</Touchable>
 					<Touchable
-						onPress={() => setPhone(!phone)}
-						style={[style.iconCallContainer, phone && style.enabledBackground]}
+						onPress={() => setMic(!mic)}
+						style={[style.iconCallContainer, mic && style.enabledBackground]}
 						hitSlop={BUTTON_HIT_SLOP}
 					>
-						<CustomIcon name={phone ? 'microphone' : 'microphone-disabled'} size={16} color={handleColor(phone)} />
+						<CustomIcon name={mic ? 'microphone' : 'microphone-disabled'} size={16} color={handleColor(mic)} />
```
diegolmello, line 56: Not related to this PR, but these icons are 24 on Figma.

diegolmello (changes requested):

app/views/RoomInfoView/components/UserInfoIcon.tsx
GleidsonDaniel: The button will be used for the other buttons on the screen.

app/containers/UIKit/VideoConferenceBlock/components/StartACallActionSheet.tsx
GleidsonDaniel: Dude, I'm sure I put 16 in the past for a reason, but I don't remember, I think in some daily we discussed it.
I think we can leave it like that hahaha

diegolmello (approved):

Issue #3231, opened 2021-06-21
Links do not work if protocol is not set in URL prefix


aleadco:
### Description:

Links do not work if protocol is not set in URL prefix

When you send a link in example.com format the address is highlighted and the hyperlink is created but when you try to click it does not react.

I think that's what was meant in #3226

### Environment Information:

- Rocket.Chat Server Version: 3.15
- Rocket.Chat App Version: 4.17
- Device Name: Xiaomi - Samsung
- OS Version: Android 8 and 9

### Steps to reproduce:

1. Send link in example.com format
2. Click
3. Does nothing

### Expected behavior:

Open the link

### Actual behavior:

Hyperlink function does not react

### Additional context:

The links sent with the established protocol without working correctly for example http://example.com

RcleydsonR:
Is this bug still happening ?

PR #4899, opened 2023-02-09
[FIX] Links do not work if protocol is not set in url prefix

reinaldonetof:
## Proposed changes

This PR is to do some tweaks and generate the build for https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4632

Within the screenshot's topic you can see that if the user presses the link that the URL didn't exist, the browser will try to open it anyway, this behavior is the same as the Web. We can validate if the link exists by doing a pre-fetch before opening it. 

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/3231
Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4632

## How to test or reproduce

- Send link in example.com format
- Click
- Does nothing

- If you want to test with a link with `http`: `http://web.simmons.edu/~grovesd/comm244/notes/week2/links`

#### Before

https://user-images.githubusercontent.com/47038980/217679399-c91a0482-28f5-47af-a127-453c11c61e27.mov

#### After

https://user-images.githubusercontent.com/47038980/217678998-9c88e895-5592-4700-839a-766be94fc3b3.mov

## Types of changes

## Checklist

## Further comments

[TC-395]

[TC-395]: https://rocketchat.atlassian.net/browse/TC-395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/methods/helpers/openLink.ts
```diff
@@ -36,10 +36,17 @@ const appSchemeURL = (url: string, browser: string): string => {
 	return schemeUrl;
 };
 
+const addProtocol = (url: string): string => {
```
GleidsonDaniel, line 39: Adds a test for this function:
http://www.google.com
https://www.google.com
www.google.com

GleidsonDaniel (changes requested):

app/lib/methods/helpers/openLink.ts
reinaldonetof: Sure!

app/lib/methods/helpers/addProtocol.ts
```diff
@@ -0,0 +1,8 @@
+const addProtocol = (url: string): string => {
+	if (!url.toLowerCase().startsWith('http')) {
+		return `https://${url.replace('//', '')}`;
```
diegolmello, line 3: How you know if there's a certificate on the URL?
This is going to fail on localhost envs, for example.

diegolmello (changes requested):

app/lib/methods/helpers/addProtocol.ts
reinaldonetof: the server didn't return the localhost as a link if the user writes only `localhost:3000`, is it expected to return as a link, or should I be able to treat this behavior too?
Screen Shot 2023-02-16 at 13 57 14

app/lib/methods/helpers/addProtocol.test.ts
```diff
@@ -0,0 +1,20 @@
+import addProtocol from './addProtocol';
+
+describe('Add the protocol https at the begin of the URL', () => {
+	it('return the link as original when sent with https at the begin', () => {
+		const linkHttps = 'https://www.google.com';
+		expect(addProtocol(linkHttps)).toBe(linkHttps);
```
reinaldonetof, line 6: > How you know if there's a certificate on the URL?

app/lib/methods/helpers/addProtocol.ts
```diff
@@ -1,3 +1,5 @@
+// If the link does not have the protocol at the beginning, we are inserting https as the default,
+// since by conversion the most used is the secure protocol, with the same behavior as the web.
```
diegolmello, line 2: "by conversion"? Do you mean "convention"?

app/lib/methods/helpers/addProtocol.ts
diegolmello: I don't understand what you're trying to do. Can you elaborate?

app/lib/methods/helpers/openLink.ts
```diff
@@ -37,9 +38,9 @@ const appSchemeURL = (url: string, browser: string): string => {
 };
 
 const openLink = async (url: string, theme: TSupportedThemes = 'light'): Promise<void> => {
+	url = addProtocol(url);
```
diegolmello, line 41: Any reason not to add this to `try/catch`?

app/lib/methods/helpers/addProtocol.test.ts
```diff
@@ -0,0 +1,20 @@
...
+	it('return the link as original when sent with http at the begin', () => {
+		const linkHttp = 'http://www.google.com';
+		expect(addProtocol(linkHttp)).toBe(linkHttp);
+	});
+	it("return a new link with protocol at the begin when there isn't the protocol at the begin", () => {
+		const linkWithoutProtocol = 'www.google.com';
+		expect(addProtocol(linkWithoutProtocol)).toBe('https://www.google.com');
+	});
+	it('return the link correctly when the original starts with double slash, because the server is returning that', () => {
+		const linkWithDoubleSlashAtBegin = '//www.google.com';
```
diegolmello, line 17: Is this a thing? Do people try this?

app/lib/methods/helpers/addProtocol.ts
```diff
@@ -0,0 +1,10 @@
...
+const addProtocol = (url: string): string => {
```
diegolmello, line 3: `ensureSecureProtocol`

diegolmello (changes requested):

app/lib/methods/helpers/openLink.ts
reinaldonetof: Because at line 58 (https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/fb47f8221fd7bfb6bd9703873d255c435a52a9d8/app/lib/methods/helpers/openLink.ts#L58) at the catch, it's expected to use the URL

app/lib/methods/helpers/addProtocol.test.ts
reinaldonetof: not the people, but the server is returning this double slash

app/lib/methods/helpers/addProtocol.ts
reinaldonetof: I found this link with `http` protocol, that I think it's secure to test. http://web.simmons.edu/~grovesd/comm244/notes/week2/links

you can send it on the chat without the protocol, the behavior will be the same while using the browser or the mobile app

reinaldonetof: > I don't understand what you're trying to do. Can you elaborate?

About my last answer before your question, 
> the server didn't return the localhost as a link if the user writes only localhost:3000, is it expected to return as a link, or should I be able to treat this behavior too?

I explained that the server doesn't consider the localhost:3000 as a link.

Also, we aren't able to check if there's a security certificate or not from a link without their protocol, so as the `https` is a convention to use, we add this protocol at the beginning of the link.

GleidsonDaniel (approved):

Issue #4878, opened 2023-02-01
Broken emoji picker on tablet


dusatvoj:
### Description:

There's and issue with emoji picker on tablets. I can't pick all of emojis due to it's badly wrapped. I can scroll only vertical but there are "hidden" / not wrapped emojis at I can't pick

### Environment Information:

- Rocket.Chat Server Version: 5.4.1
- Rocket.Chat App Version: 4.35.1.37443
- Device Name: Samsung Galaxy Tab S8
- OS Version: Android 13

### Steps to reproduce:

1. Try to add reaction emoji to message.

### Expected behavior:

It's wrapped correctly

### Actual behavior:

![image](https://user-images.githubusercontent.com/11560794/216028693-77bd51f5-ee85-4532-b958-538c9ad4cae4.png)

### Additional context:

N/A

PR #4913, opened 2023-02-16
Fixed issue #4878

ShantKhatri:
I had fixed this issue by managing the ratio of size of emoji with respect to whole screen of running device.


GleidsonDaniel:
Your logic is wrong for changing the emoji size, the emoji has that size for a reason.
The right thing would be to change the way this emoji is rendered, maybe change the amount that can be rendered or something like that.

Issue #3132, opened 2021-05-11
the number of users on the screen "Actions" is not updated after removing or adding a user


rocketbughunter:
### Description:

if you add or remove a user to the chat, then the number of users on the screen "Actions" is not updated immediately, but only if you exit and re-enter the screen "Actions"

### Environment Information:

- Rocket.Chat Server Version: 3.13.2
- Rocket.Chat App Version: 4.16.1.23838
- Device Name: Iphone 11 pro max
- OS Version:  ios 14.4.2

### Steps to reproduce:

1. go to the  "Actions" screen in a chat where there are 3 users (screenshot 1)
2. go to the "Members" section
3. delete one user via "Remove from room"
4. back to the screen "Actions" - shows that there are still 3 users
5. go to the chat and then go to the "Actions" screen again
6. now shows that there are already 2 users
7. go to the "Add users" and add one user
8. back to the screen "Actions" - shows that there are still 2 users
9. go to the chat and then go to the "Actions" screen again
10. now shows that there are already 3 users

### Expected behavior:

4. back to the screen "Actions" - shows that there already 2 users
8. back to the screen "Actions" - shows that there already 3 users

### Actual behavior:

4. back to the screen "Actions" - shows that there are still 3 users
8. back to the screen "Actions" - shows that there are still 2 users

rocketbughunter:
![1](https://user-images.githubusercontent.com/82801688/117832504-359c6f00-b27e-11eb-98ac-6064186a0afc.jpeg)

PR #4955, opened 2023-03-07
[FIX] Update the number of members from a room

reinaldonetof:
## Proposed changes

> if you add or remove a user to the chat, then the number of users on the screen "Actions" is not updated immediately, but only if you exit and re-enter the screen "Actions"

The number of members got only when the view is mounted and doesn't know when to update again. Also, the property `usersCount` exists on the server's room object and we can implement it on the watermelon too.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/3132

## How to test or reproduce

- go to the "Actions" screen in a room
- go to the "Members" section
- delete one user via "Remove from room"
- back to the screen "Actions"
- go to the chat and then go to the "Actions" screen again
- go to the "Add users" and add one user
- back to the "Actions" screen again

#### Before

https://user-images.githubusercontent.com/47038980/223559428-1d5f6153-1419-425e-b338-85af2dcdccd0.mov

#### After
https://user-images.githubusercontent.com/47038980/223559086-91250d3d-801f-4a7c-a762-32b7d4616a05.mov

## Types of changes

## Checklist

## Further comments

[TC-391]

### DIFF
--- a/app/definitions/ISubscription.ts
+++ b/app/definitions/ISubscription.ts
@@ -102,6 +102,7 @@ export interface ISubscription {
 	onHold?: boolean;
 	source?: IOmnichannelSource;
 	hideMentionStatus?: boolean;
+	usersCount?: number;
 	// https://nozbe.github.io/WatermelonDB/Relation.html#relation-api
 	messages: RelationModified<TMessageModel>;
 	threads: RelationModified<TThreadModel>;
--- a/app/lib/database/model/Subscription.js
+++ b/app/lib/database/model/Subscription.js
@@ -137,5 +137,7 @@ export default class Subscription extends Model {
 
 	@field('on_hold') onHold;
 
+	@field('users_count') usersCount;
+
 	@json('source', sanitizer) source;
 }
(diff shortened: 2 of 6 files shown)


GleidsonDaniel (approved):

GleidsonDaniel (changes requested):
add e2e test

GleidsonDaniel (changes requested):
The test does not include the fix, it only checks if there are users in the room...

reinaldonetof:
> The test does not include the fix, it only checks if there are users in the room...

Hey Dani, actually the e2e test check if the number of members was updated here: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4955/files#diff-fdda96c19399ad1d5f16f866077da90498784437d384de8af5d37d34179031daR342-R344

because currently, if we add some members when back to RoomActions, the number doesn't update, is like if was 1 member, you add a new member, back to RoomActions will continue with 1 member, will not change to 2 members

GleidsonDaniel (approved):

Issue #4975, opened 2023-03-14
Console error generated by workspaces list in login screen (VirtualizedLists should never be nested inside plain ScrollViews with the same orientation)


floels:
### Description:

Displaying the list of workspaces in the login screen triggers a console error:

![Capture d’écran 2023-03-14 à 10 19 42](https://user-images.githubusercontent.com/79283128/224954460-563b7216-ce6f-414b-9f3c-36f231074f36.png)

This is because we use a `<FlatList />` inside a plain `<ScrollView />` when rendering the `<ServerInput />` component:

https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/4dfc9c70f3ccd00016edf9c11701fda2b3b5f162/app/views/NewServerView/ServerInput/index.tsx#L75-L82

This is not necessary, we could simply `.map()` on `serversHistory` instead:

```typescript
{serversHistory.map(item => (
	<View key={item.id}>
		<Item item={item} theme={theme} onPress={() => onPressServerHistory(item)} onDelete={onDelete} />
		<Separator />
	</View>
))}
```

It would yield the same result and it would avoid the console error.

I can open a PR to make this change once I get confirmation.

### Environment Information:

- Rocket.Chat Server Version: any
- Rocket.Chat App Version: develop
- Device Name: iPhone 13 (simulated)
- OS Version: iOS 16.2

### Steps to reproduce:

1. Go to launch screen.
2. Tap in server input field:

![Simulator Screen Recording - iPhone 13 - 2023-03-14 at 10 25 50](https://user-images.githubusercontent.com/79283128/224956057-37976e9e-f04c-4582-921d-6799578aa672.gif)

### Expected behavior:

No console error.

### Actual behavior:

Console error:

```
VirtualizedLists should never be nested inside plain ScrollViews with the same orientation because it can break windowing and other functionality - use another VirtualizedList-backed container instead. 
    at VirtualizedList (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=chat.rocket.reactnative:61891:36)
    at FlatList (http://localhost:8081/index.bundle?platform=ios&dev=true&minify=false&modulesOnly=false&runModule=true&app=chat.rocket.reactnative:61324:36)
    at RCTView
    at View
    at RCTView
    at View
[...]

PR #4989, opened 2023-03-17
fix: Use plain list instead of <FlatList/> to render servers history dropdown

floels:
## Proposed changes

As discussed in issue #4975, the goal of this PR is to switch the rendering of the servers history dropdown from a `<FlatList />` to a plain list of items, `map()`-ing over the `serversHistory` array, so as to get rid of the "VirtualizedLists should never be nested inside plain ScrollViews with the same orientation" console error when tapping into the server input.

This change doesn't have any visible impact.

I used the occasion to:

- standardize the pluralization to `serversHistory` instead of `serverHistory` everywhere in the codebase, to make it more consistent (both pluralizations were used, leading to weird code),
- move all the database querying logic from `NewServerView/index.tsx` down to `NewServerView/ServerInput/index.tsx`, since `ServerInput` is the only component that needs the corresponding data.

Those three changes are made in separate commits.

## Issue(s)	

#4975 (includes reproduction instructions)

### Before

ezgif-2-8fe554e4f5

### After

ezgif-2-acbddfbe83

## Types of changes

## Checklist

[N/A] I have added tests that prove my fix is effective or that my feature works (if applicable)
[N/A] I have added necessary documentation (if applicable)
[N/A] Any dependent changes have been merged and published in downstream modules


app/views/NewServerView/ServerInput/index.tsx
```diff
@@ -72,14 +74,15 @@ const ServerInput = ({
...
+					{/*  https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/5aaf7af5f4004e8b0b9386da579ae4c8aed3a9a0/app/views/NewServerView/index.tsx#L160 */}
+					{/* But we repeat the filtering here, since having more than three items would break the display pretty badly. */}
+					{serversHistory.slice(0, MAX_ITEMS_HISTORY).map(item => (
+						<View key={item.id}>
 							<Item item={item} theme={theme} onPress={() => onPressServerHistory(item)} onDelete={onDelete} />
-						)}
-						ItemSeparatorComponent={List.Separator}
-						keyExtractor={item => item.id}
-					/>
+							<Separator />
```
diegolmello, line 83: This way you're doubling border bottom
image

app/views/NewServerView/ServerInput/index.tsx
```diff
@@ -72,14 +74,15 @@ const ServerInput = ({
 						{ backgroundColor: themes[theme].backgroundColor, borderColor: themes[theme].separatorColor }
 					]}
 				>
-					<FlatList
-						data={serversHistory}
-						renderItem={({ item }) => (
+					{/* NB: in theory the .slice(0, MAX_ITEMS_HISTORY) is useless because the number of items in the history is already limited to three when querying the database: */}
+					{/*  https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/5aaf7af5f4004e8b0b9386da579ae4c8aed3a9a0/app/views/NewServerView/index.tsx#L160 */}
+					{/* But we repeat the filtering here, since having more than three items would break the display pretty badly. */}
```
diegolmello, line 79: You have a point.
That's a weird colocation imo.
What if you write a custom hook to fetch servers history?
You could remove it from NewServerView/index and place it here only.
It only matters here really.

The max items logic would be only inside the custom hook.
wdyt?

diegolmello (changes requested):

diegolmello:
@floels btw we just started to use conventional commits on the company.

app/views/NewServerView/ServerInput/index.tsx
floels: @diegolmello true. So you would move all the logic around the servers history down in the `<ServerInput />` component?

floels: @diegolmello I'll give it a try on this PR. I'll add a separate commit.

floels:
ℹ️ Keeping this as draft while solving my issue with the local Android build (#4999)

app/views/NewServerView/ServerInput/index.tsx
floels: Hi @diegolmello,

I moved the logic to the <ServerInput /> component, as discussed. The PR is ready for a new review.

floels:
> @floels btw we just started to use conventional commits on the company. Also make sure to run e2e tests on your machine.

@diegolmello noted for the conventional commits.

I wasn't able to run E2E tests on my machine unfortunately, the E2E build fails on both iOS and Android 😕 

Let's see what the CI says? 🙈

PR #5064, opened 2023-05-08
fix: quote showing with leading empty space on the rooms list view

reinaldonetof:
## Proposed changes

The last message of a chat when is a quote message is showing with an empty space before the text on the rooms list.

#### Before
Screen Shot 2023-05-08 at 10 49 54

#### After
Screen Shot 2023-05-08 at 10 48 34

## Types of changes

## Checklist

## Further comments

[TC-408]

### DIFF
--- a/app/containers/markdown/Preview.tsx
+++ b/app/containers/markdown/Preview.tsx
@@ -26,9 +26,9 @@ const MarkdownPreview = ({ msg, numberOfLines = 1, testID, style = [] }: IMarkdo
 	let m = formatText(msg);
 	m = formatHyperlink(m);
 	m = shortnameToUnicode(m);
+	m = removeMarkdown(m);
 	// Removes sequential empty spaces
 	m = m.replace(/\s+/g, ' ');
-	m = removeMarkdown(m);
 	m = m.replace(/\n+/g, ' ');
 	return (
 		<Text


app/containers/markdown/Preview.tsx
```diff
@@ -26,9 +26,9 @@ const MarkdownPreview = ({ msg, numberOfLines = 1, testID, style = [] }: IMarkdo
 	let m = formatText(msg);
 	m = formatHyperlink(m);
 	m = shortnameToUnicode(m);
+	m = removeMarkdown(m);
```
diegolmello, line 29: Tests?

diegolmello (changes requested):

app/containers/markdown/Preview.tsx
diegolmello: The fact you didn't have to update tests doesn't mean you shouldn't write new regression tests to prevent this from happening in the future.
Also, there's a chance we moved that line around a few years ago, so it's worth checking.

reinaldonetof: I moved those functions together to another file and add unit test for it

GleidsonDaniel (approved):

app/views/RoomsListView/index.tsx
```diff
@@ -154,7 +154,8 @@ const getItemLayout = (data: ISubscription[] | null | undefined, index: number,
 	offset: height * index,
 	index
 });
-const keyExtractor = (item: ISubscription) => item.rid;
+// We need to pass the isSearching to handle the room if it's a room returned from search (without observable) or from query subscription (with observable)
```
diegolmello, line 157: This is wrong. The PR is correctly fixing the bug, but because of a different reason :)

When we load FlatList, `chats` contains the list of rooms.
When we go to search, `search` starts empty and we first render the list empty. When `search` has data, we show them.
When we go back to `chats`, we clean `search`, but `chats` already has the previous content, so it reuses the same cells with the same key from `keyExtractor`.
There's no reason to unmount the cell, which doesn't trigger `useEffect` again, hence there's no observe.

app/views/RoomsListView/index.tsx
```diff
@@ -899,7 +900,7 @@ class RoomsListView extends React.Component<IRoomsListViewProps, IRoomsListViewS
 				ref={this.getScrollRef}
 				data={searching ? search : chats}
 				extraData={searching ? search : chats}
-				keyExtractor={keyExtractor}
+				keyExtractor={item => keyExtractor(item, !!searching)}
```
diegolmello, line 903: If you fix `searching` type, you can avoid `!!`.
I don't think `searching` is ever `undefined`.

diegolmello (dismissed):
Left a few comments.
Try to split this PR next time, please.
It's fixing 2 unrelated issues and makes the title too big.
Also if we need to rollback, we'll have to either rollback both fixes or remove manually on code.

GleidsonDaniel (approved):

GleidsonDaniel (approved):

PR #5094, opened 2023-05-31
fix: capability to dismiss the keyboard on drag for iOS on RoomsListView

reinaldonetof:
## Proposed changes

When searching for a room on the Rooms List View on iOS, the keyboard keeps opening and hiding the last room of the search. To fix this issue, I'm adding the prop `keyboardDismissMode: 'on-drag'` only for iOS.

#### Before

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/def39251-e2c4-4b7e-af9e-952c1c48bdb8

#### After

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/4ffa7067-531c-4051-a5ad-d1f8eaf83c27

## Types of changes

## Checklist

## Further comments

[TC-636]

[TC-636]: https://rocketchat.atlassian.net/browse/TC-636?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


GleidsonDaniel (dismissed):
I don't think this solution is correct.
What she asked for on the card was:
_It should be possible to scroll with the keyboard open_

In this solution you are closing the keyboard...

The correct thing would be to adapt the lower padding when the keyboard is open so that the list can be scrolled to the end.

reinaldonetof:
> I don't think this solution is correct. What she asked for on the card was: _It should be possible to scroll with the keyboard open_
> 
> In this solution you are closing the keyboard...
> 
> The correct thing would be to adapt the lower padding when the keyboard is open so that the list can be scrolled to the end.

Actually, It was my mistake, I didn't update the description of the task.

This task I discussed this with the mobile team and we noticed that the expected behavior for iOS is to dismiss the keyboard when scrolling the view, the same behavior as happens with other apps.

app/views/RoomsListView/index.tsx
```diff
@@ -913,6 +925,7 @@ class RoomsListView extends React.Component<IRoomsListViewProps, IRoomsListViewS
 				windowSize={9}
 				onEndReached={this.onEndReached}
 				onEndReachedThreshold={0.5}
+				{...DISMISS_KEYBOARD_ON_DRAG}
```
GleidsonDaniel, line 928:
```suggestion
				{... keyboardDismissMode}
```

GleidsonDaniel: The default is already none, I don't think you need to create an auxiliary const, just add the check and leave the code clean.

diegolmello (approved):

diegolmello:
@reinaldonetof Fix your PR title to be more clear. It's giving the impression of changing something app wide and it's actually a tiny change.

PR #5111, opened 2023-06-20
feat: Remove Experimental

diegolmello:
## Proposed changes

We're removing Rocket.Chat Experimental app. We identified a few issues having it alongside main app:
- Reduces productivity
- Doesn't give us useful feedbacks we wouldn't have with Official app only
- Generates unnecessary builds on CI
- Makes whitelabel very confusing
- 2 apps on the stores
- 2 certificates to keep active on Apple
- 2 apps on Bugsnag
- 2 apps to explain on our docs. Generates confusion to the reader. Official, Experimental, Official beta, Experimental beta.  Which beta is the most beta?
- It probably makes builds slower. I can't be sure how much, but it does extra work

The original idea of Experimental was to try new things there, but we've never did it.
Right now the only benefit is to have the two versions installed at the same time.
Frankly, the only reason some might have them both installed is because one of them is broken, but we can fix that without the need for Experimental build to exist.

### CI
Previously CI workflows would need both Experimental and Official jobs. This is how it would look like while working on PRs:
Screenshot 2023-06-22 at 14 25 00

This is how it looks after this change:
Screenshot 2023-06-22 at 14 23 15

When a PR landed develop branch, it would look like this:
Screenshot 2023-06-22 at 14 23 55

After this PR, develop workflow is the following:
Screenshot 2023-06-22 at 14 22 38

## Types of changes

## Checklist

## Further comments

[NATIVE-351]

[NATIVE-351]: https://rocketchat.atlassian.net/browse/NATIVE-351?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


android/app/build.gradle
```diff
@@ -180,19 +170,11 @@ android {
...
-        experimentalPlayDebug {
-            java.srcDirs = ['src/main/java', 'src/play/java']
-            manifest.srcFile 'src/play/AndroidManifest.xml'
-        }
-        experimentalPlayRelease {
-            java.srcDirs = ['src/main/java', 'src/play/java']
-            manifest.srcFile 'src/play/AndroidManifest.xml'
-        }
-        officialPlayDebug {
+        playDebug {
```
GleidsonDaniel, line 173: Wouldn't it be just debug and release, instead of PlayDebug?
What does Play mean in these cases?

GleidsonDaniel (approved):

android/app/build.gradle
diegolmello: We still have Foss version, so it's either Foss or Play for every task.

GleidsonDaniel:
😢

PR #5130, opened 2023-07-05
fix: handle the attachment when the value is null

reinaldonetof:
## Proposed changes

Fix this error: https://app.bugsnag.com/rocket-dot-chat-1/mobile-1/errors/642f0f1c8ebe9f0007a932b8

Screen Shot 2023-07-05 at 18 51 27

The error is happening only in the production

```
TypeError • app/lib/methods/helpers/normalizeMessage.ts:13
Cannot read property 'fields' of null 
```
However, I was unable to reproduce it in production or in debug mode. Considering what is described, I believe that the `att` is passing as null

## Types of changes

## Checklist

## Further comments

[TC-802]

[TC-802]: https://rocketchat.atlassian.net/browse/TC-802?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/methods/helpers/normalizeMessage.ts
```diff
@@ -10,6 +10,11 @@ function normalizeAttachments(msg: TMsg) {
 		msg.attachments = [];
 	}
 	msg.attachments = msg.attachments.map(att => {
+		// The att could be as null and returning type error
+		// To fix this: https://app.bugsnag.com/rocket-dot-chat-1/mobile-1/errors/642f0f1c8ebe9f0007a932b8?pivot_tab=event
```
diegolmello, line 14: Remove these comments from code

app/lib/methods/helpers/normalizeMessage.ts
```diff
@@ -9,14 +9,17 @@ function normalizeAttachments(msg: TMsg) {
...
-		att.fields = att.fields || [];
-		if (att.ts) {
-			att.ts = moment(att.ts).toDate();
-		}
-		att = normalizeAttachments(att as TMsg);
-		return att;
-	});
+
+	msg.attachments = msg.attachments
+		.filter(att => !!att)
```
GleidsonDaniel, line 14:
```suggestion
		.filter(att => att)
```
This is not necessary because this error only occurs when the field is null or undefined.

GleidsonDaniel (changes requested):

app/lib/methods/helpers/normalizeMessage.ts
reinaldonetof: Correct, but we don't know where or since when this error started to occur, so we can't update the types of `msg.attachments` to (IAttachment | null | undefined)[], then, as we don't have this difference, in my opinion, this double-check `!!` shows quick during the read that the att could be different from object IAttachment.

GleidsonDaniel (approved):

app/lib/methods/helpers/normalizeMessage.ts
Lao-Ax: As we know, this error could be reproduced by pinning a message with an attachment.
Just FYI :) Probably we'll find a way to fix it a bit later on server side.

Issue #4889, opened 2023-02-02
[BUG] When the Light Mode Status Bar is not Readable


omertahaoztop:
### Description:

When the entering any channel on dark mode status bar is readable, but when using light mode Status bar is not readable. This behavior is same in single server and official app.

#### Dark Mode:

![darkmode](https://user-images.githubusercontent.com/26715379/216463401-216933ed-7582-4c10-8858-c960e75350b7.jpg)

#### Light Mode:

![lightmode](https://user-images.githubusercontent.com/26715379/216463511-573272b3-8ba9-4f20-816c-9ddcb3a187ab.jpg)

### Environment Information:

- Rocket.Chat Server Version: 4.8.7
- Rocket.Chat App Version: 4.35.1 Single Server
- Device Name:  Android
- OS Version:  Android 12

diegolmello:
I'm not able to reproduce it. Can you give us the steps to do it?

omertahaoztop:
In my friend's phone(Android 12) if you switch to light mode Status bar is not readable like screenshot.

But in my phone (Android 11) everything is ok and readable. 

What is your Android version?

thgoebel:
I have the same issue on Android 13 (with LineageOS 20).
On a clean install, not even logging in to the server, right after opening RocketChat for the first time.

Steps:

1. OS in light mode.
2. Open RocketChat => status bar text is light (but should be dark)
3. Change OS to dark mode via Quick Settings Tile
4. RocketChat status bar text is now light ✅ 
5. Change OS to light mode
6. RocketChat status bar text is now dark ✅ 
7. Close RocketChat and remove it from Recent Apps
8. Open RocketChat again
9. RocketChat status bar text is now light again (but should be dark)

Android 13
RocketChat 4.37.1 from F-Droid.

thgoebel:
I still have this issue. The steps from my previous comment still work, exactly the same buggy behaviour. 

- Android 13
- RocketChat 4.40.0 with the apk downloaded [from the Github release](https://github.com/RocketChat/Rocket.Chat.ReactNative/releases/tag/4.40.0).

PR #5132, opened 2023-07-07
fix: Android status bar theme on the first open isn't working properly

reinaldonetof:
## Proposed changes

I discovered that the status bar was not re-rendering correctly! So the solution was to add a useEffect to set the background and barStyle. The behavior is the same as expected when added to the props, and it fixes the bug on Android while still functioning correctly on iOS. The difference is that the useEffect re-renders one extra time after it is mounted.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4889

## How to test or reproduce

This error it's happening on the first open at my Samsung Galaxy A31. It's working properly with Xiaomi Redmi 9 and with iOS devices.

>  - OS in light mode.
>  - Open RocketChat => status bar text is light (but should be dark)
>  - Change OS to dark mode via Quick Settings Tile
>  - RocketChat status bar text is now light ✅
>  - Change OS to light mode
>  - RocketChat status bar text is now dark ✅
>  - Close RocketChat and remove it from Recent Apps
>  - Open RocketChat again
>  - RocketChat status bar text is now light again (but should be dark)

#### Before

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/71a00fa5-920f-4a04-8c1a-acd945de657b

#### After

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/d1edcf32-572d-403e-a7be-cad6dc6b1f38

## Types of changes

## Checklist

## Further comments

[TC-691]

[TC-691]: https://rocketchat.atlassian.net/browse/TC-691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/StatusBar.tsx
```diff
@@ -15,14 +14,20 @@ interface IStatusBar {
...
+	useEffect(() => {
+		if (!barStyle) {
+			barStyle = 'light-content';
+			if (theme === 'light') {
+				barStyle = 'dark-content';
+			}
 		}
-	}
-	return <StatusBarRN backgroundColor={backgroundColor ?? themes[theme].headerBackground} barStyle={barStyle} animated />;
+		StatusBarRN.setBackgroundColor(backgroundColor ?? colors.headerBackground);
```
GleidsonDaniel, line 26: Are these methods only for android?
Why do you set them here without doing a platform check, but the fix is for android.
Have you checked to see if anything happens on iOS?

GleidsonDaniel (changes requested):

app/containers/StatusBar.tsx
reinaldonetof: the method setBackgroundColor is only for Android, the same as passing the backgroundColor by props. Works the same as props, and doesn't affect the iOS. As before, we don't need to check the platform. wdyt?

GleidsonDaniel: these method**S**
The method marked above is only for android, but the one below is not.

reinaldonetof: Sure, the below`StatusBarRN.setBarStyle` works for both and is the same as passing barStyle as props, also fix the Android bug. 

I decided to do on this way just to keep the same pattern for both, instead of separating the StatusBar component for each platform

GleidsonDaniel (approved):

PR #4662, opened 2022-11-08
[FIX] Playing multiple voice messages at the same time 

gitstart:
## Proposed changes

Add an event listener to pause playing audio sounds in `Audio.tsx`
## Issue(s)	

#1654

## How to test or reproduce

1. Go to a room with more than one audio message
2. Play one audio message
3. Whilst its still playing play another, note that the first played audio message will automatically stop.

## Screenshots
Issue 
https://www.loom.com/share/d4b084c5f47346908fce430db61d9f04
FIX
https://www.loom.com/share/67f4c442d29e4ac7ab1466065924d121
## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

The part about `Also it would be neat if it will automatically play the next voice message in the view.` was not attended to because it required major code changes.

PR #5160, opened 2023-08-07
feat: new audio player

reinaldonetof:
## Proposed changes

This PR is improving the Audio Player. The main reason was to change the layout and add the capability to increase/decrease the audio's velocity (rate).

Beyond the layout stuff, I changed how we handle the audio. Also, I refactor the approach implemented in that PR https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4662 to avoid the necessity of EventListener. Added to this new approach, we are able to control and unload the audio from other places besides the RoomView and the Component Audio, this could be discussed later.

## How to test or reproduce

- Send multiple audios in a single channel to test the play/pause/rate
- Send audio in a thread 
    - The audio from a channel will pause when opening the thread
    - If the audio from a thread is playing, then back to the channel, this audio from the thread should stop
 - Background listening
 - Check if screen is always on while playing an audio and if can go off after is paused
 - Play the audio from all clients origin on all clients Android/iOS.

## Screenshots

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/87c08f06-d281-40b6-990f-d56ce67d625e

## Types of changes

## Checklist

## Further comments

[TC-781]

[TC-781]: https://rocketchat.atlassian.net/browse/TC-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/message/Components/Audio/PlayButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+	onPress: () => void;
+	cached: boolean;
+}
+
+const BUTTON_HIT_SLOP = { top: 12, right: 12, bottom: 12, left: 12 };
+
+const PlayButton = React.memo(({ loading, paused, onPress, disabled, cached }: IButton) => {
+	const { colors } = useTheme();
+
+	let customIconName: 'arrow-down' | 'play-shape-filled' | 'pause-shape-filled' = 'arrow-down';
```
GleidsonDaniel, line 22: create a type for it

app/lib/constants/colors.ts
```diff
@@ -254,6 +268,13 @@ export const colors = {
 		gray100: '#CBCED1',
 		n900: '#FFFFFF',
 		overlayColor: '#1F2329B2',
+		audioPlayerPrimary: '#3976D1',
+		audioPlayerSecondary: '#404754',
+		audioRateBackground: '#353B45',
+		audioRateText: '#E4E7EA',
+		audioTimeText: '#E4E7EA',
+		audioBorderColor: '#2F343D',
```
GleidsonDaniel, line 276: In these cases use the name that is there in figma, this color for example its name is: Theme/Stroke/extra-light

app/lib/methods/handleAudioMedia.ts
```diff
@@ -0,0 +1,102 @@
...
+	allowsRecordingIOS: false,
+	playsInSilentModeIOS: true,
+	staysActiveInBackground: true,
+	shouldDuckAndroid: true,
+	playThroughEarpieceAndroid: false,
+	interruptionModeIOS: InterruptionModeIOS.DoNotMix,
+	interruptionModeAndroid: InterruptionModeAndroid.DoNotMix
+};
+
+class HandleAudioMedia {
```
GleidsonDaniel, line 14: Why a class?

GleidsonDaniel (changes requested):
The times of the audios are coming with a value of zero, on the web they show the duration of the audio.
The audio speed multiplier does not work correctly if you have more than one audio in the room.
Play also doesn't work correctly if I have more than one audio in the room, it plays another audio.
The touch area of the slider is small, you have to keep clicking several times until you can scroll (tested on a physical device).
I don't know if it's in this PR but the auto download was activated and even then it didn't download the audios automatically.

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/18687097/60e18e4d-71eb-4217-bbd5-6e38aec36fed

reinaldonetof:
> The times of the audios are coming with a value of zero, on the web they show the duration of the audio. The audio speed multiplier does not work correctly if you have more than one audio in the room. Play also doesn't work correctly if I have more than one audio in the room, it plays another audio. The touch area of the slider is small, you have to keep clicking several times until you can scroll (tested on a physical device). I don't know if it's in this PR but the auto download was activated and even then it didn't download the audios automatically.
> 
>  RPReplay_Final1692306115.MP4

Yeah, the problem that you faced is related to this: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5171

Probably you uploaded an Audio with the same name, this could be happened by uploading the audio from the web because always the name of the audio is `Audio Record.mp3`

Sorry for that inconvenience, but will be fixed as soon as the PR #5171 is merged

app/lib/methods/handleAudioMedia.ts
reinaldonetof: I decided to create a class to unify the data and behaviors inside it because probably we wouldn't reuse the same methods in another type of file. Also, IMHO this is easier to understand 😁

reinaldonetof: To do without a singleton will look like the handleMediaDownload (that I did too).

app/containers/message/Components/Audio/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+	};
+
+	useDerivedValue(() => {
+		if (onEndGestureHandler.value) {
+			runOnJS(wrapper)(currentTime.value);
+		}
+	});
+
+	const formatTime = (ms: number) => {
+		'worklet';
```
GleidsonDaniel, line 94: Add the link to why you are using this property.

GleidsonDaniel:
CI is failing

app/containers/message/Components/Audio/Slider.tsx
reinaldonetof: https://docs.swmansion.com/react-native-reanimated/docs/2.x/fundamentals/worklets/

We need the "worklet" because this function is being called inside the worklet useDerivedValue. The function is to don't repeat the same code twice at lines 107 and 111. 

If there isn't an explicit "worklet" at formatTime function, this function will only run on the JS side, not on the UI Thread created, and won't be capable of running synchronously with the useDerivedValue

GleidsonDaniel: You didn't understand my point, you didn't leave any information in the CODE about why you used this property, a person who doesn't understand the use of this prop when reading the code won't understand why it was used, when adding very specific things try always add comments to the code.

GleidsonDaniel: Before merging this PR, add this information to the code.

app/containers/message/Components/Audio/Slider.tsx
```diff
@@ -0,0 +1,150 @@
...
+}
+
+const BUTTON_HIT_SLOP = {
+	top: 8,
+	right: 8,
+	bottom: 8,
+	left: 8
+};
+
+const Slider = ({ currentTime, duration, loaded = false, onChangeTime }: ISlider) => {
```
diegolmello, line 33: Can you rename this to `Seek`? It's the standard name for this.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#events

app/containers/message/Components/Audio/AudioRate.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Text } from 'react-native';
+
+import styles from './styles';
+import { useTheme } from '../../../../theme';
+import Touchable from '../../Touchable';
+
+const AudioRate = ({
```
diegolmello, line 8: PlaybackSpeed

app/containers/message/Components/Audio/index.tsx
```diff
@@ -0,0 +1,212 @@
...
+		return null;
+	}
+	return (
+		<>
+			<Markdown msg={msg} style={[isReply && style]} username={user.username} getCustomEmoji={getCustomEmoji} />
+			<View style={[styles.audioContainer, { backgroundColor: colors.surfaceTint, borderColor: colors.strokeExtraLight }]}>
+				<PlayButton disabled={isReply} loading={loading} paused={paused} cached={cached} onPress={onPress} />
+				<Slider currentTime={currentTime} duration={duration} loaded={!isReply && cached} onChangeTime={setPosition} />
+				<AudioRate onChange={onChangeRate} loaded={!isReply && cached} rate={rate} />
+			</View>
```
diegolmello, line 207: Can you export this to be used outside `message`?
We're going to need it on new message composer.

app/containers/message/Components/Audio/Loading.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { useEffect } from 'react';
+import Animated, { Easing, useAnimatedStyle, useSharedValue, withRepeat, withTiming } from 'react-native-reanimated';
+
+import { CustomIcon } from '../../../CustomIcon';
+import { useTheme } from '../../../../theme';
+
+const Loading = () => {
```
diegolmello, line 7: I'm not sure about this animation.
It's far too different from the usual circle animation, because there's no "collapsed" point.
It could be just me, but this is what I expect from it (and that's why I didn't want to use the ActivityIndicator from RN) https://m2.material.io/components/progress-indicators#circular-progress-indicators
You can also take a look at the favicon from Chrome loading.

app/lib/constants/colors.ts
```diff
@@ -102,6 +102,14 @@ export const colors = {
 		statusBackgroundWarning: '#FFECAD',
 		statusFontOnWarning: '#B88D00',
 		overlayColor: '#1F2329B2',
+		buttonBackgroundPrimaryDefault: '#156FF5',
```
diegolmello, line 105: Can you make sure to apply the same naming from here? https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/feat.new-composer/app/lib/constants/colors.ts#L24

Not asking to apply all colors, just trying to prevent differences like `buttonFontOnPrimary`

diegolmello (changes requested):

app/lib/constants/colors.ts
reinaldonetof: just the dark/black colors of buttonBackgroundSecondaryDefault are different and I used the same color as you set on new-composer.

```
- buttonBackgroundSecondaryDefault: '#353B45',
to
+ buttonBackgroundSecondaryDefault: '#2F343D',
```

app/containers/AudioPlayer/index.tsx
```diff
@@ -124,6 +127,16 @@ const AudioPlayer = ({
...
 
+	useEffect(() => {
+		const unsubscribeFocus = navigation.addListener('focus', () => {
+			audioPlayer.setOnPlaybackStatusUpdate(audioUri.current, onPlaybackStatusUpdate);
+		});
+
+		return () => {
+			unsubscribeFocus();
+		};
+	}, [navigation]);
```
reinaldonetof, line 138: Need this to keep the `onPlaybackStatusUpdate` updated according to the audio player component in focus.

ex:

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/121abf5d-bd1f-488b-9bb5-c7349341e31a

- The sound starts to play in the main room. 
- a thread is created on this audio's message
- when we open the thread, the audio pauses (this behavior is expected when open a thread)
- then we can see that the audio component shows the current time the same as the pause of the main room
- then we can play the audio again from where it paused, then stop and back to the main room
- In the main room again, the focus listener calls the function to `setOnPlaybackStatusUpdate` and calls the `onPlaybackStatusUpdate` function of the audio component from the main room and the current time is the expected time from when paused in the thread room
- Also, this approach is a possible solution for when the user go to the Files view and play an audio

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const x = useSharedValue(0);
+	const current = useSharedValue('00:00');
+	const scale = useSharedValue(1);
+	const isHandlePan = useSharedValue(false);
+	const onEndGestureHandler = useSharedValue(false);
+	const isTimeChanged = useSharedValue(false);
+
+	const styleLine = useAnimatedStyle(() => ({
+		width: x.value,
+		zIndex: 2
```
diegolmello, line 46: I'm curious if this is actually doing anything.
`zIndex` rarely worked for me on RN tbh.

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const styleThumb = useAnimatedStyle(() => ({
+		transform: [{ translateX: x.value }, { scale: scale.value }]
+	}));
+
+	const onLayout = (event: LayoutChangeEvent) => {
+		const { width } = event.nativeEvent.layout;
+		maxWidth.value = width - 12;
+	};
+
+	const gestureHandler = useAnimatedGestureHandler({
```
diegolmello, line 58: Matches the name of the function.

```suggestion
	const onGestureEvent = useAnimatedGestureHandler({
```

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+	right: 8,
+	bottom: 8,
+	left: 8
+};
+
+const Seek = ({ currentTime, duration, loaded = false, onChangeTime }: ISeek) => {
+	const { colors } = useTheme();
+
+	const maxWidth = useSharedValue(1);
+	const x = useSharedValue(0);
```
diegolmello, line 37: What's X?

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+	useDerivedValue(() => {
+		if (isHandlePan.value) {
+			const cTime = (x.value * duration.value) / maxWidth.value || 0;
+			currentTime.value = cTime;
+			current.value = formatTime(cTime);
+		} else {
+			const xTime = (currentTime.value * maxWidth.value) / duration.value || 0;
+			x.value = xTime;
+			current.value = formatTime(currentTime.value);
```
diegolmello, line 112: Besides `cTime` and `xTime` naming, can you explain what's happening here?
They seem to do the same thing.

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+		<View style={styles.seekContainer}>
+			<AnimatedTextInput
+				defaultValue={'00:00'}
+				editable={false}
+				style={[styles.duration, { color: colors.fontDefault }]}
+				animatedProps={getCurrentTime}
+			/>
+			<View style={styles.seek} onLayout={onLayout}>
+				<View style={[styles.line, { backgroundColor: colors.strokeLight }]} />
+				<Animated.View style={[styles.line, styleLine, { backgroundColor: colors.buttonBackgroundPrimaryDefault }]} />
```
diegolmello, line 142: The background one is a View.
Can't you just wrap it around foreground one, so you don't have to do it absolute?

app/containers/AudioPlayer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+	const [rate, setRate] = useState(1);
+
+	const duration = useSharedValue(0);
+	const currentTime = useSharedValue(0);
+
+	const { colors } = useTheme();
+
+	const audioUri = useRef<string>('');
+
+	const navigation = useNavigation();
```
diegolmello, line 40:
```suggestion
	const [paused, setPaused] = useState(true);
	const [rate, setRate] = useState(1);
	const duration = useSharedValue(0);
	const currentTime = useSharedValue(0);
	const { colors } = useTheme();
	const audioUri = useRef<string>('');
	const navigation = useNavigation();
```

app/containers/AudioPlayer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+			if (currentSecond <= durationSeconds) {
+				currentTime.value = currentSecond;
+			}
+			setRate(data.rate);
+		}
+	};
+
+	const onEnd = (data: AVPlaybackStatus) => {
+		if (data.isLoaded) {
+			if (data.didJustFinish) {
```
diegolmello, line 72: On `onPlaying` and `handlePlaybackStatusUpdate` you do a compound `if`, but here you chose to do two.
Unless I'm being naive, you could just

```suggestion
		if (data.isLoaded && data.didJustFinish) {
```

app/containers/AudioPlayer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+	disabled?: boolean;
+	onPressCallback?: Function;
+}
+
+const AudioPlayer = ({
+	fileUri,
+	disabled = false,
+	loading = true,
+	isReadyToPlay = false,
+	onPressCallback = () => {}
```
diegolmello, line 28: This component has lots of buttons.
This is a callback to press what exactly?

app/containers/AudioPlayer/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+			audioPlayer.setOnPlaybackStatusUpdate(audioUri.current, onPlaybackStatusUpdate);
+		});
+
+		return () => {
+			unsubscribeFocus();
+		};
+	}, [navigation]);
+
+	return (
+		<>
```
diegolmello, line 141: Why do you need this fragment, if there's only one view being returned?

app/containers/AudioPlayer/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const styles = StyleSheet.create({
+	audioContainer: {
+		flex: 1,
+		flexDirection: 'row',
+		alignItems: 'center',
+		height: 56,
+		borderWidth: 1,
+		borderRadius: 4,
+		marginBottom: 6
```
diegolmello, line 13: Not multiple of 4

app/containers/AudioPlayer/styles.ts
```diff
@@ -0,0 +1,69 @@
...
+		marginHorizontal: 12,
+		fontVariant: ['tabular-nums'],
+		fontSize: 14,
+		...sharedStyles.textRegular
+	},
+	thumbSeek: {
+		height: 12,
+		width: 12,
+		borderRadius: 6,
+		zIndex: 3
```
diegolmello, line 52: Same zIndex curiosity.

app/lib/methods/audioPlayer.ts
```diff
@@ -0,0 +1,119 @@
...
+	constructor() {
+		this.audioQueue = {};
+		this.audioPlaying = '';
+	}
+
+	async loadAudio(uri: string): Promise<Sound> {
+		if (this.audioQueue[uri]) {
+			return this.audioQueue[uri];
+		}
+		const { sound } = await Audio.Sound.createAsync({ uri }, { androidImplementation: 'MediaPlayer' });
```
diegolmello, line 27: What changes on this androidImplementation?

app/lib/methods/audioPlayer.ts
```diff
@@ -0,0 +1,119 @@
+import { AVPlaybackStatus, Audio, InterruptionModeAndroid, InterruptionModeIOS } from 'expo-av';
+import { Sound } from 'expo-av/build/Audio/Sound';
+
+const mode = {
```
diegolmello, line 4:
```suggestion
const AUDIO_MODE = {
```

app/lib/methods/getSettings.ts
```diff
@@ -160,7 +160,8 @@ export async function getSettings(): Promise<void> {
 			/* eslint-disable no-await-in-loop */
 			const response = await fetch(
 				`${sdk.current.client.host}/api/v1/settings.public?query={"_id":{"$in":${JSON.stringify(settingsParams)}}}
-				&offset=${offset}`);
+				&offset=${offset}`
```
diegolmello, line 163: Prettier?

diegolmello (changes requested):

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+const AnimatedTextInput = Animated.createAnimatedComponent(TextInput);
+
+interface ISeek {
+	duration: SharedValue<number>;
+	currentTime: SharedValue<number>;
+	loaded: boolean;
+	onChangeTime: (time: number) => Promise<void>;
+}
+
+const BUTTON_HIT_SLOP = {
```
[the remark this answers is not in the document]
reinaldonetof: > The touch area of the slider is small, you have to keep clicking several times until you can scroll

I used it before this comment from @GleidsonDaniel. Then, I increase the value! Actually, I can use this same hit slop for Seek and PlayButton, wdyt?

app/lib/methods/getSettings.ts
reinaldonetof: yes

app/containers/AudioPlayer/styles.ts
reinaldonetof: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5160/files#diff-1b3b3e8ec8b635efad40ed7cc2db80c2ed04b26285046232b525cb3614a6d88cL74

I only use the same marginBottom as before

app/lib/methods/audioPlayer.ts
reinaldonetof: https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/51756965550228597b4534fd50573a4ddfb0d4b5/app/containers/message/Audio.tsx#L306

https://docs.expo.dev/versions/latest/sdk/av/#avplaybacksource

support the same formats as Android's MediaPlayer

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,151 @@
...
+		zIndex: 2
+	}));
+
+	const styleThumb = useAnimatedStyle(() => ({
+		transform: [{ translateX: x.value }, { scale: scale.value }]
+	}));
+
+	const onLayout = (event: LayoutChangeEvent) => {
+		const { width } = event.nativeEvent.layout;
+		maxWidth.value = width - 12;
```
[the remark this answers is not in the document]
reinaldonetof: the thumb width

app/containers/AudioPlayer/Seek.tsx
reinaldonetof: I changed the name of variables, the cTime is timeSelected, xTime is timeInProgress, current is timeLabel and x is timePosition.

- The first one is when you hold the thumb icon and move it through the line changing his position. To calculate the time selected based on the position, then changes the currentTime that is the same time as the Audio. 
- The other is the timeInProgress, to calculate the position based on the time received from the Audio. Then changes the timePosition to move the seek to the correct position.
- Both changes the value of timeLabel(current), because they are changing the time label

app/lib/methods/audioPlayer.ts
diegolmello: I can see that. I meant why do you need this implementation over the default one.

app/containers/AudioPlayer/styles.ts
diegolmello: You have a new design on Figma. We have new standards.

app/containers/AudioPlayer/index.tsx
diegolmello: - Main room <-> thread: audio stops
- Main room <-> files: audio keeps playing

diegolmello (changes requested):

diegolmello:
I added some commits to `feat.new-audio-player-2` branch. You can cherry pick them.

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,124 @@
...
+			scale.value = 1;
+			runOnJS(onChangeTime)(Math.round(currentTime.value * 1000));
+		}
+	});
+
+	useDerivedValue(() => {
+		const timeInProgress = (currentTime.value * maxWidth.value) / duration.value || 0;
+		translateX.value = timeInProgress;
+		timeLabel.value = formatTime(currentTime.value);
+	}, [translateX, maxWidth, duration, currentTime]);
```
reinaldonetof, line 86: https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/105f83a6-0d14-4964-8953-478eaa5991e0

This is the problem that isPanning was fixing

app/lib/methods/audioPlayer.ts
reinaldonetof: Because this implementation permits the Audio to reproduce .aac audios.

diegolmello:
Oct-31-2023 16-54-11

app/containers/AudioPlayer/PlayButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+	disabled?: boolean;
+	onPress: () => void;
+	audioState: TAudioState;
+}
+
+type TCustomIconName = 'arrow-down' | 'play-shape-filled' | 'pause-shape-filled';
+
+export type TAudioState = 'loading' | 'paused' | 'to-download' | 'playing';
+
+const IconToRender = ({ audioState, disabled }: { audioState: TAudioState; disabled: boolean }) => {
```
diegolmello, line 20: Redundant. If it's a component, the goal is to render.
You can just call it `Icon` on this scope without interfering other places.

app/containers/AudioPlayer/PlaybackSpeed.tsx
```diff
@@ -0,0 +1,35 @@
...
+}) => {
+	const { colors } = useTheme();
+
+	const onPress = () => {
+		const nextSpeedIndex = speedIndex + 1 >= AVAILABLE_SPEEDS.length ? 0 : speedIndex + 1;
+		onChange(AVAILABLE_SPEEDS[nextSpeedIndex]);
+	};
+
+	return (
+		<TouchableOpacity
```
diegolmello, line 25: You shouldn't use TouchableOpacity outside iOS.
Use BorderlessButton from RNGH

app/containers/AudioPlayer/PlaybackSpeed.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Text, TouchableOpacity } from 'react-native';
+
+import styles from './styles';
+import { useTheme } from '../../theme';
+import { AVAILABLE_SPEEDS } from './constants';
+
+const PlaybackSpeed = ({
```
diegolmello, line 8: To review with design: changing this value in one audio is supposed to change on all audios?
Should it have an MMKV item for that?

app/containers/AudioPlayer/index.tsx
```diff
@@ -0,0 +1,152 @@
...
+				await audioPlayer.pauseAudio(audioUri.current);
+			} else {
+				await audioPlayer.playAudio(audioUri.current);
+			}
+		} catch {
+			// Do nothing
+		}
+	};
+
+	const onChangeRate = async (value = 1.0) => {
```
diegolmello, line 91: This default value should be somewhere else.
Probably `AVAILABLE_SPEEDS[1]`

app/lib/methods/audioPlayer.ts
```diff
@@ -0,0 +1,119 @@
+import { AVPlaybackStatus, Audio, InterruptionModeAndroid, InterruptionModeIOS } from 'expo-av';
+import { Sound } from 'expo-av/build/Audio/Sound';
```
diegolmello, line 2: Is this import correct/needed?
I'm pretty sure you can do Audio.Sound, but it could be a type, so idk

diegolmello (commented):
Almost there.

app/containers/AudioPlayer/PlaybackSpeed.tsx
reinaldonetof: I talked with Jose and he thinks it's a great idea to do this approach

app/containers/AudioPlayer/PlaybackSpeed.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+	const onPress = () => {
+		const speedIndex = AVAILABLE_SPEEDS.indexOf(playbackSpeed);
+		const nextSpeedIndex = speedIndex + 1 >= AVAILABLE_SPEEDS.length ? 0 : speedIndex + 1;
+		setPlaybackSpeed(AVAILABLE_SPEEDS[nextSpeedIndex]);
+	};
+
+	return (
+		<NativeButton
+			disabled={!(audioState === 'playing')}
```
GleidsonDaniel, line 23:
```suggestion
			disabled={audioState !== 'playing'}
```

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -0,0 +1,132 @@
...
+	useAnimatedStyle,
+	useDerivedValue,
+	useSharedValue
+} from 'react-native-reanimated';
+
+import styles from './styles';
+import { useTheme } from '../../theme';
+import { SEEK_HIT_SLOP, THUMB_SEEK_SIZE } from './constants';
+
+const DEFAULT_TIME_LABEL = '00:00';
```
GleidsonDaniel, line 18: move to constants

app/containers/AudioPlayer/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+interface IAudioPlayerProps {
+	fileUri: string;
+	disabled?: boolean;
+	onPlayButtonPress?: Function;
+	downloadState: TDownloadState;
+}
+
+const AudioPlayer = ({ fileUri, disabled = false, onPlayButtonPress = () => {}, downloadState }: IAudioPlayerProps) => {
+	const loading = downloadState === 'loading';
+	const isReadyToPlay = downloadState === 'downloaded';
```
GleidsonDaniel, line 28:
```suggestion
	const isDownloaded = downloadState === 'downloaded';
```

GleidsonDaniel (changes requested):

GleidsonDaniel (approved):

PR #5177, opened 2023-08-20
feat: properly fetch all settings from server

### LINKED ISSUE #5218: Android app does not handle pagination when requesting /api/v1/settings.public
### Description:

Android client does not handle pagination when requesting /api/v1/settings.public, leading to missed settings (and in my case, no login button shown in the app).

### Environment Information:

- Rocket.Chat Server Version: 6.3.6
- Rocket.Chat App Version: 4.40.0.47250
- Device Name: Cubot pocket
- OS Version: Android 11

### Steps to reproduce:

1. Configure 10 or less "Default Count"  in Settings -> General -> REST API
2. Try to open the application and log in
3. Even if Accounts_ShowFormLogin is enabled on the server, the Login button won't show up (as that's usually below the rest)

### Expected behavior:

Login button to show. (the app to follow the pagination)

### Actual behavior:

No login button shows up.

### Additional context:

The query sent by the App is:

```
https://URL/api/v1/settings.public?query={%22_id%22:{%22$in%22:[%22API_Gitlab_URL%22,%22CAS_enabled%22,%22CAS_login_url%22,%22Accounts_EmailVerification%22,%22Accounts_ManuallyApproveNewUsers%22,%22Accounts_ShowFormLogin%22,%22Site_Url%22,%22Accounts_RegistrationForm%22,%22Accounts_RegistrationForm_LinkReplacementText%22,%22Accounts_EmailOrUsernamePlaceholder%22,%22Accounts_PasswordPlaceholder%22,%22Accounts_PasswordReset%22,%22Accounts_iframe_enabled%22,%22Accounts_Iframe_api_url%22,%22Accounts_Iframe_api_method%22]}}
```

The return result, if the default count is less than 10, ends with:

```
  "count": 10,
  "offset": 0,
  "total": 15,
  "success": true
```

The app does not try to get the rest, but just proceeds with whatever was returned.

At some point, the default count was 10, and it's possible that during some upgrade its increase was missed, thus leading to the issue.

paviad:
## Proposed changes

There is a limitation in the code where if the mobile client needs more than a page full of settings (determined by default paging settings) - only the first page will be retrieved and some settings will be left out.

This PR iterates over the responses if more pages are available, and collects all results to provide a complete settings picture to the client.

## Issue(s)	

closes #5218

## How to test or reproduce

Add more settings to `app\lib\constants\defaultSettings.ts` such that their count is much greater than the default page length of queries. Use a network inspector to confirm that only the first page of settings is requested.

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/getSettings.ts
+++ b/app/lib/methods/getSettings.ts
@@ -152,7 +152,7 @@ export async function getSettings(): Promise<void> {
 		// RC 0.60.0
 		let offset = 0;
 		let remaining;
-		let settings = [];
+		let settings: IData[] = [];
 
 		// Iterate over paginated results to retrieve all settings
 		do {
--- a/app/lib/methods/getSettings.ts
+++ b/app/lib/methods/getSettings.ts
@@ -150,16 +150,31 @@ export async function getSettings(): Promise<void> {
 		const db = database.active;
 		const settingsParams = Object.keys(defaultSettings).filter(key => !loginSettings.includes(key));
 		// RC 0.60.0
-		const result = await fetch(
-			`${sdk.current.client.host}/api/v1/settings.public?query={"_id":{"$in":${JSON.stringify(settingsParams)}}}&count=${
-				settingsParams.length
-			}`
-		).then(response => response.json());
+		let offset = 0;
+		let remaining;
+		let settings = [];
 
-		if (!result.success) {
-			return;
-		}
-		const data: IData[] = result.settings || [];
+		// Iterate over paginated results to retrieve all settings
+		do {
+			// TODO: why is no-await-in-loop enforced in the first place?
+			/* eslint-disable no-await-in-loop */
+			const response = await fetch(
+				`${sdk.current.client.host}/api/v1/settings.public?query={"_id":{"$in":${JSON.stringify(settingsParams)}}}
+				&offset=${offset}`);
+
+			const result = await response.json();
+
+			if (!result.success) {
+				return;
+			}
+
+			offset += result.settings.length;
+			settings = [...settings, ...result.settings];
+			remaining = result.total - settings.length;
+			/* eslint-enable no-await-in-loop */
+		} while(remaining > 0);
+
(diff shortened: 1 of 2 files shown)


app/lib/methods/getSettings.ts
```diff
@@ -146,16 +146,31 @@ export async function getSettings(): Promise<void> {
...
+			// TODO: why is no-await-in-loop enforced in the first place?
+			/* eslint-disable no-await-in-loop */
+			const response = await fetch(
+				`${sdk.current.client.host}/api/v1/settings.public?query={"_id":{"$in":${JSON.stringify(settingsParams)}}}
+				&offset=${offset}`);
+
+			const result = await response.json();
+
+			if (!result.success) {
+				return;
```
djorkaeffalexandre, line 164: if you return in that case, without throwing an error, how can we know that the settings were not fetched? in this case, if one of the request fails, we should retry it until we get the result, and after a few retries, if it doesn't succeed, we should throw an error.

paviad: > if you return in that case, without throwing an error, how can we know that the settings were not fetched? in this case, if one of the request fails, we should retry it until we get the result, and after a few retries, if it doesn't succeed, we should throw an error.

You are correct, but that was the behavior also prior to my PR. I want to focus on just one thing per PR.

paviad: Can we resolve this conversation?

diegolmello: Can you fix it? Although the issue was already there, you're refactoring the whole thing, so it's worth refactoring to a working state.

diegolmello: Ok, no worries. We'll get there.
Gonna review your PR soon.
Thanks for your contribution 👊🏼

diegolmello (approved):
LGTM

diegolmello:
Thanks for your contribution, @paviad!
Awesome work! 👏🏼 👏🏼 👏🏼

PR #5185, opened 2023-08-24
feat: LTS

diegolmello:
## Types of changes

## Checklist


app/containers/SupportedVersions/SupportedVersionsExpired.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { View, Text } from 'react-native';
+
+import { useTheme } from '../../theme';
+import { CustomIcon } from '../CustomIcon';
+import Button from '../Button';
+import { styles } from './styles';
+
+export const SupportedVersionsExpired = () => {
+	const { colors } = useTheme();
+	// TODO: i18n
```
GleidsonDaniel, line 11: 👀

app/containers/SupportedVersions/useSupportedVersionMessage.ts
```diff
@@ -0,0 +1,46 @@
+import moment from 'moment';
+
+import { useAppSelector } from '../../lib/hooks';
+
+const applyParams = (message: string, params: Record<string, unknown>) => {
```
GleidsonDaniel, line 5: I think it's worth a test or even a description of how this function works.

app/sagas/selectServer.ts
```diff
@@ -16,75 +24,127 @@ import log, { logServerVersion } from '../lib/methods/helpers/log';
...
+		return validVersion;
+	}
+	const coercedVersion = coerce(version);
+	if (coercedVersion) {
+		validVersion = valid(coercedVersion);
+	}
+	if (validVersion) {
+		return validVersion;
+	}
+	throw new Error('Server version not found');
```
GleidsonDaniel, line 61: This would be to generate the logs in bugsnag, right?
But will this information alone be enough to come to any conclusions?

app/sagas/selectServer.ts
```diff
@@ -130,30 +194,27 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
...
-		}
-
 		// Return server version even when offline
-		const serverVersion = (serverInfo && serverInfo.version) || version;
+		const serverVersion = (serverInfo && serverInfo.version) || (version as string);
 
 		// we'll set serverVersion as metadata for bugsnag
 		logServerVersion(serverVersion);
-		yield put(selectServerSuccess(server, serverVersion));
+		yield put(selectServerSuccess({ server, version: serverVersion, name: serverInfo?.name || 'Rocket.Chat' }));
```
GleidsonDaniel, line 202: Is this Rocket.Chat really fixed?

app/views/RoomsListView/index.tsx
```diff
@@ -456,17 +479,33 @@ class RoomsListView extends React.Component<IRoomsListViewProps, IRoomsListViewS
 							: // @ts-ignore
 							  () => navigation.toggleDrawer()
 					}
-					badge={() => (notificationPresenceCap ? <HeaderButton.BadgeWarn /> : null)}
+					badge={() => getBadge()}
```
GleidsonDaniel, line 482:
```suggestion
					badge={getBadge}
```

GleidsonDaniel (commented):
Apart from the console.log and the i18n issue, the PR is fine.

app/sagas/selectServer.ts
diegolmello: It should never get there. It's just theoretically possible, since semver throws.
If we find any issues, then we can improve it, if needed.
Unless you have anything in mind.

app/sagas/selectServer.ts
diegolmello: Also a ts thing. It can be undefined, but it shouldn't.
In case it is, we fallback to that.

app/views/RoomsListView/index.tsx
diegolmello: It's a choice, really.
It doesn't change anything and I've been thinking in the cases when the callback has a param, like `onPress` and its `nativeEvent`.
Of course we use it only if needed, but it's more like an explicit indicator of no props.
Again, just a choice 😂

app/containers/SupportedVersions/useSupportedVersionMessage.ts
diegolmello: Yeah, tested this hook here https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5185/files#diff-ef24f5e6b36852d882d4b1c3c8e5e36add01749f3a3bb743b8ce96d2c1c60cf7R70

It's very weird to have the hook tested on `SupportedVersionsWarning`, but it's per their documentation: https://github.com/testing-library/react-hooks-testing-library#when-not-to-use-this-library
> Your hook is defined alongside a component and is only used there
> Your hook is easy to test by just testing the components using it

Still, I started writing tests ignoring docs for the sake of readability of `hook.test.ts`, only to realize it's a *lot* easier to test using the RNTL rather than RHTL.
It makes too much use of Redux, so with a mounted component, it's just a matter of having a Provider and dispatching changes.
With the hooks lib, I'd have to recreate the whole state for every test.

app/lib/services/getServerInfo.ts
```diff
@@ -0,0 +1,114 @@
+import RNFetchBlob from 'rn-fetch-blob';
+import { settings as RocketChatSettings } from '@rocket.chat/sdk';
+import { KJUR } from 'jsrsasign';
+
+import { TCloudInfo, IServerInfo, ISupportedVersions, ISupportedVersionsData, IApiServerInfo } from '../../definitions';
+import { selectServerFailure } from '../../actions/server';
+import { store } from '../store/auxStore';
+import I18n from '../../i18n';
+
+const PUBLIC_KEY = `-----BEGIN PUBLIC KEY-----
```
GleidsonDaniel, line 10: I think this would be better in another file, even for visibility issues.

app/lib/services/getServerInfo.ts
```diff
@@ -0,0 +1,114 @@
...
+		success: false,
+		message: I18n.t('Not_RC_Server', { contact: I18n.t('Contact_your_server_admin') })
+	};
+}
+
+export const getCloudInfo = async (): Promise<TCloudInfo | null> => {
+	const uniqueId = store.getState().settings.uniqueID;
+	const response = await RNFetchBlob.fetch(
+		'GET',
+		`https://releases.rocket.chat/v2/server/supportedVersions?workspaceId=${uniqueId}&source=mobile`
```
GleidsonDaniel, line 111: This url could also be in another file, since it is fixed, constant or something like that.

app/lib/services/index.ts
```diff
@@ -1,7 +1,9 @@
 import * as connect from './connect';
 import * as restApi from './restApi';
+import { getServerInfo } from './getServerInfo';
 
 export const Services = {
 	...connect,
-	...restApi
+	...restApi,
+	getServerInfo
```
GleidsonDaniel, line 8: I don't think this function is a "Service", it is a repo method that has an API call.

GleidsonDaniel (changes requested):

app/lib/services/index.ts
diegolmello: @GleidsonDaniel I was afraid of doing so and making it harder to review.
It's not a new file and moving it could mess up git to the point it would think it's a new file.
Tried it anyway in hope of being smart enough to understand it was just moved.
Did it and it's a new file 🤡 
Some git nerds could say I'm the one to blame by not knowing it enough to predict it, but idk...

app/lib/methods/checkSupportedVersions-enforcement.test.ts
diegolmello, line 1: Duplicated test file because all of these are temp only and it changes logic based on enforcementStartDate.
I'm going to remove it on next release.

PR #5198, opened 2023-09-01
feat: troubleshoot notification

reinaldonetof:
## Proposed changes

Create the notifications button in the RoomListView and redirect the user to the troubleshooting page! Also the View's layout is done too.

## Screenshots
Screen Shot 2023-09-18 at 17 55 29

## Types of changes

## Checklist

## Further comments

[TC-922]

[TC-922]: https://rocketchat.atlassian.net/browse/TC-922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/constants/colors.ts
```diff
@@ -12,6 +12,15 @@ export const SWITCH_TRACK_COLOR = {
 	true: '#2de0a5'
 };
 
+const pushTroubleshooting = {
+	pushConsumptionOnSuccess: '#148660',
```
GleidsonDaniel, line 16: use figma colors name

GleidsonDaniel (changes requested):

app/lib/constants/colors.ts
reinaldonetof: I understand, but there are already `successColor` and `dangerColor` in this file, if I just added `onSuccess` or `onDanger` it would be weird. The `dangerColor` is used in other places of the app and the difference between the names will be minimal, but there is a difference between the colors. WDYT?

GleidsonDaniel: It's not strange, the right thing is to correct the mistake of the past and not make the same mistake again because it was wrong in the past.

GleidsonDaniel: Even a good thing would be to check the colors used in figma, Jose is using some deprecated colors, you could ask him to change and use the colors that are already in the new design system.

GleidsonDaniel (approved):

PR #5219, opened 2023-09-21
fix: flaky test at auto-translate e2e test

reinaldonetof:
#### Before

Here we are expecting that the translation from `cat - fzaY1kiic8` should be `gato - fzaY1kiic8`, but we receive `gato -fzaY1kiic8` without the last space between the `hyphen(-)` and the `f`.
Screen Shot 2023-09-20 at 19 22 09

Screen Shot 2023-09-20 at 19 22 17

#### After

Screen Shot 2023-09-21 at 12 37 01

Screen Shot 2023-09-21 at 12 36 48

## Types of changes

## Checklist


GleidsonDaniel (approved):

e2e/tests/room/11-autoTranslate.spec.ts
```diff
@@ -171,35 +171,35 @@ describe('Auto Translate', () => {
 
 	it('should see new message translated', async () => {
 		const randomMatcher = random();
-		const data = await sendMessageOnTranslationTestRoom(`${newMessage[languages.default]} - ${randomMatcher}`);
-		await searchMessage(`${newMessage[languages.default]} - ${randomMatcher}`, textMatcher); // will scroll the messages list to the last one
-		await waitForVisibleTextMatcher(`${newMessage[languages.translated]} - ${randomMatcher}`, textMatcher);
+		const data = await sendMessageOnTranslationTestRoom(`${newMessage[languages.default]}. Rc${randomMatcher}`);
```
diegolmello, line 174: Correct me if I'm wrong, but if you're creating a new room for this test, you don't need a random string on each message. That makes it easier to translate the content.

diegolmello (commented):
I don't think it's a chore, nor a tweak. It's a fix.

e2e/tests/room/11-autoTranslate.spec.ts
reinaldonetof: We aren't creating a new room for this. We are using the same room for that.
So, if we don't use the "random" content, then run multiple tests in parallel one will interfere with another because the label will be the same

PR #5250, opened 2023-10-02
feat: create notification in room view

reinaldonetof:
## Screenshots
Screen Shot 2023-10-02 at 12 31 58

## Types of changes

## Checklist

## Further comments

[TC-928]

[TC-928]: https://rocketchat.atlassian.net/browse/TC-928?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/RoomView/RightButtons.tsx
```diff
@@ -52,9 +55,12 @@ interface IRigthButtonsState {
 	tunreadGroup: string[];
 }
 
+const notificationIsEnabled = true;
```
GleidsonDaniel, line 58: Why is this const fixed?

app/views/RoomView/RightButtons.tsx
```diff
@@ -288,6 +297,31 @@ class RightButtonsContainer extends Component<IRightButtonsProps, IRigthButtonsS
...
+		const { rid, navigation, isMasterDetail } = this.props;
+
+		if (!rid || !room) {
+			return;
+		}
+		if (notificationIsEnabled && room) {
+			if (isMasterDetail) {
+				navigation.navigate('ModalStackNavigator', {
+					screen: 'NotificationPrefView',
+					params: { rid, room }
```
GleidsonDaniel, line 311: I think we talked about this type of approach a time ago.
Here you are passing all the subscription via params to another room, I'm almost sure you don't need all the props, and even if it were necessary the "correct" thing would be to pass only the `asPlain` object.

app/views/RoomView/RightButtons.tsx
```diff
@@ -346,6 +380,12 @@ class RightButtonsContainer extends Component<IRightButtonsProps, IRigthButtonsS
 		}
 		return (
 			<HeaderButton.Container>
+				<HeaderButton.Item
+					color={notificationIsEnabled ? themes[theme].headerTintColor : themes[theme].fontDanger}
```
GleidsonDaniel, line 384: use `withTheme` and use `colors`

GleidsonDaniel (changes requested):

app/views/RoomView/RightButtons.tsx
reinaldonetof: to test the layout and create a variable with a name as expected

app/views/RoomView/RightButtons.tsx
reinaldonetof: Why? Isn't the best way to keep the same pattern for LeftButtons and RightButtons? Then when we decide to refactor both components, the developer can understand the expected behavior.

app/views/RoomView/RightButtons.tsx
reinaldonetof: I understood what you commented, however, at the `NotificationPreferencesView` is expected the room with observe. Otherwise, I will need to refactor the `NotificationPreferencesView` too.

Add to this, the `NotificationPrefView` is only enable to open when there is the room provided by the database

GleidsonDaniel: I replied private

app/views/RoomView/RightButtons.tsx
GleidsonDaniel: Will this be removed then?

app/views/RoomView/RightButtons.tsx
GleidsonDaniel: Not if the pattern is wrong.
The right thing to do is to tidy things up little by little as you move, if you are adding something new, correct the mistakes of the past.
Like migrating to hooks, something you did in the audio player.

app/views/RoomView/RightButtons.tsx
reinaldonetof: https://rocketchat.atlassian.net/browse/TC-991

GleidsonDaniel:
CI is failing

_From now on, it's required from the PR's author to have E2E tests passing on CI before asking for review._

GleidsonDaniel (approved):

PR #5259, opened 2023-10-06
feat: handle image base64

reinaldonetof:
## How to test or reproduce

- The easy way to test is to search for the channel: image-base64-channel
- join the channel
- look at the preview of a docs.google

## Screenshots
Screen Shot 2023-10-09 at 11 34 06

## Types of changes

## Checklist

## Further comments

[TC-985]

[TC-985]: https://rocketchat.atlassian.net/browse/TC-985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/methods/handleBase64.ts
diegolmello, line 1: This file is named `handleBase64` and the test is `handleIMAGEBase64`

app/lib/methods/handleBase64.ts
```diff
@@ -0,0 +1,21 @@
...
+		return result.replace(imageBase64RegExp, '');
+	}
+}
+
+export function getBase64MimeType(data: string): string | undefined {
+	const mimeType = data.match(mimeTypeBase64RegExp);
+	if (mimeType) {
+		return mimeType[1];
+	}
+}
```
diegolmello, line 21: Where did you get this from?

app/lib/methods/handleMediaDownload.ts
```diff
@@ -199,18 +221,23 @@ export function downloadMediaFile({
...
 				return resolve(result.uri);
 			}
-			reject();
-		} catch {
-			reject();
+			return reject();
+		} catch (e) {
+			return reject();
 		} finally {
 			delete downloadQueue[downloadKey];
```
diegolmello, line 242: I'm _pretty_ sure `finally` runs even with an early `return` like the one you added, but please test to confirm it's actually removing from queue.

diegolmello (changes requested):

app/lib/methods/handleImageBase64.test.ts
```diff
@@ -0,0 +1,49 @@
+import { sanitizeLikeString } from '../database/utils';
+import { getBase64MimeType, isImageBase64, valueOfFirst120CharactersOfImageBase64 } from './handleBase64';
+
+const base64 =
```
GleidsonDaniel, line 4: If the image is the same, why is it duplicated in the code?
Creates a constant with this image in another file.

app/lib/methods/handleBase64.ts
```diff
@@ -0,0 +1,21 @@
+export const imageBase64RegExp = new RegExp(/^data:image\/[bmp,gif,ico,jpg,png,svg,webp,x\-icon,svg+xml]+;base64,/);
+const regExpOf120Characters = new RegExp(/^data:image\/[bmp,gif,ico,jpg,png,svg,webp,x\-icon,svg+xml]+;base64,([\w+\/=]{1,120})/);
+const mimeTypeBase64RegExp = new RegExp(/^data:(.+);base64,/);
+
+export function isImageBase64(data: string): boolean {
+	return !!data && imageBase64RegExp.test(data);
+}
+
+export function valueOfFirst120CharactersOfImageBase64(imageBase64: string): string | undefined {
```
GleidsonDaniel, line 9: I don't think you need this function, you just use it to set the base64 name, a New Date() already solves the problem and is less expensive.

GleidsonDaniel: This would also make your PR simpler and easier to understand, as it would remove two regex and several tests.

GleidsonDaniel (changes requested):

app/lib/methods/handleMediaDownload.ts
reinaldonetof: Yes, it is working. This `return` fixes that infinite loading when occurs any problem and the file keeps the state of downloading, instead back the state to `to download` or `downloaded`

app/lib/methods/handleBase64.ts
reinaldonetof: The regExp I found here: https://regex101.com/library/XAQEkg?filterFlavors=javascript&filterFlavors=golang&orderBy=MOST_RECENT&search=

I just modified a bit to get the 120 first characters to avoid testing all the data as react-native-ui-lib does here: https://github.com/wix/react-native-ui-lib/blob/32ad53c9d864c34d0df35c853b5c9f82c234dff3/src/utils/imageUtils.ts#L14-L18

Add to this, I selected 120 characters after the header of base64 to differentiate between the images, but this value is in my opinion.

app/lib/methods/handleBase64.ts
reinaldonetof: But if I create a New Date() only, every time the image component renders on the view, will download new data, because we can't track properly his filename or path

app/lib/methods/handleImageBase64.test.ts
reinaldonetof: IMHO isn't needed, because I will need to export a const from a `file.test`, right? And this is just a small data, inside 2 different tests files

app/lib/methods/handleBase64.ts
reinaldonetof: A long time ago I discussed this with @diegolmello and he asked me to not work with messageId.

Issue #5237, opened 2023-09-28
Image Attachments are not using CDN_PREFIX


nmagedman:
### Description:

The RC server has a setting called `CDN_PREFIX` which, on the web client, causes assets such as FileUploads (image attachments) to be served via a CDN rather than directly from the RC servers.  The mobile app does not honor `CDN_PREFIX` and requests the images directly from RC.  This puts unnecessary load on the RC servers and leaves the CDN under-utilized.

### Environment Information:

- Rocket.Chat Server Version: 6.3.3
- Rocket.Chat App Version: 4.40.0.47248
- Device Name: iPhone SE
- OS Version: iOS 17.0 (21A329)

### Steps to reproduce:

This can be demonstrated in two ways:
1. iOS App Privacy Report
2. Add logging to the RC server 

#### iOS App Privacy Report

iOS supports some (very limited) monitoring of an app's network activity.  You can get a list of which domain names an app has accessed recently.  Launch the RC mobile app and view some messages that have image attachments and then look at the list of domains that have been recently accessed.  We should expect that both the RC server domain and the CDN domain will be in the list.  Instead we only see the former.

To access the list (instructions based on iOS 17), go to:

1. Settings App
1. Privacy & Security
1. App Privacy Report (found all the way down the page)
1. § APP NETWORK ACTIVITY
1. Show All
1. Rocket.Chat
1. § DOMAINS CONTACTED DIRECTLY BY APP
1. See that the RC server (e.g. `rc.seekingalpha.com`) is present but the CDN (e.g. `rc-cdn.seekingalpha.com`) is not

#### Add logging to the RC server 

1. The RC server currently does not log requests to `/file-upload/:upload_id/:filename`.  We will add some simple logging of the request and its headers.  Using RC 6.3.3 as a base, apply the following patch

    ```diff
    diff --git a/apps/meteor/app/file-upload/server/lib/requests.ts b/apps/meteor/app/file-upload/server/lib/requests.ts
    index 91bb50b179..c88ee45fb2 100644
    --- a/apps/meteor/app/file-upload/server/lib/requests.ts
    +++ b/apps/meteor/app/file-upload/server/lib/requests.ts
    @@ -2,8 +2,22 @@ import { WebApp } from 'meteor/webapp';
    import { Uploads } from '@rocket.chat/models';

    import { FileUpload } from './FileUpload';
    +import { Logger } from '../../../../server/lib/logger/Logger';

[...]

PR #5266, opened 2023-10-11
feat: add the support for CDN_PREFIX

reinaldonetof:
## Proposed changes

There are three properties for CDN:
 - CDN_PREFIX: when there is a param, this will replace the server for the cdn_prefix
 - CDN_PREFIX_ALL: flag to Use CDN Prefix for all assets (this property doesn't affect the mobile, because if this is false, enables the CDN Prefix for JS/CSS to add the URL)
 - CDN_JSCSS_PREFIX: CDN Prefix for JS/CSS

Considering this and the mobile doesn't use cdn for JS/CSS, we only need to know if there is the CDN_PREFIX, then use it.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/5237

## Screenshots
Screen Shot 2023-10-11 at 12 02 22

- Avatar, Image, Audio, Video
Screen Shot 2023-10-11 at 14 11 33

## Types of changes

## Checklist

## Further comments

[TC-975]

[TC-975]: https://rocketchat.atlassian.net/browse/TC-975?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/message/Attachments.tsx
```diff
@@ -58,6 +59,7 @@ const Attachments: React.FC<IMessageAttachments> = React.memo(
 	({ attachments, timeFormat, showAttachment, style, getCustomEmoji, isReply, author }: IMessageAttachments) => {
 		const { theme } = useTheme();
 		const { translateLanguage } = useContext(MessageContext);
+		const cdnPrefix = useAppSelector(state => state.settings.CDN_PREFIX as string);
```
GleidsonDaniel, line 62: In the three different components you use cdnPrefix as a prop, and you only use this cdnPrefix in one function.
The best thing would be to do the same as we do in other functions and use the auxStore to access the redux of that function.
cdnPrefix is not something that will change, so it is not really necessary to listen to redux changes.
In addition to making your code MUCH cleaner, since you changed 3 different files using the same property.

GleidsonDaniel: In audio, put the redux in the component and that's it.

GleidsonDaniel (changes requested):

GleidsonDaniel:
CI is failing

_From now on, it's required from the PR's author to have E2E tests passing on CI before asking for review._

GleidsonDaniel (approved):

Issue #5318, opened 2023-11-01
Missing translations admin-NOT_CONFIGURED-header, admin-NOT_CONFIGURED-body


paulchen:
### Description:

Some translations are missing.

### Environment Information:

- Rocket.Chat Server Version: 6.4.4
- Rocket.Chat App Version: 4.42.0.50842
- Device Name: ELE-L29 (Huawei P30)
- OS Version: EMUI 12.0.0 (Android 11)

### Steps to reproduce:

1. Use an instance of Rocket.Chat without an app for conference calls being configured.
2. Login with a user having the "Call Management" permission.
3. Tap the phone icon in any channel.

### Expected behavior:

Some message about no app being configured is shown.

### Actual behavior:

![image](https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/404840/59895712-5142-4070-b472-c65427c6cc63)

### Additional context:

Translations are missing for at least English and German.

PR #5332, opened 2023-11-10
fix: fixes error checking related to error translations when starting a call

GleidsonDaniel:
Fixes error checking related to error translations when starting a call

## Issue(s)	

resolve: #5318

## How to test or reproduce

Use an instance of Rocket.Chat without an app for conference calls being configured.
Login with a user having the "Call Management" permission.
Tap the phone icon in any channel.

## Screenshots
Before:
Simulator Screenshot - iPhone 14 - 2023-11-10 at 14 14 34

After:
Simulator Screenshot - iPhone 14 - 2023-11-10 at 14 14 21

## Types of changes

## Checklist

## Further comments

Jira: [VC-156]


app/lib/hooks/useVideoConf/index.tsx
```diff
@@ -45,7 +45,7 @@ export const useVideoConf = (
 				return true;
 			} catch (error: any) {
 				const isAdmin = !!user.roles?.includes('admin');
-				handleErrors(isAdmin, error?.error || 'NOT_CONFIGURED');
+				handleErrors(isAdmin, error?.data?.error || availabilityErrors.NOT_CONFIGURED);
```
reinaldonetof, line 48: My suggestion is inside the handleErrors, maybe instead of simple call the `i18n.t(`${key}-body`), i18n.t(`${key}-header`)`, I think you should test whether the error is already translate with i18n.isTranslated, then you show the error. This approach will avoid show to the user `[missing en.{key} translation]`. 

WDYT?

reinaldonetof (changes requested):
```
const handleErrors = (isAdmin: boolean, error: keyof typeof availabilityErrors) => {
	const key = isAdmin ? `admin-${error}` : error;
	showErrorAlert(i18n.t(`${key}-body`), i18n.t(`${key}-header`));
};
```

Considering that you had problem

reinaldonetof (approved):
lgtm

PR #5363, opened 2023-11-23
improve: handle attachment actions in a quote and how to jump to message

reinaldonetof:
## Proposed changes

> So, when quoting a message the default value should be to show/play/download the message/attachment. Only when long pressing on the message you’ll get the option to go to the message, if this message is in a context on which the user cannot read it, they should be alerted as such.

## How to test or reproduce

- Share a video/audio message link from a DM to a group
- Quote an attachment (video, image, audio) 

## Screenshots

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/d8b861cc-d485-4e3b-99fb-1439765b84a9

## Types of changes

## Checklist

## Further comments

[TC-284]


app/views/RoomView/index.tsx
```diff
@@ -1018,8 +1018,12 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
...
 				this.cancelJumpToMessage();
 			}
-		} catch (e) {
-			log(e);
+		} catch (error: any) {
+			if (isFromReply && error.data?.errorType === 'error-not-allowed') {
+				showErrorAlert(I18n.t('The_room_does_not_exist'), I18n.t('Room_not_found'));
+			} else {
+				log(error);
+			}
```
reinaldonetof, line 1026: I don't have enough information to determine if every time the error is 'error-not-allowed' in this case, 'The_room_does_not_exist' should appear.

That's why I decided to add the 'isFromReply' condition.

app/containers/MessageActions/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable complexity */
```
GleidsonDaniel, line 1: I don't think it's interesting to hide this "problem".
The warning is there for a reason, ignore it and it is not the best solution in most cases, especially since there is not even a TODO specifying that we should resolve this problem later.

app/containers/message/Video.tsx
```diff
@@ -93,26 +99,12 @@ const Video = ({ file, showAttachment, getCustomEmoji, style, isReply, msg }: IM
...
-						cancelDownload(video);
-					}
+				const isVideoCached = await handleGetMediaCache();
+				if (isVideoCached) {
 					return;
 				}
-				if (isReply) {
-					setLoading(false);
+				if (isDownloadActive(video)) {
+					await handleResumeDownload();
```
GleidsonDaniel, line 107:
```suggestion
				handleResumeDownload();
```

GleidsonDaniel (approved):
Don't forget to solve those two details that I mentioned, they are not impediments however.

GleidsonDaniel:
Another detail, maybe you forgot about this conversation here:

Unless it's for testing, don't run the CI build before the PR is approved, you don't know if it will go back to changes requested.

PR #5377, opened 2023-12-01
fix: Cannot read property 'protocol' of undefined

reinaldonetof:
## Proposed changes

Since the server version is 6.5.0, is popping up some errors at bugsnag complain: Cannot read property 'protocol' of undefined

image

And looking what was modified at the server, we found this PR: https://github.com/RocketChat/Rocket.Chat/pull/30228 that was responsible for remove the "parsedUrl" object from the result. Also, to fix the behavior at app, we decided to mirror the `buildImageURL` from the frontend https://github.com/RocketChat/Rocket.Chat/blob/5c145e3170f04e341be93a2a60f09b6cbdc46c73/apps/meteor/client/components/message/content/urlPreviews/buildImageURL.ts#L3

## Types of changes

## Checklist

## Further comments

[TC-1045]

[TC-1045]: https://rocketchat.atlassian.net/browse/TC-1045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/methods/helpers/parseUrls.ts
```diff
@@ -16,9 +17,19 @@ export default (urls: IUrlFromServer[]): IUrl[] =>
 			tmp.image = decodedOgImage || meta.twitterImage || meta.oembedThumbnailUrl;
 			if (tmp.image) {
 				if (tmp.image.indexOf('//') === 0) {
-					tmp.image = `${url.parsedUrl.protocol}${tmp.image}`;
-				} else if (tmp.image.indexOf('/') === 0 && url.parsedUrl && url.parsedUrl.host) {
-					tmp.image = `${url.parsedUrl.protocol}//${url.parsedUrl.host}${tmp.image}`;
+					// Remotion of parsedUrl from backend 6.5.0
```
diegolmello, line 20:
```suggestion
					// parsedUrl was removed on 6.5.0
```

app/lib/methods/helpers/buildImageURL.ts
```diff
@@ -0,0 +1,20 @@
+export const isValidLink = (link: string): boolean => {
```
diegolmello, line 1: I can see you reused from backend, but can you move to its own file?
You're exporting, so it might be useful to have it outside image's context.

Can you also rename it to `isValidUrl`?
It makes use of `URL` api, so calling it url instead of link makes more sense.

diegolmello (approved):
Ask for priority QA after the changes please.

PR #5435, opened 2023-12-26
feat: automatically send failed messages on app reconnect

### LINKED ISSUE #4471: Messages are not sent when the Internet is bad from a mobile phone.
При плохом интернете на мобильном телефоне, высвечивается или ошибка или сообщение темное, не отправлено. В это время другие мессенджеры отправляют сообщения, например, Telegram, whatsapp, как можно увеличить количество попыток отправки сообщений с мобильной версии приложений?

jsathu07:
## Proposed changes

The feature to automatically send failed messages when app reconnects

## Issue(s)	

closes #4471 

## Screenshots

After

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/61021881/05a08fb1-021b-4715-85c0-50dcfc538250

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/sendMessage.ts
+++ b/app/lib/methods/sendMessage.ts
@@ -1,5 +1,6 @@
 import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
 import { Model, Q } from '@nozbe/watermelondb';
+// import moment from 'moment';
 
 import database from '../database';
 import log from './helpers/log';
@@ -11,6 +12,8 @@ import { E2E_MESSAGE_TYPE, E2E_STATUS, messagesStatus } from '../constants';
 import { sendFileMessage } from './sendFileMessage';
 import { store } from '../store/auxStore';
 import { getUserSelector } from '../../selectors/login';
+
+// console.log(moment.unix(0).toDate());
 
 export async function resendFailedMessages(): Promise<void> {
 	try {
--- a/app/lib/methods/sendMessage.ts
+++ b/app/lib/methods/sendMessage.ts
@@ -35,8 +35,9 @@ export async function resendFailedMessages(): Promise<void> {
 			resendMsgs.push(resendMessage(message));
 		});
 
-		threadRecords.forEach((message: TMessageModel) => {
-			resendMsgs.push(resendMessage(message, message.rid));
+		threadRecords.forEach(async (message: TMessageModel) => {
+			const msg = await db.get('messages').find(message.id); // tshow only exists on this db
+			resendMsgs.push(resendMessage(message, message.rid, msg.tshow));
 		});
 
 		await Promise.allSettled(resendMsgs);
@@ -99,7 +100,7 @@ async function sendMessageCall(message: any) {
 	return changeMessageStatus(_id, messagesStatus.ERROR, tmid);
 }
 
-export async function resendMessage(message: TMessageModel, tmid?: string) {
+export async function resendMessage(message: TMessageModel, tmid?: string, tshow?: boolean) {
 	const db = database.active;
 	try {
 		await db.write(async () => {
@@ -111,7 +112,8 @@ export async function resendMessage(message: TMessageModel, tmid?: string) {
 			_id: message.id,
 			rid: message.subscription ? message.subscription.id : '',
 			msg: message.msg,
-			...(tmid && { tmid })
+			...(tmid && { tmid }),
+			...(tshow && { tshow })
 		} as IMessage);
 
 		await sendMessageCall(m);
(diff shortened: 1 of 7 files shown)


jsathu07:
@diegolmello I have implemented the feature. Kindly review the code at your convenience sir, Thank you.

jsathu07:
@diegolmello Should I also consider resending attachments too?

GleidsonDaniel:
> @diegolmello Should I also consider resending attachments?

I don't think so, it's better to deal with it in another PR and another function.
Looking at the method, I don't think there's anything wrong, but @diegolmello  has more knowledge of all the APP's flows, so I leave this double check to him.

app/lib/methods/sendMessage.ts
```diff
@@ -9,6 +9,42 @@ import { E2EType, IMessage, IUser, TMessageModel } from '../../definitions';
...
+
+		const resendMsgs: Promise<void>[] = [];
+
+		messageRecords.forEach((message: TMessageModel) => {
+			resendMsgs.push(resendMessage(message));
+		});
+
+		threadRecords.forEach((message: TMessageModel) => {
+			resendMsgs.push(resendMessage(message, message.rid));
+		});
```
diegolmello, line 40: I'm wondering if we should unify these arrays and sort by ts again, to make sure messages are in order.
Can you test?

app/sagas/login.js
```diff
@@ -240,6 +241,7 @@ const handleLoginSuccess = function* handleLoginSuccess({ user }) {
 			yield put(inviteLinksRequest(inviteLinkToken));
 		}
 		yield showSupportedVersionsWarning(server);
+		yield resendFailedMessages();
```
diegolmello, line 244: Move to below `yield fork(fetchUsersRoles);` with a fork, so it doesn't `await` this saga.

diegolmello (changes requested):

diegolmello:
> @diegolmello Should I also consider resending attachments?

> > @diegolmello Should I also consider resending attachments?
> 
> I don't think so, it's better to deal with it in another PR and another function. Looking at the method, I don't think there's anything wrong, but @diegolmello has more knowledge of all the APP's flows, so I leave this double check to him.

@jsathu07 I'd do it on the same function, because of where it's called and timestamps.
I think it's better to do it on this PR as well, because It's going to look weird if it resends only half of failing messages.

app/lib/methods/sendMessage.ts
jsathu07: @diegolmello Yes you're correct that would be safer option. but in the real scenario the probability is very low as people have decent internet connection and no one would send many msgs at once

app/sagas/login.js
jsathu07: Sure

jsathu07:
> > @diegolmello Should I also consider resending attachments?
> 
> > > @diegolmello Should I also consider resending attachments?
> > 
> > 
> > I don't think so, it's better to deal with it in another PR and another function. Looking at the method, I don't think there's anything wrong, but @diegolmello has more knowledge of all the APP's flows, so I leave this double check to him.
> 
> @jsathu07 I'd do it on the same function, because of where it's called and timestamps. I think it's better to do it on this PR as well, because It's going to look weird if it resends only half of failing messages.

Sure sir I'll implement that too

jsathu07:
@diegolmello How the app should behave if it needs to resend msgs and attachments (doesn't have ts)

diegolmello:
@jsathu07 Can you elaborate your question?

jsathu07:
@diegolmello Sir, if someone sends text messages and uploads images, the current code doesn't provide information about when the uploads were started. To address this, when resending messages, should I initiate resending of all uploads first and then the messages or do it the other way around. Another option is to include a "created at" field in the uploads database, sort all messages based on timestamps, and then initiate the resending process.

diegolmello:
> @diegolmello Sir, if someone sends text messages and uploads images, the current code doesn't provide information about when the uploads were started. To address this, when resending messages, should I initiate resending of all uploads first and then the messages or do it the other way around. Best option is to include a "created at" field in the uploads database, sort all messages based on timestamps, and then initiate the resending process. Please share your thoughts whether I can modify the db

@jsathu07 If what's missing is a date column to sort, I say we create `Uploads.ts` column to do it. I didn't find any `created_at`, so I think `ts` is better just for the sake of consistency. 
Don't forget to add a migration.

diegolmello:
> @diegolmello Sir, if user has any attachments which needs to be resend, which were initiated before migration how should I handle it? Should I leave it without resending so that user can initiate resending if they want

@jsathu07 Yes, you can do it like that.

jsathu07:
@diegolmello, I've completed the feature sir. Could you review it when you have time, Thanks

jsathu07:
@diegolmello tell me if there is anything I can do to improve or change in this pr sir. Thanks

diegolmello:
@jsathu07 In a few days. We have other priorities rn.

BoutPure:
+1. Really wait for this feature because very tired of copypasting gray messages to resend them.

PR #5447, opened 2023-12-28
fix: crashing the app - Cannot read property 'find' of null at fetchRole method

reinaldonetof:
## Proposed changes

It is crashing the app when the roomRoles is null. This is possible because the roomRoles state starts as null. And only change to an array after called the `fetchRoomMembersRoles`, but this method can fail and don't update correctly the roomRoles state. Maybe, because of this it's occurring this problem. However, I wasn't able to reproduce the same error/crash running the happy path, but I was able to reproduce in debug mode setting the roomRoles as null inside the method and the crash occurred.

To fix and improve the readability I added the appropriate type to roomRoles and added the optional mark.

[Bugsnag Link](https://app.bugsnag.com/rocket-dot-chat-1/mobile-1/errors/639aebd1bea15b0008e97567?filters[event.since]=30d&filters[error.status]=open&pivot_tab=event)

## Types of changes

## Checklist

## Further comments

[TC-1041]

[TC-1041]: https://rocketchat.atlassian.net/browse/TC-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/views/RoomMembersView/index.tsx
```diff
@@ -50,7 +50,7 @@ interface IRoomMembersViewState {
 	members: TUserModel[];
 	room: TSubscriptionModel;
 	end: boolean;
-	roomRoles: any;
+	roomRoles: IGetRoomRoles[] | null;
```
GleidsonDaniel, line 53:
```suggestion
	roomRoles?: IGetRoomRoles[];
```

app/views/RoomMembersView/helpers.ts
```diff
@@ -18,9 +18,9 @@ const handleGoRoom = (item: TGoRoomItem, isMasterDetail: boolean): void => {
 	goRoom({ item, isMasterDetail, popToRoot: true });
 };
 
-export const fetchRole = (role: string, selectedUser: TUserModel, roomRoles: any): boolean => {
-	const userRoleResult = roomRoles.find((r: any) => r.u._id === selectedUser._id);
-	return userRoleResult?.roles.includes(role);
+export const fetchRole = (role: string, selectedUser: TUserModel, roomRoles: IGetRoomRoles[] | null): boolean => {
```
GleidsonDaniel, line 21:
```suggestion
export const fetchRole = (role: string, selectedUser: TUserModel, roomRoles?: IGetRoomRoles[]): boolean => {
```

GleidsonDaniel (commented):
I know that null is the value it could be coming from, but in this case, as it is a forced type, like setting it as optional, it becomes less verbose.
I'll send it to QA, then update this PR with the changes before merging.

app/views/RoomMembersView/index.tsx
```diff
@@ -93,7 +93,7 @@ const RoomMembersView = (): React.ReactElement => {
 			members: [],
 			room: params.room || ({} as TSubscriptionModel),
 			end: false,
-			roomRoles: null,
+			roomRoles: undefined,
```
GleidsonDaniel, line 96:
```suggestion
			roomRoles,
```

reinaldonetof: No value exists in scope for the shorthand property 'roomRoles'. Either declare one or provide an initializer.

GleidsonDaniel (approved):

Issue #5424, opened 2023-12-22
Ring does not stop


Lao-Ax:
### Description:

Ring does not stop even after moving to JITSI screen.
See /app/containers/Ringer/index.tsx - first useEffect starts acync operation. It could be done quite **after** the moment when the component was unmounted. On my android device it happens very often.

Excuse me, that I do not propose a PR. As a simple solution could be the follwoing:
1. introduce new ref: `const mounted = useRef(false);` Set it to `true` in the first hook and set it `false` on unmount.
2. Wrap 'play' functions using the const:
```
if (mounted.current) {
  sound.current = expo.sound;
  await sound.current.playAsync();  
  await sound.current.setIsLoopingAsync(true);
}
```
### Environment Information:

- Rocket.Chat Server Version: 6.4.0
- Rocket.Chat App Version: 4.40+; 4.44.0
- Device Name: Huawei P10
- OS Version: Android 9
- videconf provider: jitsi (type = videoconf)

### Steps to reproduce:

1. Go to any chat
2. Start a call
3. Press "call"

### Expected behavior:

Ring stops when jitsi meet screen is opened

### Actual behavior:

Ring constatnly rings.

brf153:
@Lao-Ax you get a ring when someone makes a call in the room?

brf153:
I don't get any

Lao-Ax:
@brf153 
I do have an instant ring when I make a call to anybody. I didn't test a case for incoming call. Will do later.

GleidsonDaniel:
> @Lao-Ax you get a ring when someone makes a call in the room?

There is only a ringer when the call is direct, in group ou team rooms there is no ringer because it is not part of the planning to call several people at the same time, imagine a room with 100 people, 100 people receiving a call at the same time.

brf153:
@GleidsonDaniel I believe Rocket Chat may have deprecated the feature for creating direct chats when using the open Rocket Chat server. Is there a workaround to initiate direct chats? I attempted to resolve this by cloning the server repository and concurrently running the backend and mobile repositories, but my laptop became unresponsive.

Lao-Ax:
@brf153 
Hi, you asked me:
> get a ring when someone makes a call in the room

No, I don't have any. I've also tested the latest version #44: there is also no ring on an incoming call in direct messages neither in channels.

Looks like it is another issue which is worth to fix someday.

GleidsonDaniel:
[...]

PR #5453, opened 2024-01-03
fix: ringer audio playing on background

GleidsonDaniel:
## Proposed changes

This PR fixes the problem with the audio playing in the background when accepting the video call notification.

## Issue(s)	

resolve: #5424

## How to test or reproduce

- With the app open, accept, decline or "ignore" a call from within the app.
- The audio needs to stop playing when you do any of these three actions.

## Types of changes

## Checklist

## Further comments

Jira: [VC-169]

[VC-169]: https://rocketchat.atlassian.net/browse/VC-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/containers/InAppNotification/IncomingCallNotification/index.tsx
```diff
@@ -88,7 +98,7 @@ const IncomingCallHeader = React.memo(
 						<Text style={styles.buttonText}>{i18n.t('accept')}</Text>
 					</Touchable>
 				</View>
-				<Ringer ringer={ERingerSounds.RINGTONE} />
+				{audio ? <Ringer ringer={ERingerSounds.RINGTONE} /> : false}
```
reinaldonetof, line 101:
```suggestion
				{audio ? <Ringer ringer={ERingerSounds.RINGTONE} /> : null}
```

app/containers/Ringer/index.tsx
```diff
@@ -1,41 +1,34 @@
...
-	const stopSound = () => {
-		sound?.current?.unloadAsync();
-	};
+		return () => {
+			sound.current?.unloadAsync();
+		};
+	}, []);
 
-	return <View />;
+	return null;
```
reinaldonetof, line 31: Considering that the Ringer don't need to be a view to reproduce the sound, can change this behavior? wdyt?

reinaldonetof (changes requested):

app/containers/Ringer/index.tsx
GleidsonDaniel: No, because when using a hook its instance will always exist, being a component its instance ceases to exist when the component is disassembled.

reinaldonetof: Ok

reinaldonetof (approved):

PR #5476, opened 2024-01-11
feat: WatchOS app

djorkaeffalexandre:
## Proposed changes

These changes add a Watch app to Rocket.Chat, it will have the core functionality for messages, and stays on pooling the server for new messages and updates on rooms and subscriptions, every 5 seconds.

## Screenshots
| Workspaces | Room List | Message List |
|----|----|----|
| Workspaces | Rooms | Room |
## Types of changes

## Checklist


ios/Shared/RocketChat/Storage.swift
```diff
@@ -1,60 +1,24 @@
...
-      }
-    }
-
-    return nil
-  }
-
-  func getPrivateKey(server: String) -> String? {
-    return self.mmkv?.string(forKey: "\(server)-RC_E2E_PRIVATE_KEY")
-  }
+final class Storage {
```
diegolmello, line 8: Made some changes here https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5596
Can you review?

app/lib/methods/helpers/sslPinning.ts
```diff
@@ -67,6 +67,7 @@ const RCSSLPinning = Platform.select({
 					certificate = persistCertificate(name, certificate.password);
 				}
 				UserPreferences.setMap(extractHostname(server), certificate);
+				SSLPinning?.setCertificate(server, certificate.path, certificate.password);
```
diegolmello, line 70: What's going to happen to the current logged in users making use of pinning?
Are they going to be logged out?
Is this a breaking change?
If it is, should we refactor it to a better approach removing the need for a RN fork?

ios/fastlane/Fastfile
```diff
@@ -57,6 +57,7 @@ platform :ios do
     match(type: "appstore")
     get_provisioning_profile(app_identifier: "chat.rocket.reactnative.ShareExtension")
     get_provisioning_profile(app_identifier: "chat.rocket.reactnative.NotificationService")
+    get_provisioning_profile(app_identifier: "chat.rocket.reactnative.watchkitapp")
```
diegolmello, line 60: Future work of course, but we'll have to do it on Official app as well.

diegolmello (changes requested):
Overall, really good job! :)
Left a few comments in code and here are a few thing I've noticed.

- list of servers' order is random
  * sometimes open can appear before mobile and vice versa
- It should display `Rocket.Chat` on Watch's apps list view
- I couldn't find i18n. It's only `en` currently, right? We should have at least `en`, `pt-BR` and `jp`
- It couldn't load SSL certificate server

ios/Shared/RocketChat/Storage.swift
djorkaeffalexandre: Everything seems to be right on your implementation, as soon as you merge your changes I'll make sure to update my code to stop using this shared instance of Storage, for the WatchApp it should not be too relevant, since it is using only a single instance to read the information when requested by the Apple Watch, and if anything fails we have a chance to offer the user a retry, different of a push notification decryption error.

app/lib/methods/helpers/sslPinning.ts
djorkaeffalexandre: It's not changing anything to the final user, there is no breaking change, I'm just creating a "copy" of the ssl pinning certificate on mmkv, currently our main app logic depends on a path to the certificate, for every place that a request is made we read this file, this can be changed in a future project. I had some issues with this logic based on the path, since for the Watch Connectivity the files can't be read if the app is not in foreground, so I created a code that will copy current certificates as raw data to this new key in our `key-value` storage, and new certificates will be also copied to this new key.

djorkaeffalexandre:
> Overall, really good job! :) Left a few comments in code and here are a few thing I've noticed.
> 
> * list of servers' order is random
>   
>   * sometimes open can appear before mobile and vice versa
> * It should display `Rocket.Chat` on Watch's apps list view
> * I couldn't find i18n. It's only `en` currently, right? We should have at least `en`, `pt-BR` and `jp`
> * It couldn't load SSL certificate server

diegolmello (approved):

PR #5486, opened 2024-01-16
feat: right swipe to reply in quote

### LINKED ISSUE #5478: Feat: Right/Left swipe the message to reply in quote
### Description:

I suggest having a feature like left/right swipe the message to reply in quote. Because we generally use reply in quote in most of times. So it would be better if we have a shortcut for it.

### Environment Information:

- Rocket.Chat Server Version: 6.6.0
- Rocket.Chat App Version: 4.44.2.54511
- Device Name: Realme 9
- OS Version: Android

### Steps to reproduce:

1. Go to Chats

### Expected behavior:

Right/Left swipe the message to reply in quote

### Actual behavior:

Right/Left swipe the message does nothing

sanjaysargam:
## Proposed changes
I suggest having a feature like left/right swipe the message to reply in quote. Because we generally use reply in quote in most of times. So it would be better if we have a shortcut for it.

## Issue(s)	
Fixes #5478

## How to test or reproduce

1. Go to Chats

## Screenshots

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/65113071/54aa9827-1e35-44a0-bb42-ea80643d82f7

## Types of changes

## Checklist

### DIFF
--- a/app/containers/message/index.tsx
+++ b/app/containers/message/index.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Keyboard, ViewStyle } from 'react-native';
+import { Keyboard, ViewStyle, Animated } from 'react-native';
+import { PanGestureHandler, State } from 'react-native-gesture-handler';
 
 import Message from './Message';
 import MessageContext from './Context';
@@ -45,6 +46,7 @@ interface IMessageContainerProps {
 	errorActionsShow?: (item: TAnyMessageModel) => void;
 	replyBroadcast?: (item: TAnyMessageModel) => void;
 	reactionInit?: (item: TAnyMessageModel) => void;
+	replyInit?: (message: TAnyMessageModel, mention: boolean) => void;
 	fetchThreadName?: (tmid: string, id: string) => Promise<string | undefined>;
 	showAttachment?: (file: IAttachment) => void;
 	onReactionLongPress?: (item: TAnyMessageModel) => void;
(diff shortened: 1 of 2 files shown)


app/containers/message/index.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import { Keyboard, ViewStyle } from 'react-native';
+import { Keyboard, ViewStyle, Animated } from 'react-native';
```
diegolmello, line 2: Why would you use stock RN's Animated instead of reanimated?

app/containers/message/index.tsx
```diff
@@ -401,85 +425,96 @@ class MessageContainer extends React.Component<IMessageContainerProps, IMessageC
...
-					navToRoomInfo={navToRoomInfo}
-					handleEnterCall={handleEnterCall}
-					blockAction={blockAction}
-					highlighted={highlighted}
-					comment={comment}
-					isTranslated={isTranslated}
-					isPreview={isPreview}
-				/>
-			</MessageContext.Provider>
+			<PanGestureHandler
```
diegolmello, line 428: You'll have to evaluate and show us if this new gesture is impacting the other buttons inside Message.

diegolmello (commented):
You can't increase the complexity of a component without evaluating the impact of work on the codebase.
Message already has [Touchable](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/containers/message/Touchable.tsx) component.
There you'll see we make use of an already deprecated lib and we need to migrate away from it https://github.com/rnc-archive/react-native-platform-touchable

You should unify and consider the migration to RNGH on all Message's components.

app/containers/message/index.tsx
SanjaySargam: Ok I will use Animated from 'react-native-reanimated'

PR #5534, opened 2024-02-02
fix: the download's key for files with the same filename

reinaldonetof:
## Proposed changes

The error occurred when there were more than 1 file with the same filename, for example:

`https://open.rocket.chat/file-upload/ChJDEKZhJHc359rm8/Screenshot%202024-02-14%20at%2019.54.07.png`

`https://open.rocket.chat/file-upload/4roCoCiHzsxzZzFXv/Screenshot%202024-02-14%20at%2019.54.07.png`

The `mediaDownloadKey` function is responsible for determining the key related to the download object, but it was only capturing the filename from the URL. Thus, when there were 2 files with the same filename, the key was associated with only one, causing the download of the other to appear as if it was also in progress, when in fact, it was referencing the download of the first file. To fix this, I decided not to separate only the filename to use as the key for the object. Instead, I used the complete link as the object key, thus differentiating between each file.

## How to test or reproduce

- Upload two different files with the same name
- When them appears in the screen, you need to click one by one almost at the same time
- Then after the download is done and you click to open the file, you will notice that you are opening the same file
- One point to consider is: if you exit the channel and reopen it, you will notice that one of the files wasn't downloaded. The download icon will appear, and after clicking to download it, the file should be downloaded as expected.

#### Before

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/97c666cb-ca4e-4a91-9991-f513322a1148

#### After

https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/47038980/5b773be2-fb99-465e-a971-9df9baf75d84

## Types of changes

## Checklist

## Further comments

[CORE-55]

[CORE-55]: https://rocketchat.atlassian.net/browse/CORE-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


reinaldonetof:
> You might want to change `serverUrlParsedAsPath` to `sanitizeLikeString` as well, just for the sake of semantics
> 
> https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/b189f59950807b2359c818a9b051140550a00362/app/lib/methods/handleMediaDownload.ts#L26
> 
> 
> At least I can't see a reason to do that on `serverUrl`.
> I'm also missing a clear description of why `sanitizeLikeString` is better than `sanitizeString` and a string as example of what was causing the issue.


The error occurred when there were more than 1 file with the same filename, for example:

`https://open.rocket.chat/file-upload/ChJDEKZhJHc359rm8/Screenshot%202024-02-14%20at%2019.54.07.png`

`https://open.rocket.chat/file-upload/4roCoCiHzsxzZzFXv/Screenshot%202024-02-14%20at%2019.54.07.png`

The `mediaDownloadKey` function is responsible for determining the key related to the download object, but it was only capturing the filename from the URL. Thus, when there were 2 files with the same filename, the key was associated with only one, causing the download of the other to appear as if it was also in progress, when in fact, it was referencing the download of the first file. To fix this, I decided not to separate only the filename to use as the key for the object. Instead, I used the complete link as the object key, thus differentiating between each file.

diegolmello:
> > You might want to change `serverUrlParsedAsPath` to `sanitizeLikeString` as well, just for the sake of semantics
> > https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/b189f59950807b2359c818a9b051140550a00362/app/lib/methods/handleMediaDownload.ts#L26
> > 
> > At least I can't see a reason to do that on `serverUrl`.
> > I'm also missing a clear description of why `sanitizeLikeString` is better than `sanitizeString` and a string as example of what was causing the issue.
> 
> The error occurred when there were more than 1 file with the same filename, for example:
> 
> `https://open.rocket.chat/file-upload/ChJDEKZhJHc359rm8/Screenshot%202024-02-14%20at%2019.54.07.png`
> 
> `https://open.rocket.chat/file-upload/4roCoCiHzsxzZzFXv/Screenshot%202024-02-14%20at%2019.54.07.png`
> 
> The `mediaDownloadKey` function is responsible for determining the key related to the download object, but it was only capturing the filename from the URL. Thus, when there were 2 files with the same filename, the key was associated with only one, causing the download of the other to appear as if it was also in progress, when in fact, it was referencing the download of the first file. To fix this, I decided not to separate only the filename to use as the key for the object. Instead, I used the complete link as the object key, thus differentiating between each file.

I see now. You're calling the url to the attachment on backend `serverUrl`.
This naming is used on other places of the app as workspace url, like `https://open.rocket.chat`
https://github.com/search?q=repo%3ARocketChat%2FRocket.Chat.ReactNative%20serverurl&type=code

GleidsonDaniel:
The solution is simple, the explanation was unnecessarily complex.
Previously, the `sanitizeString` function was used to get the name of the file (whoever has a name has nothing to do with what it actually does) and use that name as the download key.
With this change, using `sanitizeLikeString` takes the entire file url as the download key.

diegolmello (approved):

Issue #5644, opened 2024-04-08
Chats with disabled notifications appear in Unread.


shiryov:
### Description:

If all notifications for chat is disabled, then it anyway appear in Unread list when have new messages.

### Environment Information:

- Rocket.Chat Server Version: 6.5.2
- Rocket.Chat App Version: 4.47.2
- Device Name: Poco X3 Pro
- OS Version: Android 12

### Steps to reproduce:

1. Disable all notifications for chat A
2. Send message to chat B in which notification enabled
3. Send message to chat A
4. Send message to chat B

### Expected behavior:

Chat A is not appear in Unread list, and not in bold

### Actual behavior:

Chat A is in Unread list

### Additional context:

![Screenshot_2024-04-08-19-22-22-323_chat rocket android](https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/666827/b7542b33-f7fa-4483-a23d-e153ceed4b51)
![Screenshot_2024-04-08-19-33-51-398_chat rocket android_](https://github.com/RocketChat/Rocket.Chat.ReactNative/assets/666827/e66fefe5-572f-455e-92e8-e41e09e41439)

PR #5649, opened 2024-04-10
regression: checks if the `hideUnreadStatus` property is coming from the subscription

GleidsonDaniel:
## Proposed changes

Checks if the `hideUnreadStatus` property is coming from the subscription
Sometimes the property does not come from the backend, so it is false, which is not the expected behavior.

## Issue(s)	

resolve: #5644

## How to test or reproduce

- set these settings in a room:
image
- send a message in that room with another user.

## Types of changes

## Checklist

## Further comments

Jira: [CORE-253]

[CORE-253]: https://rocketchat.atlassian.net/browse/CORE-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


diegolmello (commented):
This behavior might have changed on a recent version of backend, but it could work for older servers.
Ideally we should identify when it changed, but we need to check at least on 3.11, v4 and v5.

GleidsonDaniel:
> This behavior might have changed on a recent version of backend, but it could work for older servers. Ideally we should identify when it changed, but we need to check at least on 3.11, v4 and v5.

It hasn't changed, it's still the same, we just didn't check it correctly.
And now with some servers migrating and leaving some problematic fields, the problem has become more evident.
Looking at the history of changes, nothing has changed in the last year, for example, the rules are mostly rules with more than 3 years of implementation.

diegolmello (approved):

PR #5651, opened 2024-04-10
feat: change default browser label

GleidsonDaniel:
## Proposed changes

Currently on iOS the system default option is labeled Safari, however this does not reflect reality, given that the system default can be any browser, and as we do not have access to which browser is the system default, the update of the label is necessary to avoid misunderstandings.

Note: Safari does not have a deeplink.
See Apple's Documentation: https://developer.apple.com/library/archive/featuredarticles/iPhoneURLScheme_Reference/Introduction/Introduction.html

## How to test or reproduce

Settings > Default browser

## Screenshots

After:
3C84B3F7-1C7E-42B1-85F2-428AC034894C

Before:
C761A24B-DA27-4A05-84F5-794E955CC820

## Types of changes

## Checklist

## Further comments

Jira: [CORE-255]

[CORE-255]: https://rocketchat.atlassian.net/browse/CORE-255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/DefaultBrowserView/index.tsx
+++ b/app/views/DefaultBrowserView/index.tsx
@@ -25,7 +25,7 @@ const DEFAULT_BROWSERS: IBrowsersValues[] = [
 		value: 'inApp'
 	},
 	{
-		title: isIOS ? 'Safari' : 'Browser',
+		title: 'System_default',
 		value: 'systemDefault:'
 	}
 ];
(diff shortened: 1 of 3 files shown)


app/views/DefaultBrowserView/index.tsx
```diff
@@ -25,7 +25,7 @@ const DEFAULT_BROWSERS: IBrowsersValues[] = [
 		value: 'inApp'
 	},
 	{
-		title: isIOS ? 'Safari' : 'Browser',
+		title: 'System_default',
```
diegolmello, line 28: I don't know if it's possible, but should it have an option to open Safari then?

diegolmello (approved):
Should it

PR #5869, opened 2024-09-16
chore: Color and contrast improvements

yash-rajpal:
## Types of changes

## Checklist

## Further comments

[MA-74](https://rocketchat.atlassian.net/browse/MA-74)

### DIFF
--- a/app/containers/Button/index.tsx
+++ b/app/containers/Button/index.tsx
@@ -9,7 +9,7 @@ import ActivityIndicator from '../ActivityIndicator';
 interface IButtonProps extends PlatformTouchableProps {
 	title: string;
 	onPress: () => void;
-	type?: string; // primary | secondary
+	type?: 'primary' | 'secondary'; // primary | secondary
 	backgroundColor?: string;
 	loading?: boolean;
 	color?: string;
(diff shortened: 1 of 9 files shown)


GleidsonDaniel:
👀

app/containers/Button/index.tsx
```diff
@@ -9,7 +9,7 @@ import ActivityIndicator from '../ActivityIndicator';
 interface IButtonProps extends PlatformTouchableProps {
 	title: string;
 	onPress: () => void;
-	type?: string; // primary | secondary
+	type?: 'primary' | 'secondary'; // primary | secondary
```
diegolmello, line 12: If we go this way, let's remove the comment, while fixing the need for `as const` as well.

app/containers/Button/index.tsx
```diff
@@ -64,7 +64,7 @@ const Button: React.FC<IButtonProps> = ({
 		style
 	];
 
-	const textStyle = [styles.text, { color: isDisabled ? colors.fontDisabled : resolvedTextColor, fontSize }, styleText];
+	const textStyle = [styles.text, { color: isDisabled ? colors.fontWhite : resolvedTextColor, fontSize }, styleText];
```
diegolmello, line 67: I might be wrong, but it looks like you're bypassing semantic to make UI fit figma.
That's not the idea.
Figma says this should be `Button/Font/on-primary-disabled`.
If our colors are outdated, you should update them.

diegolmello (commented):
Also this is not a chore.
It's a accessibility improvement.
It could be `feat(a11y)`

app/containers/Button/index.tsx
yash-rajpal: Yeah, you are right, I was confused about this too, but decided since this change is only meant for the disabled button text color, so might be okay to just change the color where used.

I'll update the `fontDisabled` color to white.

diegolmello (approved):

PR #5927, opened 2024-10-22
fix: draft message not being updated

diegolmello:
## Proposed changes

On https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5827, we fixed an issue causing the draft to remain saved after the message is sent.
However, that bug can still be reproduced in some cases.

On that PR we added `forceUpdateDraftMessage` on `setInput`, to be used only on `getTextAndClear`, which is called on send message.
It bypasses the logic from `useAutoSaveDraft` to force clear it.
However, `useAutoSaveDraft(textRef.current)` uses `textRef` and `setInput` doesn't trigger a rerender.
In the end, `useAutoSaveDraft` would keep the value before `setInput` was called with `forceUpdateDraftMessage` and the draft would be saved with that value again.

Since `setInput` is going to call `setMicOrSend` (to mic icon) every time a new message is sent, I added a call to `const mic = useMicOrSend();` (even though it's not technically needed) to subscribe to changes on that value and trigger the rerender needed to update `useAutoSaveDraft`.

## How to test or reproduce

- Type Test on composer
- Send
- Enter the room again
- Draft is still there

## Types of changes

## Checklist

## Further comments

[SUP-691]

[SUP-691]: https://rocketchat.atlassian.net/browse/SUP-691?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/containers/MessageComposer/components/ComposerInput.tsx
+++ b/app/containers/MessageComposer/components/ComposerInput.tsx
@@ -6,7 +6,7 @@ import { RouteProp, useFocusEffect, useRoute } from '@react-navigation/native';
 
 import I18n from '../../../i18n';
 import { IAutocompleteItemProps, IComposerInput, IComposerInputProps, IInputSelection, TSetInput } from '../interfaces';
-import { useAutocompleteParams, useFocused, useMessageComposerApi } from '../context';
+import { useAutocompleteParams, useFocused, useMessageComposerApi, useMicOrSend } from '../context';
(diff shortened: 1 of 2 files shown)


OtavioStasiak (approved):
Everything looks good to me.
I tested and it works as expected.

PR #5928, opened 2024-10-22
fix: android navigation bar color is not displaying correctly

OtavioStasiak:
## Proposed changes

Remove the navigation bar color using react-native-navigation-bar-color and use the navigationBarColor prop of react-navigation. 

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-106

## How to test or reproduce

 - Open the app;
 
## Types of changes

## Checklist


app/lib/methods/helpers/theme.ts
```diff
@@ -1,12 +1,9 @@
 import { Appearance } from 'react-native';
-import changeNavigationBarColor from 'react-native-navigation-bar-color';
```
diegolmello, line 2: Need to uninstall it

app/lib/methods/helpers/theme.ts
```diff
@@ -47,19 +44,6 @@ export const newThemeState = (prevState: { themePreferences: IThemePreference },
...
-	if (isAndroid) {
-		const iconsLight = theme === 'light';
-		try {
-			// The late param as default is true @ react-native-navigation-bar-color/src/index.js line 8
-			await changeNavigationBarColor(themes[theme].surfaceLight, iconsLight, true);
-		} catch (error) {
-			// Do nothing
-		}
-	}
-	setRootViewColor(themes[theme].surfaceRoom);
```
diegolmello, line 61: We still need to call this.
You can see it on iOS while rotating your device portrait <-> landscape.
Without it, it should be a white flash.

app/lib/methods/helpers/theme.ts
```diff
@@ -78,5 +62,4 @@ export const subscribeTheme = (themePreferences: IThemePreference, setTheme: ()
 		unsubscribeTheme();
 	}
 	// set native components theme
```
diegolmello, line 64: Remove this comment

app/AppContainer.tsx
```diff
@@ -59,7 +60,7 @@ const App = memo(({ root, isMasterDetail }: { root: string; isMasterDetail: bool
 				}
 				Navigation.routeNameRef.current = currentRouteName;
 			}}>
-			<Stack.Navigator screenOptions={{ headerShown: false, animation: 'none' }}>
+			<Stack.Navigator screenOptions={{ headerShown: false, animation: 'none', navigationBarColor: themes[theme].surfaceLight }}>
```
diegolmello, line 63: It seems to work, but introduces a regression on Android:
- Open gallery and choose a photo
- Share to our app
- Select any room
- Composer is rendered

Make sure to reload metro and go through all the steps after a change, since fast refresh doesn't work.

diegolmello (changes requested):

app/lib/methods/helpers/theme.ts
OtavioStasiak: ok. uninstalled.

diegolmello (changes requested):
A lot better, but not entirely fixed.

diegolmello (changes requested):
image
image

Straight from CI.
I think JS code is something like that and you should try to fix it on Android, like here https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/ddf1f3ffac7e858f04f63c43485a5911c341d02e/android/app/src/main/res/values/styles.xml#L4
Maybe you could remove those lines?

Also, take a look on how to setup https://github.com/zoontek/react-native-bootsplash
There's been several releases since our last upgrade and their v5 changes the lib completely.
It fixes https://github.com/zoontek/react-native-bootsplash/issues/440, which is a known bug for us.

Still on v5 release notes, `Your Android theme status / navigation bar styles are not overwritten anymore`.

Rohit3523:
In Android 15 and above, setting the status bar and navigation bar colors is no longer supported (for 3 button navigation) and deprecated (for gesture navigation) due to enforced edge-to-edge design.
React Native has also deprecated some prop for the StatusBar component on Android.

Reference: https://developer.android.com/about/versions/15/behavior-changes-15#deprecated-apis
https://reactnative.dev/docs/statusbar#backgroundcolor-android

diegolmello (approved):

PR #5934, opened 2024-10-24
fix: pinned menu returns a blank screen for certain rooms.

OtavioStasiak:
## Proposed changes

MessagesView is breaking when you open pinned Messages on specific room menu.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-678

## Screenshots
Before:

https://github.com/user-attachments/assets/fa382fcc-6366-4465-be0b-ff8e1a890a40

After:

https://github.com/user-attachments/assets/1ce86907-36e7-49f9-acf5-a8cb7f777510

## Types of changes

## Checklist

### DIFF
--- a/app/containers/message/Urls.tsx
+++ b/app/containers/message/Urls.tsx
@@ -126,8 +126,14 @@ type TImageLoadedState = 'loading' | 'done' | 'error';
 const Url = ({ url }: { url: IUrl }) => {
 	const { colors, theme } = useTheme();
 	const { baseUrl, user } = useContext(MessageContext);
-	let image = url.image || url.url;
-	image = image?.includes('http') ? image : `${baseUrl}/${image}?rc_uid=${user.id}&rc_token=${user.token}`;
+	const getImageUrl = () => {
+		const imageUrl = url.image || url.url;
+
+		if (!imageUrl) return null;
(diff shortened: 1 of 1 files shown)


app/containers/message/Urls.tsx
```diff
@@ -127,7 +127,7 @@ const Url = ({ url }: { url: IUrl }) => {
 	const { colors, theme } = useTheme();
 	const { baseUrl, user } = useContext(MessageContext);
 	let image = url.image || url.url;
-	image = image.includes('http') ? image : `${baseUrl}/${image}?rc_uid=${user.id}&rc_token=${user.token}`;
+	image = image?.includes('http') ? image : `${baseUrl}/${image}?rc_uid=${user.id}&rc_token=${user.token}`;
```
diegolmello, line 130: If `image` doesn't have `includes`, it means it's null or undefined.
It's being used on the else of the iif, so it's still creating something like `https://open.rocket.chat/undefined...`
You got rid of the error, but it's still returning a wrong image.
If there's no image, you should not display it at all.

diegolmello (changes requested):
Before and after videos doesn't help much, since you edited the part that it's important.
You should try to reproduce on a safe channel and show the real content.

diegolmello (approved):

PR #5935, opened 2024-10-25
feat: Fetch and persist app action buttons

yash-rajpal:
## Proposed changes

Fetching and storing all Rocket.Chat app actions in watermelon DB and redux state.
Also implemented real-time handling of this stored data based on streamed data from server related to apps changes.
## Types of changes

## Checklist

## Further comments

[RCAI-23]

[RCAI-23]: https://rocketchat.atlassian.net/browse/RCAI-23?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/lib/services/sdk.ts
```diff
@@ -63,12 +64,13 @@ class Sdk {
...
 			? void
-			: Serialized<OperationParams<'POST', MatchPathPattern<TPath>>>
+			: Serialized<OperationParams<'POST', MatchPathPattern<TPath>>>,
+		apiVersion: string = 'v1'
 	): Promise<ResultFor<'POST', MatchPathPattern<TPath>>> {
 		return new Promise(async (resolve, reject) => {
 			const isMethodCall = endpoint?.startsWith('method.call/');
 			try {
-				const result = await this.current.post(endpoint, params);
+				const result = await this.current.post(endpoint, params, undefined, undefined, undefined, apiVersion);
```
diegolmello, line 73: Commenting just so we're all aware, but this is only reasonable because there's a work [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5918) that would make it possible to call `sdk.post` specifying api version in a more elegant way.
`sdk.post('/v1/...`, `sdk.post('/apps/`...

Hopefully this is only temporary 😬

app/lib/methods/getAppActions.ts
```diff
@@ -0,0 +1,126 @@
...
+			const appActionButtons = (await sdk.get('actionButtons', undefined, 'apps')) as unknown as IAppActionButton[];
+
+			if (appActionButtons && appActionButtons.length) {
+				await db.write(async () => {
+					const appActionButtonsCollection = db.get('app_actions_buttons');
+					const allAppActionButtonsRecords = await appActionButtonsCollection.query().fetch();
+
+					const filteredAppActionButtonsToCreate = appActionButtons.filter(
+						(i1: IAppActionButton) =>
+							!allAppActionButtonsRecords.find(i2 => `${i1.appId}/${i1.actionId}` === `${i2.appId}/${i2.actionId}`)
```
diegolmello, line 72: Can we create a helper function to create appId + actionId?
Having to write/read `${i1.appId}/${i1.actionId}` everywhere is error prone.

app/lib/methods/getAppActions.ts
```diff
@@ -0,0 +1,126 @@
...
+							})
+						);
+					});
+
+					const appActionButtonsToDelete = filteredAppActionButtonsToDelete.map(action => action.prepareDestroyPermanently());
+
+					const allRecords = [...appActionButtonsToCreate, ...appActionButtonsToUpdate, ...appActionButtonsToDelete];
+
+					try {
+						await db.batch(...allRecords);
```
diegolmello, line 111: It should be possible to send the whole array now https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5987
```suggestion
						await db.batch(allRecords);
```

app/sagas/login.js
```diff
@@ -229,6 +239,7 @@ const handleLoginSuccess = function* handleLoginSuccess({ user }) {
 		const server = yield select(getServer);
 		yield put(encryptionInit());
 		yield put(roomsRequest());
+		yield fork(fetchAppActionsFork);
```
diegolmello, line 242: This fetch is "soft blocking" `getPermissions` now, so there's a race condition.
Soft blocking because it's only preventing from settings the permissions on redux and not from fetching from server.
`fork` from redux-saga makes all these fetch to work in parallel, so although they're in sequence, it's not really behaving like that.

To fix it without overcomplicating, I think we can move the logic from `fetchAppActionsFork` to `fetchPermissionsFork`, before `yield getPermissions();`.

app/actions/actionsTypes.ts
```diff
@@ -104,3 +104,4 @@ export const VIDEO_CONF = createRequestTypes('VIDEO_CONF', [
 export const TROUBLESHOOTING_NOTIFICATION = createRequestTypes('TROUBLESHOOTING_NOTIFICATION', ['INIT', 'SET']);
 export const SUPPORTED_VERSIONS = createRequestTypes('SUPPORTED_VERSIONS', ['SET']);
 export const IN_APP_FEEDBACK = createRequestTypes('IN_APP_FEEDBACK', ['SET', 'REMOVE', 'CLEAR']);
+export const APP_ACTION_BUTTON = createRequestTypes('APP_ACTION_BUTTON', ['SET', 'UPDATE', 'REMOVE', 'REMOVE_BY_APPID']);
```
diegolmello, line 107: Safe to remove, since we're not using those.
```suggestion
export const APP_ACTION_BUTTON = createRequestTypes('APP_ACTION_BUTTON', ['SET', 'REMOVE_BY_APPID']);
```

diegolmello (changes requested):

app/lib/methods/getAppActions.ts
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+}
+
+export const getAppActions = () => {
+	const db = database.active;
+
+	return new Promise<void>(async resolve => {
+		try {
+			const appActionButtons = (await sdk.get('actionButtons', undefined, 'apps')) as unknown as IAppActionButton[];
```
diegolmello, line 63: Do we need to fetch everything all the time?
Maybe there's an `updatedSince` param?

yash-rajpal: https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/ee/server/apps/communication/endpoints/actionButtonsHandler.ts

There isn't any such param. Will have to fetch all everytime.

diegolmello (approved):

Issue #5942, opened 2024-10-30
Multiple image attachments in one message do not work correctly


a1grim:
### Description:

Multiple image attachments in one message do not work correctly.

### Environment Information:

- Rocket.Chat Server Version: 6.12.0
- Rocket.Chat App Version: 4.53.0
- Device Name: iPhone 12 mini / OnePlus 12
- OS Version: iOS 18.1 / Android 14

### Steps to reproduce:

Send a message with several images attached - https://developer.rocket.chat/apidocs/post-message)

### Expected behavior:

![desktop](https://github.com/user-attachments/assets/33539710-d8fe-4870-b7fe-3540f4a77614)

### Actual behavior:

![mobile](https://github.com/user-attachments/assets/67d93d41-a11d-499c-b072-de10a3a9ca6d)

### Additional context:

Rocket.Chat App 4.52.0 is working correct

Rohit3523:
Unable to reproduce this bug in app version `4.55.0.999999999` 

<img width="371" alt="Screenshot 2024-11-08 at 2 29 42 AM" src="https://github.com/user-attachments/assets/95de7779-d81d-46ed-be62-ded587739f24">

PR #5966, opened 2024-11-07
Fix: Single image shown multiple times

Rohit3523:
## Proposed changes
This pull request fixes the issue where if we attach multiple image URLs in a single message payload, the App only showed the last image multiple times. The root cause was in attachment caching logic. While mapping through the attachments array, we were incorrectly updating all attachment URLs to the same value. To fix this, I modified the code to compare the download URL with each attachment's image URL during the mapping process:

```
attachments?.map(att => ({
    ...att,
    title_link: downloadUrl === att.image_url ? uri : att.title_link,
    e2e: encryption ? 'done' : undefined
}));
```

## Issue(s)	
https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/5942

## How to test or reproduce
To test this issue please use the following api
```
URL: https://********.rocket.chat/api/v1/chat.postMessage
Method: POST
Header: {
     'X-Auth-Token': '**********',
     'X-User-Id': '**********'
}
body: {
  "roomId": "",
  "attachments": [
    {
      "image_url": "https://i.pinimg.com/474x/c4/23/63/c423639b72feb5c93a14144fb50008c3.jpg" //or any random image url
    },
    {
      "image_url": "https://i.pinimg.com/474x/fc/98/6e/fc986eb80193639da37bf34cc5f6ef57.jpg" //or any random image url
    }
  ]
}
```

## Types of changes

## Checklist

### DIFF
--- a/app/containers/message/hooks/useMediaAutoDownload.tsx
+++ b/app/containers/message/hooks/useMediaAutoDownload.tsx
@@ -136,10 +136,10 @@ export const useMediaAutoDownload = ({
 			urlToCache: url
 		});
 		if (result?.exists && !isEncrypted) {
-			if(!currentFile?.title_link) {
-				persistMessageWithCacheFile(id, result.uri, file.encryption, file.image_url || url);
+			if (!currentFile?.title_link) {
+				await persistMessageWithCacheFile(id, result.uri, file.encryption, file.image_url || url);
 			}
-			
+
 			updateCurrentFile(result.uri);
 		}
 		return result?.exists;
--- a/app/containers/message/hooks/useMediaAutoDownload.tsx
+++ b/app/containers/message/hooks/useMediaAutoDownload.tsx
@@ -9,6 +9,7 @@ import {
 	getMediaCache,
 	isDownloadActive,
 	MediaTypes,
+	persistMessageWithCacheFile,
 	TDownloadState
 } from '../../../lib/methods/handleMediaDownload';
 import { emitter } from '../../../lib/methods/helpers';
@@ -135,6 +136,10 @@ export const useMediaAutoDownload = ({
 			urlToCache: url
 		});
 		if (result?.exists && !isEncrypted) {
+			if(!currentFile.title_link) {
+				persistMessageWithCacheFile(id, result.uri, file.encryption, file.image_url || url);
+			}
+			
 			updateCurrentFile(result.uri);
 		}
 		return result?.exists;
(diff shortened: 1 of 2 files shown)


Rohit3523:
Before:

Rohit3523:
After: 
Screenshot 2024-11-08 at 4 01 37 AM
Screenshot 2024-11-08 at 4 01 25 AM

(I have to attach 2 screenshots because emulator don't have scroll screenshot option)

Rohit3523:
@diegolmello AttachmentView is showing correct image

diegolmello:
@Rohit3523 Ok. @OtavioStasiak is going to review this and we can plan it for 4.55.0 later this month.

Rohit3523:
This pull request also fixes an issue where, if we attempt to send the same images multiple times in the same channel, one of them fails to open in preview mode.


Explanation: If I send images 1 and 2 together in a single message from the API and then send the same images again, only the latest images are clickable and viewable in preview mode on the chat screen. The previous images do not open in preview mode.

app/containers/message/hooks/useMediaAutoDownload.tsx
```diff
@@ -150,7 +150,7 @@ export const useMediaAutoDownload = ({
 			download();
 			return;
 		}
-		if (!showAttachment || !currentFile.title_link || isEncrypted) {
+		if (!showAttachment || isEncrypted) {
```
diegolmello, line 153: Everything we display is downloaded locally first and the uri is persisted on `message.attachments[x].title_link` https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/ddf1f3ffac7e858f04f63c43485a5911c341d02e/app/lib/methods/handleMediaDownload.ts#L213
I don't think we should allow displaying `showAttachment` if there's no title_link here.

There could be another bug though.
There's one where the image loading indicator keeps spinning, but the file was already downloaded and tapping the indicator opens AttachmentView correctly.

diegolmello (changes requested):

app/containers/message/hooks/useMediaAutoDownload.tsx
Rohit3523: I have added few conditions to download the image before displaying it, so this condition is back :)

Rohit3523:
I have made the changes to display image from cache

OtavioStasiak (approved):
Everything works fine on both platforms, nice job!

Tested on:
iPhone 15 (**IOS 17.5**)
android Pixel 8 (**API 35**)

Rohit3523:
Thanks ^_^

diegolmello:
The rest of the work is going to be finished on #6068

diegolmello:
@Rohit3523  Send me your github email so I can add you as co-author there.
Thanks for contributing.

PR #5989, opened 2024-11-13
Fix: Non-UTF8 ID breaking the WatermelonDB roles collection.

OtavioStasiak:
## Proposed changes

The solution to fix the WMDB issue involves encoding non-UTF8 IDs to ensure they are properly handled. Before sending them to the Redux dispatch, the non-UTF8 IDs are decoded. Additionally, if the roles collection is already corrupted, the code now resets the collection and refills it, ensuring smooth operation.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-721

## How to test or reproduce

- Run the code without my updates and edit the getRoles function at line 84 to force a role with a non-UTF8 ID. Update the code to include:
```
roles.push({
    _id: 'テストDb壊れた', // Non-UTF8 characters in ID
    description: 'Admin',
    mandatory2fa: false,
    protected: true,
    scope: 'Users',
    name: 'admin'
});
```
- Restart the app. You should see an error occur due to the non-UTF8 ID.
- Apply my updates, restart the app, and observe. If the error occurs once, the roles collection will reset, and getRoles will execute again without errors.

## Types of changes

## Checklist

### DIFF
--- a/app/lib/methods/getRoles.ts
+++ b/app/lib/methods/getRoles.ts
@@ -8,13 +8,20 @@ import { store as reduxStore } from '../store/auxStore';
 import { removeRoles, setRoles as setRolesAction, updateRoles } from '../../actions/roles';
 import { TRoleModel } from '../../definitions';
 import sdk from '../services/sdk';
+import { decodeRoleIdFromStorage, encodeRoleIdForStorage } from './helpers';
 import protectedFunction from './helpers/protectedFunction';
 
 export async function setRoles(): Promise<void> {
 	const db = database.active;
 	const rolesCollection = db.get('roles');
 	const allRoles = await rolesCollection.query().fetch();
-	const parsed = allRoles.reduce((acc, item) => ({ ...acc, [item.id]: item.description || item.name || item.id }), {});
+	const parsed = allRoles.reduce(
+		(acc, item) => ({
+			...acc,
+			[decodeRoleIdFromStorage(item.id)]: item.description || item.name || decodeRoleIdFromStorage(item.id)
+		}),
+		{}
+	);
 	reduxStore.dispatch(setRolesAction(parsed));
 }
 
@@ -69,6 +76,18 @@ export async function onRolesChanged(ddpMessage: IRolesChanged): Promise<void> {
 	}
 }
 
+// if the roleID was wrote with nonUTF8 we need to reset collection because it breaks .query().fetch() of rolesCollection.
+async function resetRolesCollection() {
+	const db = database.active;
+
+	await db.write(async () => {
+		await db.adapter.unsafeExecute({
+			sqls: [['DELETE FROM roles', []]]
+		});
+	});
+	await getRoles();
+}
+
 export function getRoles(): Promise<void> {
 	const db = database.active;
 	return new Promise(async resolve => {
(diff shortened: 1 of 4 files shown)


app/lib/methods/helpers/handleRoleId.ts
```diff
@@ -0,0 +1,14 @@
+const isUTF8 = (str: string) => {
+	// eslint-disable-next-line no-control-regex
+	const regex = /^[\x00-\x7F]*([\xC2-\xDF][\x80-\xBF]|[\xE0-\xEF][\x80-\xBF]{2}|[\xF0-\xF7][\x80-\xBF]{3})*$/;
```
diegolmello, line 3: What's this trying to match? It's not easy to identify garbled text.

app/lib/methods/getRoles.ts
```diff
@@ -88,25 +107,33 @@ export function getRoles(): Promise<void> {
...
+					const filteredRolesToCreate = roles.filter(i1 => !allRolesRecords.find(i2 => encodeRoleIdForStorage(i1._id) === i2.id));
+					const filteredRolesToUpdate = allRolesRecords.filter(i1 => roles.find(i2 => i1.id === encodeRoleIdForStorage(i2._id)));
 
 					// Create
 					const rolesToCreate = filteredRolesToCreate.map(role =>
 						rolesCollections.prepareCreate(
 							protectedFunction((r: TRoleModel) => {
-								r._raw = sanitizedRaw({ id: role._id }, rolesCollections.schema);
-								Object.assign(r, role);
+								r._raw = sanitizedRaw({ id: encodeRoleIdForStorage(role._id) }, rolesCollections.schema);
```
diegolmello, line 117: Have you tried `sha256` instead of this `encodeRoleIdForStorage`? You won't need to decode it back, you can just compared to the encoded.

app/lib/methods/getRoles.ts
```diff
@@ -69,6 +76,18 @@ export async function onRolesChanged(ddpMessage: IRolesChanged): Promise<void> {
 	}
 }
 
+// if the roleID was wrote with nonUTF8 we need to reset collection because it breaks .query().fetch() of rolesCollection.
+async function resetRolesCollection() {
```
diegolmello, line 80: This fix was merged in 2021, so it's likely we're fixing corrupted data that won't be created again https://github.com/RocketChat/Rocket.Chat/pull/24052/files#diff-0d7c41e246b1f2222f40c45ae965ba22d24559e5148267d0e888e24f9925277dR204
The backend fix stops accepting the role name as the id of the collection and generates a random id, which is the default behavior everywhere else.

Based on that information, I think the reset logic here is correct, but when we do it, we should set a property on MMKV to always use the fallback id logic (I suggest sha256).

Here's my suggested flow:
- Create MMKV property: ROLES_ID_FALLBACK true|false
  * If false, saves the id from backend
  * If true, uses sha256
- Eventually, the code is going to catch and run `resetRolesCollection` and set the prop to true
  * Use logEvent(events.ROLES_ID_FALLBACK) so we can track this and remove in the future. This way we'll have data to ensure it's not affecting users anymore.
- Every role db write operation should check ROLES_ID_FALLBACK and use the appropriate id method

app/lib/methods/helpers/handleRoleId.ts
diegolmello, line 1: Move this content to getRoles file. It's too overkill to have it so distant from it.

diegolmello (changes requested):

app/lib/methods/getRoles.ts
diegolmello: It actually needs to be wmdb, because MMKV would not work if there's multiple servers logged in.

PR #6013, opened 2024-11-21
Chore a11y new discussion

OtavioStasiak:
## Types of changes

## Checklist


app/views/CreateDiscussionView/SelectChannel.tsx
```diff
@@ -55,8 +56,11 @@ const SelectChannel = ({
 		});
 
 	return (
-		<>
-			<Text style={[styles.label, { color: colors.fontTitlesLabels }]}>{I18n.t('Parent_channel_or_group')}</Text>
+		<View accessibilityLabel={`${I18n.t('Parent_channel_or_group')} - ${required ? I18n.t('Required') : ''}`}>
```
diegolmello, line 59: Here you're using `-` to separate strings, but you're using `,` on SwitchItem above.
We should use always the same technique everywhere.
I'd say `,` is better.

At the same time, is it necessary to have a separator like that?
Is it going to be spoken by the screen reader?
Is it going to be faster without a separator?

Example:
Is it going to be spoken `"Label comma required"?` or is it going to give it a pause?

app/views/CreateDiscussionView/SelectChannel.tsx
```diff
@@ -69,9 +73,9 @@ const SelectChannel = ({
 					imageUrl: getAvatar(channel)
 				}))}
 				onClose={() => getChannels('')}
-				placeholder={{ text: `${I18n.t('Select_a_Channel')}...` }}
+				placeholder={{ text: `${I18n.t('Select_a_Channel')}` }}
```
diegolmello, line 76:
```suggestion
				placeholder={{ text: I18n.t('Select_a_Channel') }}
```

app/views/CreateDiscussionView/SelectUsers.tsx
```diff
@@ -69,11 +69,11 @@ const SelectUsers = ({
 					text: { text: getRoomTitle(user) },
 					imageUrl: getAvatar(user)
 				}))}
-				placeholder={{ text: `${I18n.t('Select_Users')}...` }}
+				placeholder={{ text: `${I18n.t('Select_Users')}` }}
```
diegolmello, line 72:
```suggestion
				placeholder={{ text: I18n.t('Select_Users') }}
```

app/views/CreateDiscussionView/index.tsx
```diff
@@ -149,35 +138,39 @@ class CreateChannelView extends React.Component<ICreateChannelViewProps, ICreate
...
-							token={user.token}
-							selected={users}
-							onUserSelect={this.selectUsers}
-							blockUnauthenticatedAccess={blockUnauthenticatedAccess}
-							serverVersion={serverVersion}
-						/>
+						<Text style={[styles.description, { color: themes[theme].fontDefault }]}>{I18n.t('Discussion_Desc')}</Text>
+						<View style={{ gap: 12, paddingTop: 12 }}>
+							<SelectChannel
+								required
```
diegolmello, line 146: `SelectChannel` seems to be used only here, so no reason to overcomplicate with a prop that has always the same value.
I say we make it always required.

app/views/CreateDiscussionView/index.tsx
```diff
@@ -149,35 +138,39 @@ class CreateChannelView extends React.Component<ICreateChannelViewProps, ICreate
...
-							token={user.token}
-							initial={this.channel && { text: getRoomTitle(this.channel) }}
-							onChannelSelect={this.selectChannel}
-							blockUnauthenticatedAccess={blockUnauthenticatedAccess}
-							serverVersion={serverVersion}
-						/>
-						<FormTextInput
-							label={I18n.t('Discussion_name')}
-							testID='multi-select-discussion-name'
-							placeholder={I18n.t('A_meaningful_name_for_the_discussion_room')}
```
diegolmello, line 167: Now we don't have placeholders anymore, we need to remove them from i18n files.

app/views/CreateDiscussionView/index.tsx
```diff
@@ -188,6 +181,14 @@ class CreateChannelView extends React.Component<ICreateChannelViewProps, ICreate
 								/>
 							</>
 						) : null}
+
+						<Button
+							testID='create-discussion-submit'
+							disabled={!this.valid()}
+							style={{ marginTop: 24 }}
```
diegolmello, line 188: 36 on Figma.
Is it inheriting 12 from another component?

diegolmello (changes requested):

app/views/CreateDiscussionView/index.tsx
OtavioStasiak: adjusted.

app/views/CreateDiscussionView/index.tsx
OtavioStasiak: removed.

app/views/CreateDiscussionView/SelectUsers.tsx
OtavioStasiak: adjusted.

app/views/CreateDiscussionView/index.tsx
OtavioStasiak: it was inheriting but not 12, my bad. Adjusted the styles to be clear.

app/views/CreateDiscussionView/SelectChannel.tsx
OtavioStasiak:
```
Here you're using - to separate strings, but you're using , on SwitchItem above.
We should use always the same technique everywhere.
I'd say , is better.
```
Got it, I will extend this conversation to open.

At the same time, is it necessary to have a separator like that? on iOS it works without a separator, but on Android we need it.
Is it going to be spoken by the screen reader? No, on both platforms.
Is it going to be faster without a separator? yes, but only on android.

OtavioStasiak: we defined that accessibility separator will be ','.

diegolmello (approved):

PR #6023, opened 2024-11-25
chore(a11y): RoomInfoEditView

OtavioStasiak:
## Proposed changes

Added accessibility features to the RoomInfoEditView.

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-196

## Types of changes

## Checklist


app/containers/TextInput/FormTextInput.tsx
```diff
@@ -2,6 +2,7 @@ import { BottomSheetTextInput } from '@discord/bottom-sheet';
 import React, { useState } from 'react';
 import { StyleProp, StyleSheet, Text, TextInput as RNTextInput, TextInputProps, TextStyle, View, ViewStyle } from 'react-native';
 import Touchable from 'react-native-platform-touchable';
+import { A11y } from 'react-native-a11y-order';
```
diegolmello, line 5: My concern about this lib is
> Because of its native nature, it is important to wrap all components inside A11y.Order with A11y.Index. Components not wrapped in A11y.Index may be skipped from the reading order on iOS, or read at the end on Android. If you need to manage many elements, try to group them with a view and then control them via the ordering system. You can find an example below.

We would have to wrap the whole app on it.

OtavioStasiak: I tried to do it, but it does not work on iOS.

diegolmello (approved):

PR #6049, opened 2024-12-07
Many users were experiencing slow loading times and frequent crashes …

Jitmisra:
We identified the following key issues contributing to the performance problems:

Unoptimized Images: Large, uncompressed images were significantly slowing down page load times.
Excessive JavaScript: The application's JavaScript bundle was too large, leading to long parsing and execution times.
Inefficient DOM Manipulation: Frequent DOM manipulations were causing performance bottlenecks, especially on older devices.
To address these issues, we implemented the following solutions:

Image Optimization:
Compressed images using a suitable format (e.g., WebP) and reduced their dimensions.
Implemented lazy loading to defer the loading of images until they are about to be viewed.
JavaScript Optimization:
Minified and compressed the JavaScript code to reduce file size.
Used code splitting to break down the application into smaller chunks, loading only the necessary code for each page.
Leveraged techniques like tree-shaking to remove unused code.
DOM Optimization:
Reduced the number of DOM elements and minimized DOM manipulations.
Utilized virtual DOM libraries like React or Vue to efficiently update the DOM.
Optimized CSS selectors to improve rendering performance.

### DIFF
--- a/app/views/DirectoryView/Options.tsx
+++ b/app/views/DirectoryView/Options.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useMemo } from 'react';
 import { Text, View } from 'react-native';
 import { useSafeAreaInsets } from 'react-native-safe-area-context';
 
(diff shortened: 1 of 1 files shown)


Rohit3523:
Can you please format your code with the command `yarn prettier` and commit only this file
Thanks

app/views/DirectoryView/Options.tsx
```diff
@@ -11,73 +11,87 @@ import styles from './styles';
...
-		}
+  const { colors } = useTheme();
+  const insets = useSafeAreaInsets();
+  const [type, setType] = useState('channels');
 
-		if (itemType === 'teams') {
-			text = 'Teams';
-			icon = 'teams';
-		}
+  const renderItem = useMemo(() => {
```
Rohit3523, line 48: useMemo should be used to memoize values not a function.

PR #6080, opened 2024-12-30
chore: started parser changes

OtavioStasiak:
## Types of changes

## Checklist

### DIFF
--- a/app/containers/markdown/index.tsx
+++ b/app/containers/markdown/index.tsx
@@ -135,8 +135,8 @@ class Markdown extends PureComponent<IMarkdownProps, any> {
 		});
 
 	get isNewMarkdown(): boolean {
-		const { md, enableMessageParser } = this.props;
-		return (enableMessageParser ?? true) && !!md;
+		const { md } = this.props;
+		return !!md;
 	}
 
 	renderText = ({ context, literal }: { context: []; literal: string }) => {
--- a/app/lib/methods/sendMessage.ts
+++ b/app/lib/methods/sendMessage.ts
@@ -1,4 +1,5 @@
 import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
+import { parse } from '@rocket.chat/message-parser';
 import { Model } from '@nozbe/watermelondb';
 
 import database from '../database';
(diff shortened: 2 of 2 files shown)


app/containers/markdown/components/Strike.tsx
```diff
@@ -2,7 +2,10 @@ import React from 'react';
 import { StyleSheet, Text } from 'react-native';
 import { Strike as StrikeProps } from '@rocket.chat/message-parser';
 
-import { Bold, Italic, Plain, Link } from './components';
+import Bold from './Bold';
+import Italic from './Italic';
+import Plain from './Plain';
+import Link from './Link';
```
diegolmello, line 8: Had to go back in the history to remember why I put all of these inside `components` instead of importing them directly.
It turns out importing like you did creates a require cycle.

Bold > Italic > Bold ...

Importing through `./components` break this cycle https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5754

I came up with this file structure to help you.

```
- mentions
	- AtMention
	- Hashtag
- code
	- code
	- codeline
	- inlinecode
- emoji
	- bigemoji
	- emoji
- list
	- orderedlist
	- tasklist
	- unorderedlist
- inline
	- bold
	- italic
	- link
	- strike
- heading
- image
- katex
- linebreak
- paragraph
- plain
- preview
- quote
```

This is more in line to what we do on web as well https://github.com/RocketChat/Rocket.Chat/tree/develop/packages/gazzodown/src

app/containers/markdown/index.tsx
```diff
@@ -48,307 +28,37 @@ interface IMarkdownProps {
...
-	};
-
-	renderAtMention = ({ mentionName }: { mentionName: string }) => {
-		const { username = '', mentions, navToRoomInfo, useRealName, style } = this.props;
-		return (
-			<MarkdownAtMention
-				mentions={mentions}
-				mention={mentionName}
-				useRealName={useRealName}
+			<Body
```
diegolmello, line 49: I had to open the code to understand what's `Body`.
While it works, I think we should put its content inside here, because that's a relevant part of how our markdown works.

Instead of 

```jsx
return <Body />
```

We would do

```jsx
return (
	<MarkdownContext.Provider
		value={{
			mentions,
			channels,
			useRealName,
			username,
			navToRoomInfo,
			getCustomEmoji,
			onLinkPress
		}}>
		{tokens?.map(block => {
			switch (block.type) {
				case 'BIG_EMOJI':
   ...
```

diegolmello (changes requested):

diegolmello (approved):

Issue #6138, opened 2025-02-17
bug: Missing Name Prompt for New Users Signing in via Google Auth


jjinendra3:
#### **Issue Summary**  
When a new user logs in via Google OAuth for the first time (without having signed up separately), there is no screen prompting the user to enter their name. However, there is a screen asking for a username. This results in an incomplete profile, where the main profile section lacks a name, creating a void in the chat display.  

#### **Steps to Reproduce**  
1. Log in to the Rocket.Chat React Native app using Google OAuth as a new user.  
2. Observe that the app prompts the user to choose a username.  
3. Notice that there is no screen asking for the user's name.  
4. Go to the main profile section – the name field remains empty.  

#### **Expected Behavior**  
- If a user logs in for the first time using Google OAuth and has not previously signed up, they should be prompted to confirm or enter their name before proceeding to select a username.  
- The name field should be prefilled using the data retrieved from Google OAuth, allowing the user to edit it if needed.  
- After confirming their name, the user should proceed to the username selection screen.  

#### **Proposed Fix**  
- Detect if the user is new (i.e., has not signed up before).  
- Introduce an additional screen that asks for the user’s name before the username selection screen.  
- Prefill this field with the name retrieved from Google OAuth.  
- Allow the user to edit their name before proceeding.  

#### **Additional Context**  
This issue creates inconsistencies in user profiles, as names are missing from the main profile section and potentially in chat-related displays. Implementing the proposed fix will improve user experience and profile completeness.

jjinendra3:
@diegolmello , would love to know your thoughts on this, if you approve i can take this up :))

diegolmello:
What's the current behavior on web client?

jjinendra3:
There isn’t a dedicated screen for this, but on the web client, the name field is automatically populated with the name from the user's Google account upon signing in with Google OAuth. However, in the React Native app, the **Name** field remains empty.  

![Image](https://github.com/user-attachments/assets/a028f07d-c466-4234-8b6a-fe5b898f23b9) An account created using Google OAuth on the web client, where the name is correctly autofilled.  

![Image](https://github.com/user-attachments/assets/75c000c5-6adf-4a63-9d5c-75dd40377806) The **Profile Edit** screen in the React Native app, showing an empty **Name** field in the account i just created using Google Auth

[...]

PR #6144, opened 2025-02-19
fix: add Name field to SetUsernameView

jjinendra3:
## Proposed changes

Asks the name from the user during signup using Social Auth so to not leave the name slot empty.

## Issue(s)	

#6138 

## How to test or reproduce

1. Use a new email address not used in Rocket Chat.
2. Sign up using Social Auth.
3. There is no field to ask for the name of the user and as a result

## Types of changes

## Checklist

## Further comments

I thoroughly debugged the SocialOAuth flow, checking all params in sdk.ts and all auth related functions such as  login() [connect.ts], loginTOTP() [connect.ts], also tracing requests and their responses from and to @rocket.chat/sdk, and verifying backend auth routes. I found that in the result.me section ([screenshot](https://github.com/user-attachments/assets/a826fa0d-fa4a-4013-8d9c-aafbef107b59)), key values like name and username are coming as undefined, seemingly from the `@rocket.chat/sdk`.

To cross-check, I tried calling users.wholeInfo with users.getUsernameSuggestion to check the user details there, but since the backend doesn’t store name from Social OAuth, it doesn’t return anything substantial. 

I also analyzed how the **web client** handles this and found that it doesn’t rely on the SDK but has a different embedded indigenous auth system.

After 15+ test signups and deletions in the React Native app analyzing each through different methods, I couldn’t automate fetching the name. 

So, as a **fix**, I propose adding a controlled text input below the username field in app/views/SetUsernameView.tsx, allowing users to enter their name manually. This can be further improved by introducing a `users.getNameSuggestion` route, similar to `getUsernameSuggestion`, both of which can be called in the existing useEffect to enhance UX while keeping flexibility for users to modify their name.

Let me know if I’ve missed something or if there’s a better approach. Thanks!

### DIFF
--- a/app/views/SetUsernameView.tsx
+++ b/app/views/SetUsernameView.tsx
@@ -121,7 +121,7 @@ const SetUsernameView = () => {
 						placeholder={I18n.t('Name')}
 						returnKeyType='send'
 						onSubmitEditing={handleSubmit(submit)}
-						testID='set-username-view-input'
+						testID='set-name-view-input'
 						clearButtonMode='while-editing'
 						containerStyle={sharedStyles.inputLastChild}
 					/>
--- a/app/views/SetUsernameView.tsx
+++ b/app/views/SetUsernameView.tsx
@@ -32,10 +32,12 @@ const styles = StyleSheet.create({
 
 interface ISubmit {
 	username: string;
+	name: string;
 }
 
 const schema = yup.object().shape({
-	username: yup.string().required()
+	username: yup.string().required(),
+	name: yup.string().required()
 });
 
 const SetUsernameView = () => {
@@ -67,13 +69,13 @@ const SetUsernameView = () => {
 		init();
 	}, []);
 
-	const submit = async ({ username }: ISubmit) => {
+	const submit = async ({ username, name }: ISubmit) => {
 		if (!isValid) {
 			return;
 		}
 		setLoading(true);
 		try {
-			await Services.saveUserProfile({ username });
+			await Services.saveUserProfile({ username, name });
 			dispatch(loginRequest({ resume: token }));
 		} catch (e: any) {
 			showErrorAlert(e.message, I18n.t('Oops'));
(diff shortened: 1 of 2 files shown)


app/views/SetUsernameView.tsx
```diff
@@ -103,6 +105,26 @@ const SetUsernameView = () => {
...
+						{I18n.t('Name')}
+					</Text>
+					<ControlledFormTextInput
+						control={control}
+						name='name'
+						autoFocus
+						placeholder={I18n.t('Name')}
+						returnKeyType='send'
+						onSubmitEditing={handleSubmit(submit)}
+						testID='set-username-view-input'
```
Rohit3523, line 124: The test ID 'set-username-view-input' is used for the username TextInput and also used for the name field, which is incorrect. This field is for the name so use a test ID that matches it like 'set-name-view-input'

jjinendra3: I see, its fixed now, thanks a lot!

diegolmello:
I think this change might not cover the ideal scenario, so please test it.

IIRC SetUsernameView is only going be called when the backend was not able to use the username informed during register.
OAuth register > (username already exists) > SetUsernameView

Since the goal of this task is to fill Name correctly, this PR is aiming to fix the edge case, not the main flow.

@jjinendra3 Can you test and confirm this? It's been a long time since this flow was written in this app. Thanks!

jjinendra3:
Hey @diegolmello,  

I tested this in two ways:  

1. **Conventional Registration** – If a username is already taken, an alert notifies the user, and the flow never reaches the `SetUsernameView` screen. This confirms that the conventional registration process is working as expected.  ([Screen recording](https://github.com/user-attachments/assets/98a4c6d6-4b88-422b-afed-ef611f18cd47))

2. **OAuth Signup** – The user selects an OAuth provider and completes authentication in the browser. Upon returning to the app, they proceed through two-factor authentication. After successful 2FA, the user is **always** redirected to the `SetUsernameView` screen because no name or username is retrieved or sent to the backend. The `getUsernameSuggestion`  used in mounting of `SetUsernameView` is a backend function that generates a username based on the email prefix (e.g., `<name>@gmail.com`).  


([Entire Flow Test When Signing up using OAuth](https://github.com/user-attachments/assets/332dfe35-a63e-4700-bc21-76d7f7e48736))



### Conclusion:  
- Conventional registration is functioning correctly, handling duplicate usernames as expected through alerts.  
- OAuth signup, however, **never retrieves the username from the provider**, causing users to always land on the `SetUsernameView` screen. This confirms that the issue affects the main OAuth flow, aligning with the goal of this PR.  

Thanks for your patience and guidance—it means a lot!

diegolmello (approved):

pierre-lehnen-rc:
The name is loaded properly when logging in with google on the web client, but not on mobile. I compared the code of both and noticed that the web client requests both `email` and `profile` scopes when running the oauth login, but the mobile app only requests `email`, so that's probably why the user's name is nowhere to be seen. 

I would try changing the scope [here](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/containers/LoginServices/serviceLogin.ts#L52) to `email,profile` to see if maybe that is enough to load the name directly from google.

diegolmello:
@jjinendra3 can you try that? 🔝

PR #4866, opened 2023-01-26
Chore: Update Detox to 20.1.2

diegolmello:
## Proposed changes

Updates Detox and makes it run on CI.

## Screenshots
image

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

[NATIVE-72]

[NATIVE-72]: https://rocketchat.atlassian.net/browse/NATIVE-72?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6145, opened 2025-02-19
chore(deps): remove unused ts-node dependency

cardoso:
This dev dependency seems to be unused since mocha was removed in https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/4866

More details: `ts-node` pulls in tons of transitive dependencies which is causing duplication and affecting module-resolution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
 	"name": "rocket-chat-reactnative",
 	"version": "4.58.0",
 	"private": true,
+	"packageManager": "yarn@1.22.22",
 	"scripts": {
 		"start": "react-native start",
 		"test": "jest",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -2,6 +2,7 @@
 	"name": "rocket-chat-reactnative",
 	"version": "4.58.0",
 	"private": true,
+	"packageManager": "yarn@1.22.22",
```
diegolmello, line 5: Does it make sense to add this?

diegolmello (changes requested):
Nice catch!
Left a small comment.

diegolmello (approved):

Issue #6211, opened 2025-03-07
bug: Text Input Field Not Visible by Default in New Direct Messages


jjinendra3:
### Describe the Bug

When starting a direct message with a user for the first time, the chat screen opens, but the text input field is not visible by default. However, clicking on the blank space where the input field should be does bring it up, and it functions normally.

### Steps to Reproduce

1. Go to any thread, chat, or community.
2. Click on the icon of a user you have never contacted before.
3. Click on the "Message" button on the user's profile.
4. Observe that the text input field is missing initially.
5. Click on the blank space where the input should be, and it appears.

https://github.com/user-attachments/assets/5a8df21f-cde7-4982-841b-919b123d67af

### Expected Behavior

The Text Input should be shown normally upon going to an previosuly unvisted direct message.

[Like this](https://github.com/user-attachments/assets/9e780f94-cc21-421a-92e4-a135c4bfc883)

### Rocket.Chat Server Version

7.4.0

### Rocket.Chat App Version

4.59

### Device Name

S23

### OS Version

14

### Additional Context

This issue only occurs when messaging a user for the first time. Once a conversation has been initiated (even if no messages are exchanged), the text input field appears as expected in subsequent interactions.

jjinendra3:
Hey @diegolmello , could you please share your thoughts on this? Thanks.

Rohit3523:
Unable to reproduce this issue, getting toast message 

https://github.com/user-attachments/assets/b101d6b6-4d0e-453e-aca7-fb5d16a078da

jjinendra3:
> Unable to reproduce this issue, getting toast message
> 
>  Screenrecording_20250307_150401.mp4

Your direct message access is restricted it seems.

diegolmello:
It's a valid bug.
I've seen it happening, but couldn't find a way to reproduce it.
Can you fix it?

jjinendra3:
Sure! I will take this up.

PR #6214, opened 2025-03-08
fix: text Input Field Not Visible on First time direct message

jjinendra3:
## Proposed changes

This PR resolves an issue where the text input field is not visible when initiating a direct message with a user for the first time in the app. The problem stemmed from improper rendering of the text input field, which was caused by the useSubscription hook not functioning correctly. As a result, the placeholder was not being set, leading to errors and ultimately preventing the text input field from being displayed.
## Issue(s)	

#6211 
## How to test or reproduce

1. Go to any thread, chat, or community.
2. Click on the user you have never contacted before.
3. Click on the "Message" button on the user's profile.
4. The text input appears normally as it should 🚀

## Screenshots

https://github.com/user-attachments/assets/60a26f76-5996-454d-8679-b566d7babf8f

## Types of changes

## Checklist


app/containers/MessageComposer/hooks/useSubscription.ts
```diff
@@ -4,19 +4,30 @@ import { getSubscriptionByRoomId } from '../../../lib/database/services/Subscrip
 import { ISubscription } from '../../../definitions';
 
 // TODO: Not reactive. Should we work on an official version?
+// Please let me know if you'd like me to remove the above comment?
```
diegolmello, line 7: It's still not reactive. No reason to remove the comment.

app/containers/MessageComposer/hooks/useSubscription.ts
```diff
@@ -4,19 +4,30 @@ import { getSubscriptionByRoomId } from '../../../lib/database/services/Subscrip
...
+
 export const useSubscription = (rid?: string) => {
 	const [subscription, setSubscription] = useState<ISubscription>();
+
 	useEffect(() => {
+		// eslint-disable-next-line no-undef
+		let interval: NodeJS.Timeout;
```
diegolmello, line 14:
```suggestion
		let interval: ReturnType<typeof setTimeout> | null;
```

app/containers/MessageComposer/hooks/useSubscription.ts
```diff
@@ -4,19 +4,30 @@ import { getSubscriptionByRoomId } from '../../../lib/database/services/Subscrip
...
 			}
 			const result = await getSubscriptionByRoomId(rid);
 			if (result) {
 				setSubscription(result);
 			}
 		};
+
 		loadRoom();
+		if (rid) {
+			interval = setInterval(loadRoom, 50);
```
diegolmello, line 28: Is this the actual fix or or is it the `return` change on lines 17~18?

diegolmello (changes requested):
Can you write unit tests for this?

app/containers/MessageComposer/hooks/useSubscription.ts
jjinendra3: This is the fix actually, it works because it prevents multiple overlapping API calls when switching rooms. It ensures that the previous interval is cleared before starting a new one, avoiding unnecessary requests. Additionally, it allows time for the `rid` to be properly set before making the API call, reducing race condition.

jjinendra3:
> Can you write unit tests for this?

Aplogies, didnt see this message earlier, sure, on it!

jjinendra3:
Hey @diegolmello , added some tests for the hook, please let me know if there is an edge case i have missed. Thanks!

jjinendra3:
Hey @diegolmello, I’ve moved the tests into a single file for better modularity. Let me know if I missed anything or if there’s anything that could be improved. Thanks!

jjinendra3:
image

app/containers/MessageComposer/hooks/useSubscription.ts
diegolmello: Can you elaborate which API calls?
I don't understand why we need such interval.
It looks like there should be a way to do it without having to delay this execution.

jjinendra3: Hey @diegolmello ,

The interval is specifically there to handle a race condition with **getSubscriptionByRoomId** API call. When switching to a new direct message, the new room ID (rid) isn’t immediately available. This means that without the interval, we risk calling the API too early—fetching data with an outdated or undefined rid, which leads to an **undefined** subscription.

The 50ms interval helps in two main ways:
- **Ensuring correct data:** It gives the new rid just enough time to be properly set before the API call runs, ensuring that we get the correct subscription data.
- **Preventing overlap:** It makes sure that any previous intervals are cleared before a new one starts, so we avoid overlapping API calls.

Also, this interval only fires a few times during the mounting phase. Once the component is mounted, it **doesn’t** keep firing continuously. It’s not affecting the ongoing chatting or other operations on the page.

I feel this is a pragmatic fix that resolves the timing issue we are facing. Please let me know if you'd like me to explore any additional methods or if there is any approach that you think might yield results in a better way.

diegolmello: > The interval is specifically there to handle a race condition with getSubscriptionByRoomId API call. When switching to a new direct message, the new room ID (rid) isn’t immediately available. This means that without the interval, we risk calling the API too early—fetching data with an outdated or undefined rid, which leads to an undefined subscription.

We can check if there's `rid` and don't do anything until we have it.
Add `rid` as dependency.

jjinendra3: Hey @diegolmello , 

Apologies for the delay—I was testing various methods to fix the new User DM issue. 

1. **Using useEffect alone:**  
   I tried relying solely on useEffect with `rid` in the dependency array and an early return when `!rid`, but this approach didn’t work.

2. **Using setTimeout with various delays:**  
   I experimented with setTimeout delays (100ms, 500ms, 1000ms) to give `rid` enough time to be set with rid as dependency and `!rid` check. Unfortunately, none of these delays solved the issue.

3. **Current method (using a short interval):**  
   The current solution employs a short-interval timeout (50ms) along with the `!rid` check and `rid` as a dependency (line 33). This method, runs a minimal number of times, is the only one that reliably handles the race condition and ensures that the correct subscription data is fetched. 
   
   

Let me know if you’d like me to explore further alternatives.

jjinendra3: Hey @diegolmello,

If there’s an alternative direction you'd like me to explore, I’d be happy to give it a shot. Otherwise, if my initial approach is still not preferable, I’d appreciate any insights on what adjustments could make it more acceptable.

Please let me know your thoughts on the same, thank you.

PR #6253, opened 2025-03-21
Use open app settings

### LINKED ISSUE #6252: chore: Unify usage of openAppSettings
Search for `app-settings:` and you'll find this

Image

We have `openAppSettings` though, so we should use it in those places.
https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/b1105102f9b2a13b40970b31a6fc011ca9bad158/app/lib/methods/helpers/openAppSettings.ts

Also notice it's a promise. We should handle it.

Rohit3523:
## Proposed changes
We already have `openAppSettings` function which handle the linking of app-settings, instead of using app-settings in multiple places, we can just call the app settings function.

## Issue(s)	
Closes: https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/6252

## How to test or reproduce
N/A

## Screenshots
N/A

## Types of changes

## Checklist

## Further comments
N/A


app/views/PushTroubleshootView/components/DeviceNotificationSettings.tsx
```diff
@@ -1,26 +1,21 @@
...
 	}));
 
-	const goToNotificationSettings = () => {
-		if (isIOS) {
-			Linking.openURL('app-settings:');
-		} else {
-			notifee.openNotificationSettings();
-		}
+	const goToNotificationSettings = async () => {
+		await openAppSettings();
```
diegolmello, line 18: Can you test this on Android to understand if it's opening correctly?
Technically, the previous implementation is better because it redirects to notification settings, but I don't mind going to `app-settings`, since `notifications` is one tap away.

diegolmello (changes requested):

app/views/PushTroubleshootView/components/DeviceNotificationSettings.tsx
Rohit3523: It's working in android but i guess showing notification setting will be a great user experience, Should i revert this?

diegolmello: I agree opening the notifications directly is more optimal.
Can you try the suggestion from https://stackoverflow.com/questions/44582694/react-native-open-settings-through-linking-openurl-in-ios?

Rohit3523: I tested few deeplinks for iOS but they are limited to app settings only, can't really test on iOS 18 because the app list seems to be broken in 18.2 (https://discussions.apple.com/thread/255884009)

I have added the notifee function back for android and will track the ios one

Rohit3523:
Requesting a review, Thanks :)

PR #6027, opened 2024-11-27
feat(a11y): RegisterView

OtavioStasiak:
## Proposed changes

Added accessibility features to the RegisterView.

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-197

## How to test or reproduce

- Open the app;
- Go to RegisterView;
- Turn on the screen reader;
- navigate through the screen;
## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [X] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [X] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [X] I have added necessary documentation (if applicable)
- [X] Any dependent changes have been merged and published in downstream modules

PR #6270, opened 2025-03-25
regression: opacity animation issue on touch and incorrect height of hide message button

OtavioStasiak:
## Proposed changes

When touching the RoomItem, the opacity animation was not working due to a regression introduced in [PR #6027](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/6027). This also fixes the incorrect height of the HideMessage button on the RoomItem.
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-827

## How to test or reproduce

- Open the app;
- Click on RoomItem on RoomsListView;
- Slide left the roomItem to show the HideMessage button;

### Touch Opacity (RoomItem)
| Before | After |
| -----  | -----  |
| ezgif-3f23ad181b1832 | ezgif-34a0440cfd1eb1 |

### Height of Hide Message (RoomItem)
| Before | After |
| -----  | -----  |
| Simulator Screenshot - iPhone 15 - 2025-03-25 at 17 30 31 | Simulator Screenshot - iPhone 15 - 2025-03-25 at 17 31 06 |

## Types of changes

## Checklist


app/containers/RoomItem/Actions.tsx
```diff
@@ -141,6 +141,7 @@ export const RightActions = React.memo(({ transX, favorite, width, toggleFav, on
 					styles.actionRightButtonContainer,
 					{
 						width: width * 2,
+						height: '100%',
```
diegolmello, line 144: Hide and fav actions should be almost the same.
Have you tried adding `viewHeight` (and removing `isCondensed && { height: rowHeightCondensed },`) instead of `height 100%`?

app/containers/Touch.tsx
```diff
@@ -14,16 +14,45 @@ export interface ITouchProps extends RectButtonProps {
 const Touch = React.forwardRef<React.ElementRef<typeof RectButton>, ITouchProps>(
 	({ children, onPress, underlayColor, accessible, accessibilityLabel, style, ...props }, ref) => {
 		const { colors } = useTheme();
-
+		// The background color must be applied to the RectButton, not the View.
```
diegolmello, line 17: It's a shame we have to do this, but it can be a temp fix.
What about other styles though, like border radius, align items, etc?

#### Explaining why it's a temp fix

We shouldn't use this pattern of introducing margin/padding to the component itself, but to the parent component that uses it.
It makes easier to compose components like that.

`backgroundColor` and `borderRadius` should be a normal prop, not part of style.
`alignItems/justifyContent` can be handled on `children`.

We can fixthis in the future when we have more time though.

diegolmello (changes requested):

app/containers/Touch.tsx
OtavioStasiak: Got it, I created a task to improve this component soon.

diegolmello (approved):

PR #6250, opened 2025-03-19
fix: improve load from push notification flow

diegolmello:
## Proposed changes

I was not able to reproduce the same behavior mentioned on #6057, but I could implement a few improvements that could fix issues related to it.

- added a check to navigate from deep link only when react-navigation is ready
- added a check to avoid calling on foreground event when navigation isn't ready
- added a flag to identify if the push is configured already (that's important on Android)
- delayed hide of splash screen on local authentication
- added additional filters on message database updates to avoid unnecessary updates

## Issue(s)	

Closes #6057

#### Generic
- Push notifications
- Screen lock
- Message load

#### From issue
- Receive a push notification (repeat the test for main room and thread)
- Tap it to open the app from push (repeat the test with the app open and closed on background)
- Message should load correctly

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

[SUP-751]

[SUP-751]: https://rocketchat.atlassian.net/browse/SUP-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6272, opened 2025-03-27
fix: Message is not updated in some cases if last message

diegolmello:
## Proposed changes

On #6250 we added additional filters on message database updates to avoid operations, but there's an issue with `subscription.lastMessage`.
To deliver a better UX, `subscription.lastMessage` is saved on messages table, so we never show a full loading screen. There's always going to exist a message already saved, even if it's just last message from sub.
However, `subscription.lastMessage` doesn't have all the columns from an actual message and comparing `_updatedAt` would prevent from updating the message when the room is open.

That would cause a UI glitch in certain scenarios, like a message with thread, edited, discussion, etc. They would just appear like a regular message, when they should instead update and display the extra data.

This PR removes additional filter introduced on #6250.
Theoretically, it doesn't impact too much on the solution introduced on that PR.

## How to test or reproduce

- Make sure the last message from a subscription is a thread (or edited, discussion, etc)
- Open the room
- Message should render correctly

## Types of changes

## Checklist

## Further comments

#6250 never landed production.

[NATIVE-829]

[NATIVE-829]: https://rocketchat.atlassian.net/browse/NATIVE-829?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OtavioStasiak (approved):
LGTM!
I tested creating threads and editing message and it works as expected.

Tested on iPhone 13 (iOS 18.3.2)

PR #6274, opened 2025-03-27
fix: markdown custom emoji size

cardoso:


diegolmello (changes requested):
Missing PR description template and before/after screenshots.
It's also a regression, so we should add the commit that introduced the bug.

diegolmello (approved):

PR #6281, opened 2025-03-31
feat(a11y): avoid ellipsized texts 

OtavioStasiak:
## Proposed changes

Some parts of user interface cause texts to be ellipsized (shortened with ...) and make them difficult or impossible to understand for some users, adjusted some components to break the line or reorganize the component to avoid ellipsized.

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-169
## Screenshots
- increase your fontScale in the device;
- Open the app;
- Navigate through the app;
- Go to SettingsView | PreferencesView | MediaAutoDownloadView | RoomView | RoomsListView;

### SettingsView
| Before | After |
|--------|-------|
| Before | After |

| Before | After |
|--------|-------|
| Before | After |

### PreferencesView
| Before | After |
|--------|-------|
| Before | After |

### MediaAutoDownloadView
| Before | After |
|--------|-------|
| Before | After |

### RoomItem (condensed)
| Before | After |
|--------|-------|
| Before | After |

### RoomItem (expanded)
| Before | After |
|--------|-------|
| Before | After |

### RoomHeader
| Before | After |
|--------|-------|
| Before | After |

## Types of changes

## Checklist


app/containers/message/Message.tsx
diegolmello, line 1: Missing stories

app/containers/RoomItem/RoomItem.tsx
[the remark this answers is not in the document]
OtavioStasiak: The roomItem stories aren't working properly, and they won't work as expected because the font scale depends on the device's font size setting. I created [this task](https://rocketchat.atlassian.net/browse/NATIVE-887) to keep track of it and refactor it in the future.

app/containers/message/Message.tsx
OtavioStasiak: It won’t work perfectly as expected because the font scale depends on the device’s font size settings. However, the Message stories are working, and I believe it’s important to preserve the layout change when the font size increases, so I created a large font version for all message types. (The font won't change, but the layout will.)

app/lib/hooks/useResponsiveFontScale.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+interface IResponsiveFontScaleProviderProps {
+	children: React.ReactNode;
+}
+
+export const ResponsiveFontScaleContext = createContext({} as IResponsiveFontScaleContextData);
+
+const ResponsiveFontScaleProvider = ({ children }: IResponsiveFontScaleProviderProps) => {
+	const { fontScale } = useWindowDimensions();
```
diegolmello: Isn't it too similar to RowHeightProvider and should be unified instead?
We could argue different contexts cause less re-renders, but since they're both tied to `useWindowDimensions' fontScale`, both are going to re-render.

app/containers/List/ListItem.tsx
```diff
@@ -89,20 +100,21 @@ const Content = React.memo(
...
 		heightContainer,
+		rightContainerStyle = {},
 		styleTitle,
 		additionalAcessibilityLabel,
 		additionalAcessibilityLabelCheck,
 		accessibilityRole,
 		accessibilityLabel
 	}: IListItemContent) => {
-		const { fontScale } = useDimensions();
+		const { fontScale } = useWindowDimensions();
```
diegolmello: When should we use this directly when we should be using the new responsive fontScale hook?
It doesn't look very clear.

app/containers/CustomIcon/index.tsx
```diff
@@ -21,13 +21,15 @@ export interface ICustomIcon extends IconProps {
 const CustomIcon = memo(({ name, size, color, style, ...props }: ICustomIcon): React.ReactElement => {
 	const { colors } = useTheme();
 	const { fontScale } = useWindowDimensions();
+	const fontScaleLimit = 1.4;
```
diegolmello: Is it intentional to have it 1.4 here but 1.3 everywhere else?
Also, it should be created uppercase outside of the component.

diegolmello (changes requested):

app/containers/CustomIcon/index.tsx
OtavioStasiak: Yes, that's intentional. 
In my tests, I found that we can allow a bit more for icons without breaking the layout but if you think it's important to keep the same scale limit across all components, I can adjust it.

app/lib/hooks/useResponsiveFontScale.tsx
OtavioStasiak: I was thinking about it, and I’ve decided to unify the hooks into a `useResponsiveLayout`.  I think it's a good name, and we can include all layout cases that occur when the font size increases.

app/containers/CustomIcon/index.tsx
diegolmello: I think consistency creates better UI, but I think you should discuss with design.

OtavioStasiak: https://rocketchat.atlassian.net/browse/MA-169?focusedCommentId=54924

app/lib/hooks/mockedFontScaleProvider.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+import { FONT_SCALE_LIMIT, ResponsiveLayoutContext } from './useResponsiveLayout';
+
+type Props = {
+	children: React.ReactNode;
+	fontScale?: number;
+};
+
+export const MockResponsiveFontScaleProvider = ({ children, fontScale = 1.4 }: Props) => {
```
diegolmello: This file is strange.
It's not clear what's being mocked since there's no FontScaleProvider to match a mockedFontScaleProvider.
I can see it's being used on message stories, so it's probably more useful to use `jest.mock` inside the stories file.
Mocking `useResponsiveLayout` is also another option.

If you ever need to create a whole file to mock something, you should either create it on `__mocks__` folder or use `.mock.ts` prefix.

app/containers/message/Message.stories.tsx
```diff
@@ -80,13 +81,37 @@ export const Message = (props: any) => (
 	/>
 );
 
+// The large font components are not perfect because the text's font scale increases only with the device's font size setting.
+export const MessageLargeFont = (props: any) => (
+	<MockResponsiveFontScaleProvider fontScale={1.4}>
```
diegolmello: This is just a mock and there's a prop with 1.4 and a default value on your provider with 1.4
Either one should be used, but not both.

app/lib/hooks/useResponsiveLayout.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const ResponsiveLayoutContext = createContext({} as IResponsiveLayoutContextData);
+
+export const FONT_SCALE_LIMIT = 1.3;
+
+const ResponsiveLayoutProvider = ({ children }: IResponsiveFontScaleProviderProps) => {
+	const { fontScale } = useWindowDimensions();
+	const isLargeFontScale = fontScale > FONT_SCALE_LIMIT;
+	const fontScaleLimited = isLargeFontScale ? 1.3 : fontScale;
+	const rowHeight = 75 * fontScale;
+	const rowHeightCondensed = 60 * fontScale;
```
diegolmello: Now there's `FONT_SCALE_LIMIT` close to this, these values should also be constants for consistency.

app/lib/hooks/useResponsiveLayout.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+	return (
+		<ResponsiveLayoutContext.Provider value={{ fontScale, isLargeFontScale, fontScaleLimited, rowHeight, rowHeightCondensed }}>
```
diegolmello: The goal and difference of `fontScale` and `fontScaleLimited` isn't clear.
Can you add some comments to have some kind of documentation?

app/views/MediaAutoDownloadView/ListPicker.tsx
```diff
@@ -10,7 +10,27 @@ import sharedStyles from '../Styles';
...
+	leftTitle: {
+		...sharedStyles.textMedium,
+		fontSize: 16
+	},
+	rightContainer: {
+		flex: 1
+	},
+	rightTitle: {
+		...sharedStyles.textRegular,
+		fontSize: 16
```
diegolmello: lineHeight

app/views/MediaAutoDownloadView/ListPicker.tsx
```diff
@@ -10,7 +10,27 @@ import sharedStyles from '../Styles';
...
 const styles = StyleSheet.create({
-	title: { ...sharedStyles.textRegular, fontSize: 16 }
+	leftTitleContainer: {
+		flex: 1,
+		justifyContent: 'center',
+		alignItems: 'flex-start'
+	},
+	leftTitle: {
+		...sharedStyles.textMedium,
+		fontSize: 16
```
diegolmello: lineHeight

app/containers/List/ListItem.tsx
```diff
@@ -89,20 +91,21 @@ const Content = React.memo(
 		showActionIndicator = false,
 		theme,
 		heightContainer,
+		rightContainerStyle = {},
```
diegolmello, line 110: Missing stories update

app/containers/List/ListItem.tsx
```diff
@@ -129,28 +132,35 @@ const Content = React.memo(
...
-							/>
+				{title || subtitle ? (
+					<View style={styles.textContainer}>
+						<View style={styles.textAlertContainer}>
+							{typeof title === 'string' ? (
+								<Text style={[styles.title, styleTitle, { color: color || themes[theme].fontDefault }]}>
+									{translateTitle && title ? I18n.t(title) : title}
+								</Text>
+							) : (
+								title()
```
diegolmello: `title || subtitle`, so if `!!subtitle` and `!title`, `title()` is going to be called and the app is going to crash.

app/views/MediaAutoDownloadView/ListPicker.tsx
```diff
@@ -61,9 +81,18 @@ const ListPicker = ({
...
+				<View style={styles.leftTitleContainer}>
+					<Text style={[styles.leftTitle, { color: colors.fontDefault }]}>{title}</Text>
+				</View>
+			)}
+			right={() => (
+				<View style={styles.rightTitleContainer}>
+					<Text style={[styles.rightTitle, { color: colors.fontHint }]}>{label}</Text>
+				</View>
+			)}
+			rightContainerStyle={styles.rightContainer}
```
diegolmello, line 97: I agree with the changes, but I think the use case grow up in size, so I prefer to officialize this ListPicker as part of `containers/List` and add stories.

app/containers/RoomItem/RoomItem.tsx
```diff
@@ -57,10 +58,12 @@ const RoomItem = ({
 	hideMentionStatus,
 	accessibilityDate
 }: IRoomItemProps) => {
+	const { isLargeFontScale } = useResponsiveLayout();
 	const memoizedMessage = useMemo(
 		() => formatLastMessage({ lastMessage, username, useRealName, showLastMessage, alert, type }),
 		[lastMessage, username, useRealName, showLastMessage, alert, type]
 	);
+
```
diegolmello: Remove

app/lib/hooks/useResponsiveLayout.tsx
```diff
@@ -0,0 +1,40 @@
...
+			{children}
+		</ResponsiveLayoutContext.Provider>
+	);
+};
+
+export const useResponsiveLayout = () => {
```
diegolmello: Can you write tests for this hook?
Looks quite simple to do it.

app/views/SelectListView.tsx
```diff
@@ -179,7 +179,7 @@ class SelectListView extends React.Component<ISelectListViewProps, ISelectListVi
 			<>
 				<List.Separator />
 				<List.Item
-					title={item.name}
+					title={item.name ?? ''}
```
diegolmello: Reviewed `something ?? ''` a lot this week :eyes:
Besides the wrong usage of `??`, I think it's a bad practice to enforce an empty string to make ts happy, if that's the reason.

app/i18n/locales/nl.json
```diff
@@ -585,7 +585,7 @@
   "sending_email_confirmation": "e-mailbevestiging aan het verzenden",
   "Sending_to": "Verzenden naar",
   "Server": "Server",
-  "Server_version": "Server versie: {{version}}",
+  "Server_version": "Workspace: {{versie}}",
```
diegolmello: I don't speak dutch, but `Server versie` looks more dutch than `workspace`.
Also you're translating the param to `versie`, so it's not going to work.

app/i18n/locales/pt-PT.json
```diff
@@ -406,6 +406,7 @@
   "Send_audio_message": "Enviar mensagem de áudio",
   "Send_message": "Enviar mensagem",
   "Server": "Servidor",
+  "Server_version": "Espaço de trabalho: {{versão}}",
```
diegolmello: versao

app/i18n/locales/te-IN.json
```diff
@@ -681,7 +681,7 @@
   "sending_email_confirmation": "ఇమెయిల్ ధృవీకరణను పంపబడుతోంది",
   "Sending_to": "పంపిస్తోంది",
   "Server": "పనితనం",
-  "Server_version": "పనితనం వెర్షన్: {{version}}",
+  "Server_version": "वर्कस्पेस: {{संस्करण}}",
```
diegolmello: {{संस्करण}}

app/i18n/locales/es.json
```diff
@@ -315,7 +315,7 @@
   "Send_message": "Enviar mensaje",
   "Send_to": "Enviar a..",
   "Server": "Servidor",
-  "Server_version": "Versión servidor: {{version}}",
+  "Server_version": "Espacio de trabajo: {{versión}}",
```
diegolmello: {{versión}}",

app/i18n/locales/hi-IN.json
```diff
@@ -682,7 +682,7 @@
   "sending_email_confirmation": "ईमेल पुष्टिकरण भेज रहा है",
   "Sending_to": "भेजा जा रहा है",
   "Server": "कार्यस्थान",
-  "Server_version": "कार्यस्थान संस्करण: {{version}}",
+  "Server_version": "कार्यस्थान: {{संस्करण}}",
```
diegolmello: {{संस्करण}}",

app/i18n/locales/it.json
```diff
@@ -507,7 +507,7 @@
   "Send_to": "Invia a...",
   "Sending_to": "Invio a",
   "Server": "Server",
-  "Server_version": "Versione server: {{version}}",
+  "Server_version": "Area di lavoro: {{versione}}",
```
diegolmello: {{versione}}",

app/i18n/locales/pt-BR.json
```diff
@@ -740,7 +740,7 @@
   "sending_email_confirmation": "enviando email de confirmação",
   "Sending_to": "Envio para",
   "Server": "Workspace",
-  "Server_version": "Versão da workspace: {{version}}",
+  "Server_version": "Espaço de Trabalho: {{version}}",
```
diegolmello: Besides the regression of not using sentence case (it should be "Espaço de trabalho" with a lower t), you're confusing users here.
On "server" string, you're calling it "workspace".
On "server_version", you're calling it "Espaço de trabalho".

We should use the same terms across the whole app.
It's not an easy task to do, because we don't have control over other languages, like finish, but we should be careful with en, es and pt-BR.
Talk to design team to understand if there's any official term for workspace in pt-BR, but I don't know why you changed these strings on this PR, since it doesn't look related to ellipised text.

diegolmello (changes requested):
There's conflicts

app/containers/RoomItem/RoomItem.stories.tsx
```diff
@@ -17,6 +17,16 @@ const updatedAt = {
 	date: '10:00'
 };
 
+jest.mock('../../lib/hooks/useResponsiveLayout/useResponsiveLayout', () => ({
```
diegolmello: This mock is duplicated on Message.
You should mock it on jest.setup.js, so it's reused.

app/views/MediaAutoDownloadView/ListPicker.tsx
OtavioStasiak: I agree that we should officialize this ListPicker, but we currently have two other ListPicker components, and this PR is already large. Unifying them now would be a bit out of scope. Can I create a task to reevaluate these components and improve it separately?

Captura de Tela 2025-07-01 às 17 55 36

diegolmello: Agreed

app/containers/Avatar/__snapshots__/Avatar.test.tsx.snap
```diff
@@ -403,11 +403,11 @@ exports[`Story Snapshots: Children should match snapshot 1`] = `
       [
         {
           "color": "#D40C26",
-          "fontSize": 48,
+          "fontSize": NaN,
```
diegolmello: Looks wrong...

app/containers/List/ListItem.tsx
diegolmello: I still have this question.

app/lib/hooks/useResponsiveLayout/useResponsiveLayout.tsx
```diff
@@ -0,0 +1,44 @@
...
+export const FONT_SCALE_LIMIT = 1.3;
+export const BASE_ROW_HEIGHT = 75;
+export const BASE_ROW_HEIGHT_CONDENSED = 60;
+
+const ResponsiveLayoutProvider = ({ children }: IResponsiveFontScaleProviderProps) => {
+	// `fontScale` is the current font scaling value of the device.
+	const { fontScale } = useWindowDimensions();
+	const isLargeFontScale = fontScale > FONT_SCALE_LIMIT;
+	// `fontScaleLimited` applies the `FONT_SCALE_LIMIT` to prevent layout issues on large font sizes.
+	const fontScaleLimited = isLargeFontScale ? 1.3 : fontScale;
```
diegolmello: If it prevents layout issues, why do we have to return both?
Is there any use cases where we want to allow a font scale bigger than 1.3, even if it's a calculation of something?

Otherwise, it's a lot better to just return one `fontScale` capped to 1.3.

diegolmello:
```suggestion
	// `fontScale` is the current font scaling value of the device.
	const { fontScale } = useWindowDimensions();
	const isLargeFontScale = fontScale > FONT_SCALE_LIMIT;
	// `fontScaleLimited` applies the `FONT_SCALE_LIMIT` to prevent layout issues on large font sizes.
	const fontScaleLimited = isLargeFontScale ? FONT_SCALE_LIMIT : fontScale;
```

diegolmello (changes requested):

app/lib/hooks/useResponsiveLayout/useResponsiveLayout.tsx
OtavioStasiak: The `isLargeFontScale` is used to reposition the layout of the Message component and the RoomItem.

jest.setup.js
```diff
@@ -92,6 +92,24 @@ jest.mock('./app/lib/database/services/Message', () => ({
...
+
+jest.mock('./app/lib/hooks/useResponsiveLayout/useResponsiveLayout', () => ({
+	useResponsiveLayout: jest.fn(() => ({
+		fontScale: 1,
+		isLargeFontScale: false,
+		fontScaleLimited: 1,
+		rowHeight: 75,
+		rowHeightCondensed: 60
+	})),
+	FONT_SCALE_LIMIT: 1.25
```
diegolmello: Just wondering if this is a constant of 1.3, why it's 1.25 here 😬

PR #6291, opened 2025-04-08
feat(a11y): inline errors RegisterView

OtavioStasiak:
## Proposed changes

Create inline error on avatar url input in registerView.

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-236

## Types of changes

## Checklist


app/views/RegisterView/index.tsx
```diff
@@ -138,6 +150,13 @@ const RegisterView = ({ navigation, route }: IProps) => {
 			if (error.data?.errorType === 'username-invalid') {
 				return dispatch(loginRequest({ user: email, password }));
 			}
+
+			if (error.data.error === 'Username is already in use') {
+				setError('username', { message: `${I18n.t('Username_is_already_in_use')}`, type: 'validate' });
+				AccessibilityInfo.announceForAccessibility('Username is alredy in use');
```
diegolmello:
```suggestion
				AccessibilityInfo.announceForAccessibility('Username is already in use');
```

Typo there, but it should actually use i18n

app/views/RegisterView/index.tsx
```diff
@@ -106,7 +112,13 @@ const RegisterView = ({ navigation, route }: IProps) => {
...
 		logEvent(events.REGISTER_DEFAULT_SIGN_UP);
 
-		const { name, email, password, username } = data;
+		const { name, email, password, username, confirmPassword } = data;
+
+		if (password !== confirmPassword) {
+			setError('confirmPassword', { message: I18n.t('Passwords_do_not_match'), type: 'validate' });
+			AccessibilityInfo.announceForAccessibility(I18n.t('Passwords_do_not_match'));
+			return;
+		}
```
diegolmello: Can't you do something like this?

https://stackoverflow.com/a/63187464/5447468
https://github.com/jquense/yup/issues/682

diegolmello (changes requested):

app/lib/hooks/useVerifyPassword.ts
```diff
@@ -7,9 +7,10 @@ export interface IPasswordPolicy {
 	name: string;
 	label: string;
 	regex: RegExp;
+	validateConfirmPassword?: boolean;
 }
 
-const useVerifyPassword = (password: string, confirmPassword: string) => {
+const useVerifyPassword = (password: string, confirmPassword: string, validateConfirmPassword = true) => {
```
diegolmello: After you used the correct yup schema, this shouldn't be necessary anymore, right?

app/views/RegisterView/index.tsx
```diff
@@ -168,65 +184,79 @@ const RegisterView = ({ navigation, route }: IProps) => {
...
 						returnKeyType='next'
+						error={errors.confirmPassword?.message}
 						required
 						label={I18n.t('Password')}
 						secureTextEntry
 						textContentType={isAndroid ? 'newPassword' : undefined}
 						autoComplete={isAndroid ? 'password-new' : undefined}
 						onSubmitEditing={() => setFocus('confirmPassword')}
 						containerStyle={styles.inputContainer}
+						showErrorMessage={false}
```
diegolmello, line 248: Do you really need this?
Isn't it a matter of not sending the `error` prop?

OtavioStasiak: Yes, because we changed the error layout. In the password case, the validation message will only appear on the last one.

PR #6070, opened 2024-12-17
feat: Support API_Embed setting

diegolmello:
## Proposed changes

Admin > Message > Embed Link Previews

image

## Screenshots

|  Turned on | Turned off  |
|---|---|
| Screenshot 2024-12-17 at 15 19 57  | Screenshot 2024-12-17 at 15 20 27  |

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

[NATIVE-742]

[NATIVE-742]: https://rocketchat.atlassian.net/browse/NATIVE-742?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #6296, opened 2025-04-10
fix: API_Embed logic on useEffect dependency

diegolmello:
## Proposed changes

On #6070 we added support to `API_Embed` to prevent from displaying preview, but `useEffect` was missing `API_Embed` setting on its dependency array.
So although our tests worked because preview was rendering only when the setting was enabled, `useEffect` would still run and do a HEAD fetch.
This PR fixes this behavior and, if the setting is disabled, the URL should never be called.

## How to test or reproduce

- Create a simple express server to track URL requests, like below
```
const express = require('express')
const app = express()
const port = 3000

app.get('/', (req, res) => {
  console.log(new Date().toISOString())
  res.send('Hello World!')
})

app.listen(port, () => {
  console.log(`Example app listening on port ${port}`)
})
```
- Send a message with the URL http://localhost:3000 (use IP address instead of localhost if testing in Android)
- When `API_Embed` is enabled, opening the room should trigger a log with that date
- When disabled, nothing should happen

## Types of changes

## Checklist

## Further comments

[NATIVE-847]

[NATIVE-847]: https://rocketchat.atlassian.net/browse/NATIVE-847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


OtavioStasiak (approved):
Looks good to me!

PR #6317, opened 2025-04-23
feat(a11y): inline errors ChangeAvatarView

OtavioStasiak:
## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-237

## Types of changes

## Checklist


app/views/ChangeAvatarView/index.tsx
```diff
@@ -112,27 +123,36 @@ const ChangeAvatarView = () => {
...
 			});
 		}
-
-		setRawImageUrl(value);
 	}, textInputDebounceTime);
 
+	const onChangeText = (value: string) => {
+		setValue('rawImageUrl', value);
+		validateImage(value);
+	};
```
diegolmello, line 140: It could be wrong and it could be right, but it looks conceptually wrong.
Previously, `onChangeText` was wrapped in a debounce, so validation and set state would be triggered only after the user stopped typing for a few hundred ms.

On this PR, we're validating with a debounce, but dispatching a setValue every time.
Maybe setValue is based on a ref and it works because of that?
I don't know how react-hook-form handles this, but again... it looks wrong 😬

app/views/ChangeAvatarView/index.tsx
```diff
@@ -112,27 +123,36 @@ const ChangeAvatarView = () => {
...
-		const result = await isImageURL(rawImageUrl);
+	const validateImage = useDebounce(async (value: string) => {
+		const result = await isImageURL(value);
 
 		if (!result || !value) {
 			dispatchAvatar({
-				type: AvatarStateActions.RESET_USER_AVATAR,
-				payload: { resetUserAvatar: `@${username}` }
+				type: AvatarStateActions.CHANGE_AVATAR,
+				payload: { url: '', data: '', service: 'url' }
```
diegolmello, line 132: I don't know what's the impact of this, but it could work.

app/containers/TextInput/FormTextInput.tsx
```diff
@@ -14,13 +14,19 @@ import { A11yContainer, A11yElement } from '../A11yFlow';
 
 const styles = StyleSheet.create({
 	error: {
-		...sharedStyles.textAlignCenter,
-		paddingTop: 5
+		...sharedStyles.textRegular,
+		fontSize: 14
```
diegolmello, line 19: Missing line height.

app/views/ChangeAvatarView/index.tsx
OtavioStasiak: The idea here is just to clean the state. If I use resetUserAvatar with a payload like we did before, it will revert to the default initial avatar instead of keeping the last image. (By the way, it's an issue that's happening in production.)

app/views/ChangeAvatarView/index.tsx
OtavioStasiak: It might look a bit weird at first glance, but let me try to explain:
- If I want to use onChangeText, I have to update the state manually.
- If I apply a debounce directly inside onChangeText, the state and the input value won’t be updated consistently on each keystroke.

So, I had to use this approach to keep validating the URL and updating the state in real time.

diegolmello: The input is set on native, just like the composer, so there shouldn't be no need to update it again.
But if you thought about it, I'll trust your judgement :)

OtavioStasiak: thanks 🙏🏻

PR #6358, opened 2025-05-14
fix(a11y): add accessibility label to icons

OtavioStasiak:
## Types of changes

## Checklist


app/views/LoginView/index.tsx
```diff
@@ -23,7 +24,9 @@ const LoginView = () => {
 	useLayoutEffect(() => {
 		navigation.setOptions({
 			title: title ?? 'Rocket.Chat',
-			headerRight: () => <HeaderButton.Legal testID='login-view-more' navigation={navigation} />
+			headerRight: () => (
+				<HeaderButton.Legal accessibilityLabel={i18n.t('More')} testID='login-view-more' navigation={navigation} />
```
diegolmello: `HeaderButton.Legal` is always going to have the same icon and we'll always going to call it `More`.
This change should be applied on the component instead of as a prop.
The same way you did on `HeaderButton.Drawer`.

app/views/RoomView/RightButtons.tsx
```diff
@@ -476,17 +521,30 @@ class RightButtonsContainer extends Component<IRightButtonsProps, IRigthButtonsS
 						disabled={hasE2EEWarning}
 					/>
 				) : null}
-				{rid ? <HeaderCallButton rid={rid} disabled={hasE2EEWarning} /> : null}
+				{rid ? (
+					<HeaderCallButton
+						accessibilityLabel={i18n.t('Call_room_name', { roomName: accessibilityRoomName })}
```
diegolmello: `HeaderCallButton` is only used here, so the accessibility label should be on the component as well.
That said, we need the roomName there and perhaps it's more performant to just keep it as prop, since the data is already here.

app/views/RoomView/RightButtons.tsx
```diff
@@ -476,17 +521,30 @@ class RightButtonsContainer extends Component<IRightButtonsProps, IRigthButtonsS
 						disabled={hasE2EEWarning}
 					/>
 				) : null}
-				{rid ? <HeaderCallButton rid={rid} disabled={hasE2EEWarning} /> : null}
+				{rid ? (
```
diegolmello: There's a `!rid` with return null right above.
It's not part of this task's scope, but remove the `rid` check here and see if there's any issue.
I think it's worth trying.

app/containers/TextInput/FormTextInput.tsx
```diff
@@ -115,7 +115,9 @@ export const FormTextInput = ({
 			<A11yElement order={1}>
 				<View
 					accessible
-					accessibilityLabel={`${label} ${accessibilityLabelRequired} ${accessibilityInputValue}. ${accessibilityInputError}`}
+					accessibilityLabel={`${
+						accessibilityLabel ?? label
```
diegolmello: It's not going to change much, but I'm going to mention since I'm here :)
`??` comparison is going to make `label` be used only if `accessibilityLabel` is `null` or `undefined`.
So if it's `''`, `label` won't be used.
That's probably not expected, so `||` comparison is better on this case.

image

app/containers/RoomHeader/RoomHeader.tsx
```diff
@@ -160,7 +160,7 @@ const Header = React.memo(
 		let scale = 1;
 		const isMasterDetail = useAppSelector(state => state.app.isMasterDetail);
 		const subtitleAccessibilityLabel = tmid ? parentTitle : subtitle;
-		const accessibilityLabel = `${statusAccessibilityLabel} ${title} ${subtitleAccessibilityLabel}.`;
+		const accessibilityLabel = `${statusAccessibilityLabel} ${title} ${subtitleAccessibilityLabel ?? ''}.`;
```
diegolmello: Similar issue.
If `subtitleAccessibilityLabel === ''`, `??` isn't going to use the second argument either way.
You may want to use `||` instead.

app/containers/MessageComposer/components/RecordAudio/RecordAudio.tsx
```diff
@@ -127,12 +127,12 @@ export const RecordAudio = (): ReactElement | null => {
 					<AudioPlayer fileUri={recordingRef.current?.getURI() ?? ''} rid={rid} downloadState='downloaded' />
 				</View>
 				<View style={styles.buttons}>
-					<CancelButton onPress={cancelRecording} />
+					<CancelButton onPress={cancelRecording} accessibilityLabel='Delete_recording' />
```
diegolmello: `CancelButton` is only used here.
Adding a new prop for a11y is overcomplicating things.
Add it directly to the component instead.

diegolmello (changes requested):

app/views/RoomView/RightButtons.tsx
OtavioStasiak: it works! removed.

app/views/RoomView/RightButtons.tsx
OtavioStasiak: Makes sense! I will keep it as a prop for now.

diegolmello (approved):

Issue #1659, opened 2020-02-05
[BUG] Wrong message order makes app incompatible with screen readers (for visually impaired)


grzebo:
## Describe the bug
Android phones come with "TalkBack", a built-in screen reader, which is an accessibility app, that reads the content of the screen to visually impaired people. Rocket.chat app is incompatible with TalkBack, because when a user selects more than one message in Rocket.chat with TalkBack, they are read in reverse order (last message is read first). This causes Rocket.chat to be impractical to visually impaired people, since they need to select each message separately if they want to hear the conversation in correct order.

## To Reproduce
1. Enable TalkBack on your Android device.
2. Open Rocket.chat, go to any channel.
3. Using a TalkBack function make a selection that contains more than one message.
4. TalkBack reads messages in reverse order: first the message at the bottom of the screen (latest), then the message above it (earlier one), then one above it (even earlier one) and so on.

## Expected behavior
TalkBack should read messages in the correct chronological order (top to bottom). This is a problem with Rocket.Chat, as all other chat apps work correctly with TalkBack (FB Messenger, Slack, SMS apps, Signal etc.)

## Screenshots
Correct behavior in Slack:
User selects some messages:
![bug1](https://user-images.githubusercontent.com/47195640/73798081-188f0b80-47b2-11ea-9b1d-6b09faedd915.png)
TalkBack starts reading from the top one:
![bug2](https://user-images.githubusercontent.com/47195640/73798087-1cbb2900-47b2-11ea-9cae-f4658d54e4dd.png)

Wrong behavior in Rocket.chat:
User selects some messages:
![bug3](https://user-images.githubusercontent.com/47195640/73798091-204eb000-47b2-11ea-9075-d0b9f55e674a.png)
TalkBack starts reading the lowest message first, and then reads the top one (reverse order):
![bug4](https://user-images.githubusercontent.com/47195640/73798095-22b10a00-47b2-11ea-8b6f-caf4a88f4a8c.png)

## Devices and Versions
Bug exists in latest stable and experimental version of the app, independent of device version.

kapiwko:
My NGO is migrating to Rocket.Chat, but we have a few visually impaired members, who find it nearly impossiible to use the app with a screen reader. Is it possible for you to fix this? It's an important accessibility issue.

grzebo:
Just FYI, problem still persists on the current latest server (3.2.2) and Android apps (regular and experimental).

jsalatiel:
[...]

PR #6370, opened 2025-05-21
FIXED: #1659 [BUG] Wrong message order makes app incompatible with screen readers (for visually impaired)

HenkaOlof:
[Fixes ticket #1659](https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/1659)

The problem was that the list was being built "upside down" and then rendered in inverse order. TalkBack doesn't "see" the elements, so the inversed rendering had no effect on the order it was finding the components.

The solution was to reverse the list of messages that acts as indata for the room's FlatList of the messages. In order for the solution to take hold, the affected element (the FlatList) had to be given the attribute ImportantForAccesibility (set to 'yes').


diegolmello (changes requested):
[open a room.webm](https://github.com/user-attachments/assets/3a1dabe7-8feb-4e42-82af-728f5ddc8125)

It's not working properly.

- The list should always start close to the message composer. You can see on video it starts with the scrollbar at the top.
- Date separator isn't working anymore, since the messages are now inverted. Easy to fix though.
- onEndReached is being triggered on the wrong end of the list.
- FAB scroll down is displaying when the list is already down.

PR #6426, opened 2025-06-18
feat(a11y): Inline errors ReportUserView

OtavioStasiak:
## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-243

## Types of changes

## Checklist

### DIFF
--- a/app/i18n/locales/ar.json
+++ b/app/i18n/locales/ar.json
@@ -372,6 +372,7 @@
   "Read_Only": "قراءة فقط",
   "Read_only_hint": "يمكن لمالكي القنوات فقط إرسال رسائل جديدة",
   "Read_Receipt": "قراءة المستلم",
+  "Reason_required": "السبب مطلوب",
   "Receive_Group_Mentions": "تلقي إشارات المجموعة",
   "Receive_Group_Mentions_Info": "تلقي @all و @here للإشعارات",
   "Receive_Notification": "استلام الإشعار",
--- a/app/views/ReportUserView/index.tsx
+++ b/app/views/ReportUserView/index.tsx
@@ -34,7 +34,7 @@ interface ISubmit {
 }
 
 const schema = yup.object().shape({
-	description: yup.string().trim().required()
+	description: yup.string().trim().required(I18n.t('Reason_required'))
 });
 
 const ReportUserView = () => {
--- a/app/i18n/locales/bn-IN.json
+++ b/app/i18n/locales/bn-IN.json
@@ -547,6 +547,7 @@
   "Read_Only": "শুধুমাত্র পড়া",
   "Read_only_hint": "কেবল চ্যানেল মালিকরা নতুন বার্তা প্রেরণ করতে পারেন",
   "Read_Receipt": "পড়া স্বীকৃতি",
+  "Reason_required": "কারণ প্রয়োজন",
   "Receive_Group_Mentions": "গ্রুপ উল্লেখ প্রাপ্ত করুন",
   "Receive_Group_Mentions_Info": "@all এবং @here উল্লেখ প্রাপ্ত করুন",
   "Receive_Notification": "বিজ্ঞপ্তি প্রাপ্ত করুন",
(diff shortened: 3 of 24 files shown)


app/lib/hooks/useA11yErrorAnnouncement.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface IUseA11yErrorAnnouncement {
+	errorMessage: string | undefined;
+}
+
+const useA11yErrorAnnouncement = ({ errorMessage }: IUseA11yErrorAnnouncement) => {
+	const [errorAnnounced, setErrorAnnounced] = useState(false);
+
+	useEffect(() => {
+		if (errorMessage && !errorAnnounced) {
+			const message = errorMessage ?? '';
```
diegolmello: I don't get the reason for this.
Talked about the `??` vs `||` the other day.
`errorMessage` is tested above, so it's not undefined or falsy.

app/lib/hooks/useA11yErrorAnnouncement.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { AccessibilityInfo } from 'react-native';
+
+interface IUseA11yErrorAnnouncement {
+	errorMessage: string | undefined;
+}
+
+const useA11yErrorAnnouncement = ({ errorMessage }: IUseA11yErrorAnnouncement) => {
+	const [errorAnnounced, setErrorAnnounced] = useState(false);
+
+	useEffect(() => {
```
diegolmello: Since it looks like we're going to be using this on all forms, I think we should check how performant this is.

- It doesn't have a return, so I'm not sure it triggers a re-render on ReportUserView when `errorAnnounced` changes. I couldn't get what makes `errorAnnounced` change.
- I think we could remove `useEffect` and use `usePrevious` instead, which makes use of a ref
  - We would be comparing `errorMessage` and `previousErrorMessage`
  - `errorAnnounced` could also be a ref
  - `const shouldAnnounce = errorMessage && errorMessage !== previousMessageRef.current && !errorAnnounced;` <- This would be the first if condition
- What if the user changes the input's content too fast? Is it going to announce several times? Perhaps we should do a debounce?

app/lib/hooks/useA11yErrorAnnouncement.tsx
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useState } from 'react';
+import { AccessibilityInfo } from 'react-native';
+
+interface IUseA11yErrorAnnouncement {
+	errorMessage: string | undefined;
+}
+
+const useA11yErrorAnnouncement = ({ errorMessage }: IUseA11yErrorAnnouncement) => {
+	const [errorAnnounced, setErrorAnnounced] = useState(false);
```
diegolmello:
```suggestion
const useA11yErrorAnnouncement = ({ error }: IUseA11yErrorAnnouncement) => {
	const [announced, setAnnounced] = useState(false);
```

I'd rename for simplicity

diegolmello (changes requested):

app/i18n/locales/ar.json
```diff
@@ -403,6 +403,7 @@
   "Read_Only": "قراءة فقط",
   "Read_only_hint": "يمكن لمالكي القنوات فقط إرسال رسائل جديدة",
   "Read_Receipt": "قراءة المستلم",
+  "Reason_required": "السبب مطلوب",
```
diegolmello: i18n keys are shared with all features.
"Reason required" is too vague imo.
Perhaps "Report reason required" is enough.

app/views/ReportUserView/index.tsx
```diff
@@ -34,23 +35,29 @@ interface ISubmit {
...
-		formState: { isValid }
-	} = useForm<ISubmit>({ mode: 'onChange', resolver: yupResolver(schema), defaultValues: { description: '' } });
+		formState: { errors }
+	} = useForm<ISubmit>({
+		mode: 'onChange',
+		resolver: yupResolver(schema),
+		defaultValues: { description: '' }
+	});
+
+	// Accessibility: announce field error to screen reader
```
diegolmello: Unnecessary comment imo.
Function name is very descriptive, the usage is clear and this hook is going to be called in a few places.
It can be helpful now, but it'll be overkill very fast.

diegolmello (approved):

PR #6448, opened 2025-06-23
feat(a11y): inline error TwoFactorModal

OtavioStasiak:
## Proposed changes

Create inline errors in TwoFactor component.

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-247

## How to test or reproduce

- Open the app;
- Try to log in to an account with two-factor authentication enabled;

## Types of changes

## Checklist


app/containers/TwoFactor/index.tsx
```diff
@@ -58,14 +61,29 @@ const methods: IMethods = {
 	}
 };
 
+const schema = yup.object().shape({
+	code: yup.string().required(I18n.t('Code_required'))
+});
```
diegolmello: It's declared outside of the component.
Did you test this behavior btw?
You should make it fail, change the language and try again.
If it works outside of the component, all of them should be declared like that to make the component simpler.

app/containers/TwoFactor/index.tsx
```diff
@@ -123,6 +142,12 @@ const TwoFactor = React.memo(() => {
 		setData({});
 	};
 
+	useEffect(() => {
+		if (data.invalid) {
+			setError('code', { message: I18n.t('Invalid_code'), type: 'validate' });
+			AccessibilityInfo.announceForAccessibility(I18n.t('Invalid_code'));
+		}
+	}, [data.invalid]);
```
diegolmello: Is it really necessary to put data on a useEffect's dependency, so it can trigger an error?
Isn't it possible to do it right next to where data is actually set?

diegolmello (changes requested):

app/containers/TwoFactor/index.tsx
OtavioStasiak: I tested and It does not works outside of the component.

diegolmello (approved):

PR #6475, opened 2025-07-01
[FIX]: App crashes when search * (asterisk) in member list (#6427)

### LINKED ISSUE #6427: bug: App crashes when search * (asterisk) in member list
### Describe the Bug

When you go inside channel members view and type * in the search box the app crashes

### Steps to Reproduce

1. go to any channel
2. Open the members list
3. Try to search * (asterisk)

### Expected Behavior

The app should not crash atleast 

### Actual Behavior

The app crashes 

7.8.0

### Rocket.Chat App Version

4.61.0.80987

### Device Name

One plus nord ce3

### OS Version

Android 15

### Additional Context

https://github.com/user-attachments/assets/eb03c1bd-1fdf-4a93-b022-5bbf37381c29

luckyjsx:
## Proposed changes

This pull request fixes a bug where the app crashes when an asterisk (*) is entered into the member list search bar. The crash was due to the * being treated as an unescaped regex character in the search logic. The solution involves sanitizing the input to properly handle special characters.

## Issue(s)	

Closes #6427

## How to test or reproduce

a)Open any channel in the app.
b)Tap on the member list to view channel members.
c)Enter an asterisk (*) into the search bar.
d)The app should not crash, and input should be handled safely.

## Types of changes

- [x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)


app/lib/methods/helpers/helpers.ts
```diff
@@ -7,6 +7,8 @@ export function isGroupChat(room): boolean {
 	return ((room?.uids && room.uids.length > 2) || (room?.usernames && room.usernames.length > 2)) ?? false;
 }
 
+export const escapeRegExp = (string: string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
diegolmello: Can you test if `sanitizeLikeString` fix it?
https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c45cb0aacdd33b41be454b25651b02947b39b0e6/app/lib/database/utils.ts#L6

If it doesn't, we need to move this to the same place and write tests

app/views/RoomMembersView/index.tsx
```diff
@@ -380,12 +380,15 @@ const RoomMembersView = (): React.ReactElement => {
 		}
 	};
 
-	const filteredMembers =
+		const filteredMembers =
```
diegolmello: Looks like unnecessary change.
Run `yarn prettier-lint`

diegolmello (dismissed):
Thanks for your contribution!

Looking good, but I miss more information about the bug itself.
Why does it happen?

app/lib/methods/helpers/helpers.ts
luckyjsx: it's working good with sanitizeLikeString

app/views/RoomMembersView/index.tsx
luckyjsx: thanks for this suggestion

luckyjsx:
@diegolmello please review this pr

app/views/RoomMembersView/index.tsx
```diff
@@ -382,10 +383,10 @@ const RoomMembersView = (): React.ReactElement => {
 
 	const filteredMembers =
 		state.members && state.members.length > 0 && state.filter
-			? state.members.filter(
-					m =>
-						m.username.toLowerCase().match(state.filter.toLowerCase()) || m.name?.toLowerCase().match(state.filter.toLowerCase())
-			  )
+			? state.members.filter(m => {
+					const filter = sanitizeLikeString(state.filter.toLowerCase()) ?? '';
+					return m.username.toLowerCase().match(filter) || m.name?.toLowerCase().match(filter);
```
OtavioStasiak:
```suggestion
			? state.members.filter(m => {
					const filter = sanitizeLikeString(state.filter.toLowerCase()) || '';
					return m.username.toLowerCase().match(filter) || m.name?.toLowerCase().match(filter);
```

`||` handles all falsy values (including empty string), while `??` only handles null or undefined. In this case, we want to cover all falsy cases.

app/views/RoomMembersView/index.tsx
```diff
@@ -382,10 +383,10 @@ const RoomMembersView = (): React.ReactElement => {
 
 	const filteredMembers =
 		state.members && state.members.length > 0 && state.filter
-			? state.members.filter(
-					m =>
-						m.username.toLowerCase().match(state.filter.toLowerCase()) || m.name?.toLowerCase().match(state.filter.toLowerCase())
-			  )
+			? state.members.filter(m => {
+					const filter = sanitizeLikeString(state.filter.toLowerCase()) ?? '';
```
OtavioStasiak: We should move the `const filter` out of the `.filter()` method, moving it outside of .filter() avoids calling it repeatedly for every member. This improves performance by running the function just once, which is especially important when filtering large arrays.

luckyjsx: @OtavioStasiak I made the requested changes

OtavioStasiak (approved):
Looks good to me!
Nice job!!

PR #6477, opened 2025-07-02
chore: migrate RoomInfoEditView to hooks

OtavioStasiak:
## Proposed changes

migrate RoomInfoEditView to hooks and use react-hook-form.
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-930

## How to test or reproduce

- Open the app;
- Try to edit a room;

## Types of changes

## Checklist


app/views/RoomInfoEditView/index.tsx
```diff
@@ -1,268 +1,146 @@
...
-			(joinCodeRequired ? this.randomValue : '') === joinCode &&
-			(room.t !== 'p') === t &&
-			room.ro === ro &&
-			room.reactWhenReadOnly === reactWhenReadOnly &&
-			dequal(sysMes, systemMessages) &&
-			enableSysMes === (sysMes && sysMes.length > 0) &&
-			room.encrypted === encrypted
-		);
-	};
+	const { handleDeleteTeam, handleDeleteRoom } = useRoomDeletionActions({
```
diegolmello: I'm not a fan of calling imperative functions `handleSomething`.
It's not an event that needs to trigger anything, so there's no reason to "handle" anything.
It's the user dispatching an action, so `deleteTeam` and `deleteRoom` sounds better imo.
Open to discussion though 😬

app/views/RoomInfoEditView/hooks/useRoomDeletionActions.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+const useRoomDeletionActions = ({ navigation, room, deleteCPermission, deletePPermission }: IUseRoomDeletionActions) => {
+	const dispatch = useDispatch();
+
+	const handleVerifyTeamChannelOwner = async (teamChannels: TSubscriptionModel[]) => {
+		const permissionChecks = await Promise.allSettled(
+			teamChannels.map(async channel => {
+				const permissionType = channel.t === 'c' ? deleteCPermission : deletePPermission;
+				const permissions = await hasPermission([permissionType], channel.rid);
```
diegolmello: Lift the call to permission (close to useDispatch) and you can use `usePermissions` and get rid of `await`

app/views/RoomInfoEditView/hooks/useRoomSubscription.tsx
```diff
@@ -0,0 +1,91 @@
...
+			);
+			setPermissions({
+				'set-readonly': result[0],
+				'set-react-when-readonly': result[1],
+				'archive-room': result[2],
+				'unarchive-room': result[3],
+				'delete-c': result[4],
+				'delete-p': result[5],
+				...(room.current?.teamMain && { 'delete-team': result[6] })
+			});
```
diegolmello: This whole thing could be `usePermissions`, without the need for a `setPermissions`

app/views/RoomInfoEditView/hooks/useRoomSubscription.tsx
```diff
@@ -0,0 +1,91 @@
...
+		}
+		try {
+			const db = database.active;
+			const sub = await db.get('subscriptions').find(rid);
+			const observable = sub.observe();
+
+			querySubscription.current = observable.subscribe(data => {
+				room.current = data;
+				initializeRoomState(room.current);
+			});
```
diegolmello: I'm pretty sure it doesn't have to be reactive.
If you look at `RoomInfoEditView`'s history, you'll see we added and removed reactivity to it, because it's just annoying.
You can also test the same form on web and there it won't update when something changes on subscription while you're in the form.

app/views/RoomInfoEditView/index.tsx
```diff
@@ -1,268 +1,146 @@
...
+		encryptionEnabled: state.encryption.enabled,
+		setReadOnlyPermission: state.permissions['set-readonly'] as string[],
+		setReactWhenReadOnlyPermission: state.permissions['set-react-when-readonly'] as string[],
+		archiveRoomPermission: state.permissions['archive-room'] as string[],
+		unarchiveRoomPermission: state.permissions['unarchive-room'] as string[],
+		deleteCPermission: state.permissions['delete-c'] as string[],
+		deletePPermission: state.permissions['delete-p'] as string[],
+		deleteTeamPermission: state.permissions['delete-team'] as string[]
+	}));
+	const randomValue = random(15);
```
diegolmello: Pay attention to the methods you call on render.
This is going to be triggered on every render and generate a different value.

app/views/RoomInfoEditView/index.tsx
```diff
@@ -39,38 +38,21 @@ const MESSAGE_TYPE_VALUES = MessageTypeValues.map(m => ({
 	text: { text: I18n.t('Hide_type_messages', { type: I18n.t(m.text) }) }
 }));
 
+const RANDOM_VALUE = random(15);
```
diegolmello: Now you're calling this once when the user opens this screen and keeping the result.
Can't you just call it inside `initializeRoomState`?

app/views/RoomInfoEditView/index.tsx
```diff
@@ -1,268 +1,131 @@
...
+			joinCode: ''
 		}
-	};
+	});
 
-	init = (room: ISubscription) => {
+	const initializeRoomState = (room: ISubscription) => {
 		const { description, topic, announcement, t, ro, reactWhenReadOnly, joinCodeRequired, encrypted } = room;
 		const sysMes = room.sysMes as string[];
 		// fake password just to user knows about it
```
diegolmello: ?

app/views/RoomInfoEditView/hooks/useRoomSubscription.tsx
```diff
@@ -0,0 +1,38 @@
...
+const useRoomSubscription = ({ rid, initializeRoomState }: IUserRoomSubscription) => {
+	const room = useRef<ISubscription>({} as ISubscription);
+
+	const loadRoom = async () => {
+		if (!rid) {
+			return;
+		}
+		try {
+			const db = database.active;
+			const sub = await db.get('subscriptions').find(rid);
```
diegolmello: getSubscriptionById.
There's a very similar hook on https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/develop/app/containers/MessageComposer/hooks/useSubscription.ts

app/views/RoomInfoEditView/hooks/useRoomSubscription.tsx
```diff
@@ -0,0 +1,38 @@
...
+			log(e);
+		}
+	};
+
+	useEffect(() => {
+		loadRoom();
+	}, []);
+
+	return {
+		room: room.current
```
diegolmello: Is this returning the correct value?
You call `initializeRoomState` on useEffect, so although there's an indirection, it's going to trigger re-renders on RoomInfoEditView.

But this return might not be triggered after the first render.
`room` is being held on a ref, which doesn't trigger re-renders.

Again, `initializeRoomState` is triggering re-renders on the parent component and that's making this hook return the correct value.
But that's luck 😬

diegolmello: Btw here's a chatgpt take https://chatgpt.com/share/686c29a9-b9e8-8002-9317-da502de335a2

app/views/RoomInfoEditView/index.tsx
```diff
@@ -1,268 +1,131 @@
...
-}
+interface IRoomInfoEditViewProps extends IBaseScreen<ChatsStackParamList | ModalStackParamList, 'RoomInfoEditView'> {}
 
 const MESSAGE_TYPE_VALUES = MessageTypeValues.map(m => ({
 	value: m.value,
 	text: { text: I18n.t('Hide_type_messages', { type: I18n.t(m.text) }) }
 }));
-
-class RoomInfoEditView extends React.Component<IRoomInfoEditViewProps, IRoomInfoEditViewState> {
-	randomValue = random(15);
```
diegolmello, line 89: This used to be called on root level of a class component, similar to a `useEffect` without deps.

app/views/RoomInfoEditView/index.tsx
```diff
@@ -1,268 +1,131 @@
...
+		setValue('topic', topic || '');
+		setValue('announcement', announcement || '');
+		setValue('joinCode', joinCodeRequired ? RANDOM_VALUE : '');
+		setT(t !== 'p');
+		setReadOnly(ro);
+		setReactWhenReadOnly(reactWhenReadOnly);
+		setArchived(room.archived);
+		setSystemMessages(sysMes);
+		setEnableSysMes(sysMes && sysMes.length > 0);
+		setEncrypted(encrypted);
```
diegolmello: Shouldn't we manage all form fields using the same technique on react-hook-form?

diegolmello (changes requested):

Issue #6239, opened 2025-03-14
bug: Markdown Image not rendering correctly


diegolmello:
### Describe the Bug

Image is rendering on top of the rest of the content.

### Steps to Reproduce

- Apply the following code here https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/037253bdf26fde48e320f3a61350d8b343dfa090/app/containers/markdown/Markdown.stories.tsx#L134
- Inspect MarkdownImage on Storybook

```
const mdText = `${lineBreakText} ![alt text](https://play.google.com/intl/en_us/badges/images/badge_new.png)`

export const Image = () => (
	<View style={styles.container}>
		<Markdown msg={mdText} />
		<View style={{ height : 1, backgroundColor : 'red' }} />
		<RNText>{lineBreakText} <RNImage source={{ uri: 'https://play.google.com/intl/en_us/badges/images/badge_new.png' }} style={{ width: 100, height: 100 }} /></RNText>
	</View>
);
```

I've added a red line to display where Markdown ends and where the Text sample starts.

### Actual Behavior

<img width="461" alt="Image" src="https://github.com/user-attachments/assets/d2f91127-6b00-416f-9144-78e1aa788623" />

### Rocket.Chat Server Version

any

### Rocket.Chat App Version

4.58.0

### Device Name

any

### OS Version

any

TapItNinja:
Hi @diegolmello,
I would love to work on this issue as part of my GSoC 2025 contribution! 🚀
I have experience in React Native, UI debugging, and frontend fixes.

Please assign me this issue, and I’ll begin working on a fix. Thanks!

diegolmello:
@TapItNinja Hey! There's no assignment.
Open a PR and we're going to review it.
Thanks!

JASIM0021:
I'm unable to produce it using the storyboard. Please help me to provide more details about these bugs.

JASIM0021:
The issue has been fixed, and I have made the image height and width dynamic based on the remote image size.

Please review my PR and merge it. Here is a screenshot of the fixed screen.

![Image](https://github.com/user-attachments/assets/d0cc1abc-01e7-48d9-aaa1-b2326d22b372)

diegolmello:
It appears to be a bug on React Native, so we can't solve on our side alone.
At least not without refactoring and trying new things.
We would have to render inline images outside of a `Text` node that contains a `lineHeight` style.

[...]

PR #6478, opened 2025-07-03
chore: Enable New Architecture

diegolmello:
## Proposed changes

I had to change a few things to enable Fabric:
- Migrated RoomsListView to hooks #6502 
- Can't do `...NativeModules` anymore https://github.com/RocketChat/react-native-simple-crypto/pull/8
- Fixed a logic on draft message that would work on old arch, but failed on Fabric. It was technically wrong before, but it worked 🤷🏼 
- Disabled a few animations, but I'm going work on them right after this PR
- Fabric enabled batch updates, so I had to change a few things on DirectoryView https://react.dev/learn/queueing-a-series-of-state-updates
- Detox is also behaving a little bit differently
- Added setTimeout on two places that wasn't required before. It's not the end of the world, but I'm going to keep an eye on that.
- It doesn't fix the text cutoff issue on Android completely, but I could remove the letter spacing https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5892
- I had a few issues with images not loading properly, but it was actually my fault while trying to use `ref.measure()` incorrectly. Changed the way we get image sizes either way, because we focus everything on expo-image now.

## Issue(s)	

Depends on https://github.com/RocketChat/react-native-simple-crypto/pull/8

## Types of changes

## Checklist


diegolmello:
Apparently class components and function components behave differently in Fabric.
It took me quite some time to understand what was happening and that seems to be the reason.

Created this repro:

The idea is to replace RoomsListView with it and inspect logs.
The function component one is not going to log unmounted after Go to profile is pressed.
The class component one is going to log unmount.

```tsx
import { Component, useEffect } from 'react';
import { View, Button, Text } from 'react-native';

const Dummy = ({ navigation }) => (
	<View>
		<Text>RoomsListView</Text>
		<Button onPress={() => navigation.jumpTo('ProfileStackNavigator')} title='Go to profile' />
	</View>
);

function RoomsListView({ navigation }) {
	useEffect(() => {
		console.log('RoomsListView mounted');
		return () => {
			console.log('RoomsListView unmounted');
		};
	}, []);

	return <Dummy navigation={navigation} />;
}

// class RoomsListView extends Component {
// 	componentWillUnmount(): void {
// 		console.log('RoomsListView unmounted');
// 	}
// 	render() {
// 		const { navigation } = this.props;
// 		return <Dummy navigation={navigation} />;
// 	}
// }

export default RoomsListView;
```

I think it's time to migrate everything to function components before migrating to Fabric 👀

Rohit3523:
> I think it's time to migrate everything to function components before migrating to Fabric 👀

You are just stuck in a update loop 🤣🤣

>> Apparently class components and function components behave differently in Fabric.

Do react native team still focus on the class components?? When I started my journey I found some articles to use the functional component instead of class 🤔

diegolmello:
@Rohit3523 

> You are just stuck in a update loop 🤣🤣

Well, it's part of the job 😅 

> Do react native team still focus on the class components?? When I started my journey I found some articles to use the functional component instead of class 🤔

We started the app before React Hooks even existed.
We always focus on making a better product and there was no real benefits in migrating (it could actually create new bugs).
That's why we still have quite a few class components.
Now we have a good reason.

diegolmello:
Fabric doesn't fix inline images :(

Rohit3523:
I think it's not React Native issue, we should check the styling and code for it 🫣

diegolmello:
@Rohit3523 Are you sure? https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/6239#issuecomment-2769941920

Rohit3523:
😬😬

app/views/ShareView/index.tsx
```diff
@@ -222,10 +222,12 @@ class ShareView extends Component<IShareViewProps, IShareViewState> {
 		};
 	};
 
-	startShareView = () => {
+	startShareView = async () => {
 		const startShareView = this.props.route.params?.startShareView;
 		if (startShareView) {
 			const { selectedMessages, text } = startShareView();
+			// Synchronization needed for Fabric to work
+			await new Promise(resolve => setTimeout(resolve, 100));
```
OtavioStasiak, line 230: We also have it in RoomView. Can we create a 'delay' function in` lib/helpers` for this?

diegolmello: I don't want to make it official and this is very straightforward.

PR #6488, opened 2025-07-10
feat(a11y): inline errors profileView

OtavioStasiak:
## Proposed changes

Create inline errors in ProfileView.

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-256

## How to test or reproduce

- Open the app;
- Go to ProfileView;
- Try to edit your profile;

## Types of changes

## Checklist


Rohit3523:
I noticed one issue but not sure if this PR introduce that 

If we change email to an existing one it shows the error modal but text input keeps the new email instead of reverting to old one even if I change the screen too 

Steps:-
1. Try to change email with existing one
2. Click on save and dismiss the modal 
3. observe the email field, it keep the new one
4. Change screen
5. Go to profile again
6. Observe email text input, it shows the new one

It only revert to old if we restart the app.

OtavioStasiak:
It's happening in development as well. I'll fix it in this PR. Thanks @Rohit3523  🙏🏻 
> I noticed one issue but not sure if this PR introduce that
> 
> If we change email to an existing one it shows the error modal but text input keeps the new email instead of reverting to old one even if I change the screen too
> 
> Steps:-
> 
> 1. Try to change email with existing one
> 2. Click on save and dismiss the modal
> 3. observe the email field, it keep the new one
> 4. Change screen
> 5. Go to profile again
> 6. Observe email text input, it shows the new one
> 
> It only revert to old if we restart the app.

app/views/ProfileView/index.tsx
```diff
@@ -266,6 +280,12 @@ const ProfileView = ({ navigation }: IProfileViewProps): React.ReactElement => {
 		navigation.setOptions(options);
 	}, []);
 
+	useFocusEffect(
+		useCallback(() => {
+			reset();
+		}, [])
+	);
```
diegolmello, line 289: I think this might be wrong.
The goal seems to be to reset the form when ProfileView is focused, but I wonder what's the behavior when the user navs to `edit avatar` or `preferences` and then navs back to ProfileView.

app/views/ProfileView/index.tsx
```diff
@@ -46,9 +48,9 @@ import useVerifyPassword from '../../lib/hooks/useVerifyPassword';
 const MAX_BIO_LENGTH = 260;
 const MAX_NICKNAME_LENGTH = 120;
 const validationSchema = yup.object().shape({
-	name: yup.string().min(1).required(),
-	email: yup.string().email().required(),
-	username: yup.string().min(1).required()
+	name: yup.string().required(I18n.t('Name_required')),
+	email: yup.string().email(I18n.t('Email_must_be_a_valid_email')).required(I18n.t('Email_required')),
+	username: yup.string().required(I18n.t('Username_required'))
```
diegolmello: This is going to create a validation schema based on the current i18n and it won't update the value when the user changes the language, since it's outside of the component.

app/lib/hooks/useA11yErrorAnnouncement.tsx
```diff
@@ -1,28 +1,38 @@
...
+		if (!hasError) {
+			announced.current = {};
+			previousMessages.current = {};
+			return;
+		}
+
+		Object.entries(errors).forEach(([fieldName, error]: [string, any]) => {
+			const message = error?.message?.trim();
+
+			if (message && message !== previousMessages.current[fieldName] && !announced.current[fieldName]) {
```
diegolmello: I'm not sure it works like that, but here it goes.

- User enters an invalid email
- App announces the error
- User enters another invalid email
- Is the app going to announce it again? It should, right?

app/lib/hooks/useA11yErrorAnnouncement.tsx
```diff
@@ -1,28 +1,38 @@
...
 			}
-		} else if (!error) {
-			announced.current = false;
-		}
+			if (!message) {
+				announced.current[fieldName] = false;
+				delete previousMessages.current[fieldName];
+			}
+		});
 	}, textInputDebounceTime);
```
diegolmello: I wonder if we should increase this value.
Debounce it going to trigger announcements after 500ms.
Depending on the user's condition, we might end up announcing it too often?

Not sure it happens, just making sure we're considering such scenario and being respectful.

diegolmello (changes requested):

app/views/ProfileView/index.tsx
OtavioStasiak: I don't think so, because when we navigate to 'Edit Avatar', for example, we are taken to another form with its own component. So, as we discussed earlier, let's keep clearing the form when navigating to another screen.

app/lib/hooks/useA11yErrorAnnouncement.tsx
OtavioStasiak: Agreed.

PR #5458, opened 2024-01-04
feat: add user authentication cookies to JitsiMeetView

GleidsonDaniel:
## Proposed changes

add user authentication cookies to JitsiMeetView

## Issue(s)	

closes: #2286

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [x] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

Jira: [SUP-453]

[SUP-453]: https://rocketchat.atlassian.net/browse/SUP-453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Issue #2286, opened 2020-07-14
No avatar in Jitsi call in mobile application when Block Unauthenticated Access to Avatars is enabled


ankar84:
- Your Rocket.Chat Experimental app version: 4.8.1.18375 - 4.16.2
- Your Rocket.Chat server version: 3.7.1 - 3.14.0
- Device (or Simulator) you're running with: Xiaomi Redmi Note 5

According to https://github.com/RocketChat/Rocket.Chat/issues/18117 we need to enable **Block Unauthenticated Access to Avatars** in **Admin UI - Accounts**
Desktop application and web version after that work fine - user appears in Jitsi call with their avatars.
But mobile version shows only user initials instead of avatar.
Mobile version
![image](https://user-images.githubusercontent.com/4023037/87401767-42e7bf00-c5c3-11ea-9e34-5846c370efb9.png)
Desktop version
![image](https://user-images.githubusercontent.com/4023037/87401822-54c96200-c5c3-11ea-9a04-48135ae6bb24.png)

ankar84:
@diegolmello @djorkaeffalexandre can you take a look at that issue?

ankar84:
I think it could be resolved in 4.12
Did a couple test and it looks like there is avatars in 4.12 with **Block Unauthenticated Access to Avatar**s enabled.
But I also checked it 4.7 - and same - there is avatars. So maybe server version 3.7.1 matters in that case.

diegolmello:
@ankar84 is this still an issue?

ankar84:
> @ankar84 is this still an issue?

Hi, @diegolmello 
I need your help in reproducing steps for that issue.
Last week I did try to reproduce that on 4.12 and got strange results. 
One user saw my avatar, but other is not.
I think it could be some cache. Users, that I talked previously have my avatar in application cache maybe.
So, I need a proper way to check if that issue is resolved in 4.12.1, can you help me with that?

diegolmello:
I see.
4.12 had Jitsi issues (not only voice related, but image related) and we solved them on 4.12.1.
Can you make sure everyone is up-to-date and try again?
Thank you.

ankar84:
> Can you make sure everyone is up-to-date and try again?

Just tested 4.12.1 with **Block Unauthenticated Access to Avatars** enabled and saw users avatars.
So, I think that issue is resolved!
Thank you!

ankar84:
[...]

PR #6507, opened 2025-07-27
fix: set cookie without port in JitsiMeetView to fix #5852

### LINKED ISSUE #5852: Mobile App show green preloader on "Join" to call
### Description:
Self hosted RocketCaht and Jitsi with JWT server.
Jitsi on a custom port.
(in rocketchat Marketpalce Jitsi app setting Domain: jitsi.EXAMPLE.com:8585)

In windows app and in browsers all for fine.
When I start video call or join to existed call via mobile app It show green preloader (see screenshot below).

When I just setup RocketCaht + Jitsi all work fine but after some updates I become to see preloader.
### Environment Information:

- Rocket.Chat Server Version: 6.11.1
- Rocket.Chat App Version: 4.50.1.63100
- Device Name: Oneplus
- OS Version: Android 14

### Steps to reproduce:

1. Open direct message on channel.
2. Click on "Call" button or "Join" call.

### Expected behavior:

Join to jitsi call. 

### Actual behavior:
323228829-47713aa7-7879-45c1-9c5e-efa39338ef5b

bezaleelcheung:
Fixes #5852

### Problem

When opening a Jitsi meeting in the Rocket.Chat React Native app, users see a green preloader and never actually join the call when using a custom Jitsi server with a non-standard HTTPS port (e.g., 8443 instead of 443).

This regression was introduced by the fix for [#2286](https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/2286), specifically in [PR #5458](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/5458), which added cookie support to the JitsiMeetView component.

### Root Cause

The code uses the full meeting URL to extract the cookie domain. However, when the Jitsi server is running on a non-default HTTPS port, the extracted domain includes the port (e.g., `example.com:8443`), which is not a valid domain for setting a cookie. Per the HTTP cookie spec, the domain should **not include the port**.

As a result, setting the cookie fails silently, and the JitsiMeetView fails to join the meeting.

### Fix

This PR ensures that the cookie domain passed to `JitsiMeetView` never includes a port number, even when the meeting URL uses a custom port.

### Testing

Tested with both:
- Standard Jitsi server at `https://meet.jit.si` (port 443)
- Custom Jitsi server at `https://example.com:8443`

In both cases, meetings can now be joined successfully.


bezaleelcheung:
@OtavioStasiak I see the circleCI error log in “BUild IOS Experimental” ：
OpenSSL::PKey::ECError: [!] invalid curve name
Is there anything I can do？

OtavioStasiak:
> @OtavioStasiak I see the circleCI error log in “BUild IOS Experimental” ： OpenSSL::PKey::ECError: [!] invalid curve name Is there anything I can do？



No, I am looking at this.

bezaleelcheung:
This is my first PR in github. I am so glad!

diegolmello:
@bezaleelcheung congrats! 🎉

PR #6567, opened 2025-08-15
fix: evaluate suggested e2e key when joining room

### LINKED ISSUE #6566: bug: starting an encrypted direct conversation from mobile fails to find key
### Describe the Bug

There is a race condition when starting a direct message from mobile to a recipient on the server website. The recipient will often create the room key and create a `E2ESuggestedKey` in the sender's subscription. But the sender will not evaluate it because it does not listen to subscription changes yet (it lost the race).

### Steps to Reproduce

1. Log in with the recipient user to the server website.
2. Make sure there is no direct message room existing between the sender and recipient.
3. Use the sender's mobile device to start a direct message to the recipient.
4. You will 90% of the time see that you should wait a while to get the E2E key and that will never happen if nothing else is done.

### Expected Behavior

Starting a direct message with someone should allow sending a message immediately, without waiting for keys.

### Actual Behavior

Starting a direct message for the first time with a recipient which is logged on to the server website will result in a "wait for keys" message that will not go away on its own.

7.9.3

### Rocket.Chat App Version

4.63

### Device Name

iPhone 16 Pro, Android Pixel 7

### OS Version

iOS 16, Android API 35

### Additional Context

_No response_

paviad:
## Proposed changes

When starting a direct message from a sender to a recipient while the recipient is logged on the server website, there is a race condition which the recipient usually wins that causes the sender to end up in a stuck state, waiting for room key.

## Issue(s)	

Depends on https://github.com/RocketChat/Rocket.Chat/pull/36731
Resolves #6566 
https://rocketchat.atlassian.net/browse/NATIVE-972

## How to test or reproduce

See the resolved issue.

## Screenshots

image

## Types of changes

## Checklist

## Further comments

**WARNING** do not merge this PR without also merging the server code PR https://github.com/RocketChat/Rocket.Chat/pull/36731

### DIFF
--- a/app/definitions/rest/v1/e2e.ts
+++ b/app/definitions/rest/v1/e2e.ts
@@ -29,6 +29,12 @@ export type E2eEndpoints = {
 	'e2e.setRoomKeyID': {
 		POST: (params: { rid: string; keyID: string }) => {};
 	};
+	'e2e.setOrGetRoomKeyID': {
+		POST: (params: { rid: string; keyID: string }) => {
+			keyID: string;
+			suggestedKey?: string;
+		};
+	};
 	'e2e.fetchMyKeys': {
 		GET: () => { public_key: string; private_key: string };
 	};
--- a/app/lib/encryption/room.ts
+++ b/app/lib/encryption/room.ts
@@ -93,24 +93,33 @@ export default class EncryptionRoom {
 			}
 		}
 
+		const tryProcessSuggestedKey = async (suggestedKey: string) => {
+			if (suggestedKey && Encryption.privateKey) {
+				try {
+					try {
+						this.establishing = true;
+						const { keyID, roomKey, sessionKeyExportedString } = await this.importRoomKey(suggestedKey, Encryption.privateKey);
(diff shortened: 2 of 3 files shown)


paviad:
Undid my previous change, I think I found the real race condition, and I've worked around it (it's not a proper fix, but it's a working workaround)

paviad:
It still isn't working in all circumstances, for example when creating a channel on a device, with a user on web, it is not possible to immediately send message if the race is lost. The only way to recover is to stop the app (kill it) and relaunch.

paviad:
Ok there was no avoiding it, a change is also required in the server, I've submitted a corresponding PR there (https://github.com/RocketChat/Rocket.Chat/pull/36731)

**WARNING** do not merge this PR without also merging the server PR

paviad:
this PR doesn't really solve the issue, withdrawing.

paviad:
I closed this because the PR doesn't solve the issue, it requires deeper thought the code is riddled with race conditions - I am still investigating.

Issue #6562, opened 2025-08-14
bug: only the last message is loaded when the connection is poor


sanyappc:
### Describe the Bug

When the connection is poor, it often happens that the mobile app does not load some of the messages before the last message in the chat

### Steps to Reproduce

1. Open the app with the internet turned off
2. Wait some time
3. Turn on the internet
4. Open a chat in which the last few messages have not been downloaded before
5. A chat window will open with several missed messages. Only the last message was downloaded after connecting to the Internet, while the previous ones are missing.

### Expected Behavior

Downloading the entire chat without missing any messages

### Actual Behavior

Some messages are missing in the chat

### Rocket.Chat Server Version

7.6.0

### Rocket.Chat App Version

4.63.0

### Device Name

any

### OS Version

any

diegolmello:
@sanyappc I was able to reproduce it.
I'm going to take a look at it ASAP.

PR #6581, opened 2025-08-21
fix: only last message is loaded when the connection is poor

OtavioStasiak:
## Proposed changes

When the connection is poor, it often happens that the mobile app does not load some of the messages before the last message in the chat. 
When init runs in RoomView, it calls RoomServices.getMessages. If the request exceeds the timeout, the error is handled inside getMessages itself and no error is returned to init. Because of that, the catch block in init is never triggered, and the retry logic does not run.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-970
Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/6562

## How to test or reproduce

- Open the app with the internet turned off;
- Wait some time;
- Turn on the internet;
- Open a chat in which the last few messages have not been downloaded before;
- Turn off the internet and wait some minutes;
- Turn on the internet;

## Types of changes

## Checklist


app/views/RoomView/index.tsx
```diff
@@ -607,13 +606,11 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 
 			this.setState({ canAutoTranslate, member, loading: false });
 		} catch (e) {
+			log(e);
```
diegolmello: This logic is meant to fail when there's connectivity issues.
We shouldn't log this, because it's only going to flood our analytics.

diegolmello (approved):
While this fix is probably the way to go atm, I'll make sure to work on something better on a future migration to hooks of RoomView.

Listening to SDK connection status is probably a better strategy.
Approving this one until then.

PR #6622, opened 2025-09-03
fix: reset of e2e key not working when 2FA is enabled

OtavioStasiak:
## Proposed changes

We started using `users.resetE2EKey` instead of `method.call/e2e.resetOwnE2EKey`. This also fixes a regression introduced here https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/6448 where the TwoFactorModal was not clearing errors when attempting a new fetch.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1000

## How to test or reproduce

- Enable two factor in your account;
- Open the app;
- Try to reset your e2e key;

## Types of changes

## Checklist


app/sagas/login.js
```diff
@@ -223,6 +224,8 @@ const handleLoginSuccess = function* handleLoginSuccess({ user }) {
 	try {
 		getUserPresence(user.id);
 
+		remove2FAHeaders();
```
diegolmello: We have this logic on backend to enforce 2fa on any endpoint we want https://github.com/RocketChat/Rocket.Chat/blob/84df8b6a0835d27ca6b62375a0f0ad1b3c458d19/apps/meteor/app/2fa/server/twoFactorRequired.ts#L6
Here's an example on save profile https://github.com/RocketChat/Rocket.Chat/blob/84df8b6a0835d27ca6b62375a0f0ad1b3c458d19/apps/meteor/server/methods/saveUserProfile.ts#L237

Removing the 2fa only on login success is not the way.
IMO we should come up with a more clever solution on the SDK side.
For example, manage a 2fa state inside the SDK that is cleared after the 2fa loop is finished.

My suggestion is to create a separate task for this scope.

app/views/E2EEncryptionSecurityView/index.tsx
```diff
@@ -57,7 +57,7 @@ const E2EEncryptionSecurityView = () => {
 					 * It might return an empty object when TOTP is enabled,
 					 * that's why we're using strict equality to boolean
```
diegolmello: Isn't this comment outdated now?

app/lib/services/restApi.ts
```diff
@@ -938,12 +938,12 @@ export const emitTyping = (room: IRoom, typing = true) => {
 	return sdk.methodCall('stream-notify-room', `${room}/typing`, name, typing);
 };
 
-export function e2eResetOwnKey(): Promise<boolean | {}> {
+export function e2eResetOwnKey(): Promise<{ success?: boolean }> {
 	// {} when TOTP is enabled
 	unsubscribeRooms();
 
 	// RC 0.72.0
```
diegolmello: You changed the endpoint, but didn't update the version of the server here.
When was this new endpoint introduced?
Should it be backwards compatible or is it old enough?

app/containers/TwoFactor/index.tsx
```diff
@@ -91,7 +92,9 @@ const TwoFactor = React.memo(() => {
 	const sendEmail = async () => {
 		try {
 			if (params?.user) {
+				clearErrors();
```
diegolmello, line 95: Do you know why we didn't need this before?

OtavioStasiak: My bad, I missed this during the migration to inline errors. It only happens in an edge case: when you open a modal, get an error, then succeed, and later open it again. I only noticed it now because I first got an error on 2FA login, then succeeded, and later opened the E2EE reset key 2FA modal.

PR #6680, opened 2025-09-24
fix: cannot load previous messages when lastMessage thread exceeds 50 messages

OtavioStasiak:
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1035

## How to test or reproduce

- Open the app;
- Enter a room where the last message is a thread with more than 50 messages;
- The loading of the thread and previous messages must work;

## Screenshots
| Before | After |
| ----- | ----- |
| Simulator Screenshot - iPhone 16 - 2025-09-24 at 19 37 40 | Simulator Screenshot - iPhone 16 - 2025-09-24 at 19 42 24 |

## Types of changes

## Checklist

## Summary by CodeRabbit

- Bug Fixes
  - More reliable, complete chat history loading when browsing older messages: backfilled messages now load in batches with limits and termination when enough main messages are collected, reducing gaps and stalls in large rooms.
- Tests
  - Added an automated end-to-end test that opens a room whose last message is a thread with over 50 replies and verifies thread count display and scroll-to-first-message behavior.

### DIFF
--- a/app/lib/methods/loadMessagesForRoom.ts
+++ b/app/lib/methods/loadMessagesForRoom.ts
@@ -10,6 +10,7 @@ import updateMessages from './updateMessages';
 import { generateLoadMoreId } from './helpers/generateLoadMoreId';
 
 const COUNT = 50;
+const COUNT_LIMIT = COUNT * 10;
 
 async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t: RoomTypes }): Promise<IMessage[]> {
 	const apiType = roomTypeToApiType(t);
@@ -21,7 +22,7 @@ async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t:
 	let mainMessagesCount = 0;
 
 	async function fetchBatch(lastTs?: string): Promise<void> {
-		if (allMessages.length >= COUNT * 10) {
+		if (allMessages.length >= COUNT_LIMIT) {
 			return;
 		}
 
@@ -77,7 +78,7 @@ export function loadMessagesForRoom(args: {
 			if (data?.length) {
 				const lastMessage = data[data.length - 1];
 				const lastMessageRecord = await getMessageById(lastMessage._id as string);
-				if (!lastMessageRecord && (data.length === COUNT || data.length >= COUNT * 10)) {
+				if (!lastMessageRecord && (data.length === COUNT || data.length >= COUNT_LIMIT)) {
 					const loadMoreMessage = {
 						_id: generateLoadMoreId(lastMessage._id as string),
 						rid: lastMessage.rid,


app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -11,23 +11,52 @@ import { generateLoadMoreId } from './helpers/generateLoadMoreId';
 
 const COUNT = 50;
 
-async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t: RoomTypes }): Promise<IMessage[]> {
-	let params = { roomId, count: COUNT } as { roomId: string; count: number; latest?: string };
-	if (latest) {
-		params = { ...params, latest: new Date(latest).toISOString() };
-	}
+interface HistoryParams {
```
diegolmello: There's this non-written convention

interface: ISomething
type: TSomething

That said, it's only working because `sdk.get` accepts anything, because it should really check `definitions/rest/v1`

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -11,23 +11,52 @@ import { generateLoadMoreId } from './helpers/generateLoadMoreId';
...
+		}
+
+		const batch = data.messages as IMessage[];
+		allMessages.push(...batch);
+
+		const mainMessagesInBatch = batch.filter(message => !message.tmid);
+		mainMessagesCount += mainMessagesInBatch.length;
+
+		const needsMoreMainMessages = mainMessagesCount < COUNT;
+		const hasMoreMessages = batch.length === COUNT;
```
diegolmello: If the room has exactly 50 messages, `batch.length === COUNT` but that doesn't mean there's more messages to be loaded.

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -11,23 +11,52 @@ import { generateLoadMoreId } from './helpers/generateLoadMoreId';
...
+
+		if (needsMoreMainMessages && hasMoreMessages) {
+			const lastMessage = batch[batch.length - 1];
+			return fetchBatch(lastMessage.ts as string);
+		}
```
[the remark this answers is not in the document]
diegolmello: @OtavioStasiak we should do the first suggesting as a safety limit

OtavioStasiak:
> Shouldn't we make something similar on loadMissedMessages?

No, we shouldn’t, because loadMissedMessages already has its own logic to handle cursor-based pagination.

https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/53e26daf815711dbd16e827f1d3bf74b693f0c74/app/lib/methods/loadMissedMessages.ts#L111

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -12,22 +12,57 @@ import { generateLoadMoreId } from './helpers/generateLoadMoreId';
...
+				break;
+			case 'groups':
+				data = await sdk.get('groups.history', params);
+				break;
+			case 'im':
+				data = await sdk.get('im.history', params);
+				break;
+			default:
+				return;
+		}
```
diegolmello, line 44: Did you intentionally rewrote this?
Don't you think the old version is clear enough?

```
const data = await sdk.get(`${apiType}.history`, params);
```

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -12,22 +12,57 @@ import { generateLoadMoreId } from './helpers/generateLoadMoreId';
 const COUNT = 50;
```
diegolmello, line 14:
```suggestion
const COUNT_LIMIT = COUNT * 10;
```

I think we do this and apply it on both places below.

app/lib/methods/loadMessagesForRoom.ts
OtavioStasiak: I’ll keep it to respect the parameter types.

PR #6682, opened 2025-09-24
fix: reset form values when user changes

OtavioStasiak:
## Proposed changes

The form in ProfileView wasn’t updating when user changed because react-hook-form only applies defaultValues on the first render. This fix resets the form with the latest user values.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1036

## How to test or reproduce

- Open the app;
- Navigate to ProfileView;
- Edit Bio;
- Back to RoomsListView;
- Navigate to ProfileView;

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Profile form now reliably resets to the updated user data after a successful save (name, username, email chosen from primary address, bio, nickname) and clears the current password field.

### DIFF
--- a/app/views/ProfileView/index.tsx
+++ b/app/views/ProfileView/index.tsx
@@ -274,8 +274,19 @@ const ProfileView = ({ navigation }: IProfileViewProps): React.ReactElement => {
 
 	useFocusEffect(
 		useCallback(() => {
-			reset();
-		}, [])
+			if (user) {
+				reset({
+					name: user?.name || '',
+					username: user?.username || '',
+					email: user?.emails?.[0]?.address || '',
+					currentPassword: null,
+					bio: user?.bio || '',
+					nickname: user?.nickname || '',
+					saving: false
+				});
+				setCustomFields(user?.customFields ?? {});
+			}
+		}, [user, reset])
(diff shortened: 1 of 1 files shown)


app/views/ProfileView/index.tsx
```diff
@@ -274,8 +274,19 @@ const ProfileView = ({ navigation }: IProfileViewProps): React.ReactElement => {
 
 	useFocusEffect(
 		useCallback(() => {
-			reset();
-		}, [])
+			if (user) {
```
diegolmello: This is adding reactivity, but not fixing the actual issue.
The issue is after a submit, the values the user just entered are cleared.
That's still going to happen, but after the new user state gets to the app through websocket and redux, it's going to apply the new value on a re-render.

What you really want to do is to change the way we handle the form after the submit.

This reactivity also introduces unexpected behaviors.
Imagine you're changing something on your profile and something changes your user, like user status (a tab switch on web).
It's going to trigger reset and the current data is lost.
This is quite sensitive if you're changing your password, for example.

diegolmello (changes requested):

PR #6719, opened 2025-10-12
chore: migrate from expo-av to expo-audio and expo-video

Rohit3523:
## Issue(s)	

https://rocketchat.atlassian.net/browse/CORE-2035

## Screenshots
| Senerio | Android | iOS |
| --- | --- | --- |
| Opening a channel displays a play button in the audio message & playing it | <video src="https://github.com/user-attachments/assets/c71f36c3-45d7-4671-be21-0817969d50f4" controls width="640" /> | <video src="https://github.com/user-attachments/assets/4c7cadda-2fc4-4612-8e97-321d178adea6" controls width="640" /> |
| seek bar, playback speed, and pause | <video src="https://github.com/user-attachments/assets/cc45d09f-8699-4896-829f-9cf018015271" controls width="640" /> | <video src="https://github.com/user-attachments/assets/9b7e082a-6275-45ae-9950-78e4fd441314" controls width="640" /> |
| Video message | <video src="https://github.com/user-attachments/assets/ebbd0515-ac73-4993-8cd2-a8f97db38c2b" controls width="640" /> | <video src="https://github.com/user-attachments/assets/53c2284a-042a-4bf8-8995-67d4bba193a8" controls width="640" /> |
| Video view: everything is displayed within the safe area | Screenshot 2026-09-02 at 9 10 34 PM | Screenshot 2026-09-02 at 9 10 22 PM |
| Video ShareView | <video src="https://github.com/user-attachments/assets/775629c2-79d6-40db-b2f6-e4acc2b42d3c" controls width="640" /> | <video src="https://github.com/user-attachments/assets/4cb54984-6c67-43bc-a773-c02f05c16947" controls width="640" />

## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added video playback with fullscreen, picture-in-picture, native controls, loading states, and playback error handling.
  * Added audio recording, playback, seeking, looping, and background playback support.
  * Added localized video playback error messages across supported languages.

* **Bug Fixes**
  * Improved audio seek recovery and playback-position restoration.
  * Improved handling of video playback failures and recording permission issues.

* **Tests**
  * Added coverage for audio management, video playback, recording behavior, and playback error scenarios.


Rohit3523:
Recording is not working... idk why...

Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

app/containers/AudioPlayer/index.tsx
```diff
@@ -47,35 +47,34 @@ const AudioPlayer = ({
...
-			duration.value = durationSeconds > 0 ? durationSeconds : 0;
-			const currentSecond = data.positionMillis / 1000;
-			if (currentSecond <= durationSeconds) {
-				currentTime.value = currentSecond;
-			}
+	const handlePlaybackStatusUpdate = (data: AudioStatus) => {
+		if (!data.isLoaded) return;
+		const durationSeconds = data.duration ?? 0;
+		duration.value = durationSeconds > 0 ? durationSeconds : 0;
+		const currentSecond = data.currentTime ?? 0;
```
diegolmello: Why `??` instead of `||`?

app/containers/MessageComposer/components/RecordAudio/RecordAudio.tsx
```diff
@@ -36,45 +43,44 @@ export const RecordAudio = (): ReactElement | null => {
 	const permissionToUpload = useCanUploadFile(rid);
 	useKeepAwake();
 
+	async function doRecording() {
```
diegolmello: Any reason to rename and move away from the effect?

app/containers/Ringer/index.tsx
```diff
@@ -1,32 +1,27 @@
...
+	[ERingerSounds.DIALTONE]: require('./dialtone.mp3'),
+	[ERingerSounds.RINGTONE]: require('./ringtone.mp3')
+} as const;
+
 const Ringer = React.memo(({ ringer }: { ringer: ERingerSounds }) => {
-	const sound = useRef(new Audio.Sound());
+	const player = useAudioPlayer(RINGER_SOUND_FILES[ringer]);
 
 	useEffect(() => {
-		const loadAndPlay = async () => {
```
diegolmello, line 13: Any reason to remove the conditional loading?

app/lib/constants/audio.ts
```diff
@@ -1,48 +1,22 @@
...
 		extension: RECORDING_EXTENSION,
-		outputFormat: Audio.AndroidOutputFormat.AAC_ADTS,
-		audioEncoder: Audio.AndroidAudioEncoder.AAC,
-		// Settings related to audio quality.
-		sampleRate: Audio.RecordingOptionsPresets.LOW_QUALITY.android.sampleRate,
-		numberOfChannels: Audio.RecordingOptionsPresets.LOW_QUALITY.android.numberOfChannels,
-		bitRate: Audio.RecordingOptionsPresets.LOW_QUALITY.android.bitRate
+		outputFormat: 'aac_adts',
+		audioEncoder: 'aac',
+		sampleRate: 44100
```
diegolmello: This magic number is wrong. They probably have defaults, no?

app/lib/methods/AudioManager.ts
```diff
@@ -1,122 +1,158 @@
...
+	function addAudioRendered(audioKey: string) {
+		audiosRendered.add(audioKey);
+	}
 
-const getAudioKey = ({ msgId, rid, uri }: { msgId?: string; rid: string; uri: string }) => `${msgId}-${rid}-${uri}`;
+	function removeAudioRendered(audioKey: string) {
+		audiosRendered.delete(audioKey);
+	}
 
-class AudioManagerClass {
```
diegolmello, line 14: Why did you refactor this? What's the benefit?
You also refactored, but didn't add any tests.

app/views/AttachmentView.tsx
```diff
@@ -18,13 +19,58 @@ import I18n from '../i18n';
...
 import { useAppNavigation, useAppRoute } from '../lib/hooks/navigation';
-import { formatAttachmentUrl, isAndroid, fileDownload, showErrorAlert } from '../lib/methods/helpers';
+import { formatAttachmentUrl, isAndroid, fileDownload } from '../lib/methods/helpers';
 import EventEmitter from '../lib/methods/helpers/events';
 import { getUserSelector } from '../selectors/login';
 import { type TNavigation } from '../stacks/stackType';
 import { useTheme } from '../theme';
 import { LOCAL_DOCUMENT_DIRECTORY, getFilename } from '../lib/methods/handleMediaDownload';
 
+const VideoContent = ({
```
diegolmello: Different file with stories and tests

diegolmello (changes requested):
Besides this, make sure to resolve code rabbit

Rohit3523:
@coderabbitai review

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -69,23 +71,23 @@ const Seek = ({ currentTime, duration, loaded = false, onChangeTime }: ISeek) =>
 		.onStart(() => {
 			isPanning.value = true;
 			contextX.value = translateX.value;
+			savedTranslateX.value = translateX.value;
+			savedCurrentTime.value = currentTime.value;
```
Rohit3523, line 75: We are temporary storing the current play time, so if the seek fail, we can revert the time

app/i18n/locales/zh-TW.json
```diff
@@ -222,6 +222,7 @@
   "Enter_E2EE_Password_description": "輸入您的E2EE密碼以查看和發送加密訊息。\n\n每個設備都需要輸入密碼。",
   "Enter_manually": "手動輸入",
   "Enter_username_or_number": "輸入使用者名稱或號碼",
+  "Error": "Error",
```
OtavioStasiak: new I18N labels are not being translated

app/containers/MessageComposer/components/RecordAudio/RecordAudio.tsx
```diff
@@ -61,20 +66,25 @@ export const RecordAudio = (): ReactElement | null => {
 				}
 			}
 		};
+		numberOfTriesRef.current = 0;
```
OtavioStasiak: Is this reset necessary? Since this effect only runs once ([]), numberOfTriesRef.current should already be 0 when record() is first called. Was this added to address a specific scenario where the retry counter persists across recording attempts?

app/containers/AudioPlayer/Seek.tsx
```diff
@@ -69,23 +71,23 @@ const Seek = ({ currentTime, duration, loaded = false, onChangeTime }: ISeek) =>
...
 		.onFinalize((_, didSucceed) => {
-			if (isPanning.value && !didSucceed) {
-				translateX.value = contextX.value;
-				currentTime.value = (contextX.value * duration.value) / maxWidth.value || 0;
-			}
-
 			isPanning.value = false;
 			scale.value = withTiming(1, { duration: 150 });
+			if (didSucceed) {
+				scheduleOnRN(onChangeTime, currentTime.value);
```
OtavioStasiak, line 86: why now we must to schedule it again? I do not understand the new approach.

OtavioStasiak (changes requested):

app/containers/AudioPlayer/Seek.tsx
Rohit3523: Gesture.Pan() callbacks are worklets, they run on the UI thread, not the JS thread. `onChangeTime` is a plain JS function (it ends up in AudioManager.setPositionAsync), so it can't be called directly from there.

scheduleOnRN(onChangeTime, currentTime.value) marshals the call across the thread boundary: it copies the argument and queues onChangeTime to run on the JS thread. It's the `react-native-worklets` replacement for Reanimated's old runOnJS, same mechanism, new name.

app/containers/MessageComposer/components/RecordAudio/RecordAudio.tsx
```diff
@@ -61,20 +66,25 @@ export const RecordAudio = (): ReactElement | null => {
...
-				recordingRef.current?.stopAndUnloadAsync();
+				audioRecorder.stop();
 			} catch {
 				// Do nothing
 			}
 		};
 	}, []);
 
+	useEffect(() => {
+		durationRef.current?.onRecordingStatusUpdate?.(recorderState);
```
OtavioStasiak: The `durationRef` + `useImperativeHandle` pattern made sense with expo-av: `setOnRecordingStatusUpdate` was a callback outside React, so pushing
  status into `Duration` kept `RecordAudio` from re-rendering.

  That no longer holds. `useAudioRecorderState(audioRecorder)` polls and `setState`s in `RecordAudio` itself, so the parent re-renders every tick
  anyway — and this effect then triggers a *second* render in `Duration`. We kept the ref ceremony and lost the benefit.

  Suggestion: pass `audioRecorder` down and call `useAudioRecorderState` inside `Duration`. That actually restores the isolation (only `Duration`
  re-renders) and lets us drop `IDurationRef`, `forwardRef`, `useImperativeHandle`, `durationRef` and this effect. `RecordAudio` doesn't need
  `recorderState` for anything else if we use `audioRecorder.uri` for the review player instead of `recorderState.url` — which `sendAudio` already
  does, and which avoids reading a value that can be up to one poll interval stale right after `stop()`.

app/containers/VideoPlayer/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+interface IVideoPlayerProps {
+	attachment: IAttachment;
+	user: { id: string; token: string };
+	baseUrl: string;
+	setLoading: Dispatch<SetStateAction<boolean>>;
+}
+
+const VideoPlayer = ({ attachment, user, baseUrl, setLoading }: IVideoPlayerProps): ReactElement => {
+	const navigation = useAppNavigation();
+	const hasHandledErrorRef = useRef(false);
```
OtavioStasiak, line 20: hasHandledErrorRef never resets. It's keyed to the component instance, not to uri. Today AttachmentView is route-scoped so attachment can't change
  under a mounted player — but that's an invariant one screen away from the component, not something this file enforces. Reset it in an effect on
  [uri], or the guard silently stops working the day someone reuses this.

app/containers/VideoPlayer/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+	});
+
+	useEffect(() => {
+		const blurSub = navigation.addListener('blur', () => {
+			player.pause();
+		});
+		return () => {
+			blurSub();
+		};
+	}, [navigation, player]);
```
OtavioStasiak, line 51: Pause on blur, but never resume on focus (:40-47). If that's intentional, fine — but it's the kind of asymmetry worth a comment, since
  allowsPictureInPicture is on and the interaction between PiP and the blur pause isn't obvious from reading this.

OtavioStasiak (changes requested):

patches/expo-audio+0.4.9.patch
Rohit3523, line 1: On Android, we were previously using `MediaPlayer` through `expo-av`

https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/2edc50b432b4d9aba0dc792129bd7be28103855b/app/lib/methods/AudioManager.ts#L39
Note: Above code is from develop branch

With the migration to `expo-audio`, the Android implementation switched to ExoPlayer and seeking stopped working for some audio files.

This patch enables constant-bitrate seeking in ExoPlayer, which fixes the seeking behavior for those files.

app/containers/VideoPlayer/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+const attachment = {
+	title_link: 'https://open.rocket.chat/video.mp4',
+	video_url: 'https://open.rocket.chat/video.mp4'
+};
+
+const user = { id: 'user-id', token: 'token' };
+
+const setLoading = () => undefined;
+
+export const Basic = (): ReactElement => (
```
OtavioStasiak, line 18: Story with no snapshot test — app/containers/VideoPlayer/index.stories.tsx.
  De-facto convention (.rnstorybook/generateSnapshots.tsx + BackgroundContainer/index.test.tsx, UnreadBadge, Markdown, ReactionsList, UIKit, LoginServices): every *.stories.tsx has a companion test whose body
  is generateSnapshots(stories). VideoPlayer/index.test.tsx never imports it, so the new stories are never exercised.

app/containers/VideoPlayer/index.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+	<VideoPlayer attachment={attachment as any} user={user} baseUrl='https://open.rocket.chat' setLoading={setLoading} />
+);
+
+const Themed = ({ theme }: { theme: TSupportedThemes }): ReactElement => (
```
OtavioStasiak: Same file, broken theme stories: export { Themed as Light, Themed as Dark } exports one component twice with no theme prop bound — both render theme === undefined.
  BackgroundContainer/index.stories.tsx:29-35 binds the theme per story (<ThemeStory theme='dark' …/>).

app/i18n/locales/en.json
```diff
@@ -920,6 +921,7 @@
   "The_user_wont_be_able_to_type_in_roomName": "The user won't be able to type in {{roomName}}",
   "Theme": "Theme",
   "There_was_an_error_while_action": "There was an error while {{action}}!",
+  "There_was_an_error_while_playing_video": "There was an error while playing the video.",
```
OtavioStasiak, line 923: 5. i18n duplication. Error_play_video ("There was an error while playing this video") is now unreferenced in app/ but left in all locale files, while a near-identical There_was_an_error_while_playing_video
  was added to all of them. One of the two should go.

OtavioStasiak (changes requested):

Rohit3523:
@coderabbitai review

app/lib/constants/audio.ts
```diff
@@ -1,48 +1,41 @@
...
-		// Settings related to audio encoding.
-		extension: RECORDING_EXTENSION,
-		audioQuality: Audio.IOSAudioQuality.MEDIUM,
-		outputFormat: Audio.IOSOutputFormat.MPEG4AAC,
-		// Settings related to audio quality.
-		sampleRate: Audio.RecordingOptionsPresets.LOW_QUALITY.ios.sampleRate,
-		numberOfChannels: Audio.RecordingOptionsPresets.LOW_QUALITY.ios.numberOfChannels,
-		bitRate: Audio.RecordingOptionsPresets.LOW_QUALITY.ios.bitRate
+		audioQuality: AudioQuality.MEDIUM,
+		outputFormat: IOSOutputFormat.MPEG4AAC,
```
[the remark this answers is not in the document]
Rohit3523: in the previous code, it was using the same format

app/containers/VideoPlayer/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+	baseUrl: string;
+	setLoading: Dispatch<SetStateAction<boolean>>;
+}
+
+const VideoPlayer = ({ attachment, user, baseUrl, setLoading }: IVideoPlayerProps): ReactElement => {
+	const navigation = useAppNavigation();
+	const hasHandledErrorRef = useRef(false);
+	const { bottom } = useSafeAreaInsets();
+
+	const url = formatAttachmentUrl(attachment.title_link || attachment.video_url, user.id, user.token, baseUrl);
```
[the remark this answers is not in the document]
Rohit3523: Pre-existing pattern moved verbatim from AttachmentView.tsx:78 on develop — out of scope for this migration; flag it against develop if you want it changed everywhere.

package.json
```diff
@@ -63,7 +63,7 @@
 		"eslint-import-resolver-typescript": "^4.4.4",
 		"expo": "^54.0.0",
 		"expo-apple-authentication": "~8.0.8",
-		"expo-av": "~16.0.8",
+		"expo-audio": "~0.4.9",
```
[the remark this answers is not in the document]
Rohit3523: Valid catch, but deferring deliberately — not in this PR:

- `0.4.9` wasn't picked by accident: this branch started on Expo SDK 53 (`"expo": "53"` at migration commit `8f79396`), where `0.4.9` was the bundled `expo-audio`. It drifted when the branch merged up to SDK 54. (Side note: `0.4.9` does exist — it's installed from npm in this branch; it was SDK 53's line, not an alpha.)
- Upgrading is not a one-line bump: `patches/expo-audio+0.4.9.patch` (the Android ExoPlayer constant-bitrate-seeking fix) is written against 0.4.9's `AudioModule.kt` and has to be regenerated and re-validated against 1.1.1's native code; the JS surface across AudioManager / RecordAudio / Ringer / VoIP / tests needs re-verification; and all audio/video QA evidence on this PR was captured on 0.4.9, so an upgrade invalidates it and requires full re-QA on both platforms.
- `expo install --check` flags many pre-existing drifts repo-wide (screens, webview, file-system, …) — strict alignment isn't the enforced bar, and none of those are in scope here either.

app/lib/methods/AudioManager.ts
```diff
@@ -153,20 +213,23 @@ class AudioManagerClass {
 		if (!rid) {
 			return;
 		}
-		const regExp = new RegExp(rid);
-		const roomAudioKeysLoaded = Object.keys(this.audioQueue).filter(audioKey => regExp.test(audioKey));
+		const roomAudioKeysLoaded = Object.keys(this.audioQueue).filter(audioKey => this.audioMeta[audioKey]?.rid === rid);
```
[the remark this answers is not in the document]
Rohit3523: Verified the mechanism — `onEnd` releases the native player but leaves the JS entries (`audioUris`/`audioPositions`/`audioRates`/`audioCallbacks`/`audioMeta`) behind, and this function skips those keys. Declining anyway: the retained entries are a few hundred bytes of plain objects per played message with no native-resource leak and no observed symptom. Churning the unload path for that in a migration PR is not worth the re-QA surface. Happy to take it as a separate cleanup if you want it tracked.

app/lib/methods/AudioManager.ts
```diff
@@ -153,20 +213,23 @@ class AudioManagerClass {
...
 		try {
-			await Promise.all(
-				roomAudiosLoaded.map(async audio => {
-					await audio?.stopAsync();
-					await audio?.unloadAsync();
-				})
-			);
-		} catch {
-			// Do nothing
+			await Promise.all(roomAudiosLoaded.map(audio => audio?.release()));
```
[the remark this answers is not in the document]
Rohit3523: Declining: per your own analysis `release()` is sync and not documented to throw, and the codebase calls it bare in 6 other sites (`onEnd:152`, Ringer, `playCallEndedSound`, Dialpad) — isolating just this one is inconsistent. The existing try/catch covers the documented (async) failure mode. Happy to revisit with a crash report or repro.

patches/expo-audio+0.4.9.patch
```diff
@@ -0,0 +1,24 @@
...
+ import androidx.media3.exoplayer.source.ProgressiveMediaSource
+ import expo.modules.interfaces.permissions.Permissions
+@@ -553,7 +554,10 @@ class AudioModule : Module() {
+       CONTENT_TYPE_SS -> SsMediaSource.Factory(factory)
+       CONTENT_TYPE_DASH -> DashMediaSource.Factory(factory)
+       CONTENT_TYPE_HLS -> HlsMediaSource.Factory(factory)
+-      CONTENT_TYPE_OTHER -> ProgressiveMediaSource.Factory(factory)
++      CONTENT_TYPE_OTHER -> ProgressiveMediaSource.Factory(
++        factory,
++        DefaultExtractorsFactory().setConstantBitrateSeekingEnabled(true)
```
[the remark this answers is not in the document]
Rohit3523: The patch is scoped to the regressed files by design: the seek failures vs `MediaPlayer` were on progressive MP3/ADTS voice-message attachments, which is exactly what `setConstantBitrateSeekingEnabled` covers. MP4/M4A extraction is intentionally untouched (those carry their own seek index). A native regression fixture is not feasible here — this repo has JVM unit tests only (`android/app/src/test`), no `androidTest` instrumentation harness to host an ExoPlayer fixture. If you have a specific failing M4A attachment, share it and I will verify it against this patch.

app/views/ShareView/Preview.tsx
```diff
@@ -16,6 +17,30 @@ import { type IShareAttachment } from '../../definitions';
...
+const VideoPreview = memo(({ uri, width, height }: { uri: string; width?: number; height?: number }) => {
+	const player = useVideoPlayer(uri, player => {
+		player.pause();
+	});
+	const hasHandledErrorRef = useRef(false);
+
+	useEventListener(player, 'statusChange', ({ status }) => {
+		if (status === 'error' && !hasHandledErrorRef.current) {
+			hasHandledErrorRef.current = true;
+			Alert.alert(I18n.t('Error'), I18n.t('There_was_an_error_while_playing_video'));
```
OtavioStasiak: Raw Alert.alert instead of showErrorAlert — app/containers/VideoPlayer/index.tsx:38, app/views/ShareView/Preview.tsx:29. The removed code in AttachmentView.tsx
     used showErrorAlert(I18n.t('Error_play_video')). app/lib/methods/helpers/info.ts:6 is the established helper (only 16 raw Alert.alert remain in app/).

app/containers/Ringer/index.tsx
```diff
@@ -1,31 +1,30 @@
...
-
 	useEffect(() => {
-		const loadAndPlay = async () => {
+		const soundFile = ringer === ERingerSounds.DIALTONE ? require('./dialtone.mp3') : require('./ringtone.mp3');
+		const player = createAudioPlayer(soundFile);
+		try {
+			player.loop = true;
+			player.play();
+		} catch (error) {
+			console.error('Error playing sound:', error);
```
OtavioStasiak: console.error instead of log() — app/containers/Ringer/index.tsx:20,28. The diff rewrote these exact lines and kept console.error, while AudioManager.ts in the
     same diff correctly added import log from './helpers/log'. Same for the silent catch {} at AudioManager.ts:~78 next to a log(error) at ~224.

app/lib/methods/AudioManager.ts
```diff
@@ -30,67 +43,115 @@ class AudioManagerClass {
...
 		try {
-			await this.audioQueue[audioKey]?.setPositionAsync(time);
+			await this.audioQueue[audioKey]?.seekTo(time);
 		} catch {
 			// Do nothing
 		}
 	}
 
-	async setRateAsync(audioKey: string, value = 1.0) {
+	setRateAsync(audioKey: string, value = 1.0): void {
```
OtavioStasiak: 3. Async suffix on a synchronous method — AudioManager.setRateAsync() now returns void (app/lib/methods/AudioManager.ts:~118); callers still await it.

app/lib/methods/AudioManager.ts
```diff
@@ -30,67 +43,115 @@ class AudioManagerClass {
...
+		this.audioCallbacks[audioKey] = callback;
+		this.audioSubscriptions[audioKey]?.();
+		const sub = this.audioQueue[audioKey]?.addListener('playbackStatusUpdate', status => {
+			this.onPlaybackStatusUpdate(audioKey, status, callback);
+		});
+		if (sub) this.audioSubscriptions[audioKey] = () => sub.remove?.();
 	}
 
-	async onEnd(audioKey: string, status: AVPlaybackStatus) {
+	getCurrentStatus(audioKey: string): AudioStatus | null {
```
OtavioStasiak: ** Dead code — AudioManager.getCurrentStatus() (AudioManager.ts:139) has zero call sites.**

app/containers/VideoPlayer/index.tsx
```diff
@@ -0,0 +1,70 @@
+import { useVideoPlayer, VideoView } from 'expo-video';
+import { useEffect, useRef, type Dispatch, type SetStateAction, type ReactElement } from 'react';
+import { Alert, View } from 'react-native';
+import { useEventListener } from 'expo';
+
+import { styles } from './styles';
```
OtavioStasiak, line 7: Import grouping — VideoPlayer/index.tsx:10: react-native-safe-area-context imported after the local ./styles/../../lib block. Not lint-enforced, but universal
     here, including sibling Preview.tsx.

OtavioStasiak (changes requested):

diegolmello (commented):
Reviewed the migration at ee140b055f. Two playback regressions below have focused fixes. The completed-audio cleanup finding already has an inline thread; additional reproduction evidence is posted there. No broader AudioManager redesign is needed. Validation used the pinned native source and the current manager with native/database stubs; local Jest could not run because of dependency errors.

app/lib/methods/AudioManager.ts
```diff
@@ -30,72 +43,116 @@ class AudioManagerClass {
...
+		this.audioRates[audioKey] = value;
+		if (!audioKey || this.audioPlaying !== audioKey) {
+			return;
+		}
+
 		try {
-			await this.audioQueue[audioKey].setRateAsync(value, true);
-		} catch {
-			// Do nothing
+			this.audioQueue[audioKey]?.setPlaybackRate(value);
```
diegolmello: **[P2 — Medium] Preserve pitch when changing playback speed**

**Explain like I'm 5:** Making a voice message faster should sound like the same person talking quickly. This change also makes their voice higher, like a cartoon character. We lost the setting that keeps the voice sounding normal.

**Details:** At reviewed commit `ee140b055f`, the old `setRateAsync(value, true)` has become `setPlaybackRate(value)`. The old `true` enabled pitch correction. In the pinned expo-audio 0.4.9 Android source, `AudioPlayer.preservesPitch` defaults to `false`; `AudioModule.setPlaybackRate` then passes the selected rate as both speed and pitch to `PlaybackParameters`. Nothing in this manager enables `shouldCorrectPitch`. The iOS implementation also needs an explicit correction quality: when it is omitted, this version selects `.varispeed`.

**Severity:** P2 / medium. Audio still plays, but changing speed alters the speaker's voice and regresses the previous listening experience. This is not data loss or a playback outage.

**Likelihood:** High whenever an Android user selects a non-default speed such as 1.5× or 2×. The native code deterministically couples pitch to speed with the current settings. Frequency across users is unknown; 1× listeners are unaffected by this particular issue.

**Steps to reproduce:**

1. On Android, open a downloaded voice message with recognizable speech.
2. Play it at 1×, then select 1.5× or 2×.
3. Compare the speaker's pitch with the same message at 1× or with the pre-migration implementation.

Expected: speech becomes faster while the speaker's pitch stays natural. Actual, according to the pinned native implementation: pitch rises with speed. Device listening validation remains to be done; this finding was verified from source.

**Proposed change:** Keep this in `setRate`, after the existing active-player guard. This path also applies to recreated players. Explicit `'medium'` quality avoids the iOS `.varispeed` fallback.

```diff
 		try {
-			this.audioQueue[audioKey]?.setPlaybackRate(value);
+			const player = this.audioQueue[audioKey];
+			if (!player) return;
+			player.shouldCorrectPitch = true;
+			player.setPlaybackRate(value, 'medium');
 		} catch (error) {
 			log(error);
 		}
```

**Validation after the fix:** Check that both initial playback and replay after completion enable correction before applying the rate. Listen at 1×, 1.5× and 2× on Android and iOS, and confirm loading an idle player still does not start playback.

app/lib/methods/AudioManager.ts
```diff
@@ -30,72 +43,116 @@ class AudioManagerClass {
...
+			return;
+		}
+
 		if (status.isLoaded && status.didJustFinish) {
 			try {
-				await this.audioQueue[audioKey]?.stopAsync();
+				this.audioSubscriptions[audioKey]?.();
+				delete this.audioSubscriptions[audioKey];
+				this.audioQueue[audioKey].release();
+				delete this.audioQueue[audioKey];
```
diegolmello, line 152: **[P2 — Medium] Keep completed messages eligible for sequential replay**

**Explain like I'm 5:** Imagine two songs on a playlist. After playing them once, we put the second song's player away. When you replay the first song, the app thinks the second song is gone—even though it still knows where the song is and could get its player back out.

**Details:** At reviewed commit `ee140b055f`, `onEnd` releases and deletes the finished player from `audioQueue`, retaining its source for later recreation. But `playNextAudioInSequence` still requires `audioQueue[nextAudioInSeqKey]` before it calls `playAudio`. That guard rejects a completed next message, so `playAudio` never gets the chance to recreate its player. The retained source and live player now have different lifetimes, while sequence eligibility still treats them as equivalent.

**Severity:** P2 / medium. Automatic advancement breaks when replaying a previously completed sequence. Messages remain available and can be played manually; there is no data loss.

**Likelihood:** Deterministic for the sequence below while both messages remain rendered and their components have not reloaded the players. User-level frequency is unknown and depends on replaying adjacent voice messages. A first playback of freshly loaded messages does not exercise this condition.

**Steps to reproduce:**

1. Open a room with two consecutive downloaded voice messages, A followed by B, with no intervening message. Keep both rendered.
2. Play A and allow automatic playback to advance to B.
3. Let B finish completely.
4. Without leaving the room, remounting the messages or scrolling them away, replay A and let it finish.

Expected: B starts automatically again. Actual: playback stops after A because B's player has been removed from `audioQueue`.

**Validation performed:** Executed the current manager with native/database stubs. After completing B, supplied B as A's next message while B remained in `audiosRendered`. Its source was retained, but `playAudio` received zero calls. This validates the JavaScript control-flow defect; the full device flow remains to be exercised.

**Proposed change:** Check the retained source instead of requiring a live player. Keep the rendered-message guard and let `playAudio` recreate the released player.

```diff
-			if (nextAudioInSeqKey && this.audioQueue?.[nextAudioInSeqKey] && this.audiosRendered.has(nextAudioInSeqKey)) {
+			if (nextAudioInSeqKey && this.audioUris[nextAudioInSeqKey] && this.audiosRendered.has(nextAudioInSeqKey)) {
 				await this.playAudio(nextAudioInSeqKey);
 			}
```

**Validation after the fix:** Add a regression case for completing A → B and replaying A, asserting that B is recreated and played. Also verify that unrendered messages do not auto-play and room cleanup removes their eligibility.

app/lib/methods/AudioManager.ts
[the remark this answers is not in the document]
diegolmello: **[P2 — Medium, cumulative impact] Completed-audio data survives room cleanup**

Adding reproduction details to this existing thread, rather than opening a duplicate finding. The native player is released; the defect concerns retained JavaScript state.

**Explain like I'm 5:** When we finish listening, we put the music player away but keep a note about it. When we leave the room, cleanup only looks for music players, so it misses all those leftover notes. Each newly finished message can leave another note behind.

**Details:** At reviewed commit `ee140b055f`, `onEnd` deletes `audioQueue[audioKey]` but keeps `audioUris`, `audioPositions`, `audioRates`, `audioCallbacks` and `audioMeta` for recreation. `unloadRoomAudios` discovers keys using only `Object.keys(this.audioQueue)`. Completed entries are therefore absent from its input and never reach the existing deletion block. The singleton retains their data and the registered component callback closure after room teardown. No eviction or size bound for those completed entries is present in this path.

**Severity:** P2 / medium as a cumulative JavaScript retention defect. There is no demonstrated native-resource leak, crash or user-visible slowdown. The amount retained per entry and real-world memory impact have not been measured, so this is less urgent than the two playback regressions. This evidence does not establish an immediate outage.

**Likelihood:** The retention is deterministic when a loaded message finishes and its player is not recreated before room unload. Growth depends on the number of distinct completed messages in the session. A noticeable memory symptom is unmeasured; replaying the same key does not create a new key each time.

**Steps to reproduce and inspect:**

1. Start with a fresh manager in a unit-test harness. Load an audio entry for a room and register its playback callback.
2. Deliver a loaded status with `didJustFinish: true` and await completion handling. Stub the next-message lookup to return no next message.
3. Call `unloadRoomAudios` for that room.
4. Inspect the manager's stored keys in the test/debugger.

Expected: no player, callback, URI or metadata remains for that room. Actual in the executed stubbed reproduction: `audioQueue` contained zero entries, but `audioUris`, `audioCallbacks` and `audioMeta` each retained one. The corresponding app path is listening to a message through completion, then leaving the room before replaying it; the retained data is observable through instrumentation, not necessarily visible on screen.

**Proposed change:** Discover room entries through retained metadata. The existing optional `audio?.release()` handles entries whose native player is already gone; the existing deletion block then removes their stored data.

```diff
-		const roomAudioKeysLoaded = Object.keys(this.audioQueue).filter(audioKey => this.audioMeta[audioKey]?.rid === rid);
+		const roomAudioKeysLoaded = Object.keys(this.audioMeta).filter(audioKey => this.audioMeta[audioKey].rid === rid);
 		const roomAudiosLoaded = roomAudioKeysLoaded.map(key => this.audioQueue[key]);
```

Keep the existing release and per-key deletion code below this unchanged. No broader map consolidation is required.

**Validation after the fix:** Add a completion → room-unload regression case. Verify that retained state is removed for the target room, still-live players in that room are released, and another room's entries remain intact. If this is deferred, the deferral should acknowledge cumulative JavaScript retention rather than assume the cache has a bound.

app/lib/methods/AudioManager.ts
diegolmello: Fixed in d72d2a5d10. `setRate` now enables `shouldCorrectPitch` before applying the rate with explicit `medium` quality. Coverage checks initial playback, active rate changes, recreated players, and the existing idle-player guard.

Formatting and targeted lint pass. Regression tests were added, but Jest could not run because the local dependency tree fails with `TypeError: LRU is not a constructor`. TypeScript reports errors outside the changed files. Device validation remains pending.

app/lib/methods/AudioManager.ts
diegolmello: Fixed in d72d2a5d10. Sequential playback now checks the retained URI and keeps the rendered-message guard, allowing `playAudio` to recreate completed players. Added coverage for completing A → B, replaying A → B, and excluding unrendered or unloaded messages.

Formatting and targeted lint pass. Regression tests were added, but Jest could not run because the local dependency tree fails with `TypeError: LRU is not a constructor`. TypeScript reports errors outside the changed files. Device validation remains pending.

app/lib/methods/AudioManager.ts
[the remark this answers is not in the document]
diegolmello: Fixed the completed-audio cleanup finding in d72d2a5d10. Room cleanup now enumerates retained metadata so completed entries reach the existing deletion block. Added coverage for retained-state removal, live-player release, and preserving another room’s entries.

Formatting and targeted lint pass. Regression tests were added, but Jest could not run because the local dependency tree fails with `TypeError: LRU is not a constructor`. TypeScript reports errors outside the changed files. Device validation remains pending.

OtavioStasiak (approved):
LGTM! NICE JOB!

PR #6775, opened 2025-11-07
fix: unique key prop warnings for all components

### LINKED ISSUE #4943: Each child in a list should have a unique "key" prop
### Description:
Each child in a list should have a unique "key" prop in inline at the paragraph.tsx
It is giving error after running the app first time.

Environment Information:
Rocket.Chat Server Version: 4.8.7
Rocket.Chat App Version: 4.35.1 Single Server
Device Name: Android
OS Version: Android 10

### Steps to reproduce:

It is giving error after running the app first time. in logs

Screenshot
image

divyanshu-patil:
## Proposed changes
This PR resolves the React warning `Each child in a list should have a unique key prop.`

- Added unique key props to list-rendered components that were missing them.
- added the `_id` to each parsed token in `app/containers/markdown/index.tsx` 
- the key is generated by taking random values + current date/ time
- the numeric values are converted into base-36 to make generated keys container 0-9 and a-z letters

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/4943

### files changed
- `app/containers/markdown/components/code/Code.tsx`
- `app/containers/markdown/components/emoji/BigEmoji.tsx‎` 
- `app/containers/markdown/components/inline/Bold.tsx` 
- `app/containers/markdown/components/inline/Italic.tsx`
- `app/containers/markdown/components/inline/Strike.tsx`
- `app/containers/markdown/components/list/UnorderedList.tsx` 
- `app/containers/markdown/components/list/TaskList.tsx`
- `app/containers/markdown/components/Inline.tsx` 
- `app/containers/markdown/index.tsx` 
- `app/containers/UIKit/MessageBlock.tsx`

- `app/containers/markdown/index.tsx` : generate unique id and assigns them to each parsed token so now every block contains `._id` field

## Screenshots
image
image

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents empty message blocks from rendering and reduces UI flicker.
  * More stable markdown rendering: inline formatting, lists, code blocks, quotes, emojis, mentions and big-emoji now preserve item identity to avoid visual glitches and React key warnings.
  * Lists, tasks and quoted content update more reliably.

* **Chores**
  * No changes to public APIs or exports.


diegolmello (commented):
Using index as key is never a good idea https://www.epicreact.dev/why-react-needs-a-key-prop

divyanshu-patil:
Updated the PR 
All index-based keys have been replaced with well formed keys
also added keys for components that i missed before

app/containers/markdown/components/code/Code.tsx
```diff
@@ -22,10 +23,11 @@ const Code = ({ value }: ICodeProps): React.ReactElement => {
 					borderColor: colors.strokeLight
 				}
 			]}>
-			{value.map(block => {
+			{value.map((block, index) => {
+				const key = `${block.type}-${getBlockValueString(block.value)}-${index}`;
```
diegolmello: You're still relying on map's indexes.
That doesn't fix the bug that may happen currently and it may introduce new ones.
It would also hide the warning, which can be a problem to debug if needed.

diegolmello (changes requested):

app/containers/markdown/components/code/Code.tsx
divyanshu-patil: if we can't really use the `index` then should i modify block with `random Id` generated by bcrypt, crypto or uuid ? 
cause if the block value is empty or the `getBlockvalueString()` returned the 'empty' or '' then the two keys can collide with each other leading to unknown bugs

divyanshu-patil:
updated the PR,
now none of the keys relies on the index

every direct or nested block now contains `_id` field which we can use as key
keys are generated using helper function `generateID()`

app/containers/markdown/components/code/Code.tsx
```diff
@@ -22,10 +24,11 @@ const Code = ({ value }: ICodeProps): React.ReactElement => {
 					borderColor: colors.strokeLight
 				}
 			]}>
-			{value.map(block => {
+			{value.map(b => {
+				const block = b as TCodeLineWithID;
 				switch (block.type) {
 					case 'CODE_LINE':
-						return <CodeLine value={block.value} />;
+						return <CodeLine key={block._id} value={block.value} />;
```
diegolmello, line 31: where is _id coming from?

diegolmello (changes requested):

app/containers/markdown/components/code/Code.tsx
divyanshu-patil: in `app/containers/markdown/index.tsx`
after the tokens are parsed we add `_id` field in it

[assigning ids here](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/6775/files#diff-cf2f80983cb151b2d53d9fbcb3ebdc2e195934907a1858b657c5716bc0877b9eR81)

[generating ids](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/6775/files#diff-cf2f80983cb151b2d53d9fbcb3ebdc2e195934907a1858b657c5716bc0877b9eR37)

i also tested in my local `_id` field always contains unique 16 char value

divyanshu-patil: image


ids are assigned to nested blocks also, so even if the block internally maps the tokens again like `BigEmoji` works perfectly with this keys

divyanshu-patil:
hello maintainers,
any updates on the changes?

Rohit3523:
I have some questions
1. will it assign a new id to each view every single time a component rerender?
2. why are you not using value as key?

divyanshu-patil:
> I have some questions
> 
> 1. will it assign a new id to each view every single time a component rerender?
> 2. why are you not using value as key?

ans to que 1
- it will only assign new id when the markdown parses the message
- ie. when ever `Markdown` component is going to render 
- until the tokens doesn't changes the new ID isn't assigned
- **cases** :
    id is only going to change when ever
1. user enters a new room or leaves, as it unmounts the `Markdown` component
ex. leaving and entering same room 
image
image
image

2. upon re-rendering of whole markdown component
3. when new tokens are loaded within same chat

ex: scrolling in same chat up and down changes id cause value of token changes eventually creating new id
image
image
image


ans to que 2
because if any 2 token contains similar value then it will generate duplicate keys
as you can see above

> `${block.type}-${getBlockValueString(block.value)}-${index}`
  or 
> BOLD-Hello will be the key if we used the value as key

divyanshu-patil:
hello sir, 
any updates on this?

app/containers/markdown/components/Inline.tsx
```diff
@@ -18,11 +18,14 @@ interface IParagraphProps {
...
 
+type TInlineWithID = InlinesType & { _id: string };
+
 const Inline = ({ value, forceTrim }: IParagraphProps): React.ReactElement | null => {
 	const { useRealName, username, navToRoomInfo, mentions, channels } = useContext(MarkdownContext);
 	return (
 		<Text style={styles.inline}>
-			{value.map((block, index) => {
+			{value.map((b, index) => {
+				const block = b as TInlineWithID;
```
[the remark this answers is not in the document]
divyanshu-patil: _id values are assigned only when tokens are created, not on every re-render. They change only when tokens are re-parsed (e.g., due to virtualization), which already causes unmount/remount and resets reconciliation state. In that context, regenerating keys is expected and does not affect performance. Content-based keys are unsafe due to potential duplication.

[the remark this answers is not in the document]
divyanshu-patil: in `app/containers/markdown/index.tsx`
```ts
tokens = tokens.map(assignIds);
```
```ts
const assignIds = (block: any) => {
	if (!block._id) {
		block._id = generateId();
	}
	...
	return block;
};
```

this code guarantees that every block have _id even if its deeply nested
so this comment is invalid

divyanshu-patil:
hello @diegolmello can you take a look at this pr and it's approach or should i Update it?
or this pr is not plan of the plan currently

PR #6797, opened 2025-11-14
fix: profile avatar cache busting and refresh mechanism

### LINKED ISSUE #6794: bug: After updating the profile avatar and going back, the old cached image still shows instead of the new one.
### Describe the Bug

After I update a profile avatar and go back, it still shows the old image from cache. The new avatar only appears if I hit "Save Changes" again or refresh the screen manually.

### Steps to Reproduce

Steps to Reproduce
Open the Profile view
Tap the edit to open image change
Select a new image (camera, gallery, or URL)
Save the changes
Navigate back to the Profile view
you will se the old avatar is still displayed

### Expected Behavior

After saving a new profile avatar and going back, the Profile screen should instantly show the updated avatar, not the old cached one.

### Actual Behavior

When I go back after changing the avatar, the Profile screen still shows the old cached image. The new avatar only appears if I tap “Save Changes” again, refresh the app, or restart it. This makes it look like the avatar didn’t update properly.

Any version (client-side caching issue)

### Rocket.Chat App Version

4.67.0

### Device Name

All devices (iOS and Android)

### OS Version

All supported versions

### Additional Context

_No response_

deepak0x:
I fixed an issue where updating a profile avatar wasn’t showing the new image immediately after going back to the ProfileView. The old cached image would keep showing until the user saved again or manually refreshed.

Issue(s)
Closes #6794

### Problem

When a user updates their avatar and goes back to the ProfileView, the old avatar still appears. This happens because:

* **expo-image** caches images very aggressively
* The `avatarETag` wasn't updating correctly across databases
* The `useAvatarETag` hook didn’t re-run when the value changed
* There was no proper cache-busting or refresh mechanism

### What I Did

I added a few coordinated fixes to make sure the avatar updates instantly:

1. **useAvatarETag**

   * Made it always listen for changes (removed the old condition)
   * Set the initial value right away
   * Fixed the effect dependencies so it re-runs correctly

2. **Avatar component**

   * Added a cache-busting `_cb=${avatarETag}` query param to the image URL
   * Added a `key` based on `avatarETag` so the image remounts when updated
   * Used `useMemo` to generate the final avatar URL

3. **ProfileView**

   * Added a refresh key that updates when returning to the view
   * Used it to force the avatar container to remount
   * Cleared the in-memory image cache on focus

4. **connect.ts**

   * When updating the avatar, I now also sync the logged-in user’s `avatarETag` in Redux
   * This triggers the saga to update the database, so the hook immediately detects changes

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable avatar updates: avatar images now remount when their source changes to ensure fresh display.
  * Improved avatar synchronization: avatar change events are propagated to remote servers to keep records consistent.
* **Performance**
  * Enhanced image caching using a memory-disk policy for faster load and reduced network use.
* **Stability**
  * More robust avatar observation and cleanup to avoid stale updates after unmount.

### DIFF
--- a/app/containers/Avatar/Avatar.tsx
+++ b/app/containers/Avatar/Avatar.tsx
@@ -84,12 +84,14 @@ const Avatar = React.memo(
 
 			image = (
 				<Image
+					key={`${uri}-${avatarETag || ''}`}
 					style={avatarStyle}
 					source={{
 						uri,
 						headers: RocketChatSettings.customHeaders
 					}}
 					priority='high'
+					cachePolicy='memory-disk'
 				/>
 			);
 		}
--- a/app/containers/Avatar/useAvatarETag.ts
+++ b/app/containers/Avatar/useAvatarETag.ts
@@ -20,10 +20,9 @@ export const useAvatarETag = ({
 }) => {
 	const [avatarETag, setAvatarETag] = useState<string | undefined>('');
 
-	const isDirect = () => type === 'd';
-
 	useEffect(() => {
-		let subscription: Subscription;
+		let subscription: Subscription | undefined;
+		let isSubscribed = true;
 		const observeAvatarETag = async () => {
 			const db = database.active;
 			const usersCollection = db.get('users');
@@ -36,7 +35,7 @@ export const useAvatarETag = ({
 					const userCollections = serversDB.get('users');
 					const user = await userCollections.find(id);
 					record = user;
-				} else if (isDirect()) {
+				} else if (type === 'd') {
 					const [user] = await usersCollection.query(Q.where('username', text)).fetch();
 					record = user;
 				} else if (rid) {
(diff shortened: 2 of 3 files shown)


Rohit3523:
This change is done by AI and i don't think you have done the testing in iOS and Android.

1. You have added `react-native-skeleton-placeholder` but package.json, package-lock.json, ios/Podfile.lock files are missing from the PR.
Screenshot 2025-11-18 at 3 16 09 AM

2. Please explain why we have to clear all images from the memory cache after one user avatar changes
```
Image.clearMemoryCache().catch(() => {
    // Ignore errors if cache clearing fails
});
```
Screenshot 2025-11-18 at 3 21 18 AM

3. Can you provide a recording of this feature as working

deepak0x:
It was showing the old avatar due to cache... so i clear cache after avatar change so that it request fresh fetch of avatar from server... 

https://github.com/user-attachments/assets/d6be4946-abd8-464f-a8e3-f4e37696a39f


here's the recording 

Thanks

Rohit3523:
Thanks for your contribution! We really appreciate it. We have already merged a PR that resolves same issue: https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7430

PR #6815, opened 2025-11-27
fix: prevent message text from being cut off in channel list

### LINKED ISSUE #4740: Channel's card crops the messages inside it
### Description:
When we join a channel and see we see that messages inside the channels are cropped by the card of the channel this should not happen. Extra length of message should be replace by the three dots (...) 

### Environment Information:

- Rocket.Chat App Version: 4.33.1
- Device Name: POCO X2 
- OS Version: Android 11

### Steps to reproduce:

1. Go to any channel and look at the cards and messages 

### Expected behavior:

This should not crop the message either it should display full message or full screen width message, if message is longer than screen width then further message will be '...'
For example:
Rocket is great organization, community is really helpful... (if there is more)

### Actual behavior:

Screenshot_2022-12-12-09-45-20-324_chat rocket android

deepak0x:
### Fixes #4740

In the channel list, long message previews were cut off mid-word or mid-sentence without an ellipsis, especially when the device font size was increased.

## Before vs After Comparison

| Scale | Before | After |
|-------|--------|--------|
| **0.85** | before0.85 |  |
| **0.95** | before0.95 |  |
| **1.00** | before1 |  |
| **1.15** | before1.15 |  |

---

### What We Fixed
1. Added ellipsis truncation (`...`) instead of cutting text abruptly.
2. Increased channel row height to support larger text at higher font scales.
3. Improved layout constraints to ensure proper text truncation within the container.

### Result
- Messages now show ellipsis when too long.  
- No more mid-word or mid-sentence clipping.  
- Behaves correctly across all font sizes.  
- Text remains readable and visually consistent.

This ensures message previews are always readable and properly truncated at any device font size.

## Summary by CodeRabbit

* **Style**
  * Message previews now occupy full width, wrap more reliably, and shrink to avoid overflow.
  * Tightened spacing and layout constraints for compact views; updated base row sizing and font-scaling rules for better proportions.
  * Subtle divider styling adjusted for more consistent row separators.

* **Bug Fixes**
  * Fixed cross-platform alignment to ensure consistent rendering on Android and other platforms.

* **Notes**
  * Markdown preview behavior unchanged.


Rohit3523:
Screenshot for before and after seems same

Also can you please add images like this

| Scale | Before | After |

It will be easy to compare the changes

deepak0x:
The major issue was with smallest font...
have added the  ellipsis truncation and improved layout so that there won't be any cut of text in any font size...

Rohit3523:
I have noticed some points 

1. Your PR is adding extra space in bottom for <= 1x scale font size which hide 1 channel from the list. can you revert this behaviour because showing extra space between two channels is not a good idea

Screenshot_20251127_192811
2. Just check this screenshot, here user is trying to say that text is cutting into half because of fixed height.

deepak0x:
have tested and fixed all issues mentioned by u...

PR #6861, opened 2025-12-22
fix: chat fails to load older messages when scrolling to the top

Rohit3523:
## Proposed changes
This PR fixes an issue where scrolling to the top of a DM or Room chat would get stuck and fail to load older messages.

To fix this, I observed how the web handles message loading and found that it only loads room level messages while excluding thread messages. Thread messages are fetched only when navigating into a thread. I applied the same behavior in the app to ensure consistent pagination and prevent message loading from getting stuck.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1701

## How to test or reproduce
1. Go to an old room or user dm with whom you chat the most
2. Scroll to top
3. Observe

# Expected Behaviour
It should load previous messages

# Actual Behaviour
It just stuck on top

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Chore**
  * Message loading now uses smaller per-batch loads, causing more frequent incremental fetches when entering rooms.
  * Threaded messages are no longer fetched by default, reducing initial clutter in message lists.
  * The "Load more" prompt now appears only when a full batch of messages is returned, making progressive loading more predictable.

### DIFF
--- a/app/actions/room.ts
+++ b/app/actions/room.ts
@@ -39,9 +39,14 @@ interface IForwardRoom extends Action {
 	rid: string;
 }
 
+type IUserTypingArgs = {
+	tmid?: string;
+};
+
 interface IUserTyping extends Action {
 	rid: string;
 	status: boolean;
+	args?: IUserTypingArgs;
 }
 
 export interface IRoomHistoryRequest extends Action {
@@ -109,11 +114,12 @@ export function removedRoom(): Action {
 	};
 }
 
-export function userTyping(rid: string, status = true): IUserTyping {
+export function userTyping(rid: string, status = true, args?: IUserTypingArgs): IUserTyping {
 	return {
 		type: ROOM.USER_TYPING,
 		rid,
-		status
+		status,
+		args
 	};
 }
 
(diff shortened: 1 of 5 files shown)


app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -9,7 +9,6 @@ import updateMessages from './updateMessages';
 import { generateLoadMoreId } from './helpers/generateLoadMoreId';
 
 const COUNT = 50;
-const COUNT_LIMIT = COUNT * 10;
```
OtavioStasiak, line 12: We've done this for a reason…
A few months ago, we had an issue with large threads being used as the last message.
If you test a thread with more than 100 replies as the last message in develop and in this branch, you'll see that this change rolls back the fix.

| develop | your branch |
| ----- | ----- |
| Simulator Screenshot - iPhone 16 - 2025-12-22 at 19 36 54 | Simulator Screenshot - iPhone 16 - 2025-12-22 at 19 35 38 |

We must find a smart way to make it work without overloading in long-thread cases, if you want we can try to find a solution together on DM.

PR #6878, opened 2026-01-04
fix: reset total state when switching directory filters

### LINKED ISSUE #6879: Directory view shows empty results when switching from empty Teams filter to Users or Channels
## Description
When navigating in the Directory view, switching from the Teams filter (which has no data) to Users or Channels filters incorrectly displays empty results, even though Users and Channels have data available.

## Steps to Reproduce
1. Open the Directory view
2. Navigate to the Teams filter (assuming Teams is empty/has no data)
3. Switch to Users filter
4. Observe that Users shows empty results (even though users exist)
5. Switch to Channels filter
6. Observe that Channels also shows empty results (even though channels exist)

## Expected Behavior
When switching between directory filters (Users, Teams, Channels), each filter should display its respective data correctly, regardless of the previous filter's state or data availability.

## Actual Behavior
When switching from an empty Teams filter to Users or Channels, those filters incorrectly display as empty even when they contain data.

deepak0x:
Fixed issue where switching from empty teams filter to users or channels would show empty results. The problem was that changeType and toggleWorkspace were only resetting data array but not the total count, causing the load logic to think all data was already loaded.

Now both data and total are reset when switching filters, ensuring proper state cleanup before new searches.

Closes #6879

Before (Issue)

<video src="https://github.com/user-attachments/assets/3eda08e1-12fb-4f85-8e6a-eb4d731f0d57" controls width="100%"></video>

After (Fix)

<video src="https://github.com/user-attachments/assets/554687d9-6a0f-4e28-81bf-2ec05773e2a0" controls width="100%"></video>

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state management when changing directory filters and workspace view options to ensure data resets properly and searches execute correctly.


OtavioStasiak (changes requested):
The main problem is that when you click to change the filter, sometimes `newSearch` is triggered as false. This is probably related to the debounce, so you should investigate it to really fix the root cause.

deepak0x:
ok checking again...

deepak0x:
@OtavioStasiak can u pls review again have addressed your changes...

PR #6908, opened 2026-01-09
fix(iOS): app crashing on render image

OtavioStasiak:
## Proposed changes

The app crashes when using avatars encoded in Base64 because the getAvatarUrl function does not properly handle Base64 images.
This also explains why the issue only occurs on iOS starting from version 18.6.
In iOS 18.6, Apple introduced stricter validation for URL(fileURLWithPath:), which causes fatal crashes in production TestFlight builds when HTTP or HTTPS URLs are incorrectly passed to this API.

The root cause is that the isFileUrlPath() function in expo-modules-core was incorrectly identifying HTTP/HTTPS URLs as file paths. This happened because:
The function checks whether a URL has no scheme by attempting to parse it with URL(string:)
If URL parsing fails or returns nil (for example, due to malformed URLs), the function assumes the input is a file path
As a result, the code calls URL(fileURLWithPath:) with HTTP/HTTPS URLs.
iOS 18.6 detects this API misuse and terminates the app with a fatal error.

Error details:
```
API MISUSE: URL(filePath:) called with an HTTP URL string. Use URL(string:) instead.
Fatal error: API MISUSE: URL(filePath:) called with an HTTP URL string. URL(string:) returned nil.
```
References:
- **Apple Foundation URL Documentation**: [init(fileURLWithPath:relativeTo:)](https://developer.apple.com/documentation/foundation/url/init(fileurlwithpath:relativeto:))
- **Swift Forums - URL String Behavior Changes**: [URL(string:) behavior changed with Xcode 15.0 beta 5](https://forums.swift.org/t/url-string-behavior-changed-with-xcode-15-0-beta-5/66570)

I also opened a PR to expo to avoid this kind of error.
https://github.com/expo/expo/pull/42168

## Issue(s)	

https://rocketchat.atlassian.net/browse/SUP-962

## How to test or reproduce

- Open the app;
- Open a room with a user with base64 avatar;

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of inline (base64/data URI) avatar images so they display reliably.

* **New Features**
  * Added a demo/story showing an avatar rendered from a base64 image.

* **Tests**
  * Expanded unit tests for avatar URL handling and formatting across many scenarios (data URIs, external URLs, server/version branches, auth/etag params, CDN/trailing-slash cases).

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -5024,8 +5024,8 @@
     tldts "~5.7.112"
 
 "@rocket.chat/mobile-crypto@RocketChat/rocket.chat-mobile-crypto":
-  version "0.2.0"
-  resolved "https://codeload.github.com/RocketChat/rocket.chat-mobile-crypto/tar.gz/06bab0cb2e329822911c80d1b937219c15154faa"
+  version "0.2.1"
+  resolved "https://codeload.github.com/RocketChat/rocket.chat-mobile-crypto/tar.gz/b75e261282bc0c25a3b8fde3230d1c1e01809e00"
 
 "@rocket.chat/sdk@RocketChat/Rocket.Chat.js.SDK#mobile":
   version "1.3.3-mobile"


app/containers/UIKit/Image.tsx
```diff
@@ -23,9 +24,15 @@ const ThumbContext = (args: IThumb) => (
...
-);
+export const Thumb = ({ element, size = 88 }: IThumb) => {
+	const normalizedUrl = normalizeImageUrl(element?.imageUrl);
+
+	if (!normalizedUrl) {
+		return null;
+	}
+
+	return <ExpoImage style={[{ width: size, height: size }, styles.image]} source={{ uri: normalizedUrl }} />;
+};
```
diegolmello: This would be a nice implementation of our wrapper of expo-image to normalize urls

PR #6956, opened 2026-01-29
fix: collapsible attachment with text duplicating the content

OtavioStasiak:
## Proposed changes

Fix collapsible attachment with text duplicating the content.

## Issue(s)	

https://rocketchat.atlassian.net/browse/SUP-982

## How to test or reproduce

- Post the message;
```
curl -X POST 'https://YOUR_SERVER/api/v1/chat.postMessage' \
  -H 'Content-Type: application/json' \
  -H 'X-Auth-Token: YOUR_AUTH_TOKEN' \
  -H 'X-User-Id: YOUR_USER_ID' \
  -d '{
    "roomId": "ROOM_ID",
    "text": "This is the main message body.",
    "attachments": [
      {
        "collapsed": true,
        "title": "Collapsed attachment block",
        "text": "This attachment text should NOT appear as plain text above the message or duplicate before the block.",
        "description": "Attachment description that might also leak as duplicate plain text.",
        "color": "#2c3e50",
        "fields": [
          { "title": "Field 1", "value": "Value 1", "short": true },
          { "title": "Field 2", "value": "Value 2", "short": true },
          { "title": "Long field", "value": "This field value could also contribute to duplicate text when expanded.", "short": false }
        ]
      }
    ]
  }'
```
- open the app;
- navigate to the room with the message;

## Screenshots
| Before | After |
| ----- | ----- |
| Simulator Screenshot - iPhone 16 - 2026-01-29 at 20 09 23 | Simulator Screenshot - iPhone 16 - 2026-01-29 at 20 09 09 |

## Types of changes

- [X] Documentation update (if none of the other choices apply)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Collapsed attachments are no longer treated as quote attachments, preventing incorrect quoted content.

* **New Features**
  * Nested attachments now surface as related/reply content rather than being ignored.
  * Added an avatar-suggestion endpoint and client support to fetch service-based avatar suggestions.

* **Documentation**
  * Added stories demonstrating collapsible attachments (standard and large-font) in collapsed and expanded states.


app/containers/message/Components/Attachments/Quote.tsx
```diff
@@ -11,6 +11,8 @@ import { getMessageFromAttachment } from '../../utils';
 const isQuoteAttachment = (file?: IAttachment): boolean => {
 	if (!file) return false;
 
+	if (file.collapsed) return false;
+
 	if (!file.color && !file.text && (file.image_url || file.audio_url || file.video_url || file.collapsed)) {
```
Rohit3523: we can remove `file.collapsed` from here because we have added a check in line 14

PR #6957, opened 2026-01-30
feat: WatchOS Quick Replies

divyanshu-patil:
**Linked PR: https://github.com/RocketChat/Rocket.Chat/pull/38911**

## Proposed changes

- **Apple Watch Quick Replies support**
  - Added a new setting `Apple_Watch_Quick_Actions` with related constants and localization keys.
  - Introduced storage keys (`FIRST_START`, `WATCHOS_QUICKREPLIES`) and default settings for managing quick replies.

- **Quick Replies Management UI**
  - Added a new **UserWatchOSQuickRepliesView** where users can add, remove, and manage Apple Watch quick replies.
  - Integrated the entry into **User Preferences** with conditional visibility based on watch availability.

- **Watch Sync & Status Handling**
  - Implemented helper methods to:
    - Check Apple Watch availability.
    - Sync quick replies with the watch via the native bridge.
    - Retrieve persisted state from MMKV.
  - Added automatic synchronization after login.

- **User Preferences Integration**
  - Extended `UserPreferences` with `getArray` and `setArray`.
  - Included quick replies in settings retrieval and cleared them on logout.

- **Native iOS Bridge**
  - Implemented `RCTWatchModule` TurboModule exposing APIs to:
    - Check watch connectivity.
    - Sync quick replies with the Apple Watch.
  - Integrated MMKV storage and `WCSession` communication.

- **Watch App Updates**
  - Added database and storage support for quick replies.
  - Updated Watch session and loaders to receive and apply synced quick replies.
  - Display quick reply pills in the **message composer** on the watch.

- **UI Improvements**
  - Added scroll-to-bottom anchor in `MessageListView`.
  - Updated watch message composer to support quick reply buttons.

- **Localization**
  - Added English localization strings for Apple Watch quick replies.
 
### iOS
https://github.com/user-attachments/assets/31aca833-5c88-476f-b5f3-29bc0db5206e

### Android
image

### Admin settings
image

## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Apple Watch Quick Replies: add, edit, and persist quick replies per server; dedicated preferences section and management screen.
  * Composer & Watch UI: quick-reply pills in the message composer and watch app syncing for immediate use.
  * Automatic sync: quick replies synchronize with a paired Apple Watch when available.
  * Localization: added English strings for Watch quick replies.


ios/WatchBridge.m
```diff
@@ -0,0 +1,16 @@
+#import <React/RCTBridgeModule.h>
+
+@interface RCT_EXTERN_MODULE(WatchBridge, NSObject)
+
+RCT_EXTERN_METHOD(
```
diegolmello: We should not write old arch anymore.
Please migrate it to turbo modules.

ios/WatchBridge.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import React
+import WatchConnectivity
+
+@objc(WatchBridge)
+final class WatchBridge: NSObject {
```
diegolmello: I think you can make it more simple than this.
Instead of relying on this native module, we rename [this one](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/e1af1902f73f6cc5b272fe341edf4beabc658a0e/ios/SSLPinning/SSLPinning.swift) to `WatchModule`.

Also, everything is going to be done relying on MMKV.
Similar to `WatchModule.getCertificate`, you'll create `WatchModule.getQuickReplies` to read from MMKV.

On `WatchConnection`, you'll create a new method `getQuickReplies` as well.

`WatchConnection.getQuickReplies` will be read on the Watch app for the composer.
When a quick reply is tapped, it should send the message automatically.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { WATCHOS_QUICKREPLIES } from '../../lib/constants/keys';
+import { syncWatchOSQuickReplies } from '../../lib/methods/WatchOSQuickReplies/syncReplies';
+import { checkWatch } from '../../lib/methods/WatchOSQuickReplies/getWatchStatus';
+
+interface IUserPreferencesViewProps {
+	navigation: NativeStackNavigationProp<ProfileStackParamList, 'UserPreferencesView'>;
+}
+
+const UserPreferencesView = ({ navigation }: IUserPreferencesViewProps): JSX.Element => {
+	const [quickreplies, setQuickreplies] = useUserPreferences<string[]>(WATCHOS_QUICKREPLIES, []);
```
diegolmello:
```suggestion
	const [quickReplies, setQuickReplies] = useUserPreferences<string[]>(WATCHOS_QUICKREPLIES, []);
```

diegolmello: Create a setting on [Admin/Mobile](https://github.com/RocketChat/Rocket.Chat/blob/2003802793b0ab6b043bce6457cfc0c3576afe95/apps/meteor/server/settings/mobile.ts#L9) like `ScreenLock` section.

Admin should look like this:
- Apple Watch
  - Quick actions: string

The string should be comma separated and the default ones should be:
- OK
- Yes
- No
- On my way
- Will follow up shortly

diegolmello (changes requested):

app/lib/native/NativeWatchModule.ts
```diff
@@ -0,0 +1,16 @@
...
+	syncQuickReplies(): void;
+	isWatchSupported(): boolean;
+	isWatchPaired(): boolean;
+	isWatchAppInstalled(): boolean;
+
+	// debug
+	getCurrentServerFromNative(): string;
+	getkey(): string;
+	getReplies(): string;
+}
```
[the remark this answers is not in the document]
divyanshu-patil: all three debug methods are removed and no longer present in spec

ios/RocketChat Watch App/Loaders/WatchSession.swift
```diff
@@ -1,98 +1,152 @@
...
+                    WatchMessage.self,
+                    from: data
+                )
+
+                completionHandler(.success(message))
+            } catch {
+                completionHandler(.failure(.undecodable(error)))
+            }
+        }
+    }
```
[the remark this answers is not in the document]
divyanshu-patil: not related to PR changes

ios/RocketChat Watch App/Loaders/WatchSession.swift
```diff
@@ -1,98 +1,152 @@
...
+            guard let self else {
+                return
+            }
+
+            switch result {
+            case .success(let message):
+                completionHandler(.success(message))
+            case .failure where self.retries > 0:
+                self.session.sendMessage(completionHandler: completionHandler)
+            case .failure(let error):
```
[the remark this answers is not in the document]
divyanshu-patil: not related to pr changes

ios/RocketChat Watch App/Loaders/WatchSession.swift
```diff
@@ -1,98 +1,152 @@
...
+    init(delay: TimeInterval = 1, session: WatchSessionProtocol) {
+        self.delay = delay
+        self.session = session
+    }
+
+    func sendMessage(
+        completionHandler:
+            @escaping (Result<WatchMessage, ServersLoadingError>) -> Void
+    ) {
+        Timer.scheduledTimer(withTimeInterval: 1, repeats: false) {
```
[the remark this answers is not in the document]
divyanshu-patil: not related to pr changes

diegolmello:
Resolve CodeRabbit comments

ios/RocketChat Watch App/Views/MessageComposerView.swift
```diff
@@ -1,34 +1,80 @@
...
+                                    .padding(.vertical, 8)
+                                    .background(
+                                        RoundedRectangle(cornerRadius: 10)
+                                            .fill(Color.white.opacity(0.12))
+                                    )
+                                    .contentShape(Rectangle())
+                                    .onTapGesture {
+                                        message = reply
+                                        send()
+                                    }
```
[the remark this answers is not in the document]
divyanshu-patil: using Button here causes issues while scrolling like sending text to channel
keeping text here so that this issues are prevented

app/sagas/selectServer.ts
```diff
@@ -196,6 +199,14 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
 			yield put(appStart({ root: RootEnum.ROOT_OUTSIDE }));
 		}
 
+		const isFirstServerLogin = UserPreferences.getBool(`${server}-${FIRST_START}`);
+		if (isFirstServerLogin == null) {
+			UserPreferences.setBool(`${server}-${FIRST_START}`, false);
+			yield put(setIsFirstServerLogin(true));
+		} else {
+			yield put(setIsFirstServerLogin(isFirstServerLogin));
+		}
```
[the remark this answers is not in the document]
divyanshu-patil: removed the isFirstServerLogin completely so this isnt valid anymore

divyanshu-patil:
@coderabbit review

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+	useEffect(() => {
+		syncWatchOSQuickReplies();
+	}, []);
+
+	const removeQuickReply = (reply: string) => {
+		const newReplies = quickReplies?.filter(quickreply => quickreply !== reply);
+		setQuickReplies(newReplies);
+		syncWatchOSQuickReplies();
+	};
```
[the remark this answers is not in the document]
divyanshu-patil: `syncWatchOSQuickReplies` have void return type

ios/RCTWatchModule.mm
```diff
@@ -0,0 +1,177 @@
...
+NSString *mmkvCurrentServerKey = @"currentServer";
+
+#pragma mark - initialisation
+
+- (id)init {
+    if (self = [super init]) {
+        _session = WCSession.defaultSession;
+    }
+    return self;
+}
```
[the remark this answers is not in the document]
divyanshu-patil: WCSession is activated in `Watch/WatchConnection` so no need to activate again here

app/i18n/locales/en.json
```diff
@@ -29,6 +29,7 @@
   "Add_Channel_to_Team": "Add channel to team",
   "Add_Existing": "Add existing",
   "Add_Existing_Channel": "Add existing channel",
+  "Add_Quick_Reply": "Add Quick Reply",
```
diegolmello: Use Sentence case, like this
```suggestion
  "Add_Quick_Reply": "Add quick reply",
```

app/lib/methods/getSettings.ts
```diff
@@ -185,6 +188,15 @@ export async function getSettings(): Promise<void> {
 
 		reduxStore.dispatch(addSettings(parsedSettings));
 
+		if (shouldShowWatchAppOptions()) {
+			try {
+				// case: when we switched to another server which is already added
+				syncWatchOSQuickRepliesWithServer(reduxStore.getState());
```
diegolmello: You're calling it too often and in random places.
Can you try adding a `fork` to login saga, like the others?
Remove it from the other places.

diegolmello (changes requested):

app/lib/methods/getSettings.ts
divyanshu-patil: resolved

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+// import { useAppSelector } from '../../lib/hooks/useAppSelector';
+
+interface IUserWatchOSQuickRepliesViewProps {
+	navigation: NativeStackNavigationProp<ProfileStackParamList, 'UserWatchOSQuickRepliesView'>;
+}
+
+const UserWatchOSQuickRepliesView = ({ navigation }: IUserWatchOSQuickRepliesViewProps): JSX.Element => {
+	const currentServer = userPreferences.getString(CURRENT_SERVER);
+	const [quickReplies, setQuickReplies] = useUserPreferences<string[]>(`${currentServer}-${WATCHOS_QUICKREPLIES}`, []);
+	const [input, setInput] = useState<string>('');
```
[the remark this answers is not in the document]
divyanshu-patil: this case have very low chances of happening,
but still adding guard instead of storing it under different key

divyanshu-patil:
@CodeRabbit review

app/views/UserWatchOSQuickRepliesView/index.tsx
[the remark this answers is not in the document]
divyanshu-patil: function return type is made as string now and logging if error and returning `false`
created a wrapper around it that returns true or false, but as syncing is automatic without a dedicated optino, ui have no effect if the sync fails

ios/RocketChat Watch App/Views/MessageListView.swift
```diff
@@ -1,125 +1,160 @@
...
+                    VStack {
+                        Text("No messages")
+                            .font(.caption.italic())
+                            .foregroundStyle(Color.secondaryInfo)
+                            .frame(maxHeight: .infinity)
+                    }
+                    Spacer()
+                }
+            } else {
+                ScrollViewReader { proxy in
```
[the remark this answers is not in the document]
divyanshu-patil: remove else block added in previous coderabbit fix because composer is a part of scrollview

divyanshu-patil:
@CodeRabbit review

ios/Libraries/VoipModule.mm
```diff
@@ -2,7 +2,7 @@
 #import <React/RCTEventEmitter.h>
 #import <PushKit/PushKit.h>
 
-#import <SSLPinning/SSLPinning.h>
+#import <RocketChatSpec/RocketChatSpec.h>
```
divyanshu-patil, line 5: as turbomodules are used for multiple purposes this PR renamed Spec Name from SSLPinning to RocketChatSpec to avoid ambiguity

package.json
```diff
@@ -225,11 +225,17 @@
 		"node": ">=18"
 	},
 	"codegenConfig": {
-		"name": "SSLPinning",
+		"name": "RocketChatSpec",
```
divyanshu-patil, line 228: as turbomodules are used for multiple purposes this PR renamed Spec Name from SSLPinning to RocketChatSpec to avoid ambiguity

divyanshu-patil:
hello @diegolmello i have resolved all conflicts in this PR, can you take a look ?

app/lib/methods/WatchOSQuickReplies/getWatchOSRepliesFromMMKV.ts
```diff
@@ -0,0 +1,4 @@
+import { FIRST_START } from '../../constants/keys';
+import UserPreferences from '../userPreferences';
+
+export const getWatchOSRepliesForServer = (server: string) => UserPreferences.getBool(`${server}-${FIRST_START}`);
```
diegolmello: 🔴 **Dead gate + misleading name.** This reads `${server}-${FIRST_START}`, but `FIRST_START` is never written anywhere in the codebase (grep-confirmed), so this always returns falsy. As a result `syncWatchOSRepliesWithServer` re-seeds from server settings on every login instead of only the first. Also, the file/function name reads as "get replies" but it returns a first-start boolean — rename to reflect what it does (per `UBIQUITOUS_LANGUAGE.md`).

app/lib/methods/logout.ts
```diff
@@ -23,6 +30,7 @@ function removeServerKeys({ server, userId }: { server: string; userId?: string
 	UserPreferences.removeItem(`${server}-${E2E_PUBLIC_KEY}`);
 	UserPreferences.removeItem(`${server}-${E2E_PRIVATE_KEY}`);
 	UserPreferences.removeItem(`${server}-${E2E_RANDOM_PASSWORD_KEY}`);
+	UserPreferences.removeItem(`${server}-${WATCHOS_QUICKREPLIES}`);
```
diegolmello, line 33: The `${server}-${FIRST_START}` key is never removed on logout (and never set in the first place — see `getWatchOSRepliesFromMMKV.ts`). Once the first-login gate is actually wired up, this key needs clearing here too, alongside `WATCHOS_QUICKREPLIES`.

app/lib/methods/userPreferences.ts
```diff
@@ -145,6 +145,19 @@ class UserPreferences {
 		this.mmkv.set(key, value);
 	}
 
+	getArray<T = unknown>(key: string): T[] | null {
```
diegolmello: `getArray` is added but never called in production code — only `setArray` is used (in `syncWatchOSRepliesWithServer.ts`); the view reads replies via `useUserPreferences`. Either route the read-back through `getArray` for consistency, or drop the dead method.

app/sagas/login.js
```diff
@@ -39,6 +39,8 @@ import { getServerById } from '../lib/database/services/Server';
 import appNavigation from '../lib/navigation/appNavigation';
 import { showActionSheetRef } from '../containers/ActionSheet';
 import { SupportedVersionsWarning } from '../containers/SupportedVersions';
+import { isIOS } from '../lib/methods/helpers';
```
diegolmello: `isIOS` is imported but never used. This suggests the intended iOS-gating of `fetchWatchReplies` was dropped — it currently runs on Android too (harmless only because `shouldShowWatchAppOptions()` is false there). Either gate the sync with `isIOS` or remove the import.

app/sagas/login.js
```diff
@@ -228,6 +230,24 @@ const fetchUsersRoles = function* fetchRoomsFork() {
 	}
 };
 
+const fetchWatchReplies = function* fetchWatchRepliesFork() {
+	try {
+		// we are getting replies from server settings
+		const state = yield select(state => state);
+
+		if (!state.settings?.Apple_Watch_Quick_Actions) {
+			yield delay(1000);
```
diegolmello, line 238: Racy: when `Apple_Watch_Quick_Actions` hasn't arrived yet, this blind-waits 1s then reads state once and proceeds regardless of whether the setting actually showed up. Prefer waiting on the setting event/selector rather than a fixed delay.

app/lib/methods/WatchOSQuickReplies/syncReplies.ts
```diff
@@ -0,0 +1,18 @@
+import NativeWatchModule from '../../native/NativeWatchModule';
+import { shouldShowWatchAppOptions } from './getWatchStatus';
+
+export function syncWatchOSQuickReplies(): boolean {
+	if (!shouldShowWatchAppOptions()) return false;
+	try {
+		const result = NativeWatchModule?.syncQuickReplies();
+		if (result?.startsWith('[ERROR]')) {
+			console.error(result);
```
diegolmello: Use the methods-layer `log` helper (`./helpers/log`) instead of `console.error` here and at line 15 — consistent with `loadMessagesForRoom.ts`, `logout.ts`, and the login saga.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+	useEffect(() => {
+		navigation.setOptions({
+			title: I18n.t('WatchOS_Quick_Replies')
+		});
+	}, [navigation]);
+
+	const removeQuickReply = (reply: string) => {
+		if (!currentServer) {
+			console.error('Error: cannot set quick replies, current server is undefined');
```
diegolmello: Prefer the `log` helper over `console.error` (also at line 42), matching the rest of the methods/views code.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+		}
+		const value = input.trim();
+		if (!value) return;
+		if (!quickReplies?.includes(value)) setQuickReplies([...(quickReplies ?? []), value]);
+		setInput('');
+		syncWatchOSQuickReplies();
+	};
+
+	return (
+		<SafeAreaView testID='preferences-view'>
```
diegolmello: `testID='preferences-view'` is copied from `UserPreferencesView` — this is a different screen, so the id collides in e2e selectors. Give it a distinct testID (e.g. `watchos-quick-replies-view`).

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+			<List.Container>
+				<List.Section title='WatchOS_Quick_Replies'>
+					<>
+						{quickReplies && quickReplies.length !== 0 && (
+							<ScrollView horizontal style={{ marginVertical: 8, paddingHorizontal: 4 }} showsHorizontalScrollIndicator={false}>
```
diegolmello: Inline literal style drifts from the theme/StyleSheet pattern used in neighbouring views — move to a `StyleSheet.create`/themed style.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+								{quickReplies.map((reply, index) => (
+									<Chip key={index} text={reply} onPress={() => removeQuickReply(reply)} />
```
diegolmello: `key={index}` is an anti-pattern; replies are deduped on add, so `key={reply}` is stable and correct here.

diegolmello (commented):
## Review — WatchOS Quick Replies

Two-axis review (standards + spec) against `develop` (merge-base `3961d4dea`). Inline comments below; this body holds findings that have no single diff line to anchor to.

### 🔴 Headline bug — the first-login gate is dead
`FIRST_START` is **never written** anywhere in the diff (repo-wide grep: only imported + read). So `getWatchOSRepliesForServer` is always falsy → `syncWatchOSRepliesWithServer` re-seeds from server settings on **every** login, silently overwriting the user's local quick replies, and `logout.ts` never clears the key either. See inline notes on `getWatchOSRepliesFromMMKV.ts` and `logout.ts`. This was flagged independently by both review axes.

### Scope creep — recommend splitting
The spec's only RN display item is "scroll-to-bottom anchor in MessageListView", which maps **only** to the watch app's `ios/.../Views/MessageListView.swift`. The RN `MessageListView.tsx` is not touched. Everything else in the RN message pipeline is unrelated to Apple Watch and looks like the separate "load 50 visible messages when system messages hidden" fix (commit `815ff6c3d`):
- `loadMessagesForRoom.ts` rewrite (hidden-system-message batching, async refactor)
- new `HISTORY_UI_LOADER_PUSH/POP` actions + reducer + `room.ts` creators
- `useMessages.ts` in-flight guard
- ~470 lines of new tests (`loadMessagesForRoom.test.ts`, `useMessages.test.tsx`, `room.test.ts`)

Incidental churn also in the diff: new `Rocket.Chat.Watch.xcscheme`, `VoipModule.mm` codegen rename `SSLPinning`→`RocketChatSpec`, `AppDelegate.swift` reordering, `Gemfile.lock` newline. **Suggest moving the RN message-pipeline changes to their own PR.**

### Other spec gaps (no anchor line)
- **`isWatchPaired`** is exposed by the native bridge / `checkWatch()` but never consulted in `shouldShowWatchAppOptions` (uses only `isWatchSupported && isWatchAppInstalled`).
- **Conditional visibility is non-reactive** — `UserPreferencesView` gates the row on a synchronous `shouldShowWatchAppOptions()` evaluated once at render, with no state/subscription; it won't update if pairing/install status changes without a remount.

---
*Reviewed with the two-axis `/review` flow. All line references verified against `74407994`.*

divyanshu-patil:
@diegolmello can you review the changes?

ios/RCTWatchModule.mm
```diff
@@ -0,0 +1,194 @@
...
+
+- (nonnull NSNumber *)isWatchSupported {
+    if (![self isSupported]) {
+        return @(0);
+    } else {
+        return @(1);
+    }
+}
+
+- (nonnull NSNumber *)isWatchPaired {
```
divyanshu-patil, line 177: reserved for future, if we want to perform certain tasks or show conditional UI based on watchapp

app/lib/methods/WatchOSQuickReplies/getWatchOSRepliesFromMMKV.ts
```diff
@@ -0,0 +1,4 @@
+import { WATCHOS_QUICKREPLIES } from '../../constants/keys';
+import UserPreferences from '../userPreferences';
+
+export const getWatchOSRepliesForServer = (server: string) => UserPreferences.getBool(`${server}-${WATCHOS_QUICKREPLIES}`);
```
diegolmello, line 4: 🔴 **[blocking]** This reads a key that holds a **JSON-stringified `string[]`** using `getBool`. `getBoolean()` on a string value returns `undefined`, so this is effectively `undefined ?? null` → **always `null`**. That breaks the "first login only" gate in `syncWatchOSRepliesWithServer.ts`, so the server setting overwrites the user's locally-edited replies on **every** login (data loss).

Read the real value and gate on presence instead, e.g. `UserPreferences.contains(`${server}-${WATCHOS_QUICKREPLIES}`)` (or `getString(...) != null`). The name also implies it returns replies, not a bool — consider renaming.

app/lib/methods/WatchOSQuickReplies/syncWatchOSRepliesWithServer.ts
```diff
@@ -0,0 +1,26 @@
...
+const syncWatchOSQuickRepliesWithServer = (state: IApplicationState): boolean => {
+	if (!shouldShowWatchAppOptions()) return false;
+	const { server } = state.server;
+	const appleWatchReplies = state.settings.Apple_Watch_Quick_Actions;
+	if (!server) return false;
+
+	const isRepliesAvailable = getWatchOSRepliesForServer(server);
+
+	// we use apple watch settings from server on first login
+	if (!isRepliesAvailable && appleWatchReplies && typeof appleWatchReplies === 'string') {
```
diegolmello, line 17: 🔴 **[blocking]** This is where the broken guard bites: `isRepliesAvailable` is always falsy (see `getWatchOSRepliesFromMMKV.ts`), so this block re-seeds from the server setting on every login and clobbers the user's edits. Once the guard reads a real presence check, this becomes a true first-login-only seed.

app/lib/constants/keys.ts
```diff
@@ -23,5 +23,9 @@ export const ALERT_DISPLAY_TYPE_PREFERENCES_KEY = 'RC_ALERT_DISPLAY_TYPE_PREFERE
 export const CRASH_REPORT_KEY = 'RC_CRASH_REPORT_KEY';
 export const ANALYTICS_EVENTS_KEY = 'RC_ANALYTICS_EVENTS_KEY';
 export const TOKEN_KEY = 'reactnativemeteor_usertoken';
-export const CURRENT_SERVER = 'currentServer';
 export const CERTIFICATE_KEY = 'RC_CERTIFICATE_KEY';
+export const FIRST_START = 'RC_FIRST_START';
+
+// if you change below 2 keys you will need to change them in RCTWatchModule.mm too
+export const CURRENT_SERVER = 'currentServer';
+export const WATCHOS_QUICKREPLIES = 'RC_WATCHOS_QUICKREPLIES';
```
diegolmello, line 31: 🔴 **[blocking]** This key and the `currentServer` prefix are duplicated as mutable globals in `ios/RCTWatchModule.mm` (`mmkvQuickRepliesKey`/`mmkvCurrentServerKey`), kept in sync only by the comment above. A rename on one side silently breaks the cross-language contract at runtime. Prefer a single source of truth / a `static const` the native side references.

ios/RCTWatchModule.mm
```diff
@@ -0,0 +1,194 @@
...
+#import <stdexcept>
+
+@interface RCTWatchModule ()
+@property(nonatomic, strong) WCSession *session;
+@end
+
+@implementation RCTWatchModule
+
+// sync this with key declared in keys.ts
+NSString *mmkvQuickRepliesKey = @"RC_WATCHOS_QUICKREPLIES";
```
diegolmello, line 15: 🔴 **[blocking]** Native half of the duplicated key contract (see `app/lib/constants/keys.ts`). `RC_WATCHOS_QUICKREPLIES` / `currentServer` are hard-coded here as mutable globals; keep them from drifting via a shared constant rather than a comment.

package.json
diegolmello, line 228: 🔴 **[blocking]** Renaming the codegen spec `SSLPinning` → `RocketChatSpec` is cross-cutting: this spec also backs SSLPinning (iOS) and `NativeSSLPinningAndroid.ts`. Please confirm before merge that SSLPinning still registers on iOS (it's not in `modulesProvider` below), Android codegen still builds, and a clean **dual-platform** build passes. Worth resolving the current merge conflicts first since CI is red.

ios/RocketChat Watch App/Loaders/WatchSession.swift
```diff
@@ -1,98 +1,167 @@
...
+        }
+        
+        // for debug
+        // print(serverString,replies)
+
+        // if server exists, update in DB directly
+        DispatchQueue.main.async { [weak self] in
+            guard let self else { return }
+
+            if let server = self.serversDB.server(url: URL(string: serverString)!) {
```
diegolmello, line 94: 🔴 **[blocking]** `URL(string: serverString)!` force-unwraps a value that arrives over the wire in the application context. A malformed string crashes the Watch app. Guard it: `guard let url = URL(string: serverString) else { return }`.

ios/RocketChat Watch App/Loaders/ServersLoader.swift
```diff
@@ -3,56 +3,85 @@ import Foundation
 import WatchConnectivity
 
 enum ServersLoadingError: Error, Equatable {
```
diegolmello, line 5: 🔴 **[blocking]** This file (and `WatchSession.swift`, `MessageListView.swift`, `MessageComposerView.swift`) is reindented tabs → 4-spaces, turning it into a near-total rewrite in the diff. The rest of the Watch app uses tabs. This buries the real changes, poisons `git blame`, and is a likely source of the merge conflicts. Please revert formatting-only churn and keep tabs, leaving only semantic lines.

diegolmello (commented):
## Consolidated review — WatchOS Quick Replies

Thanks for the rework — the TurboModule migration (`RCTWatchModule` + `NativeWatchModule` via `RocketChatSpec`) and the server-side `Apple_Watch_Quick_Actions` setting both address the earlier asks, and `syncQuickReplies` always returning a `string` fixes the old nullability-vs-Spec crash. 🎉

**Heads-up:** most existing bot/reviewer comments target the pre-refactor code (`WatchBridge.m/.swift`, `getReplies()/getkey()`, `SET_FIRST_LOGIN`) and are now stale.

Inline comments below, severity-labelled 🔴 blocking / 🟡 important / 🟢 nit / 💡 suggestion.

**Top blockers:** (1) `getBool` on a JSON-string key makes the first-login gate always fail → server defaults overwrite the user's replies every login (data loss); (2) whole-file tab→space reindentation of 4 Swift files buries the real diff and likely causes the conflicts; (3) `URL(string:)!` force-unwrap on wire data can crash the Watch app; (4) the codegen `SSLPinning`→`RocketChatSpec` rename needs a verified clean dual-platform build; (5) the MMKV key contract is duplicated across JS/native as mutable globals. No tests were added, and blocker #1 is unit-testable.

app/sagas/login.js
```diff
@@ -228,6 +229,24 @@ const fetchUsersRoles = function* fetchRoomsFork() {
...
```
diegolmello, line 238: 🟡 **[important]** Magic `delay(1000)` race workaround: if settings still haven't loaded after 1s this seeds empty. Prefer reacting to the settings-loaded signal rather than a fixed sleep. Also note the inconsistency — `select(state => state)` above vs `select()` here (both return full state; pick one).

app/lib/methods/WatchOSQuickReplies/getWatchStatus.ts
```diff
@@ -0,0 +1,31 @@
...
+export function checkWatch(): IWatchStatus {
+	if (isAndroid || !NativeWatchModule)
+		return {
+			isWatchSupported: false,
+			isWatchPaired: false,
+			isWatchAppInstalled: false
+		};
+	const status: IWatchStatus = {
+		isWatchSupported: NativeWatchModule.isWatchSupported(),
+		isWatchPaired: NativeWatchModule.isWatchPaired(),
```
diegolmello, line 19: 🟡 **[important]** `isWatchPaired` is fetched (an extra sync bridge call) but never consumed by `shouldShowWatchAppOptions`. Also the guard below is `if (!a || !b) return false; return true;` → simplify to `return isWatchSupported && isWatchAppInstalled;`. And since `shouldShowWatchAppOptions()` is called inline in `UserPreferencesView` render, each of these is a synchronous JSI hop per render — see that call site.

app/views/UserPreferencesView/index.tsx
```diff
@@ -160,6 +161,18 @@ const UserPreferencesView = ({ navigation }: IUserPreferencesViewProps): JSX.Ele
 						</>
 					) : null}
 				</List.Section>
+				{shouldShowWatchAppOptions() ? (
```
diegolmello, line 164: 🟡 **[important]** `shouldShowWatchAppOptions()` runs on every render and makes ~3 synchronous TurboModule (JSI) calls each time. Compute once / memoize (`'use memo'` or `useMemo`) rather than calling inline in the render body.

ios/RCTWatchModule.mm
```diff
@@ -0,0 +1,194 @@
...
+#pragma mark - helpers
++ (NSString *)stringToHex:(NSString *)string {
+    const char *utf8 = [string UTF8String];
+    NSMutableString *hex = [NSMutableString string];
+    while (*utf8) [hex appendFormat:@"%02X", *utf8++ & 0x00FF];
+    return [[NSString stringWithFormat:@"%@", hex] lowercaseString];
+}
+
+#pragma mark - mmkv
+- (MMKVBridge *)getMMKV {
```
diegolmello, line 46: 🟡 **[important]** `getMMKV` reimplements the secure-key/appGroup/rootPath logic that `MMKVBridge.build()` already does (this PR even `@objc`-exposed `build()` in `MMKV.swift` but never calls it). This duplicates crypto/path logic that will drift. Reuse `[MMKVBridge build]`. Also `getValueFromMMKV` calls `getMMKV` on each read, so a single `syncQuickReplies` constructs MMKV twice — build once and reuse.

ios/Shared/RocketChat/MMKV.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 extension MMKVBridge {
-	static func build() -> MMKVBridge {
+    @objc static func build() -> MMKVBridge {
```
diegolmello, line 4: 🟡 **[important]** `build()` was made `@objc` here but nothing calls it — `RCTWatchModule.mm` reimplements the same logic instead. Either call `[MMKVBridge build]` from the module (preferred) or drop the unused `@objc`.

ios/RCTWatchModule.mm
```diff
@@ -0,0 +1,194 @@
...
+        
+        if (error) {
+            NSString *message =
+            error ? [NSString stringWithFormat:@"%@/%@",@"[ERROR]: ",[error localizedDescription]]
+            : @"[ERROR]: JSON parse error";
+            return message;
+        } else {
+            // quick replies are stored as array of strings
+            if ([json isKindOfClass:[NSArray class]]) {
+                NSArray *array = (NSArray *)json;
```
diegolmello, line 129: 🟡 **[important]** `try { ... } catch (const std::exception &e)` wraps a pure-Objective-C API (`updateApplicationContext:error:`) that reports failure via the `NSError**` already handled below — the C++ catch is cargo-cult and can never fire here. Separately, `@"_t" : @(arc4random())` a few lines down is a 32-bit token; rare collisions mean a skipped context update. Consider a monotonic counter or `NSUUID`.

ios/AppDelegate.swift
```diff
@@ -45,6 +45,9 @@ public class AppDelegate: ExpoAppDelegate {
     reactNativeFactory = factory
     bindReactNativeFactory(factory)
 
+    // Initialize Watch Connection
+    watchConnection = WatchConnection(session: WCSession.default)
```
diegolmello, line 49: 🟡 **[important]** Nice catch moving the `WatchConnection` init before `startReactNative`. But `RCTWatchModule` assigns `WCSession.defaultSession` without activating/delegating it — it relies on this line having activated the shared session first. Worth guarding native queries on `activationState == .activated` so the ordering dependency is explicit rather than implicit.

app/lib/constants/keys.ts
```diff
@@ -23,5 +23,9 @@ export const ALERT_DISPLAY_TYPE_PREFERENCES_KEY = 'RC_ALERT_DISPLAY_TYPE_PREFERE
 export const CRASH_REPORT_KEY = 'RC_CRASH_REPORT_KEY';
 export const ANALYTICS_EVENTS_KEY = 'RC_ANALYTICS_EVENTS_KEY';
 export const TOKEN_KEY = 'reactnativemeteor_usertoken';
-export const CURRENT_SERVER = 'currentServer';
 export const CERTIFICATE_KEY = 'RC_CERTIFICATE_KEY';
+export const FIRST_START = 'RC_FIRST_START';
```
diegolmello, line 27: 🟢 **[nit]** `FIRST_START` is declared but never used anywhere in `app/`. Dead code — remove (the PR description still lists it).

app/lib/methods/WatchOSQuickReplies/syncWatchOSRepliesWithServer.ts
```diff
@@ -0,0 +1,26 @@
...
+		const quickRepliesMMKVKey = `${server}-${WATCHOS_QUICKREPLIES}`;
+
+		const replies = appleWatchReplies.split(',').map(reply => reply.trim());
```
diegolmello, line 20: 🟢 **[nit]** Splitting the server setting on `,` means a reply containing a comma can't round-trip. Since the view stores an arbitrary `string[]`, consider disallowing/escaping commas on input, or agreeing a different delimiter with the server setting.

app/lib/methods/WatchOSQuickReplies/syncReplies.ts
```diff
@@ -0,0 +1,19 @@
...
+	try {
+		const result = NativeWatchModule?.syncQuickReplies();
+		if (result?.startsWith('[ERROR]')) {
+			log(new Error(result));
+			return false;
+		}
+
+		return true;
+	} catch (e) {
+		console.error('Failed to send quick replies', e);
```
diegolmello, line 16: 🟢 **[nit]** `console.error` — repo canon is `log()` (imported just above). Use `log(e)` for consistency and crash reporting.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+			return;
+		}
+		const newReplies = quickReplies?.filter(quickreply => quickreply !== reply);
+		setQuickReplies(newReplies);
+		syncWatchOSQuickReplies();
+	};
+
+	const addQuickReply = () => {
+		if (!currentServer) {
+			console.error('Error: cannot set quick replies, current server is undefined');
```
diegolmello, line 44: 🟢 **[nit]** `console.error` here vs `log(new Error(...))` in `removeQuickReply` a few lines up — inconsistent. Use `log()` in both.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+	return (
+		<SafeAreaView testID='watchos-quick-replies-view'>
+			<List.Container>
+				<List.Section title='WatchOS_Quick_Replies'>
+					<>
+						{quickReplies && quickReplies.length !== 0 && (
+							<ScrollView horizontal style={styles.chipContainer} showsHorizontalScrollIndicator={false}>
+								{quickReplies.map(reply => (
+									<Chip key={reply} text={reply} onPress={() => removeQuickReply(reply)} />
```
diegolmello, line 62: 🟢 **[nit]** Tapping a `Chip` deletes it with no affordance or confirmation. Intentional? A delete icon / long-press would make the destructive action discoverable.

app/views/UserWatchOSQuickRepliesView/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+							</ScrollView>
+						)}
+					</>
+					<List.Separator />
+					<FormTextInput
+						value={input}
+						onChangeText={text => setInput(text)}
+						placeholder={I18n.t('Add_Quick_Reply')}
+						onSubmitEditing={addQuickReply}
+						maxLength={30}
```
diegolmello, line 73: 🟢 **[nit]** `maxLength={30}` is a magic number — extract a named constant so the limit is documented and reusable.

app/views/UserPreferencesView/index.tsx
```diff
@@ -160,6 +161,18 @@ const UserPreferencesView = ({ navigation }: IUserPreferencesViewProps): JSX.Ele
...
+					<List.Section title='WatchOS_Quick_Replies'>
+						<List.Separator />
+						<List.Item
+							title='WatchOS_Quick_Replies'
+							onPress={() => navigateToScreen('UserWatchOSQuickRepliesView')}
```
diegolmello, line 169: 🟢 **[nit]** `navigateToScreen` always fires `logEvent(events.UP_GO_USER_NOTIFICATION_PREF)`, so this Watch item logs the wrong analytics event. (CodeRabbit raised this too — still valid.) Pass the event in or use a dedicated navigate.

package.json
```diff
@@ -225,11 +225,17 @@
...
 		"type": "modules",
 		"jsSrcsDir": "./app/lib/native",
 		"android": {
 			"javaPackageName": "chat.rocket.reactnative.networking"
+		},
+		"ios": {
+			"modulesProvider": {
```
diegolmello, line 235: 🟢 **[nit]** The iOS `modulesProvider` lists only `WatchModule` and `VoipModule`. Double-check no other iOS TurboModule under this spec relies on auto-registration that this now overrides.

ios/Libraries/VoipModule.mm
diegolmello, line 5: 💡 **[suggestion]** Import updated to follow the `RocketChatSpec` rename — good. Just make sure this is the only stale `SSLPinning` header import across the iOS target.

ios/RCTWatchModule.mm
```diff
@@ -0,0 +1,194 @@
...
+        return nil;
+    }
+
+    NSString *value = [mmkv stringForKey:key];
+
+    return value;
+}
+
+#pragma mark - internal methods
+- (BOOL)isSupported {
```
diegolmello, line 83: 🟢 **[nit]** `if ([WCSession isSupported]) { return YES; } else { return NO; }` → `return [WCSession isSupported];`. Same pattern in `isWatchSupported`/`isWatchPaired`/`isWatchAppInstalled` (the `@(0)`/`@(1)` if/else) can be direct returns.

ios/RocketChat Watch App/Loaders/WatchSession.swift
```diff
@@ -1,98 +1,167 @@
...
+    ) {
+        
+        // for debug
+        // print("received Context:", applicationContext)
+        
+        guard
+            let serverString = applicationContext["server"] as? String,
+            let replies = applicationContext["quickReplies"] as? [String]
+        else {
+            print("Invalid context")
```
diegolmello, line 83: 🟢 **[nit]** Leftover `print("Invalid context")` plus several commented-out `// print(...)` in this file. Repo has a no-noise rule — strip debug prints and commented code.

ios/RocketChat Watch App/Database/Entity/Server.swift
```diff
@@ -28,6 +29,15 @@ public final class Server: NSManagedObject {
 	init(context: NSManagedObjectContext) {
 		fatalError()
 	}
+    
+    public var quickReplies: [String] {
+        get {
+            quickRepliesData as? [String] ?? []
+        }
+        set {
+            quickRepliesData = newValue as NSArray;
```
diegolmello, line 38: 🟢 **[nit]** Trailing semicolon (`newValue as NSArray;`) is non-idiomatic Swift, and this block mixes 4-space indent into a tab-indented file. Drop the `;` and match surrounding tabs.

ios/RocketChat Watch App/Database/Default.xcdatamodeld/Default.xcdatamodel/contents
```diff
@@ -12,6 +12,7 @@
         <attribute name="iconURL" optional="YES" attributeType="URI"/>
         <attribute name="name" optional="YES" attributeType="String"/>
         <attribute name="password" optional="YES" attributeType="String"/>
+        <attribute name="quickRepliesData" optional="YES" attributeType="Transformable" customClassName="NSArray"/>
```
diegolmello, line 15: 🟢 **[nit]** `Transformable` / `customClassName="NSArray"` without a named `valueTransformerName` triggers secure-coding warnings on modern iOS. Consider a `NSSecureUnarchiveFromDataTransformer`-based transformer.

Gemfile.lock
```diff
@@ -320,4 +320,4 @@ RUBY VERSION
    ruby 2.7.7p221
 
 BUNDLED WITH
-   2.1.4
+   2.1.4
```
diegolmello, line 323: 🟢 **[nit]** The only change here is a stripped trailing newline (`\ No newline at end of file`). Accidental — please revert.

PR #6984, opened 2026-02-13
fix: prevent falsy values from being coerced to null

### LINKED ISSUE #6983: bug: UserPreferences.getBool() returns null for false values due to || operator instead of ??
### Describe the Bug

`UserPreferences.getBool()` and `UserPreferences.getNumber()` in [app/lib/methods/userPreferences.ts](cci:7://file:///home/parash/Videos/files/os/org/Rocket.Chat.ReactNative/app/lib/methods/userPreferences.ts:0:0-0:0) use the `||` operator to return `null` for missing keys. However, this causes stored `false` (boolean) and `0` (number) values to also be incorrectly returned as `null`, because `false || null → null` and `0 || null → null`.

// Line 113 — getBool
return this.mmkv.getBoolean(key) || null;  // false || null → null ❌

// Line 143 — getNumber
return this.mmkv.getNumber(key) || null;   // 0 || null → null ❌

### Steps to Reproduce

1. Set any boolean preference to false (e.g., disable a notification toggle).
2. Close and reopen the app (or trigger a code path that reads the preference via UserPreferences.getBool()).
3. Observe that the preference is read back as null instead of false, reverting to its default value.

### Expected Behavior

getBool():should return false when the stored value is false, and 
getNumber(): should return 0 when the stored value is 0. Only genuinely missing keys should return null.

### Actual Behavior

Both false and 0 are coerced to null by the || operator, effectively resetting user preferences to their defaults on every read.

8.2.0

### Rocket.Chat App Version

4.70.0.999999999

### Device Name

All devices (platform-independent logic bug)

### OS Version

Android 15

### Additional Context

_No response_

Yaddalapalli-Charan-Kumar-Naidu:
## Proposed changes

The `getString()`, `getBool()`, and `getNumber()` methods in `UserPreferences` currently use the `||` operator to return `null` when a key is missing.

This unintentionally coerces valid falsy values into `null`:

- `false || null → null`
- `0 || null → null`
- `"" || null → null`

As a result, legitimate stored preferences (e.g., disabled toggles, zero counters, or empty strings) are read back as `null` instead of their actual values.

This PR replaces the `|| null` pattern with an explicit `undefined` check, ensuring:

- Only truly missing keys return `null`
- Valid falsy values (`false`, `0`, `""`) are preserved correctly

The change is minimal, safe, and non-breaking.

## Issue(s)

Closes #6983

## How to test or reproduce

1. Navigate to **Settings → Security & Privacy → Screen Lock**
2. Enable Screen Lock and set a passcode
3. Toggle **Biometry OFF**
4. Add a `console.log` inside `getBool()` to inspect the returned value

**Before fix:**  
`getBool()` returns `null` instead of `false`.

**After fix:**  
`getBool()` correctly returns `false`.

The same issue can be reproduced for:
- `getNumber()` when a stored value is `0`
- `getString()` when a stored value is `""`

## Screenshots

N/A — logic-only change with no UI impact.

## Types of changes

## Checklist

## Further comments

All three getter methods shared the same `|| null` pattern, so the fix applies consistently across them.

While some current callers use `??` fallbacks that mask this issue, the underlying behavior is still incorrect and could affect future usages where no fallback is applied.

This update ensures correct handling of falsy preference values and aligns the implementation with the expected behavior of MMKV getters (`undefined` meaning “key not found”).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected retrieval of stored user preferences so empty strings, false, and 0 are preserved rather than being treated as missing. Undefined values are now consistently mapped to null. No public APIs or settings changed; users will see more reliable behavior when reading previously saved preferences.


app/lib/methods/userPreferences.ts
```diff
@@ -98,7 +98,8 @@ class UserPreferences {
 
 	getString(key: string): string | null {
 		try {
-			return this.mmkv.getString(key) || null;
```
diegolmello, line 101: This should work.

```
return this.mmkv.getString(key) ?? null;
```

diegolmello (changes requested):

app/lib/methods/userPreferences.ts
Yaddalapalli-Charan-Kumar-Naidu: Thanks for the suggestion! That's much cleaner.

I've updated getString, getBool and getNumber to use the nullish coalescing operator (??)

PR #6991, opened 2026-02-18
chore: upgrade to react-native-mmkv v4

Rohit3523:
## Proposed changes

This PR upgrades `react-native-mmkv` from v3 to v4 to ensure compatibility with Android's 16KB page size alignment requirement.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1842

### Fresh install (sanity check)

1. Delete the app from the device.
2. Install this branch.
3. Log in to a server, create / restore E2EE keys, enable biometry and screen lock if available, add a second server.
4. Confirm all flows work end to end.

### Upgrade path (the primary risk surface for this PR)

For each scenario below, start from a **clean `develop` build with state already set up**, then install this branch over the top without uninstalling.

1. **Login persistence**
   - On `develop`: log in to a server, leave the app running.
   - Install this branch on top.
   - Expected: still logged in, no re-authentication required, server connection resumes.

2. **E2EE keys**
   - On `develop`: log into a server with E2EE, decrypt at least one encrypted message in a room.
   - Install this branch on top.
   - Expected: encrypted messages still decrypt without re-entering the password; E2EE keys remain accessible.

3. **SSL client certificates**
   - On `develop`: connect to a server that requires SSL client certificate authentication, save the certificate password so it is stored via MMKV.
   - Trigger a request that goes through the Notification Service (e.g. background push) to confirm SSL pinning works from the extension.
   - Install this branch on top.
   - Expected: SSL pinning continues to work in both the main app and the Notification Service; certificate password is still resolvable.

4. **Biometry**
   - On `develop`: enable biometric unlock (Face ID / Touch ID / fingerprint).
   - Install this branch on top.
   - Expected: biometric prompt still appears on app launch, unlocks as before.

5. **Screen lock**
   - On `develop`: set a passcode-based screen lock with a non-default auto-lock interval.
   - Install this branch on top.
   - Expected: passcode still required after the configured interval; chosen interval still in effect.

6. **Server list**
   - On `develop`: add at least two servers, switch between them.
   - Install this branch on top.
   - Expected: both servers still listed, last-active server is the one selected on launch, switching between servers still works.

### Platform coverage

- iOS: verify all six scenarios on a physical device (so the Notification Service path actually exercises SSL pinning).
- Android: verify all six scenarios; pay specific attention to the 16KB page-size alignment behavior on a Pixel 9 or other 16KB-aligned device.

## Screenshots

N/A. This PR has no user-facing UI changes. The intent is for the upgrade to be invisible to the end user.

## Types of changes

## Checklist

### Why we are using community managed MMKV Core instead of Tencent MMKV Core

We are using `io.github.zhongwuzw:mmkv:2.2.4` instead of `com.tencent:mmkv-static:1.2.10` because `react-native-mmkv` v4 is a rewrite based on Nitro Modules / JSI. Since we also use MMKV directly in native Android code, we aligned our native dependency with the same MMKVCore artifacts used internally by `react-native-mmkv` v4 to avoid runtime / ABI inconsistencies between the JS, C++, and native storage layers.

A follow-up task to switch back to Tencent's `mmkv-static` once the upstream alignment lands: <link to tracker ticket>

### iOS specifics: AppGroupIdentifier rename and dynamic frameworks

- The `react-native-mmkv` v4 upgrade guide requires renaming the Info.plist key from `AppGroup` to `AppGroupIdentifier` so that `HybridMMKVPlatformContext.swift` can resolve the App Group container. The on-disk container path is determined by the *value* (`group.ios.chat.rocket`), which is unchanged, so the rename is a key-name update only and existing data remains in place.
- MMKV / MMKVCore / NitroMmkv / NitroModules are now built as dynamic frameworks via the corrected `pre_install` block in the Podfile. The previous version of that block had a `def...end` mismatch that left the second `if` unreachable, so MMKV had been building as a static framework by default; the fix here makes the originally intended dynamic-framework declaration actually take effect.


Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

__mocks__/react-native-mmkv.js
```diff
@@ -1,20 +1,25 @@
...
 	constructor(config = {}) {
-		const { id = 'default', mode, path } = config;
+		const { id = 'mmkv.default', path, encryptionKey, mode, readOnly } = config;
 		this.id = id;
-		this.mode = mode;
 		this.path = path;
+		this.encryptionKey = encryptionKey;
+		this.mode = mode;
+		this.isReadOnly = readOnly || false;
+		this._size = 0;
```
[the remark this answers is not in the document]
Rohit3523: in testing we have single instance only

__mocks__/react-native-mmkv.js
```diff
@@ -94,106 +138,186 @@ export class MMKV {
...
+	return storageInstances.has(id);
+}
 
-	return [value, setStoredValue];
+export function deleteMMKV(id) {
+	const deleted = storageInstances.delete(id);
+	if (defaultInstance?.id === id) {
+		defaultInstance = null;
+	}
+	return deleted;
```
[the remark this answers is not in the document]
Rohit3523: If we delete it from storageIntance, it will delete all data mapped under it. Can you please check this again

android/app/build.gradle
```diff
@@ -147,7 +147,7 @@ dependencies {
     implementation project(':watermelondb-jsi')
 
     implementation "com.google.code.gson:gson:2.8.9"
-    implementation "com.tencent:mmkv-static:1.2.10"
+    implementation "io.github.zhongwuzw:mmkv:2.2.4"
```
OtavioStasiak, line 150: `io.github.zhongwuzw` is a community fork. This is supply-chain-sensitive — MMKV stores encrypted user data. The PR should justify why this fork   
  (presumably the only one with 16KB alignment) and confirm the maintainer is trusted, ideally by linking the upstream tracking issue. If Tencent
  has since published a 16KB-aligned release, prefer that.

app/lib/methods/userPreferences.ts
```diff
@@ -23,14 +23,10 @@ const getEncryptionKey = (): string | undefined => {
 
 const buildConfiguration = (): Configuration => {
 	const config: Configuration = {
-		id: 'default'
+		id: 'default',
+		mode: 'multi-process'
```
OtavioStasiak, line 27: - mode: Mode.MULTI_PROCESS                                                                                                                       
  + mode: 'multi-process'   
  v4 still exports a typed Mode enum. Using Mode.MULTI_PROCESS would catch a future rename at compile time; the bare string won't.

OtavioStasiak (changes requested):
https://github.com/RocketChat/Rocket.Chat.ReactNative/actions/runs/25510665089/job/74873027764?pr=6991 
Test 4 is failing... please check this one.

Tested:
- E2ee
- notification
- first install
- upgrade


Everything seems to be working as expected, but we have some points to improve!

app/lib/methods/userPreferences.ts
Rohit3523: The version we are using is exporting Mode as only type

Screenshot 2026-05-08 at 1 40 01 AM

To ensure type safety, I added the configuration to the object.

android/app/build.gradle
Rohit3523: I have added `io.github.zhongwuzw` because the same is being used in the package https://github.com/mrousavy/react-native-mmkv/blob/8396c70d759707a31deea21686383a3da5c9c069/packages/react-native-mmkv/android/build.gradle#L141-L142

Tencent added the support for 16kb in https://github.com/Tencent/MMKV/releases/tag/v1.3.14

Rohit3523: I have added a comment in PR description with more information

diegolmello: Author is from Margelo, so it should be ok, but we need a task to point to Tencent's.

ios/Libraries/MMKVKeyManager.mm
```diff
@@ -35,45 +35,48 @@ static void Logger(NSString *format, ...) {
...
     } @catch (NSException *exception) {
         Logger(@"MMKV initialization error: %@ - %@", exception.name, exception.reason);
     }
 }
 
 + (NSString *)initializeMMKV {
-    NSString *appGroup = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroup"];
+    NSString *appGroup = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroupIdentifier"] 
+                         ?: [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroup"];
+
```
diegolmello: Why do we need this?

ios/Libraries/MMKVKeyManager.mm
```diff
@@ -35,45 +35,48 @@ static void Logger(NSString *format, ...) {
...
         }
 
         // Verify MMKV can be opened with this key
         NSData *cryptKey = [password dataUsingEncoding:NSUTF8StringEncoding];
-        MMKVBridge *mmkv = [[MMKVBridge alloc] initWithID:@"default"
-                                                cryptKey:cryptKey
-                                                rootPath:mmkvPath];
+
+        // Ensure we use MMKVMultiProcess to match the JS side
+        MMKV *mmkv = [MMKV mmkvWithID:@"default" cryptKey:cryptKey mode:MMKVMultiProcess];
```
diegolmello: You don't need path anymore?

ios/Libraries/SecureStorage.m
```diff
@@ -64,8 +64,8 @@ - (NSMutableDictionary *)newSearchDictionary:(NSString *)identifier {
     NSMutableDictionary *searchDictionary = [[NSMutableDictionary alloc] init];
 
     // this value is shared by main app and extensions, so, is the best to be used here
-    serviceName = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroup"];
-    
+    serviceName = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroupIdentifier"];
```
diegolmello, line 67: I didn't get the rename.

ios/Shared/RocketChat/MMKVBridge.mm
```diff
@@ -21,92 +17,55 @@ - (instancetype)initWithID:(NSString *)mmapID
                  rootPath:(nullable NSString *)rootPath {
     self = [super init];
     if (self) {
-        // Initialize MMKV if needed
+        // 1. Initialize MMKV path globally if provided (e.g., for App Groups)
```
diegolmello: Remove these AI comments steps.
Add comments based on what's meaningful for us.

ios/Shared/RocketChat/MMKVBridge.mm
```diff
@@ -2,16 +2,12 @@
...
 
 #import "MMKVBridge.h"
-#import "MMKV.h"
-#import <string>
+#import <MMKV/MMKV.h>
 
-@interface MMKVBridge()
-@property (nonatomic, assign) MMKV *mmkvInstance;
+@interface MMKVBridge ()
+@property (nonatomic, strong) MMKV *mmkv;
```
diegolmello: Why did you rename?

diegolmello (changes requested):

ios/Podfile
```diff
@@ -45,6 +46,11 @@ pre_install do |installer|
         def pod.build_type;
         Pod::BuildType.static_library # >= 1.9
       end
+      if ['MMKV', 'MMKVCore', 'react-native-mmkv', 'react-native-nitro-modules'].include?(pod.name)
```
diegolmello:
```suggestion
      if ['MMKV', 'MMKVCore', 'NitroMmkv', 'NitroModules'].include?(pod.name)
```

ios/Shared/RocketChat/MMKVBridge.mm
```diff
@@ -21,92 +17,55 @@ - (instancetype)initWithID:(NSString *)mmapID
...
 
 - (NSUInteger)count {
-    if (!_mmkvInstance) return 0;
-    return _mmkvInstance->count();
+    if (!_mmkv) return 0;
+    return _mmkv.count;
 }
 
-@end
+@end
```
diegolmello: missing new line

ios/Libraries/SecureStorage.m
Rohit3523: This change is mentioned in the upgrade guide of `react-native-mmkv`
From upgrade log: "To better comply with Apple's naming, I changed the AppGroup key to AppGroupIdentifier. If you used MMKV with App Groups, make sure you change the key in your Info.plist"

https://github.com/mrousavy/react-native-mmkv/blob/main/docs/V4_UPGRADE_GUIDE.md#appgroup---appgroupidentifier

OtavioStasiak (dismissed):
SSL certificate doesn't work considering upgrades... I will send a message to discuss about it.

ios/Libraries/Challenge.mm
```diff
@@ -25,7 +25,16 @@ + (MMKVBridge *)getMMKVInstance {
...
+
+  // Use same App Group path as MMKVKeyManager.mm to access shared SSL certificates
+  NSString *appGroup = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"AppGroupIdentifier"];
+  NSString *rootPath = nil;
+  if (appGroup) {
+    NSURL *groupURL = [[NSFileManager defaultManager] containerURLForSecurityApplicationGroupIdentifier:appGroup];
+    rootPath = [groupURL path];
+  }
+
+  MMKVBridge *mmkvBridge = [[MMKVBridge alloc] initWithID:@"default" cryptKey:cryptKey rootPath:rootPath];
```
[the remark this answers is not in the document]
Rohit3523: When i add path in Challenge.mm, SSL stop working

Rohit3523:
@coderabbitai review

ios/PrivacyInfo.xcprivacy
diegolmello, line 1: These changes look unrelated.

ios/Libraries/MMKVKeyManager.mm
```diff
@@ -35,13 +36,19 @@ static void Logger(NSString *format, ...) {
...
     @try {
         NSString *mmkvPath = [self initializeMMKV];
         if (!mmkvPath) {
             Logger(@"Failed to initialize MMKV path");
             return;
         }
 
+        // Initialize MMKV against the App Group container so the main app and
+        // NotificationService share the same storage.
+        [MMKV initializeMMKV:nil groupDir:mmkvPath logLevel:MMKVLogInfo];
```
diegolmello: It was not needed before. Why it's needed now?

diegolmello (changes requested):
Apple watch is not working.
I tested with SSLPinning, but you should test without it as well.

Rohit3523:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > ios/Libraries/MMKVKeyManager.mm (1)> `83-83`: _⚠️ Potential issue_ | _🔴 Critical_ | _⚡ Quick win_
> > > **Remove manual `/mmkv` path appending—MMKV SDK adds it internally, causing double-nesting.**
> > > Line 83 appends `/mmkv` to the App Group container path before passing it to `MMKVBridge` as `rootPath` (line 62). The MMKV SDK internally appends `/mmkv` to any `rootPath` it receives, resulting in double-nesting: `<AppGroup>/mmkv/mmkv`. This breaks shared storage between the app and extensions and can break SSL certificate lookups.
> > > Pass the raw App Group container URL path directly without appending `/mmkv`:
> > > 
> > > 🛠️ Proposed fix
> > > ```diff
> > > -    NSString *mmkvPath = [[groupURL path] stringByAppendingPathComponent:@"mmkv"];
> > > -    [[NSFileManager defaultManager] createDirectoryAtPath:mmkvPath
> > > -                              withIntermediateDirectories:YES
> > > -                                               attributes:nil
> > > -                                                    error:nil];
> > > -    return mmkvPath;
> > > +    // MMKV SDK appends /mmkv internally, so pass raw App Group path
> > > +    return [groupURL path];
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > Based on learnings: "For iOS code using react-native-mmkv, when calling `[MMKV initializeMMKV:nil groupDir:groupDir logLevel:...]` (and when passing `rootPath` to MMKVBridge), pass the raw App Group container URL path only. Do NOT append `/mmkv` yourself. The MMKV SDK already appends `/mmkv` internally, so passing `<AppGroup>/mmkv` will cause double-nesting and can break shared storage/SSL cert lookups."
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > 
> > > In `@ios/Libraries/MMKVKeyManager.mm` at line 83, The code currently constructs
> > > mmkvPath by appending @"/mmkv" to the App Group container (NSString *mmkvPath =
> > > [[groupURL path] stringByAppendingPathComponent:@"mmkv"];), which causes MMKV to
> > > double-nest because the MMKV SDK appends "/mmkv" internally; remove the manual
> > > append and pass the raw App Group container path (e.g., use [groupURL path] or
> > > groupURL.path) to whatever consumes mmkvPath/rootPath (references: mmkvPath,
> > > groupURL, MMKVBridge, rootPath) so MMKVBridge receives the App Group container
> > > path only.
> > > ```
> 
> ```
> 
> Outside diff comments:
> In `@ios/Libraries/MMKVKeyManager.mm`:
> - Line 83: The code currently constructs mmkvPath by appending @"/mmkv" to the
> App Group container (NSString *mmkvPath = [[groupURL path]
> stringByAppendingPathComponent:@"mmkv"];), which causes MMKV to double-nest
> because the MMKV SDK appends "/mmkv" internally; remove the manual append and
> pass the raw App Group container path (e.g., use [groupURL path] or
> groupURL.path) to whatever consumes mmkvPath/rootPath (references: mmkvPath,
> groupURL, MMKVBridge, rootPath) so MMKVBridge receives the App Group container
> path only.
> ```
> 
> ℹ️ Review info
> ⚙️ Run configuration
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro
> 
> **Run ID**: `4d3b2bcf-96fd-4bd4-bf0a-6c8d259d8d6f`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [414b288](https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/414b288a1e194670e87e03e0b5f7a42a2cffcdb6) and [f9952cf](https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/f9952cf0796a39e75f4ed5e03bb88fa90598f389).
> 
> 📒 Files selected for processing (1)
> * `ios/Libraries/MMKVKeyManager.mm`
> 
> 📜 Review details

@coderabbitai This change is causing sync issue on watch os

ios/Shared/RocketChat/MMKVBridge.mm
```diff
@@ -21,92 +17,60 @@ - (instancetype)initWithID:(NSString *)mmapID
...
-            MMKV::initializeMMKV(rootPathStr);
-        }
-        
-        std::string mmapIDStr = [mmapID UTF8String];
-        
-        if (cryptKey && [cryptKey length] > 0) {
-            std::string cryptKeyStr((const char *)[cryptKey bytes], [cryptKey length]);
-            _mmkvInstance = MMKV::mmkvWithID(mmapIDStr, MMKV_MULTI_PROCESS, &cryptKeyStr);
-        } else {
-            _mmkvInstance = MMKV::mmkvWithID(mmapIDStr, MMKV_MULTI_PROCESS);
```
diegolmello, line 36: You removed the call without cryptKey arg.
Did you test and confirmed that sending nil as cryptKey works properly?

ios/Shared/RocketChat/MMKVBridge.mm
```diff
@@ -21,92 +17,60 @@ - (instancetype)initWithID:(NSString *)mmapID
...
+        if (resolvedRootPath.length > 0 && ![resolvedRootPath.lastPathComponent isEqualToString:@"mmkv"]) {
+            resolvedRootPath = [resolvedRootPath stringByAppendingPathComponent:@"mmkv"];
         }
+
+        NSData *finalCryptKey = (cryptKey.length > 0) ? cryptKey : nil;
+        _mmkvInstance = [MMKV mmkvWithID:mmapID
+                                cryptKey:finalCryptKey
+                                rootPath:resolvedRootPath
+                                    mode:MMKVMultiProcess
+                        expectedCapacity:0];
```
diegolmello, line 36: What's this?

ios/Shared/RocketChat/MMKVBridge.mm
```diff
@@ -21,92 +17,60 @@ - (instancetype)initWithID:(NSString *)mmapID
                  rootPath:(nullable NSString *)rootPath {
     self = [super init];
     if (self) {
-        // Initialize MMKV if needed
-        if (rootPath) {
-            std::string rootPathStr = [rootPath UTF8String];
-            MMKV::initializeMMKV(rootPathStr);
-        }
```
diegolmello, line 28: You changed this logic and the comment.
Now you're initializing MMKV every time.
Before it would initialize conditionally.

Can you elaborate on this decision?

ios/Libraries/MMKVKeyManager.mm
```diff
@@ -65,13 +58,13 @@ + (void)initialize {
...
 
-        // Ensure we use MMKVMultiProcess to match the JS side
         MMKVBridge *mmkv = [[MMKVBridge alloc] initWithID:@"default"
                                                 cryptKey:cryptKey
                                                 rootPath:mmkvPath];
 
         if (mmkv) {
-            Logger(@"MMKV initialized successfully. Keys: %lu", (unsigned long)[mmkv count]);
+            NSUInteger keyCount = [mmkv count];
+            Logger(@"MMKV initialized with encryption, %lu keys found", (unsigned long)keyCount);
```
diegolmello, line 66: `with encryption` only? What about the unencrypted use case?

ios/Libraries/Challenge.mm
diegolmello, line 37: Why'd add rootPath?

ios/NotificationService/NotificationService-Bridging-Header.h
```diff
@@ -1,10 +1,11 @@
 //
 //  Use this file to import your target's public headers that you would like to expose to Swift.
+//  NotificationService-Bridging-Header.h
```
diegolmello: ?

ios/NotificationService/NotificationService-Bridging-Header.h
```diff
@@ -1,10 +1,11 @@
...
 //
 
 #import "../Libraries/SecureStorage.h"
 #import "../Libraries/Challenge.h"
-#import "../Shared/RocketChat/MMKVBridge.h"
+#import "MMKVBridge.h"
```
diegolmello: You can rollback to relative path.

ios/RocketChatRN.xcodeproj/project.pbxproj
```diff
@@ -3027,6 +3026,7 @@
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(SRCROOT)",
+					"$(SRCROOT)/Shared/RocketChat",
```
diegolmello: If you remove this, relative path probably works.
This change looks unrelated to this PR, but a code choice.

ios/RocketChatRN.xcodeproj/project.pbxproj
diegolmello, line 1: I don't think these changes are actually needed.

ios/Shared/Extensions/FileManager+Extensions.swift
```diff
@@ -2,12 +2,11 @@ import Foundation
 
 extension FileManager {
   func groupDir() -> String {
-    let applicationGroupIdentifier = Bundle.main.string(forKey: "AppGroup")
-    
-    guard let path = containerURL(forSecurityApplicationGroupIdentifier: applicationGroupIdentifier)?.path else {
+    guard let appGroup = Bundle.main.object(forInfoDictionaryKey: "AppGroupIdentifier") as? String,
```
diegolmello: Where's `AppGroup`?
If you choose to change it to `AppGroupIdentifier` (which it's just cosmetic and makes the code harder to understand), you may need to check `AppGroup` as well.

ios/ShareRocketChatRN/Info.plist
```diff
@@ -4,7 +4,7 @@
 <dict>
 	<key>IS_OFFICIAL</key>
 	<false/>
-	<key>AppGroup</key>
+	<key>AppGroupIdentifier</key>
```
diegolmello, line 7: What's the benefit of making this change?
If you don't have a better explanation than "the docs mention it", we should revert this change.
react-native-mmkv v4 is a breaking change.
This PR should not.

diegolmello (changes requested):

ios/Libraries/MMKVKeyManager.mm
Rohit3523: I have reverted this change

ios/Libraries/Challenge.mm
Rohit3523: Before v4 we could pass `rootPath:nil` here because the old MMKV C++ API exposed `MMKV::initializeMMKV(rootPathStr)`, which MMKVKeyManager.mm called during `+initialize` to set a process wide default rootPath. Every later mmkvWithID call inherited it, so passing nil from this file worked by coincidence of init order.

The v4 Objective-C API moved rootPath from a process wide default to a per-instance parameter (see MMKV.h:122 in our installed Pod, where the default is documented as `NSDocumentDirectory/mmkv`). With nil here, the bridge resolves to the app sandbox Documents folder instead of the App Group container, and the SSL certs written by the main app are not visible from the Notification Service extension.

ios/Shared/RocketChat/MMKVBridge.mm
Rohit3523: `expectedCapacity` is an MMKV file-size hint for the initial mmap allocation. The MMKV.h header in our Pod documents it as "the file size you expected when opening or creating file" (line 96, repeated on the overloads that take it). Passing `0` means no hint, use MMKV's default starting size, and the file grows lazily as data is written.

It's here because of MMKV's API shape, not a behavior choice on our side. We need `cryptKey:`, `rootPath:`, and `mode:MMKVMultiProcess` all on the same call (encrypted store, App Group path, and shared access between the main app and the share/notification extensions). The only overload in MMKV.h that accepts all three is the 5-arg variant at line 151, and that one also requires `expectedCapacity:. 0` is the documented "no hint" sentinel.

ios/Shared/RocketChat/MMKVBridge.mm
Rohit3523: We did not remove `cryptKey`, it is still being passed as the 2nd argument in the new Obj-C overload. The old code had two separate C++ overloads (one with cryptKey, one without) selected by an if branch; the new code collapses them into a single call that accepts a nullable cryptKey, which is the v2 Obj-C API style.

ios/ShareRocketChatRN/Info.plist
Rohit3523: The rename is required by the v4 upgrade itself, it is not a separate breaking change we are introducing. From the  https://github.com/mrousavy/react-native-mmkv/blob/main/docs/V4_UPGRADE_GUIDE.md#appgroup---appgroupidentifier

v4's HybridMMKVPlatformContext.swift looks up the AppGroupIdentifier key to resolve the shared container. Without the rename that lookup returns nil and the library silently falls back to per-target sandbox storage, so the main app and the share/notification extensions stop seeing each other's data.

ios/Shared/RocketChat/MMKVBridge.mm
Rohit3523: The old conditional was a v3 C++ API quirk: `MMKV::initializeMMKV(rootPathStr)` set both the process-wide default AND the rootPath any later instance would inherit, so we only called it when we had a rootPath to register. In v4 those two responsibilities are separate. `initializeMMKV:` now just sets up MMKV's internal state (logger, default base path for instances that do not specify their own), and the actual instance rootPath is passed per call to `mmkvWithID:cryptKey:rootPath:...`. We no longer need the global init to know about our App Group path.

The reason it now runs unconditionally is that the Notification Service and Share extensions never load the JS bundle, so react-native-mmkv (which calls `initializeMMKV:` from its Nitro setup at JS startup) never fires in those targets. Without our explicit call, the first `mmkvWithID:` invocation in an extension would hit an uninitialized MMKV. In the main app the call is a no-op because react-native-mmkv has already done it, and `initializeMMKV:` is idempotent (subsequent calls return the cached base path without doing real work).

ios/ShareRocketChatRN/Info.plist
diegolmello: Can you point me to that code you're mentioning?

ios/RocketChatRN.xcodeproj/project.pbxproj
diegolmello: Did you do it?

ios/ShareRocketChatRN/Info.plist
Rohit3523: https://github.com/mrousavy/react-native-mmkv/blob/main/packages/react-native-mmkv/ios/HybridMMKVPlatformContext.swift#L34

ios/RocketChatRN.xcodeproj/project.pbxproj
Rohit3523: Yup i have reverted the changes

app/lib/methods/userPreferences.ts
```diff
@@ -134,7 +130,7 @@ class UserPreferences {
 	}
 
 	removeItem(key: string): void {
-		this.mmkv.delete(key);
+		this.mmkv.remove(key);
 	}
 
 	getNumber(key: string): number | null {
```
diegolmello, line 136: `getBool` (109) and `getNumber` (138) use `|| null`. `false || null` and `0 || null` both return `null`, so stored `false`/`0` round-trip as `null`.

```ts
// 109
return this.mmkv.getBoolean(key) ?? null;
// 138
return this.mmkv.getNumber(key) ?? null;
```

Tests:

```ts
describe('UserPreferences', () => {
  beforeEach(() => userPreferences.clearAll());

  it('getBool returns false for stored false', () => {
    userPreferences.setBool('k', false);
    expect(userPreferences.getBool('k')).toBe(false);
  });

  it('getBool returns null for unset key', () => {
    expect(userPreferences.getBool('missing')).toBeNull();
  });

  it('getNumber returns 0 for stored 0', () => {
    userPreferences.setNumber('k', 0);
    expect(userPreferences.getNumber('k')).toBe(0);
  });

  it('getNumber returns null for unset key', () => {
    expect(userPreferences.getNumber('missing')).toBeNull();
  });
});
```

Hit on real flows: `getUsersPresence.ts:141`, `RoomView/index.tsx:1363` (no fallback), `localAuthentication.ts:115`, `ScreenLockConfigView.tsx:128` (`?? default` masks the bug only while `default` equals stored `false`).

android/app/build.gradle
diegolmello, line 150: Did you open the task to switch back to Tencent's `mmkv-static`? Didn't find it.

diegolmello (commented):
Description is empty. Fill in `How to test`, screenshots, type of change. Test plan should cover upgrade from a v3 install: login, E2EE keys, SSL client certs, biometry, screen lock, server list.

ios/Podfile
```diff
@@ -45,6 +46,11 @@ pre_install do |installer|
         def pod.build_type;
         Pod::BuildType.static_library # >= 1.9
       end
+      if ['MMKV', 'MMKVCore', 'NitroMmkv', 'NitroModules'].include?(pod.name)
+        def pod.build_type;
+          Pod::BuildType.dynamic_framework
+        end
+      end
```
diegolmello: This block has never executed since it was introduced — line 48's `end` closes `def`, not `if`, so the parser sees this `if` as nested inside `if static`. MMKV/Nitro pods have built as static frameworks throughout this PR's history, and CI is green (iOS Release `Upload`, Android Experimental, 13 E2E iOS shards).

The v4 upgrade guide explicitly supports static linkage via `:modular_headers => true`, which the Podfile already does on lines 11-12. Delete:

```suggestion
```

android/app/build.gradle
Rohit3523: We have a task for this https://rocketchat.atlassian.net/browse/NATIVE-1149

app/lib/methods/userPreferences.ts
Rohit3523: done

app/lib/methods/userPreferences.ts
```diff
@@ -58,7 +54,7 @@ const getAppGroupPath = (): string => {
 	}
 };
 
-const MMKV_INSTANCE = new MMKV(buildConfiguration());
+const MMKV_INSTANCE = createMMKV(buildConfiguration());
 
 export const useUserPreferences = <T>(key: string, defaultValue?: T): [T | undefined, (value: T | undefined) => void] => {
```
diegolmello, line 59: Did you do an evaluation of the impact of these `||` -> `??` changes?

diegolmello (approved):

app/lib/methods/userPreferences.ts
Rohit3523: Yes, done. getBool has 4 production callers, getNumber has 0. Two of the getBool callers (`localAuthentication.ts:115` and `ScreenLockConfigView.tsx:128`) are unchanged because their downstream ?? false defaults match the stored value either way. The other two (`getUsersPresence.ts:141` and `RoomView/index.tsx:1363`) start respecting an explicitly-stored false, which fixes two latent bugs where user-disabled settings were being ignored. No regressions

PR #6998, opened 2026-02-20
fix(Android): icons are not showing on older version

Rohit3523:
## Proposed changes

On Android 8, all icons are invisible. The custom icon font (custom.ttf) fails to load at runtime.

Root Cause
On these devices, when Expo resolves a bundled font asset, the URI comes back as a bare filename, just "custom.ttf" — with no URI scheme (file://, asset://, etc.). This breaks two Expo modules:

expo-asset — `downloadAsync` doesn't handle `uri.scheme == null`. It skips the early-return for file:// URIs and falls through to the download/caching logic, which fails on a raw filename that's already a local native asset.

expo-font — `loadAsync` only handles two URI shapes: asset://... and absolute file paths. A bare "custom.ttf" doesn't match asset://, so it falls into the file-path branch where Uri.parse("custom.ttf") resolves to a nonexistent filesystem path. Font loading fails silently → no icons render.

This does not happen on newer Android versions, where Expo resolves the asset to a proper asset:// or file:// URI.

Fix
Two patch-package patches:

expo-asset: Early-return when `uri.scheme == null`, preventing the broken cache/download attempt on raw asset references.
expo-font: Add a third case to loadAsync that detects bare filenames (no :, no leading /) and loads them via Typeface.createFromAsset(), trying both `fonts/<name>` and `<name>` paths in the APK assets directory.

Found this log in console
```
Error: Call to function 'ExpoAsset.downloadAsync' has been rejected.

→ Caused by: Unable to download asset from url: custom.ttf
```

## Issue(s)	

https://rocketchat.atlassian.net/browse/SUP-977

## Screenshots
| Before | After |
| --- | --- |
| Screenshot 2026-02-20 at 8 26 43 PM | Screenshot 2026-02-20 at 8 31 02 PM |
| Screenshot 2026-02-20 at 8 27 57 PM | Screenshot 2026-02-20 at 8 31 54 PM |
| Screenshot 2026-02-20 at 8 28 33 PM | Screenshot 2026-02-20 at 8 32 34 PM |
| Screenshot 2026-02-20 at 8 29 05 PM | Screenshot 2026-02-20 at 8 33 10 PM |
| Screenshot 2026-02-20 at 8 29 29 PM | Screenshot 2026-02-20 at 8 30 02 PM |

## Types of changes

## Checklist

## Further comments

I did the test on Android 8, 14, 16 and can confirm that icons are loading

## Summary by CodeRabbit

* **Bug Fixes**
  * Return early for URIs with no scheme on Android 8 (Oreo) and earlier to avoid incorrect handling.
  * Improve font loading on Android 8 (Oreo) and earlier by better resolving raw asset paths and registering fonts from matching asset candidates.


app/containers/CustomIcon/index.tsx
```diff
@@ -7,7 +7,7 @@ import { useTheme } from '../../theme';
 import { useResponsiveLayout } from '../../lib/hooks/useResponsiveLayout/useResponsiveLayout';
 import icoMoonConfig from './selection.json';
 
-export const IconSet = createIconSetFromIcoMoon(icoMoonConfig, 'custom', 'custom.ttf');
+export const IconSet = createIconSetFromIcoMoon(icoMoonConfig, 'custom', require('../../static/fonts/custom.ttf'));
```
diegolmello: We would be loading fonts only from JS bundle instead of native.
We should evaluate the performance impact of that.
It could be smaller than I expect, because I think Expo does the same, but we need to test.

Also, if you're changing the code to load from JS, you should remove the fonts from native.

diegolmello (changes requested):

Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

patches/expo-font+14.0.10.patch
```diff
@@ -0,0 +1,81 @@
+diff --git a/node_modules/expo-font/android/src/main/java/expo/modules/font/FontLoaderModule.kt b/node_modules/expo-font/android/src/main/java/expo/modules/font/FontLoaderModule.kt
+index 2cd51dc..fab6cc1 100644
+--- a/node_modules/expo-font/android/src/main/java/expo/modules/font/FontLoaderModule.kt
++++ b/node_modules/expo-font/android/src/main/java/expo/modules/font/FontLoaderModule.kt
+@@ -19,7 +19,7 @@ private class FileNotFoundException(uri: String) :
+ 
+ open class FontLoaderModule : Module() {
+   private val context: Context
+-    get() = appContext.reactContext ?: throw Exceptions.ReactContextLost()
++    get() = appContext.reactContext ?: throw Exceptions.AppContextLost()
```
diegolmello: Why changing the exception here?

patches/expo-font+14.0.10.patch
```diff
@@ -0,0 +1,81 @@
...
+-          localUri.substring(ASSET_SCHEME.length + 1)
+-        )
+-      } else {
+-        val file = Uri.parse(localUri).path?.let { File(it) }
+-          ?: throw FileNotFoundException(localUri)
+-
+-        if (file.length() == 0L) {
+-          throw CodedException(
+-            "Font file for $fontFamilyName is empty. Make sure the local file path is correctly populated."
++      val context = appContext.reactContext ?: throw Exceptions.AppContextLost()
```
diegolmello: same

diegolmello (dismissed):
In both cases, you should test Android version, so it doesn't change the behavior for newer working versions.

Also, you should open a PR back to Expo (without the Android version condition). There's a chance this is a fix for the libs and we can remove the patch in the future.

Rohit3523:
> In both cases, you should test Android version, so it doesn't change the behavior for newer working versions.

I did the test on Android 8, 14 and 16

> Also, you should open a PR back to Expo (without the Android version condition). There's a chance this is a fix for the libs and we can remove the patch in the future.

Okay

package.json
```diff
@@ -183,7 +183,7 @@
 		"babel-plugin-transform-remove-console": "^6.9.4",
 		"babel-preset-expo": "~54.0.9",
 		"eslint": "^8.57.0",
-		"eslint-config-prettier": "~10.1.8",
+		"eslint-config-prettier": "10.1.8",
```
OtavioStasiak, line 186: is it really necessary? if it's not please just rollback it and you can merge.

OtavioStasiak (approved):
LGTM

package.json
Rohit3523: This change is done to refresh the CI cache to make sure that patch is being applied

PR #7005, opened 2026-02-22
fix: save button always enabled in status view

### LINKED ISSUE #7004: bug: Save button is always enabled on Edit Status screen
### Describe the Bug

On the "Edit Status" screen, the "Save" button is supposed to be disabled until the user actually types a new message or selects a new status. 

However, due to a logic error in how the app checks for changes, the app gets confused and always thinks a change has been made. As a result, the "Save" button is permanently lit up and clickable the moment you open the screen, even if you haven't touched anything. 

### Steps to Reproduce

1. Open the app and go to **Profile**.
2. Tap on **Edit Status** (Message field + Online/Busy/Away/Offline options).
3. Do not change anything in the input field or the status selection.
4. Observe the **Save** button at the bottom of the screen.

### Expected Behavior

The Save button should be disabled when the current status and status text perfectly match what is already saved.

### Actual Behavior

The Save button is instantly tappable and enabled, running an unnecessary save operation when clicked.

8.2.0

### Rocket.Chat App Version

4.70.0.9999999999

### Device Name

Realme 12+

### OS Version

Android 15

### Additional Context

_No response_

Yaddalapalli-Charan-Kumar-Naidu:
## Proposed changes

Fixes a bug in `StatusView` where the **"Save"** button remained permanently enabled even when no changes were made to the user's status or status text.

### Root Cause

The previous `isStatusChanged()` implementation contained a logic flaw:

- An early-exit guard handled `!isValid`
- However, the final return expression also used `!isValid && ...`
- Since `!isValid` was already handled earlier, it always evaluated to `false` in the final condition
- As a result, `isStatusChanged()` consistently returned `false`
- This caused the "Save" button to remain enabled regardless of actual form state

### What was changed

**Replaced `isStatusChanged()` with `isSaveDisabled` (`useMemo`)**
   - Computes disabled state deterministically
   - Returns `true` when:
     - Selected status matches the persisted user status
     - Status text matches the persisted user status text
   - Ensures the button is only enabled when there is a real change

This makes the Save button behavior consistent, predictable, and aligned with expected UX.

## Issue(s)

Fixes #7004 

## How to test or reproduce

1. Navigate to **Profile → Edit Status**
2. **Expected:** The "Save" button is disabled when the screen opens.
3. Modify either:
   - The status text  
   - Or the selected status option  
4. **Expected:** The "Save" button becomes enabled immediately.
5. Revert the values back to their original state without saving.
6. **Expected:** The "Save" button becomes disabled again.
7. Run tests:

All tests pass successfully.

## Screenshots

| Before | After |
|--------|-------|
|  |  |
| Save button enabled on initial load | Save button disabled on initial load |
| Save button remains enabled after reverting changes | Save button correctly toggles based on change state |
## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Save button now reliably disables when status content is unchanged.

* **Performance**
  * Improved efficiency of status-change detection and footer rendering for snappier UI responsiveness.

* **Accessibility**
  * Enhanced status error announcements with clearer, structured feedback when submitting changes.


Rohit3523:
Thanks so much for taking the time to work on this PR! 🙌 We really appreciate your contribution. It looks like this issue has already been fixed in https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7392

Thanks again for jumping in and helping out! 🚀

PR #7111, opened 2026-04-07
fix: resolve deep links by room id for channels and groups

Rohit3523:
## Proposed changes

Deep links to a room sometimes carry a **room id** in the path segment instead of a room name (e.g. `rocketchat://room?host=...&path=group/6997e23f362b278aeb3d369b`).

The previous implementation passed that path segment straight to the REST `groups.info` / `channels.info` endpoint as `roomName`. Those endpoints only ever match a room **name** exactly, so whenever the link contained an id for a **group**, the lookup returned "not found" and the deep link failed to open the room.

This PR makes room resolution for both `group` and `channel` deep links use the `getRoomByTypeAndName` method, which resolves **both** a room name and a room id — with no code anymore tied to "name-only" REST params.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-1857

## How to test or reproduce

1. Open the app and log in.
2. Deep link to a group using its **name** → the group opens.
3. Deep link to the same group using its **id** in the path (`path=group/<id>`) → the group now opens (previously failed).
4. Repeat for a channel using its **name** and its **id** in `path=channel/<name|id>` → the channel opens in both cases.

## Screenshots

N/A — deep-link behavior change, no visual change.

## Types of changes

## Checklist

## Further comments

The previous implementation used two different REST endpoints (`channels.info` for channels, `groups.info` for groups) with a `roomName` param, so it silently missed id-based links. We consolidated resolution on the DDP `getRoomByTypeAndName`, which the server supports for both `c` and `p` room types, matching here also used by the in-app Directory. A regression test was added and the e2e coverage was adjusted to fixed reusable fixtures (old group fixtures no longer exist on the QA server).

## Summary by CodeRabbit

- **New Features**
  - Improved channel and group deeplinks so links can open rooms by name or ID, even when a room ID is not included.
  - Added support for opening groups through the room-opening flow.

- **Bug Fixes**
  - Improved handling of already-open rooms and invalid or unresolved deeplinks.
  - Preserved direct-message and explicit room-ID opening behavior.

- **Tests**
  - Added coverage for channel and group deeplinks, successful openings, failures, and unsupported paths.
  - Added end-to-end checks for opening channels and groups by name and ID.


Rohit3523:
> 🧹 Nitpick comments (1)
> > app/lib/methods/canOpenRoom.test.ts (1)> `26-35`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚖️ Poor tradeoff_
> > > **Add explicit return types to the new async test callbacks.**
> > > Use `Promise<void>` for each added async `it` callback. Apply the same rule to the other added callback functions where applicable.
> > > 
> > > Proposed change
> > > ```diff
> > > -	it('returns room when getRoomByTypeAndName succeeds and groups.open succeeds', async () => {
> > > +	it('returns room when getRoomByTypeAndName succeeds and groups.open succeeds', async (): Promise<void> => {
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > As per coding guidelines, “add explicit type annotations to function parameters and return types.”
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > 
> > > In `@app/lib/methods/canOpenRoom.test.ts` around lines 26 - 35, Add an explicit
> > > Promise<void> return type to the async callback in the canOpenRoom test case,
> > > and apply the same annotation to every other newly added async it/test callback
> > > in this diff. Keep the test behavior and assertions unchanged.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Coding guidelines_
> 
> ```
> 
> Nitpick comments:
> In `@app/lib/methods/canOpenRoom.test.ts`:
> - Around line 26-35: Add an explicit Promise<void> return type to the async
> callback in the canOpenRoom test case, and apply the same annotation to every
> other newly added async it/test callback in this diff. Keep the test behavior
> and assertions unchanged.
> ```
> 
> ℹ️ Review info
> ⚙️ Run configuration
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro Plus
> 
> **Run ID**: `5df86ac2-5784-493b-9aae-220d7917cf72`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [41e87a8](https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/41e87a85bc6746c58e20967a2c94add6610a8cae) and [2d1a8f0](https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/2d1a8f0f57bcc49b4761c14c7ce1c09d449efd5f).
> 
> 📒 Files selected for processing (4)
> * `.maestro/tests/assorted/group-deeplink.yaml`
> * `app/definitions/rest/v1/groups.ts`
> * `app/lib/methods/canOpenRoom.test.ts`
> * `app/lib/methods/canOpenRoom.ts`
> 
> 🚧 Files skipped from review as they are similar to previous changes (1)
> * .maestro/tests/assorted/group-deeplink.yaml
> 
> 📜 Review details

Adding Promise<void> to every it() callback in canOpenRoom.test.ts would deviate from the surrounding test suite and add noise with no lint backing.

.maestro/tests/assorted/channel-deeplink.yaml
```diff
@@ -0,0 +1,57 @@
...
+---
+- runFlow:
+    file: '../../helpers/login-with-deeplink.yaml'
+    env:
+      USERNAME: ${output.account.adminUser}
+      PASSWORD: ${output.account.adminPassword}
+
+# open group using name
+- killApp
+- runFlow:
```
OtavioStasiak, line 23: The flow hardcodes four things that data-setup.js never creates: the room names discussion-with-name / discussion-with-id and the ids 6a74cb92d09f52225dd8f0af / 8Lhm98suz3bGKMuK2. They only resolve because someone created them by
  hand on mobile.qa.rocket.chat — the group id is a 24-char Mongo ObjectId while RC room ids are 17-char randoms (8Lhm98suz3bGKMuK2), which gives it away. When that server is reset or those rooms are renamed, this flow fails and the
  failure will read as a product regression in deeplinking rather than missing test data.

  data-setup.js already exposes exactly what this needs — createRandomRoom(username, password, type) returns { _id, name } — so both the name case and the id case can provision themselves, following the pattern in
  tests/room/discussion.yaml:

  onFlowStart:
    - runFlow: '../../helpers/setup.yaml'
  onFlowComplete:
    - evalScript: ${output.utils.deleteCreatedUsers()}
  ---
  - evalScript: ${output.user = output.utils.createUser()}
  - evalScript: ${output.group = output.utils.createRandomRoom(output.user.username, output.user.password, 'p')}
  - evalScript: ${output.channel = output.utils.createRandomRoom(output.user.username, output.user.password, 'c')}

  then path=group/${output.group.name} and path=group/${output.group._id}, asserting room-view-title-${output.group.name} in both cases. Same for the channel. That also drops the dependency on the shared admin account, which the four
  killApp + deeplink steps currently mutate state for.

  Two cases that are closer to the bug this PR fixes and aren't covered:

  1. A private group with a closed subscription. All four rooms are already open for the admin, so groups.open in canOpenRoom never actually does anything here — the one branch the fix reorders (resolve room → then open, instead of
  open → then fetch info) is untested end-to-end. output.utils.post('rooms.hide', user, pass, { roomId }) before the deeplink makes it real.
  2. A public channel the user is not a member of — the new code resolves it via canAccessRoomAsync rather than channels.info + view-c-room, and that's a genuine behavior difference worth pinning.

  Non-blocking, but (1) is the one I'd push for, since it's the branch the fix actually restructures.

OtavioStasiak (changes requested):
lint is failing after update with develop...
Fix it and run e2e tests, all tests must pass.

app/lib/methods/canOpenRoom.ts
```diff
@@ -26,29 +19,33 @@ async function open({ type, rid, name }: { type: ERoomTypes; rid: string; name:
...
-			try {
-				// RC 0.61.0
-				// @ts-ignore
-				await sdk.post(`${restTypes[type]}.open`, params);
-			} catch (e: any) {
-				if (!(e.data && /is already open/.test(e.data.error))) {
-					return false;
+		if (type === ERoomTypes.CHANNEL || type === ERoomTypes.GROUP) {
+			// The path segment of a deep link may hold either a room name or a room id.
+			// getRoomByTypeAndName resolves both, unlike the REST endpoints, which match
```
OtavioStasiak, line 27: we can improve the code organization like this...

```
if (type === ERoomTypes.CHANNEL || type === ERoomTypes.GROUP) {
	let roomId = rid;

	if (!roomId) {
		const roomType = type === ERoomTypes.GROUP ? 'p' : 'c';
		const room = await getRoomByTypeAndName(roomType, name);
		roomId = room?._id;
	}

	if (!roomId) {
		return false;
	}
```

OtavioStasiak (changes requested):

OtavioStasiak (approved):
LGTM

PR #7157, opened 2026-04-15
chore: enable edge to edge on android

Rohit3523:
## Proposed changes
Replace `expo-navigation-bar` with `@zoontek/react-native-navigation-bar` and add safe area bottom insets across 30+ views to enable proper edge-to-edge display on Android. This prevents content from being hidden behind the system navigation bar on devices with gesture navigation.

Key changes:
- Replaced `expo-navigation-bar` with `@zoontek/react-native-navigation-bar` for better edge-to-edge support
- Added `paddingBottom: bottom` (via `useSafeAreaInsets` / `withSafeAreaInsets`) to FlatLists, ScrollViews, and content containers across the app
- Cleaned up hardcoded `paddingBottom: 100` in `NewServerView` in favor of `FormContainer`s built-in safe area handling
- Switched `JitsiMeetView` from RNs built-in `SafeAreaView` to `react-native-safe-area-context`s
- Fixed `RoomView` footer sections to respect bottom safe area

## Issue(s)
https://rocketchat.atlassian.net/browse/CORE-1967

## How to test or reproduce
1. Build and run on Android 12+ with gesture navigation enabled
2. Verify these screens show content above the system navigation bar:
   - MessagesView / ThreadMessagesView / SearchMessagesView
   - RoomView (preview mode, read-only, blocked room footers)
   - RoomActionsView / RoomInfoEditView / RoomMembersView
   - DiscussionsView / TeamChannelsView / PickerView
   - NewMessageView / SelectedUsersView
   - SettingsView / LanguageView / StatusView / ProfileView
   - E2EEncryptionSecurityView / E2ESaveYourPasswordView
   - SelectServerView / SelectListView
   - ShareListView
3. Verify the "Jump to last message" FAB in rooms respects bottom safe area
4. Verify the "Add Server" screen (both first-run and from rooms list header) shows correctly without extra header spacing

## Screenshots
| Screen | Android 12 | Android 13 | Android 14 | Android 15 | Android 16 | Android 17 | iOS |
|--------|--------|--------|--------|--------|--------|--------|--------|
| Login |  |  |  |  |  |  |  |
| Register |  |  |  |  |  |  |  |
| Rooms List |  |  |  |  |  |  |  |
| Select Server |  |  |  |  |  |  |  |
| Room View Default |  |  |  |  |  |  |  |
| Room Keyboard |  |  |  |  |  |  |  |
| Room Emoji Keyboard |  |  |  |  |  |  |  |
| Emoji Picker Bottom |  |  |  |  |  |  |  |
| Thread Messages |  |  |  |  |  |  |  |
| Search Messages |  |  |  |  |  |  |  |
| Room Members |  |  |  |  |  |  |  |
| New Message View |  |  |  |  |  |  |  |
| Directory |  |  |  |  |  |  |  |
| Discussions |  |  |  |  |  |  |  |
| Read Only Room |  |  |  |  |  |  |  |
| Status View |  |  |  |  |  |  |  |
| Language |  |  |  |  |  |  |  |
| Default Browser |  |  |  |  |  |  |  |
| Auto Translate |  |  |  |  |  |  |  |
| Room Action Sheet |  |  |  |  |  |  |  |
| Action Sheet Delete |  |  |  |  |  |  |  |
| Media Download Images |  |  |  |  |  |  |  |
| Members Filter |  |  |  |  |  |  |  |
| Profile |  |  |  |  |  |  |  |
| Reaction Picker |  |  |  |  |  |  |  |
| Reaction Picker Bottom |  |  |  |  |  |  |  |

### Jisti
| OS | Call Screen | Members | Chat | 
| --- | --- | --- | --- |
| Android | Screenshot 2026-06-29 at 11 55 44 PM | Screenshot 2026-06-29 at 11 57 56 PM | Screenshot 2026-06-29 at 11 58 09 PM |
| iOS | Screenshot 2026-06-30 at 12 02 21 AM | Screenshot 2026-06-30 at 12 02 33 AM | Screenshot 2026-06-30 at 12 02 58 AM |

### iPhone SE Bottom sheet screenshots
| Screen | iPhone SE |
| --- | --- |
| Add new server | Screenshot 2026-06-30 at 3 58 06 AM |
| Directory Filter | Screenshot 2026-06-30 at 3 59 02 AM |
| Member filter | Screenshot 2026-06-30 at 4 00 13 AM | 
| Delete button visible in sheet | Screenshot 2026-06-30 at 4 01 57 AM |
| Last row of emoji is visible | Screenshot 2026-06-30 at 4 03 08 AM |

Note: I tested this on both a tablet and in landscape mode, and everything is working as expected.

## Summary by CodeRabbit

* **New Features**
  * Enabled edge-to-edge display on Android for a more immersive app layout.
  * Improved safe-area handling across lists, sheets, chats, profiles, and message screens to better fit devices with gesture bars or notches.
  * Updated keyboard and bottom-sheet behavior for more consistent spacing on both iOS and Android.

* **Bug Fixes**
  * Prevented content from being hidden behind the bottom system area in several views.
  * Improved Android navigation bar appearance and contrast handling.


Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

.github/workflows/e2e-build-android.yml
```diff
@@ -26,6 +26,8 @@ jobs:
     steps:
       - name: Checkout Repository
         uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+        with:
+          persist-credentials: false
```
Rohit3523: This change was suggested by coderabbit

android/app/src/main/res/values/styles.xml
```diff
@@ -3,6 +3,7 @@
         <item name="android:colorEdgeEffect">#aaaaaa</item>
         <item name="android:forceDarkAllowed">false</item>
         <item name="android:editTextBackground">@drawable/rn_edit_text_material</item> 
+        <item name="android:enforceNavigationBarContrast">false</item>
```
Rohit3523, line 6: This makes the navigation bar transparent. Setting it to true makes the navigation bar partially transparent, resulting in a color mismatch.

android/app/src/main/res/values/styles.xml
```diff
@@ -11,6 +12,7 @@
         <item name="postBootSplashTheme">@style/AppTheme</item>
         <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="android:navigationBarColor">@android:color/transparent</item>
+         <item name="android:enforceNavigationBarContrast">false</item>
```
Rohit3523: Same as above, making splashscreen navigation bar transparent 

| Before | After |
| --- | --- |
| Image | Image |

app/containers/ActionSheet/ActionSheet.tsx
```diff
@@ -34,10 +33,6 @@ const ActionSheet = memo(
 		const [contentHeight, setContentHeight] = useState(0);
 		const onCloseSnapshotRef = useRef<TActionSheetOptions['onClose']>(undefined);
 
-		// TrueSheet detects the bottom inset for Android 16 and iOS
-		// To avoid content hiding behind navigation bar on older Android versions
-		const isNewAndroid = isAndroid && Number(Platform.Version) >= 36;
-		const bottom = isIOS || isNewAndroid ? 0 : windowHeight * 0.03;
```
Rohit3523, line 41: truesheet handles the bottom insets by it's own when we have edgeToEdge enabled, so it's safe to remove it

app/containers/ActionSheet/ActionSheet.tsx
```diff
@@ -118,7 +112,7 @@ const ActionSheet = memo(
 
 		const hasOptions = !!data?.options?.length;
 		const hasSnaps = !!effectiveSnaps?.length;
-		const disableContentPanning = data?.enableContentPanningGesture === false || (!scrollEnabled && isAndroid);
```
Rohit3523, line 122: This part of the code was added for sheets with a small number of options. Without it, tapping or scrolling the sheet causes its height to shrink.

https://github.com/user-attachments/assets/a1a44fd9-1a38-4fca-806f-962fffe5e95f

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -48,17 +46,14 @@ export function useActionSheetDetents({
 	contentHeight
 }: UseActionSheetDetentsParams): { detents: SheetDetent[]; maxHeight: number; scrollEnabled: boolean } {
 	const { fontScale } = useWindowDimensions();
-	const CANCEL_HEIGHT = 48 * fontScale;
 
 	return useMemo(() => {
+		const CANCEL_HEIGHT = 48 * fontScale;
```
Rohit3523: It was CodeRabbit's suggestion to move it inside useMemo so the value stays memoized.

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -48,17 +46,14 @@ export function useActionSheetDetents({
...
 		const maxHeight = windowHeight * ACTION_SHEET_MAX_HEIGHT_FRACTION;
 		const hasOptions = optionsLength > 0;
 
 		const maxSnap = hasOptions
-			? Math.min(
-					(itemHeight + 0.5) * optionsLength + HANDLE_HEIGHT + headerHeight + bottomInset + (hasCancel ? CANCEL_HEIGHT : 0),
-					maxHeight
-			  )
+			? Math.min((itemHeight + 0.5) * optionsLength + HANDLE_HEIGHT + headerHeight + (hasCancel ? CANCEL_HEIGHT : 0), maxHeight)
```
Rohit3523: removed `bottomInset` from calculation, as it is handled by TrueSheet now

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -71,14 +66,13 @@ export function useActionSheetDetents({
 				detents = [0.5, ACTION_SHEET_MAX_HEIGHT_FRACTION];
 				scrollEnabled = true;
 			} else {
-				const measuredHeight =
-					optionsLength * itemHeight + HANDLE_HEIGHT + headerHeight + bottomInset + (hasCancel ? CANCEL_HEIGHT : 0);
+				const measuredHeight = optionsLength * itemHeight + HANDLE_HEIGHT + headerHeight + (hasCancel ? CANCEL_HEIGHT : 0);
```
Rohit3523: removed `bottomInset`

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -55,15 +54,14 @@ const EmojiCategory = ({
...
-	const { bottom } = useSafeAreaInsets();
 
 	if (!parentWidth) {
 		return null;
 	}
 
 	const numColumns = Math.trunc(parentWidth / EMOJI_BUTTON_SIZE);
 	const marginHorizontal = (parentWidth % EMOJI_BUTTON_SIZE) / 2;
-	const contentPaddingBottom = bottomSheet ? Math.max(0, MIN_BOTTOM_SHEET_BREATHING_ROOM - bottom) : undefined;
+	const contentPaddingBottom = bottomSheet ? Math.max(0, MIN_BOTTOM_SHEET_BREATHING_ROOM) : undefined;
```
Rohit3523: Removed bottom because it was making emoji hiding behind nav bar

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -55,15 +54,14 @@ const EmojiCategory = ({
...
 	const items = useEmojis(category);
-	const { bottom } = useSafeAreaInsets();
 
 	if (!parentWidth) {
 		return null;
 	}
 
 	const numColumns = Math.trunc(parentWidth / EMOJI_BUTTON_SIZE);
 	const marginHorizontal = (parentWidth % EMOJI_BUTTON_SIZE) / 2;
-	const contentPaddingBottom = bottomSheet ? Math.max(0, MIN_BOTTOM_SHEET_BREATHING_ROOM - bottom) : undefined;
```
Rohit3523, line 66: https://github.com/user-attachments/assets/8dd21f94-16a2-4361-be59-dd96a4406968

app/containers/FormContainer.tsx
```diff
@@ -27,19 +28,20 @@ export const FormContainerInner = ({
 }: {
 	children: (ReactElement | null)[];
 	accessibilityLabel?: string;
-}) => (
+}): ReactElement => (
```
Rohit3523, line 31: Suggested by coderabbit to add return type

app/containers/FormContainer.tsx
```diff
@@ -27,19 +28,20 @@ export const FormContainerInner = ({
...
 	<View accessibilityLabel={accessibilityLabel} style={[sharedStyles.container, isTablet && sharedStyles.tabletScreenContent]}>
 		{children}
 	</View>
 );
 
-const FormContainer = ({ children, testID, showAppVersion = true, ...props }: IFormContainer) => {
+const FormContainer = ({ children, testID, showAppVersion = true, ...props }: IFormContainer): ReactElement => {
```
Rohit3523, line 37: Suggested by coderabbit to add return type

app/containers/MessageComposer/hooks/useEmojiKeyboard.tsx
```diff
@@ -79,9 +78,8 @@ export const useEmojiKeyboard = () => {
 
 	const { bottom } = useSafeAreaInsets();
 	const { height } = useKeyboardAnimation();
-	const initialHeight = Platform.OS === 'ios' ? bottom : 0;
-	const keyboardHeight = useSharedValue(initialHeight);
-	const previousHeight = useSharedValue(initialHeight);
+	const keyboardHeight = useSharedValue(bottom);
+	const previousHeight = useSharedValue(bottom);
```
Rohit3523, line 82: Using `bottom` insets in both OS now to prevent message composer to hide behind navigation bar

Before on android, safe area was added by default but iOS was edge to edge

app/containers/FormContainer.tsx
```diff
@@ -17,7 +18,7 @@ interface IFormContainer extends ScrollViewProps {
 
 const styles = StyleSheet.create({
 	scrollView: {
-		minHeight: '100%'
+		flexGrow: 1
```
Rohit3523, line 21: Replaced `minHeight` with `flexGrow` because `minHeight` triggered a layout height change after mount in NewServerScreen,

| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/5027bc56-016c-49f6-9cc6-951fea9322e5" /> | <video src="https://github.com/user-attachments/assets/68bdd95d-b6b8-412a-8064-71008661d898" /> |

app/lib/methods/helpers/theme.ts
```diff
@@ -1,6 +1,6 @@
 import { Appearance } from 'react-native';
-import * as NavigationBar from 'expo-navigation-bar';
 import * as SystemUI from 'expo-system-ui';
+import { NavigationBar } from '@zoontek/react-native-navigation-bar';
```
Rohit3523, line 3: `expo-navigation-bar` doesn't change the navigation bar button color as it needs the semi transparent navigation bar. Switched to `react-native-navigation-bar`, which is backed by Expo and correctly updates the button style.

app/views/AddExistingChannelView/index.tsx
```diff
@@ -175,7 +177,7 @@ const AddExistingChannelView = () => {
 					);
 				}}
 				ItemSeparatorComponent={List.Separator}
-				contentContainerStyle={{ backgroundColor: colors.surfaceRoom }}
+				contentContainerStyle={{ backgroundColor: colors.surfaceRoom, paddingBottom: bottom }}
```
Rohit3523, line 180: To prevent last item from hiding behind nav bar

app/views/E2EEncryptionSecurityView/index.tsx
```diff
@@ -50,7 +52,7 @@ const E2EEncryptionSecurityView = () => {
 	};
 
 	return (
-		<SafeAreaView testID='e2e-encryption-security-view' style={{ backgroundColor: colors.surfaceRoom }}>
+		<SafeAreaView testID='e2e-encryption-security-view' style={{ backgroundColor: colors.surfaceRoom, paddingBottom: bottom }}>
```
Rohit3523: Added because button was hiding behind navigation bar and impossible to click, also i have not added 'bottom' in `SafeAreaView` component directly because on iOS it was adding safe bottom padding

app/views/JitsiMeetView/index.tsx
```diff
@@ -2,8 +2,9 @@ import CookieManager from '@react-native-cookies/cookies';
 import { type RouteProp, useNavigation, useRoute } from '@react-navigation/native';
 import { activateKeepAwake, deactivateKeepAwake } from 'expo-keep-awake';
 import { useCallback, useEffect, useState, type ReactElement } from 'react';
-import { ActivityIndicator, Linking, SafeAreaView, StyleSheet, View } from 'react-native';
```
Rohit3523, line 5: `SafeAreaView` from `react-native` is deprecated

app/views/NewServerView/index.tsx
```diff
@@ -53,7 +53,6 @@ const NewServerView = () => {
 	const { submit } = useConnectServer({ workspaceUrl, certificate, previousServer });
 	const phoneMarginTop = previousServer ? 32 : 84;
 	const marginTop = isTablet ? 0 : phoneMarginTop;
-	const formContainerStyle = previousServer ? { paddingBottom: 100 } : {};
```
Rohit3523, line 56: I have verified and there is no UI difference

app/views/ReadReceiptView/index.tsx
```diff
@@ -169,4 +172,6 @@ const mapStateToProps = (state: IApplicationState) => ({
 	Message_TimeAndDateFormat: state.settings.Message_TimeAndDateFormat as string
 });
 
-export default connect(mapStateToProps)(withTheme(ReadReceiptView));
+const ReadReceiptViewWithInsets = withSafeAreaInsets(ReadReceiptView);
+hoistNonReactStatics(ReadReceiptViewWithInsets, ReadReceiptView);
+export default connect(mapStateToProps)(withTheme(ReadReceiptViewWithInsets));
```
Rohit3523: Added `hoistNonReactStatics` because `withSafeAreaInsets` was removing the static navigation options, causing the header title to disappear.

app/views/RoomActionsView/index.tsx
```diff
@@ -1344,4 +1349,6 @@ const mapStateToProps = (state: IApplicationState, ownProps: Partial<Pick<IRoomA
 	};
 };
 
-export default connect(mapStateToProps)(withTheme(withActionSheet(withDimensions(withMasterDetail(RoomActionsView)))));
+const RoomActionsViewWithInsets = withSafeAreaInsets(RoomActionsView);
+hoistNonReactStatics(RoomActionsViewWithInsets, RoomActionsView);
+export default connect(mapStateToProps)(withTheme(withActionSheet(withDimensions(withMasterDetail(RoomActionsViewWithInsets)))));
```
Rohit3523: Added `hoistNonReactStatics` because `withSafeAreaInsets` was removing the static navigation options, causing the header title to disappear.

app/views/RoomActionsView/styles.ts
```diff
@@ -4,6 +4,9 @@ import { PADDING_HORIZONTAL } from '../../containers/List/constants';
 import sharedStyles from '../Styles';
 
 export default StyleSheet.create({
+	scrollContentContainer: {
+		paddingTop: 16
+	},
```
Rohit3523: For some reason, the room action view lost the top padding, so added padding top to get the old UI

app/views/SearchMessagesView/index.tsx
```diff
@@ -357,4 +361,6 @@ const mapStateToProps = (state: any) => ({
 	customEmojis: state.customEmojis
 });
 
-export default connect(mapStateToProps)(withTheme(withMasterDetail(SearchMessagesView)));
+const SearchMessagesViewWithInsets = withSafeAreaInsets(SearchMessagesView);
+hoistNonReactStatics(SearchMessagesViewWithInsets, SearchMessagesView);
+export default connect(mapStateToProps)(withTheme(withMasterDetail(SearchMessagesViewWithInsets)));
```
Rohit3523: Added `hoistNonReactStatics` because `withSafeAreaInsets` was removing the static navigation options, causing the header title to disappear.

Rohit3523:
@coderabbitai review

app/views/MessagesView/index.tsx
```diff
@@ -386,4 +389,4 @@ const mapStateToProps = (state: IApplicationState) => ({
 	useRealName: state.settings.UI_Use_Real_Name
 });
 
-export default connect(mapStateToProps)(withTheme(withActionSheet(withMasterDetail(MessagesView))));
+export default connect(mapStateToProps)(withTheme(withActionSheet(withMasterDetail(withSafeAreaInsets(MessagesView)))));
```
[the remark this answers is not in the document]
Rohit3523: [app/views/MessagesView/index.tsx](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7157/files/311a2d758035907e8adac5a5816b05bcac62144b#diff-3c6fbdd9484eef2ec4454eff0a3bf69168d01f8277ed9a7f58db10cc64da043b) has no static keyword

.github/workflows/e2e-build-android.yml
diegolmello: looks unrelated to the pr. what does it do?

app/containers/FormContainer.tsx
diegolmello: I understand why you did it, but you don't have to accept everything CodeRabbit says. This is unrelated to the PR.

Keep it though. Not worth wasting time.

app/views/E2EEncryptionSecurityView/index.tsx
diegolmello: It's weird having to add a padding bottom from the SafeArea hook to the SafeArea view, no? You shouldn't have to.

app/views/JitsiMeetView/index.tsx
```diff
@@ -2,8 +2,9 @@ import CookieManager from '@react-native-cookies/cookies';
...
+import { ActivityIndicator, Linking, StyleSheet, View } from 'react-native';
 import WebView, { type WebViewNavigation } from 'react-native-webview';
+import { SafeAreaView } from 'react-native-safe-area-context';
```
diegolmello:
```suggestion
import { SafeAreaView } from '../../containers/SafeArea';
```

app/views/ReadReceiptView/index.tsx
diegolmello: You're having to repeat yourself on other places like RoomActionsView. I'd patch `withSafeAreaInsets` in the lib instead. It should be simple to add `hoistNonReactStatics` there.

app/views/RoomActionsView/index.tsx
```diff
@@ -84,6 +86,7 @@ interface IOnPressTouch {
 
 interface IRoomActionsViewProps extends IActionSheetProvider, IBaseScreen<StackType, 'RoomActionsView'> {
 	userId: string;
+	insets: { bottom: number; top: number; left: number; right: number };
```
diegolmello:
```suggestion
	insets: EdgeInsets;
```

app/views/RoomActionsView/styles.ts
diegolmello: "For some reason" always gets me 🤔 or 🙈

app/views/SettingsView/index.tsx
```diff
@@ -136,7 +138,7 @@ const SettingsView = (): ReactElement => {
 
 	return (
 		<SafeAreaView testID='settings-view'>
-			<List.Container>
+			<List.Container contentContainerStyle={{ paddingBottom: bottom }}>
```
diegolmello: We should evaluate making this change inside `List.Container`. The component should know how to render itself.

package.json
```diff
@@ -52,6 +52,7 @@
 		"@rocket.chat/mobile-crypto": "RocketChat/rocket.chat-mobile-crypto#main",
 		"@rocket.chat/sdk": "RocketChat/Rocket.Chat.js.SDK#mobile",
 		"@rocket.chat/ui-kit": "^0.39.0",
+		"@zoontek/react-native-navigation-bar": "^1.1.1",
```
diegolmello, line 55: It's no-op on iOS, right? So no podfile changes? Same for the removal of `expo-navigation-bar`, right?

diegolmello (commented):
Almost there!
Nice job with the screenshots.

Just to be sure: we should test on landscape and on tablets.

app/views/JitsiMeetView/index.tsx
diegolmello: `SafeAreaView` is now imported from `react-native-safe-area-context` but rendered with no `edges` prop (usage at line ~108). Unlike `react-native`'s `SafeAreaView` (iOS-only, effectively a no-op on Android), this one defaults to insetting **all four edges** on both platforms. On Android edge-to-edge the full-bleed Jitsi `WebView` now gets top + bottom + left + right safe-area padding, shrinking the call UI away from every screen edge. Pass explicit edges, e.g. `<SafeAreaView style={styles.container} edges={[]}>`.

app/views/SettingsView/index.tsx
```diff
@@ -136,7 +138,7 @@ const SettingsView = (): ReactElement => {
```
diegolmello, line 136: `List.Container` spreads `{...props}` **after** its own `contentContainerStyle={styles.container}` (`paddingVertical: 16`) in `ListContainer.tsx`, so passing `contentContainerStyle` here **replaces** the default instead of merging. The list loses `paddingTop: 16` (first row touches the header) and the 16px bottom fallback (now `bottom`, which is 0 on hardware-button devices / landscape / tablet master-detail).

Quick fix: `contentContainerStyle={{ paddingTop: 16, paddingBottom: bottom }}`.

Better structural fix (kills the whole bug class): have `ListContainer` merge its default — `contentContainerStyle={[styles.container, props.contentContainerStyle]}` — so callers can't silently nuke it. `RoomActionsView` already works around this same trap by re-adding `paddingTop`; the merge would let it drop that workaround.

app/containers/FormContainer.tsx
```diff
@@ -27,19 +28,20 @@ export const FormContainerInner = ({
...
-const FormContainer = ({ children, testID, showAppVersion = true, ...props }: IFormContainer) => {
+const FormContainer = ({ children, testID, showAppVersion = true, ...props }: IFormContainer): ReactElement => {
 	const { colors } = useTheme();
+	const { bottom } = useSafeAreaInsets();
 
 	return (
 		<KeyboardAwareScrollView
 			style={[sharedStyles.container, { backgroundColor: colors.surfaceRoom }]}
-			contentContainerStyle={[sharedStyles.containerScrollView, styles.scrollView]}
+			contentContainerStyle={[sharedStyles.containerScrollView, styles.scrollView, { paddingBottom: bottom }]}
```
diegolmello: `{ paddingBottom: bottom }` is the last element of the array, so it **overrides** `sharedStyles.containerScrollView`'s `paddingBottom: 24` (RN flattens the array, last key wins). Wherever `bottom < 24` (gesture-nav ~16–24; hardware-button devices / landscape = 0) every `FormContainer` screen (NewServerView, ForgotPasswordView, WorkspaceView, SendEmailConfirmationView, …) loses bottom clearance. Consider `paddingBottom: Math.max(24, bottom)`.

app/containers/MessageComposer/hooks/useEmojiKeyboard.tsx
```diff
@@ -79,9 +78,8 @@ export const useEmojiKeyboard = () => {
 
 	const { bottom } = useSafeAreaInsets();
 	const { height } = useKeyboardAnimation();
-	const initialHeight = Platform.OS === 'ios' ? bottom : 0;
-	const keyboardHeight = useSharedValue(initialHeight);
-	const previousHeight = useSharedValue(initialHeight);
+	const keyboardHeight = useSharedValue(bottom);
```
diegolmello, line 81: This removes the previous `Platform.OS` gating: the shared value(s) now seed to `bottom` on **all** platforms (was `0` on Android), and the notch at line ~96 (`height.value === 0 ? bottom : 0`) now adds the inset on Android too (was iOS-only). On Android this gives (a) a permanent ~`bottom`px spacer below the composer at mount before any keyboard interaction, and (b) the same spacer persisting after the keyboard is dismissed — and react-native-keyboard-controller already includes the system inset in its reported Android keyboard height, so the notch double-counts. Also `useAnimatedReaction` deps omit `bottom`, so a late inset resolution (0→N) isn't picked up. Worth re-checking Android composer spacing with the keyboard closed.

diegolmello (commented):
## Code review — Edge-to-edge Android (#7157)

Consolidated from two independent review passes. Inline comments below.

**Correctness (worth fixing before merge):**
1. JitsiMeetView — `SafeAreaView` from `react-native-safe-area-context` with no `edges` insets the WebView on all four sides on Android.
2. SettingsView (+ a `ListContainer` merge fix) — `contentContainerStyle` override drops the default `paddingVertical: 16`.
3. FormContainer — `{ paddingBottom: bottom }` overrides the base `paddingBottom: 24` across all form screens.
4. useEmojiKeyboard — Android loses its `Platform` gating → composer spacer + double-counted notch.
5. useActionSheetDetents — dropped `HANDLE_HEIGHT` can clip the last row of custom-content sheets.
6. EmojiCategory — inset double-counted against the parent's `marginBottom`.
7. E2EEncryptionSecurityView — padding on the wrong element (outer frame, not scroll content).
8. BottomSheetContent — Android lost the Cancel-row bottom padding.

**Confirm intent / device-check:** ActionSheet content-panning, NavBottomFAB offset, NewServerView cert-picker clearance.

**Inset-dependent — only bites where `insets.bottom` is 0 (hardware buttons, landscape, tablet master-detail):** RoomActionsView, and the LanguageView / AutoTranslate / DefaultBrowser / SelectServer / SelectList FlatLists.

**Minor:** RoomActionsView `insets` should use the `EdgeInsets` type.

Note: CI sets `edgeToEdgeEnabled=true` both in `android/gradle.properties` and by echoing it in the two build workflows — two sources of truth, so a future toggle in `gradle.properties` won't take effect in CI.

Verified-correct, no action: the added `hoistNonReactStatics` calls, the zoontek `setBarStyle` API, the RoomView footer change (`paddingBottom` is additive to the existing `marginBottom`), and `FormContainer`'s `flexGrow` change.

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -71,14 +66,13 @@ export function useActionSheetDetents({
...
 
 				scrollEnabled = false;
 				detents = [heightToDetent(Math.round(measuredHeight), windowHeight)];
 			}
 		} else if (contentHeight > 0) {
-			const rawContentDetent = (contentHeight + bottomInset + HANDLE_HEIGHT) / windowHeight;
+			const rawContentDetent = contentHeight / windowHeight;
```
diegolmello: The content-height detent dropped `HANDLE_HEIGHT` (now `contentHeight / windowHeight`). TrueSheet's native detent calculator only adds the grabber height for the `auto` (-1) detent, not for numeric fractions — so for **custom-children** sheets (e.g. the emoji picker) the sheet ends up ~`HANDLE_HEIGHT` (28px) shorter than the measured content and the last row can clip on low/zero bottom-inset devices. Dropping the double-counted `bottomInset` looks right; dropping `HANDLE_HEIGHT` does not. Please verify the tallest custom-content sheets show their last row.

app/containers/EmojiPicker/EmojiCategory.tsx
diegolmello: Two things here: (1) this no longer subtracts `bottom`, but the parent `EmojiPicker` container still applies `marginBottom: bottom` — so in bottom-sheet mode on a nav-bar device the inset is counted twice (~68px breathing room + ~48px margin ≈ 116px of empty space below the last emoji row). (2) `Math.max(0, MIN_BOTTOM_SHEET_BREATHING_ROOM)` is dead — the constant is always positive, so it's just the constant. Restore the `- bottom` subtraction, or drop the parent's `marginBottom`.

app/views/E2EEncryptionSecurityView/index.tsx
diegolmello: Unlike the other views in this PR, `paddingBottom: bottom` here is on the **outer `SafeAreaView` frame**, not on the inner `List.Container` scroll content. That shrinks the frame by `bottom` rather than adding scroll clearance, so if the content scrolls the last item is still clipped behind the nav bar — the padding doesn't reach what it's meant to clear. Move it to `List.Container`'s `contentContainerStyle`, consistent with the rest of the PR.

app/containers/ActionSheet/BottomSheetContent.tsx
```diff
@@ -68,7 +67,7 @@ const BottomSheetContent = memo(
 					style={{ backgroundColor: colors.strokeExtraDark }}
 					keyboardDismissMode='interactive'
 					indicatorStyle='black'
-					contentContainerStyle={{ paddingBottom, backgroundColor: colors.surfaceLight }}
+					contentContainerStyle={{ paddingBottom: bottom, backgroundColor: colors.surfaceLight }}
```
diegolmello, line 70: This dropped the Android-only extra bottom padding — it used to be `isAndroid ? bottom + height : bottom`, where `height = 48 * fontScale` (the Cancel/footer row). On Android edge-to-edge with a small/zero bottom inset the Cancel row now sits flush against the nav-bar gesture area (worse at large font scale). `AttachmentActionSheet.tsx` has a comment documenting that TrueSheet on Android needs padding beyond the inset alone — same rationale applies here.

app/containers/ActionSheet/ActionSheet.tsx
```diff
@@ -118,7 +112,7 @@ const ActionSheet = memo(
...
+		const disableContentPanning = data?.enableContentPanningGesture === false;
```
diegolmello, line 116: This dropped `|| (!scrollEnabled && isAndroid)` from `disableContentPanning`, so small non-scrollable Android sheets that don't set `enableContentPanningGesture` now have `draggable=true` (content-pannable). The removed clause was an intentional Android workaround. If this is deliberate given the edge-to-edge inset changes, fine — but please confirm the drag-to-dismiss feel on small Android sheets and that content swipes don't fight the system edge-swipe.

app/views/RoomView/List/components/NavBottomFAB.tsx
```diff
@@ -28,13 +28,14 @@ const styles = StyleSheet.create({
...
 	const { colors } = useTheme();
+	const { bottom } = useSafeAreaInsets();
 
 	if (!visible) {
 		return null;
 	}
 
 	return (
-		<View style={styles.container}>
+		<View style={[styles.container, { bottom: EDGE_DISTANCE + bottom }]}>
```
diegolmello: `bottom: EDGE_DISTANCE + bottom` — this FAB is absolutely positioned inside the message `List` view, whose bottom edge is the **top of the composer** (RoomView's `SafeAreaView` uses `edges={['right','left']}`, so the list doesn't extend into the nav-bar area; the composer is a sibling below it). Adding the nav-bar inset `bottom` here looks like it offsets the FAB ~`bottom`px too high, leaving a gap above the composer on nav-bar devices. Please confirm the FAB still sits just above the composer.

app/views/RoomActionsView/index.tsx
```diff
@@ -1089,14 +1092,16 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
...
 		const { rid, t, prid, teamId } = room;
 		const isGroupChatHandler = isGroupChat(room);
 		const itsMe = t === SubscriptionType.DIRECT && !isGroupChatHandler && !!userId && getUidDirectMessage(room) === userId;
 
 		return (
 			<SafeAreaView testID='room-actions-view'>
-				<List.Container testID='room-actions-scrollview'>
+				<List.Container
+					testID='room-actions-scrollview'
+					contentContainerStyle={[styles.scrollContentContainer, { paddingBottom: insets.bottom }]}>
```
diegolmello: Same `List.Container` override trap as SettingsView: passing `contentContainerStyle` replaces the default `paddingVertical: 16`. Here `scrollContentContainer` re-adds `paddingTop: 16`, so only the bottom is affected — it becomes `insets.bottom`, which is 0 on hardware-button devices / landscape / tablet master-detail (was 16). Minor; the structural `ListContainer` merge suggested on SettingsView would remove the need for the `scrollContentContainer` workaround entirely.

app/views/RoomActionsView/index.tsx
diegolmello: Nit: `insets` is typed inline `{ bottom: number; top: number; left: number; right: number }`; the rest of the PR uses `EdgeInsets` from `react-native-safe-area-context`. Use `EdgeInsets` for consistency.

app/views/LanguageView/index.tsx
```diff
@@ -95,7 +97,7 @@ const LanguageView = () => {
 				keyExtractor={item => item.value}
 				ListHeaderComponent={List.Separator}
 				ListFooterComponent={List.Separator}
-				contentContainerStyle={List.styles.contentContainerStyleFlatList}
+				contentContainerStyle={[List.styles.contentContainerStyleFlatList, { paddingBottom: bottom }]}
```
diegolmello: `[List.styles.contentContainerStyleFlatList, { paddingBottom: bottom }]` — `contentContainerStyleFlatList` already has `paddingVertical: 16`, and the appended `{ paddingBottom: bottom }` overrides its `paddingBottom` (array flatten, last key wins). On zero-inset contexts the list drops from 16px to 0 bottom padding. Same pattern in AutoTranslateView, DefaultBrowserView, SelectServerView, SelectListView. Use `paddingBottom: Math.max(16, bottom)` or add to (not override) the base.

app/views/NewServerView/index.tsx
```diff
@@ -148,11 +147,7 @@ const NewServerView = () => {
 	}, [connecting, previousServer]);
 
 	return (
-		<FormContainer
-			style={formContainerStyle}
-			showAppVersion={showBottomInfo}
-			testID='new-server-view'
-			keyboardShouldPersistTaps='handled'>
+		<FormContainer showAppVersion={showBottomInfo} testID='new-server-view' keyboardShouldPersistTaps='handled'>
```
diegolmello, line 150: This removed the conditional `previousServer ? { paddingBottom: 100 } : {}` and now relies solely on `FormContainer`'s `paddingBottom: bottom`. When adding a second server (`previousServer` set) a `CertificatePicker` renders at the bottom; the 100px of guaranteed clearance is replaced by ≤inset (0 on hardware-button devices), so the certificate row / Connect button can end up obscured or not fully scrollable. Worth checking the add-server flow with the cert picker visible.

Rohit3523: Verified and both buttons are now visible
| Android | iOS | iPhone SE | 
| --- | --- | --- |
| Screenshot 2026-06-25 at 9 53 28 PM | Screenshot 2026-06-25 at 9 53 59 PM | Screenshot 2026-06-25 at 9 58 16 PM |

app/containers/ActionSheet/useActionSheetDetents.ts
Rohit3523: Adding handle height in calculation is adding extra space in bottom sheet

app/views/E2EEncryptionSecurityView/index.tsx
Rohit3523: List.Container is handling it now

app/containers/ActionSheet/ActionSheet.tsx
Rohit3523: We can still close the bottomsheet by tapping the action bar

package.json
Rohit3523: Yup, it's android only

app/containers/ActionSheet/BottomSheetContent.tsx
Rohit3523: I did test in multiple device but cancel button is always visible and clickable, having 0 bottom inset value is kinda rare

app/containers/FormContainer.tsx
```diff
@@ -27,19 +28,20 @@ export const FormContainerInner = ({
...
-const FormContainer = ({ children, testID, showAppVersion = true, ...props }: IFormContainer) => {
+const FormContainer = ({ children, testID, showAppVersion = true, ...props }: IFormContainer): ReactElement => {
 	const { colors } = useTheme();
+	const { bottom } = useSafeAreaInsets();
 
 	return (
 		<KeyboardAwareScrollView
 			style={[sharedStyles.container, { backgroundColor: colors.surfaceRoom }]}
-			contentContainerStyle={[sharedStyles.containerScrollView, styles.scrollView]}
+			contentContainerStyle={[sharedStyles.containerScrollView, styles.scrollView, { paddingBottom: Math.max(24, bottom) }]}
```
Rohit3523, line 44: We get 24 from sharedStyles.containerScrollView

app/views/DefaultBrowserView/index.tsx
```diff
@@ -86,7 +88,7 @@ const DefaultBrowserView = () => {
 			<FlatList
 				data={DEFAULT_BROWSERS.concat(supported)}
 				keyExtractor={item => item.value}
-				contentContainerStyle={List.styles.contentContainerStyleFlatList}
+				contentContainerStyle={[List.styles.contentContainerStyleFlatList, { paddingBottom: Math.max(16, bottom) }]}
```
Rohit3523: We get 16 from List.styles.contentContainerStyleFlatList

app/views/DirectoryView/Options.tsx
```diff
@@ -52,7 +50,7 @@ const DirectoryOptions = ({
 	};
 
 	return (
-		<List.Container contentContainerStyle={{ backgroundColor: colors.surfaceRoom, marginBottom: insets.bottom }}>
+		<List.Container contentContainerStyle={{ backgroundColor: colors.surfaceRoom }}>
```
Rohit3523: List.Container is now handling bottom insets

app/views/JitsiMeetView/index.tsx
```diff
@@ -105,7 +106,7 @@ const JitsiMeetView = (): ReactElement => {
 
 	return (
 		<SafeAreaView style={styles.container}>
-			{authModal && <JitsiAuthModal setAuthModal={setAuthModal} callUrl={callUrl} />}
+			{authModal ? <JitsiAuthModal setAuthModal={setAuthModal} callUrl={callUrl} /> : null}
```
Rohit3523, line 111: Did this because of TS error

Rohit3523:
@coderabbitai review

.maestro/tests/room/message-markdown-click.yaml
```diff
@@ -78,7 +78,7 @@ tags:
       platform: iOS
     commands:
       - tapOn:
-          point: 66%,66%
+          point: 66%,63%
```
Rohit3523, line 81: Here, we are using a point because on iOS, Maestro detects the link and the surrounding text as a single element. Because of this, it's not possible to click only the specific link within the message.

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -71,14 +68,14 @@ export function useActionSheetDetents({
...
-					optionsLength * itemHeight + HANDLE_HEIGHT + headerHeight + bottomInset + (hasCancel ? CANCEL_HEIGHT : 0);
+				const measuredHeight = optionsLength * itemHeight + HANDLE_HEIGHT + headerHeight + (hasCancel ? CANCEL_HEIGHT : 0);
 
 				scrollEnabled = false;
 				detents = [heightToDetent(Math.round(measuredHeight), windowHeight)];
 			}
 		} else if (contentHeight > 0) {
-			const rawContentDetent = (contentHeight + bottomInset + HANDLE_HEIGHT) / windowHeight;
+			// we are adding handle height in ios only because last item on iPhone SE is not fully visible.
+			const rawContentDetent = (contentHeight + (isIOS ? HANDLE_HEIGHT : 0)) / windowHeight;
```
Rohit3523: I added the handle height only for iOS because, on the iPhone SE, the last item (or part of the bottom sheet) is hidden behind the screen, making it impossible to tap. Applying the same change on Android adds unnecessary extra space at the bottom.

app/containers/List/ListContainer.tsx
```diff
@@ -13,12 +14,14 @@ interface IListContainer extends ScrollViewProps {
...
-const ListContainer = ({ children, ...props }: IListContainer) => {
+const ListContainer = ({ children, contentContainerStyle, ...props }: IListContainer) => {
 	'use memo';
 
+	const { bottom } = useSafeAreaInsets();
+
 	return (
 		<ScrollView
-			contentContainerStyle={styles.container}
+			contentContainerStyle={[contentContainerStyle || styles.container, { paddingBottom: Math.max(bottom, 16) }]}
```
Rohit3523: We get 16 from `styles.container`

app/views/JitsiMeetView/index.tsx
```diff
@@ -116,7 +119,7 @@ const JitsiMeetView = (): ReactElement => {
 					}}
 					onNavigationStateChange={onNavigationStateChange}
 					onShouldStartLoadWithRequest={onNavigationStateChange}
-					style={styles.webviewContainer}
+					style={[styles.webviewContainer, { marginBottom: bottom }]}
```
Rohit3523, line 122: Using this because SafeAreaView doesn't handle bottom inset in vertical

app/views/NewServerView/index.tsx
Rohit3523, line 56: Screenshot after removing this part

| Android | iOS | iPhone SE |
| --- | --- | --- | 
| Image | Image | Image |

app/containers/List/ListContainer.tsx
```diff
@@ -13,12 +14,14 @@ interface IListContainer extends ScrollViewProps {
...
-const ListContainer = ({ children, ...props }: IListContainer) => {
+const ListContainer = ({ children, contentContainerStyle, ...props }: IListContainer) => {
 	'use memo';
 
+	const { bottom } = useSafeAreaInsets();
+
 	return (
 		<ScrollView
-			contentContainerStyle={styles.container}
+			contentContainerStyle={[styles.container, contentContainerStyle, { paddingBottom: Math.max(bottom, 16) }]}
```
Rohit3523: I did check and only directory view option pass background color via contentContainerStyle

In old logic it completely overwrite the value of `contentContainerStyle` ignoring the padding value but now we need the bottom insets to prevent last option to hide behind nav bar

app/views/DirectoryView/Options.tsx
```diff
@@ -52,7 +50,7 @@ const DirectoryOptions = ({
 	};
 
 	return (
-		<List.Container contentContainerStyle={{ backgroundColor: colors.surfaceRoom, marginBottom: insets.bottom }}>
```
Rohit3523, line 55: List.Container now handles bottom margin

app/views/LanguageView/index.tsx
```diff
@@ -95,7 +97,7 @@ const LanguageView = () => {
 				keyExtractor={item => item.value}
 				ListHeaderComponent={List.Separator}
 				ListFooterComponent={List.Separator}
-				contentContainerStyle={List.styles.contentContainerStyleFlatList}
+				contentContainerStyle={[List.styles.contentContainerStyleFlatList, { paddingBottom: Math.max(16, bottom) }]}
```
Rohit3523: contentContainerStyleFlatList also tries to apply padding of 16, so in case bottom insets return 0, it will use 16.

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -72,7 +74,8 @@ export function useActionSheetDetents({
 				detents = [heightToDetent(Math.round(measuredHeight), windowHeight)];
 			}
 		} else if (contentHeight > 0) {
-			const rawContentDetent = contentHeight / windowHeight;
+			// we are adding handle height in ios only because last item on iPhone SE is not fully visible.
+			const rawContentDetent = (contentHeight + (isIOS ? HANDLE_HEIGHT : 0)) / windowHeight;
```
diegolmello: If the last element is hidden and you're adding handle height to the math, you're guessing the behavior. We should understand why it's not being rendered.

app/views/RoomView/List/components/NavBottomFAB.tsx
```diff
@@ -28,14 +28,13 @@ const styles = StyleSheet.create({
...
 	const { colors } = useTheme();
-	const { bottom } = useSafeAreaInsets();
 
 	if (!visible) {
 		return null;
 	}
 
 	return (
-		<View style={[styles.container, { bottom: EDGE_DISTANCE + bottom }]}>
+		<View style={styles.container}>
```
diegolmello, line 37: Looks like a regression on iOS. It can get hidden behind the keyboard depending on the device's notch.

diegolmello (changes requested):

app/containers/EmojiPicker/EmojiCategory.tsx
```diff
@@ -55,15 +54,14 @@ const EmojiCategory = ({
...
+	const contentPaddingBottom = bottomSheet ? MIN_BOTTOM_SHEET_BREATHING_ROOM : undefined;
```
Rohit3523, line 64: Removed Math.max because we do a fixed value addition in `MIN_BOTTOM_SHEET_BREATHING_ROOM` which is always > 0

app/views/DirectoryView/styles.ts
```diff
@@ -3,6 +3,9 @@ import { StyleSheet } from 'react-native';
 import sharedStyles from '../Styles';
 
 export default StyleSheet.create({
+	container: {
+		paddingVertical: 0
```
Rohit3523, line 7: Resetting padding in directory view bottom sheet to prevent extra space

app/views/UserNotificationPreferencesView/ListPicker.tsx
```diff
@@ -37,10 +36,9 @@ const ListPicker = ({
 	const { showActionSheet, hideActionSheet } = useActionSheet();
 	const { colors } = useTheme();
 	const option = value ? OPTIONS[preference].find(option => option.value === value) : OPTIONS[preference][0];
-	const insets = useSafeAreaInsets();
 
 	const getOptions = (): ReactElement => (
-		<View style={{ backgroundColor: colors.surfaceRoom, marginBottom: insets.bottom }}>
```
Rohit3523, line 43: bottom inset is now handled in sheet

app/views/StatusView/ClearAfterPicker/ClearAfterSheetContent.tsx
```diff
@@ -69,7 +67,7 @@ const ClearAfterSheetContent = ({ initialValue, initialDate, onConfirm }: IClear
 	};
 
 	return (
-		<View style={{ backgroundColor: colors.surfaceRoom, marginBottom: insets.bottom }}>
```
Rohit3523, line 72: bottom inset is now handled in sheet

app/containers/markdown/Markdown.test.tsx
```diff
@@ -1,4 +1,8 @@
 import { generateSnapshots } from '../../../.rnstorybook/generateSnapshots';
 import * as stories from './Markdown.stories';
 
+beforeAll(() => {
+	jest.useFakeTimers({ now: new Date('2026-02-01T00:00:00.000Z').getTime() });
+});
```
Rohit3523, line 6: This is unrelated to this PR but unit test was failing because relative time is rendering '2 years ago' instead of 'a year ago'

app/views/LanguageView/index.tsx
OtavioStasiak: Create something like `const paddingBottom` instead of do the Math inside of it.

app/containers/EmojiPicker/EmojiCategory.tsx
OtavioStasiak: but you also removed the `MIN_BOTTOM_SHEET_BREATHING_ROOM - bottom` ... is it not necessary? why?

OtavioStasiak (dismissed):

Rohit3523:
Regarding latest comment

The bottom sheet already handle safe area internally, `- bottom` was double-correcting and collapsing the breathing room on notch devices.
Screenshot 2026-07-03 at 3 32 09 AM

diegolmello (approved):

Issue #1, opened 2017-08-07
Better code structure


rodrigok:
There are some options to improve the code structure:

- [ ] Read the [Building Chatty](https://medium.com/react-native-training/building-chatty-a-whatsapp-clone-with-react-native-and-apollo-part-1-setup-68a02f7e11) to get ideas
- [x] Add redux
- [x] Rename file `meteor.js` to `RocketChat` to encapsulate the API to populate realm
- [x] Separate views to standalone files
- [x] Integrate [Storybook](https://medium.com/storybookjs/announcing-storybook-3-2-e00918a1764c) to be able to test state of components

PR #7175, opened 2026-04-17
refactor(voip): decouple peer autocomplete from Redux; simplify NewMediaCall

diegolmello:
## Proposed changes

Decouples `getPeerAutocompleteOptions` from Redux (`auxStore`), trims the peer-autocomplete Zustand API, inlines the single-use `NewMediaCall` container, and hardens async autocomplete against stale responses.

### Key changes
- `getPeerAutocompleteOptions` takes `username` / `sipEnabled` (via `TPeerAutocompleteAuth`) instead of reading the aux store. Service is pure; filter is trimmed once at the store boundary; `peerInfo` narrowed via discriminated-union on `type`.
- `FilterHeader` uses named selectors (`selectSipEnabled`, `selectUsername`) for login username + SIP setting and passes auth into `fetchOptions`.
- Removed `setFilter` / `clearSelection` from the store; filter updates use `setState`; clear chip uses `setSelectedPeer(null)` and clears `options`.
- `fetchOptions`: last-write-wins. Monotonic sequence counter lives in the store closure, incremented at the very top of `fetchOptions` (before the empty-filter early return) and bumped on `reset()`. Single `seq` stale-guard in both success and catch paths — overlapping requests and late errors cannot clobber newer UI state.
- Inlined former `Container.tsx` into `NewMediaCall.tsx`; removed `Container` + `Container.stories.tsx`.
- `PeerList`: `setSelectedPeer(option)` directly; `keyExtractor` uses `${type}:${value}`.
- Dropped the dead commented `setIceServers` line in `MediaSessionInstance.ts`.

### Tests
- New `getPeerAutocompleteOptions.test.ts` covering trimming, current-username exclusion, empty-filter shortcut, SIP-disabled routing, and canonical typed mocks (`mockUsersAutoCompleteResponse` helper; failure shape carries `error`).
- Extended `usePeerAutocompleteStore.test.ts` — `reset`, stale fetch, `setSelectedPeer(null)` clears options, username forwarding, trim-at-store-boundary.
- Updated `FilterHeader` / `SelectedPeer` tests; refreshed SelectedPeer story snapshots.

## How to test or reproduce
- `TZ=UTC yarn test --testPathPattern='getPeerAutocompleteOptions|usePeerAutocompleteStore|FilterHeader|SelectedPeer|PeerList|NewMediaCall|MediaSessionInstance'`
- Manual: New call sheet — type to search, select user/SIP, clear chip; confirm list and field stay coherent under quick typing.

## Screenshots
N/A (behavioral refactor)

## Types of changes

## Checklist
- [ ] I have read the CONTRIBUTING doc
- [ ] I have signed the CLA

## Further comments
Branched from `feat.voip-lib-new`.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale autocomplete results from replacing newer ones; failing requests no longer clear options unless current.

* **Refactor**
  * Autocomplete requests now include caller identity and SIP enablement; clearing a selection also clears options.
  * Simplified media-call layout and applied consistent theming.
  * Peer list keys are now type-aware to avoid collisions.

* **Tests**
  * Expanded tests for autocomplete, SIP branches, trimming, errors, and race conditions.

* **Chores**
  * Removed an obsolete Storybook story and unused container.


diegolmello (commented):
## Consolidated review — VoIP PR2: peer autocomplete + NewMediaCall

Three parallel review passes (architect / anti-slop / simplifier) plus two existing CodeRabbit comments. Overall direction is a net improvement — findings ordered by severity.

---

### Critical

**1. `app/lib/services/voip/getPeerAutocompleteOptions.test.ts:38-44` — typed mocks use fields not in `IUser` (CodeRabbit)**

Fixture `freeSwitchExtension` is not part of `IUser`, and the failure response shape `{ success: false }` is missing the required `error` field. The typed mock fails type-checks once consumed. Introduce a `mockUsersAutoCompleteResponse` helper returning the canonical shape, and use `Awaited<ReturnType<typeof ...>>` on the mocked fn so the compiler enforces the real signature.

---

### Major

**2. `app/lib/services/voip/usePeerAutocompleteStore.ts:26-56` — `fetchSeq` race: increment must precede the empty-filter early return (CodeRabbit)**

`++peerAutocompleteFetchSeq` lives below the `if (!trimmedFilter)` early return. Typing → clearing the filter fast enough lets a stale in-flight response resolve and overwrite cleared state. Move `const seq = ++peerAutocompleteFetchSeq;` to the very top of `fetchOptions` so every invocation (including the early return) invalidates prior requests.

**3. `app/lib/services/voip/usePeerAutocompleteStore.ts:30` — sequence counter at module scope (simplifier)**

`let peerAutocompleteFetchSeq = 0` is module-level mutable state. It leaks across store resets and across tests / hot-reload. Scope it to the store closure (`let seq = 0` just before `create(...)`). Same semantics, proper boundary.

**4. `app/lib/services/voip/getPeerAutocompleteOptions.ts:2,42,47,50` — Redux reads inside the VoIP service layer (architect)**

The service reaches into `reduxStore.getState().login.user.username` and `reduxStore.getState().settings.VoIP_TeamCollab_SIP_Integration_For_Internal_Calls`. Per project convention the VoIP layer is Zustand-only and should stay pure; pulling Redux in here creates hidden coupling and forces unit tests to mock Redux (see #6). Inject `currentUsername` and `sipEnabled` as arguments from the caller — the React layer already has both via `useSelector`.

**5. `app/lib/services/voip/getPeerAutocompleteOptions.ts` — service has zero unit tests for key branches (anti-slop)**

Filter trimming, current-username exclusion, empty-filter shortcut, SIP-disabled routing. Once dependencies are injected per #4 these branches become trivially testable.

**6. `app/lib/services/voip/usePeerAutocompleteStore.test.ts:110` — Redux store not mocked; filter branches silently untested (architect)**

`currentUsername` resolves to `undefined` in the test harness, so the "exclude self" filter is a no-op in every assertion. Either mock the Redux store, or (preferred) inject the username per #4.

**7. `app/containers/NewMediaCall/CreateCall.tsx:8,22` — direct `mediaSessionInstance.startCall` bypasses the Zustand store boundary (architect)**

Every other VoIP call site goes through the store. Directly importing the singleton here breaks the pattern and makes the component harder to test in isolation. Route through the store action.

---

### Minor

**8. `app/lib/services/voip/usePeerAutocompleteStore.ts:45-55` — redundant double stale-guard (simplifier)**

After `await`, both the `seq` check and the `filter/selectedPeer` check guard staleness. `seq` alone is sufficient — any filter change has already bumped the counter. The same redundant guard is duplicated in the `catch` block. Keep only the `seq` guard.

**9. `app/lib/services/voip/getPeerAutocompleteOptions.ts` — double-trim overlap with store `fetchOptions` (anti-slop)**

Both the store and the service trim the filter string. Trim once, at the store boundary.

**10. `app/containers/NewMediaCall/FilterHeader.tsx:15-18,20` — four separate `usePeerAutocompleteStore` subscriptions + 500ms debounce (architect)**

Four separate selectors on the same store per render. Collapse with a single selector + shallow compare. Separately, 500ms debounce is not sufficient on its own to prevent the race fixed by #2/#3 — the sequence counter is the real fix.

**11. `app/containers/NewMediaCall/FilterHeader.tsx:17-27` — `useCallback` inside `useDebounce` adds indirection with no benefit (simplifier)**

The inner `useCallback` is not observed by anything; `useDebounce` already produces a stable ref. Pass the inline function directly.

---

### Nits

**12. `app/lib/services/voip/MediaSessionInstance.ts:235` — commented-out `// this.instance?.setIceServers(this.iceServers);` (anti-slop)**

Pre-existing but worth cleaning: either delete or add a one-line note on why it's disabled.

**13. `app/lib/services/voip/getPeerAutocompleteOptions.ts:28-34` — repeated `'callerId' in peerInfo` guards (simplifier)**

Inline ternaries all use `peerInfo && 'x' in peerInfo ? peerInfo.x : undefined`. `TPeerItem` is already a discriminated union on `type` — narrow on `peerInfo?.type === '...'` for type-safety and readability.

**14. `app/containers/NewMediaCall/PeerList.tsx:13-16` — `handleSelectOption` is a one-line wrapper (simplifier)**

If `PeerItem`'s `onSelectOption` prop matches `(peer: TPeerItem | null) => void`, pass `setSelectedPeer` directly.

**15. `app/containers/NewMediaCall/FilterHeader.tsx:14-17` — two `useSelector` calls (simplifier)**

`Boolean(settings.VoIP_TeamCollab_SIP_Integration_For_Internal_Calls)` inline is fine but a named selector (`selectSipEnabled`) reads better.

**16. `app/containers/NewMediaCall/NewMediaCall.tsx` — inlining `Container` is the right call (simplifier)**

Confirm `styles.screen` matches the values previously in `Container`.

---

**Summary**: fix #1, #2, #3 before merge. #4–#7 and #10 are architectural — worth this PR or a fast follow-up. Rest is polish.

PR #7178, opened 2026-04-17
refactor(voip): MediaCallEvents Redux adapters and resetVoipState

diegolmello:
## Proposed changes

This PR decouples VoIP native event wiring from Redux: `MediaCallEvents` no longer imports the global store or `deepLinkingOpen`. Instead, `Root` passes `MediaCallEventsAdapters` (`getActiveServerUrl`, `onOpenDeepLink`) into `setupMediaCallEvents` and `getInitialMediaCallEvents`, so reads and dispatches happen at the composition root with current store state.

VoIP teardown after a failed native accept is centralized in new `resetVoipState()` (now also clears module-level VoIP accept dedupe sentinels, then resets native call id, then full call-store reset). The deep-linking saga calls that helper instead of duplicating `useCallStore` calls.

`resetMediaCallEventsStateForTesting()` is exported for tests, and iOS `endCall` clears module-level VoIP accept dedupe keys so a later call with the same UUID is not incorrectly suppressed. Android coverage for the same dedupe reset is provided through `resetVoipState()` (no iOS-only `endCall` path on Android).

Logging that used `console.log`/`console.error` in this module now goes through `MediaCallLogger`. `MediaCallLogger.error`/`warn` correctly route to `console.error`/`console.warn` (previously silently downgraded to `console.log`). Unit tests mock `MediaCallLogger` to avoid noisy output.

### Review fixes (latest commit)

- `MediaCallLogger.error` / `warn` now use `console.error` / `console.warn` (severity downgrade bug)
- `VoipAcceptSucceeded` sentinel write moved past the `type !== 'incoming_call'` guard so non-incoming events cannot poison dedupe for a real incoming call
- `clearVoipAcceptDedupeSentinels` exported and called from `resetVoipState()` for cross-platform sentinel reset
- Remaining `console.error` calls migrated to `mediaCallLogger.error`
- Removed unused `path?: string` field from `VoipDeepLinkParams`
- Added regression tests: sentinel poisoning, distinct callIds not suppressed, `getActiveServerUrl` undefined → deep-link path, `resetVoipState` clear-then-reset ordering

Tests: Android suite updated for adapter injection; iOS suite adds coverage for `VoipPushTokenRegistered` (registers push token), cold-start `getInitialMediaCallEvents` (answered + same vs foreign workspace), `endCall` dedupe reset, and a small `resetVoipState` ordering test. **35 tests across 3 suites pass.**

## How to test or reproduce

- `TZ=UTC yarn test --testPathPattern='MediaCallEvents|resetVoipState'`
- Smoke: cold start / incoming VoIP on iOS and Android; accept failure toast path; workspace switch during accept.
- Verify on Android: rapid-accept the same call twice → sentinel reset prevents the second accept from being suppressed.

## Screenshots

N/A (behavioral / architectural change).

## Types of changes

## Checklist

## Further comments

Cold-start `NativeVoipModule.clearInitialEvents()` coverage on answered paths matches the previous implementation (unchanged in this PR). `endCall` listener remains iOS-scoped as before. Reviewers flagged possible follow-ups for native stash lifecycle and Android parity; those can be separate if product agrees.

## Summary by CodeRabbit

* **Improvements**
  * More robust VoIP call handling and startup behavior, with reliable cold-start and server-matching for incoming calls.
  * Reduced direct state coupling for cleaner call flow and clearer diagnostics.

* **Bug Fixes**
  * End-call now reliably clears accept deduplication so repeated call flows can be retriggered after reset.

* **Tests**
  * Expanded VoIP test coverage, including idempotency, cold-start, and cross-platform event scenarios.

### DIFF
--- a/app/index.tsx
+++ b/app/index.tsx
@@ -33,7 +33,11 @@ import {
 } from './lib/methods/helpers/theme';
 import { initializePushNotifications, onNotification } from './lib/notifications';
 import { getInitialNotification, setupVideoConfActionListener } from './lib/notifications/videoConf/getInitialNotification';
-import { getInitialMediaCallEvents, setupMediaCallEvents } from './lib/services/voip/MediaCallEvents';
+import {
+	getInitialMediaCallEvents,
+	setupMediaCallEvents,
+	type MediaCallEventsAdapters
+} from './lib/services/voip/MediaCallEvents';
 import store from './lib/store';
 import { initStore } from './lib/store/auxStore';
 import { type TSupportedThemes, ThemeContext } from './theme';
@@ -109,6 +113,13 @@ export default class Root extends React.Component<{}, IState> {
 		setNativeTheme(theme);
 	}
 
+	private getMediaCallEventsAdapters(): MediaCallEventsAdapters {
+		return {
+			getActiveServerUrl: () => store.getState().server.server,
+			onOpenDeepLink: params => store.dispatch(deepLinkingOpen(params))
+		};
+	}
+
 	componentDidMount() {
 		this.listenerTimeout = setTimeout(() => {
 			Linking.addEventListener('url', ({ url }) => {
@@ -123,7 +134,7 @@ export default class Root extends React.Component<{}, IState> {
 		// Set up video conf action listener for background accept/decline
 		this.videoConfActionCleanup = setupVideoConfActionListener();
 		// Set up media call event listeners for incoming calls
-		this.mediaCallEventCleanup = setupMediaCallEvents();
+		this.mediaCallEventCleanup = setupMediaCallEvents(this.getMediaCallEventsAdapters());
 	}
 
 	componentWillUnmount() {
@@ -153,7 +164,7 @@ export default class Root extends React.Component<{}, IState> {
 			return;
 		}
 
-		const voipInitialHandled = await getInitialMediaCallEvents();
+		const voipInitialHandled = await getInitialMediaCallEvents(this.getMediaCallEventsAdapters());
 		if (voipInitialHandled) {
 			// VoIP path already dispatched navigation (or will via deep linking); do not call appInit() in parallel
 			return;
--- a/app/lib/services/voip/MediaCallEvents.ios.test.ts
+++ b/app/lib/services/voip/MediaCallEvents.ios.test.ts
@@ -1,28 +1,75 @@
 /**
  * @jest-environment node
  *
- * iOS-only mute tests: requires isIOS = true so didPerformSetMutedCallAction listener is registered.
+ * iOS-only paths: isIOS = true, NativeEventEmitter for VoIP events, CallKit listeners.
  */
-import { setupMediaCallEvents } from './MediaCallEvents';
+import type { VoipPayload } from '../../../definitions/Voip';
+import RNCallKeep from 'react-native-callkeep';
+import NativeVoipModule from '../../native/NativeVoip';
+import {
+	getInitialMediaCallEvents,
+	resetMediaCallEventsStateForTesting,
+	setupMediaCallEvents,
+	type MediaCallEventsAdapters
+} from './MediaCallEvents';
 import { useCallStore } from './useCallStore';
+import { registerPushToken } from '../restApi';
 
-const mockAddEventListener = jest.fn();
+/** Shared bucket for NativeEventEmitter / DeviceEventEmitter VoIP listeners (Jest allows `mock*` refs inside factories). */
+const mockNativeVoipListeners: Record<string, ((payload: unknown) => void)[]> = {};
+
+/** Minimal RN surface for MediaCallEvents — avoid `requireActual('react-native')` in @jest-environment node. */
+jest.mock('react-native', () => ({
+	Platform: { OS: 'ios' },
+	DeviceEventEmitter: {
+		addListener: (eventType: string, listener: (payload: unknown) => void) => {
+			mockNativeVoipListeners[eventType] = mockNativeVoipListeners[eventType] || [];
+			mockNativeVoipListeners[eventType].push(listener);
+			return {
+				remove() {
+					const list = mockNativeVoipListeners[eventType];
+					if (!list) {
+						return;
+					}
+					const idx = list.indexOf(listener);
+					if (idx >= 0) {
+						list.splice(idx, 1);
+					}
+				}
+			};
+		}
+	},
+	NativeEventEmitter: class {
+		// eslint-disable-next-line @typescript-eslint/no-useless-constructor, @typescript-eslint/no-unused-vars
+		constructor(_nativeModule?: unknown) {}
+		addListener(eventType: string, listener: (payload: unknown) => void) {
+			mockNativeVoipListeners[eventType] = mockNativeVoipListeners[eventType] || [];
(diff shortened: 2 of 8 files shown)


diegolmello (commented):
## Critical Review — PR #7178

Multi-angle review: anti-slop (code-simplifier), architecture, and test specialist agents, consolidated with the CodeRabbit findings already on this PR. Two confirmed defects, one of which is high-severity and silently introduced by this PR.

---

### 🔴 High — `MediaCallLogger.error` is a no-op for severity (silently downgrades errors)

`app/lib/services/voip/MediaCallLogger.ts:14-20`

```ts
error(...args: unknown[]): void {
    console.log(`[Media Call Error] ${JSON.stringify(args)}`);
}

warn(...args: unknown[]): void {
    console.log(`[Media Call Warning] ${JSON.stringify(args)}`);
}
```

`error` and `warn` both call `console.log`. The PR's stated intent is to migrate noisy logging through `MediaCallLogger`. Any future migration of the four remaining `console.error` sites (`MediaCallEvents.ts:98, 195, 270, 285`) to `mediaCallLogger.error` would silently demote severity, breaking Sentry/log-pipeline filtering. Fix in this PR before the planned migration runs.

```diff
-    error(...args: unknown[]): void {
-        console.log(`[Media Call Error] ${JSON.stringify(args)}`);
-    }
-    warn(...args: unknown[]): void {
-        console.log(`[Media Call Warning] ${JSON.stringify(args)}`);
-    }
+    error(...args: unknown[]): void {
+        console.error(`[Media Call Error] ${JSON.stringify(args)}`);
+    }
+    warn(...args: unknown[]): void {
+        console.warn(`[Media Call Warning] ${JSON.stringify(args)}`);
+    }
```

---

### 🔴 High — Dedupe sentinel set BEFORE type guard (also flagged by CodeRabbit, outside-diff)

`app/lib/services/voip/MediaCallEvents.ts:81-92`

```ts
function handleVoipAcceptSucceededFromNative(data: VoipPayload, adapters: MediaCallEventsAdapters) {
    const { callId } = data;
    if (callId && lastHandledVoipAcceptSucceededCallId === callId) {
        return;
    }
    if (callId) {
        lastHandledVoipAcceptSucceededCallId = callId;   // ← poisons sentinel
    }
    if (data.type !== 'incoming_call') {                 // ← guard fires AFTER
        mediaCallLogger.log(`${TAG} VoipAcceptSucceeded: not an incoming call`);
        return;
    }
    ...
}
```

A non-incoming `VoipAcceptSucceeded` with `callId="X"` writes the sentinel, then a legitimate `incoming_call` with the same `callId="X"` is silently dropped — no `setNativeAcceptedCallId`, no `onOpenDeepLink`. Move the sentinel write below the type guard.

No regression test catches this. Add: emit `outgoing_call`/non-`incoming_call` with callId X, then `incoming_call` with callId X, assert `setNativeAcceptedCallId` was called.

---

### 🟠 Medium — Android dedupe sentinels never reset

`app/lib/services/voip/MediaCallEvents.ts:115-132`

`clearVoipAcceptDedupeSentinels()` is called only from the iOS `endCall` listener, gated by `if (isIOS)`. On Android there is no equivalent reset hook. If the server ever reuses a callId across calls (or across a failure → retry), the second event is silently suppressed.

Suggested: also call `clearVoipAcceptDedupeSentinels()` from `resetVoipState()` so both platforms reset through a single teardown entry point. This further justifies the helper extraction.

---

### 🟠 Medium — Cold-start workspace check runs against a possibly-unhydrated store

`app/index.tsx:116-121` & `MediaCallEvents.ts:268`

`getInitialMediaCallEvents` is invoked from `Root.init()` which runs from the constructor — before `appInitLocalSettings` has finished hydrating `state.server.server`. `getActiveServerUrl()` can therefore return a falsy value, `isVoipIncomingHostCurrentWorkspace` returns `false`, and the same-workspace shortcut (`applyRestStateSignals`) is skipped in favor of the deep-link path. Functionally safe, but diverges from the documented intent and is testable: add a test where `getActiveServerUrl` returns `undefined` and assert the deep-link path runs.

---

### 🟠 Medium — Test coverage gaps

(from test specialist)

1. **Sentinel poisoning via type guard** — no test (see High #2).
2. **Different callId NOT suppressed** — same-id suppression is tested; the inverse is not.
3. **Sentinel reset on `endCall`** — iOS-only test; no Android symmetric test or doc explaining absence.
4. **`getActiveServerUrl` returning `undefined`/`null`/`""`** — branch at `MediaCallEvents.ts:40-41` untested.
5. **Catch-block coverage** — all four `console.error` sites (`98, 195, 270, 285`) are unexercised; the `try/catch` at lines 192-197 is the most relevant since it was added defensively.
6. **`registerPushToken` assertion is too weak** — `expect(registerPushToken).toHaveBeenCalled()` without verifying args (`MediaCallEvents.ios.test.ts:225-230`).
7. **iOS cold-start non-answered branch** — only the answered branch is tested.
8. **`resetVoipState` failure path** — only ordering tested.

---

### 🟡 Minor — Anti-slop / simplification

- **Mock duplication** in `MediaCallEvents.ios.test.ts:24-61` — `DeviceEventEmitter.addListener` and `NativeEventEmitter.addListener` bodies are byte-identical. Extract a `makeAddListener(bucket)` factory.
- **Pattern regression** — the previous `getMuteHandler()` helper was removed and replaced by 4 inline `mockAddEventListener.mock.calls.find(...)` with non-null assertions. Re-extract.
- **`resetMediaCallEventsStateForTesting`** is a one-line passthrough whose JSDoc claims it "is invoked internally on CallKit endCall" — but the internal caller invokes `clearVoipAcceptDedupeSentinels` directly, not this export. Either inline, or rename + drop the misleading JSDoc.
- **Speculative `path` field** on `VoipDeepLinkParams` (`MediaCallEvents.ts:20-27`) — exported but never set; remove or document.
- **`getMediaCallEventsAdapters()` private method** (`app/index.tsx:113-127`) — called twice in the same component; an inline `const adapters = ...` would read more directly.

---

### CodeRabbit findings already on this PR (echoed)

1. 🟡 [`MediaCallEvents.ios.test.ts:6-16`](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7178#discussion_r3103526276) — import ordering ESLint failure (CI failing).
2. 🟡 [`MediaCallEvents.ios.test.ts:42-62`](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7178#discussion_r3103526294) — empty constructor `no-empty-function` (CI failing).
3. 🔴 Outside-diff [`MediaCallEvents.ts:81-92`] — same sentinel ordering bug as High #2 above.
4. 🟢 Nitpick [`MediaCallEvents.ts:97-99, 194-196, 269-271, 284-285`] — four remaining `console.error` to migrate to `mediaCallLogger.error` (blocked by High #1 above — fix the logger first).

---

### Verdict

**Request changes.** The architectural goals (Redux decoupling, teardown centralization) are achieved cleanly, but the PR introduces a silent log-severity downgrade via `MediaCallLogger.error` and ships an existing dedupe-ordering bug uncovered by tests. Fix the logger, fix the sentinel ordering, add the missing regression tests, then this is mergeable. The iOS/Android sentinel asymmetry and cold-start hydration window can be follow-ups if scope is tight.

PR #7199, opened 2026-04-22
feat(android): create VoipCallService with FOREGROUND_SERVICE_MICROPHONE

diegolmello:
## Proposed changes

Create `VoipCallService` — an Android foreground `Service` with `foregroundServiceType="microphone"` — so VoIP audio calls keep running when the app is backgrounded. Without a foreground service, Android terminates the process and drops the active WebRTC audio session, which is the same problem iOS solves via CallKit audio retention.

### What this adds

- `VoipCallService.kt` — foreground service that starts/stops around the call lifecycle.
  - `startService(context, callId)`: starts in foreground with an ongoing notification.
  - `stopService(context)`: requests shutdown.
  - `ACTION_START` / `ACTION_STOP` intent actions for explicit control.
  - `ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE` on API 29+.
  - Low-priority notification channel and ongoing notification.
- `AndroidManifest.xml` — declares the service with `foregroundServiceType="microphone"`, `enabled="true"`, `exported="false"`.

### Review fixes applied

Addresses CodeRabbit review findings #1 (production crash) and #2 (start-command semantics):

1. `stopService()` helper now calls `context.stopService(intent)` instead of `context.startService(intent)`. On Android 8+ (API 26+), invoking `startService` from a background context throws `IllegalStateException` / `BackgroundServiceStartNotAllowedException` — which is exactly when a hangup from a backgrounded or headless-notification flow happens.
2. `onStartCommand` returns `START_NOT_STICKY` for `ACTION_START` (was `START_STICKY`). A redelivered null intent cannot recover the WebRTC peer connection, so sticky semantics would only log a warning and terminate anyway.
3. `stopSelf(startId)` replaces `stopSelf()` in the `ACTION_STOP` and unknown-action branches, so only the matching start token is released and no start counts are leaked.

## Issue(s)

N/A — internal VoIP work.

## How to test or reproduce

1. Place a VoIP call (`ACTION_START`), background the app, verify audio continues and the ongoing notification is visible.
2. Hang up from the notification (`ACTION_STOP`); verify the service stops without `IllegalStateException` in logs.
3. Send `ACTION_STOP` while the app is fully backgrounded; the service still tears down cleanly (this is the case the fix unblocks).

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue) — for the review-fix commit on top

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — N/A for foreground service lifecycle; covered by manual VoIP test plan above
- [ ] I have added necessary documentation — N/A

### Merge order

This is PR 3 of 6 in the VoIP Android integration series:

1. ~~PR-2: fix(ios) DDP cleanup — independent~~ (merged)
2. ~~PR-1: fix(ios) NSLock + timeout — independent~~ (merged)
3. **PR-3: feat(android) VoipCallService** ← this PR
4. PR-4: fix(android) service integration — depends on PR-3
5. PR-5: fix(both) null guard — independent, merge before PR-6
6. PR-6: chore(ts) cleanup — last (shares file with PR-5)

## Summary by CodeRabbit

* **New Features**
  * VoIP calling now runs reliably in the background, allowing calls to persist when switching apps.
  * Active calls show a persistent notification so users are informed and can return to the app quickly.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -147,6 +147,13 @@
 
         <service android:name="io.wazo.callkeep.RNCallKeepBackgroundMessagingService" />
 
+        <!-- VoIP foreground service for keeping audio calls alive in the background. -->
+        <service
+            android:name="chat.rocket.reactnative.voip.VoipCallService"
+            android:enabled="true"
+            android:exported="false"
+            android:foregroundServiceType="microphone" />
+
         <!-- react-native-webrtc ships MediaProjectionService (foregroundServiceType=mediaProjection)
              for screen sharing. We don't use screen sharing, and Android 15+ forbids starting
              restricted foreground service types from BOOT_COMPLETED receivers (expo-notifications
--- a/android/app/src/main/java/chat/rocket/reactnative/voip/VoipCallService.kt
+++ b/android/app/src/main/java/chat/rocket/reactnative/voip/VoipCallService.kt
@@ -1,0 +1,156 @@
+package chat.rocket.reactnative.voip
+
+import android.app.Notification
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.app.Service
+import android.content.Intent
+import android.content.pm.ServiceInfo
+import android.os.Build
+import android.os.IBinder
+import android.util.Log
+import androidx.core.app.NotificationCompat
+import chat.rocket.reactnative.MainActivity
+
+/**
+ * Foreground service that keeps the VoIP call alive when the app moves to the background.
+ * Required because Android terminates background processes without a foreground service,
+ * which would drop the active audio session.
+ *
+ * Started on call accept, stopped on hangup.
+ */
+class VoipCallService : Service() {
+
+    companion object {
+        private const val TAG = "RocketChat.VoipCallService"
+        private const val CHANNEL_ID = "voip-call-service"
+        private const val CHANNEL_NAME = "VoIP Call"
+        private const val NOTIFICATION_ID = 1
+
+        private const val ACTION_START = "chat.rocket.reactnative.voip.START_SERVICE"
+        private const val ACTION_STOP = "chat.rocket.reactnative.voip.STOP_SERVICE"
+        const val EXTRA_CALL_ID = "callId"
+
+        private var isRunning = false
+
+        @JvmStatic
+        fun startService(context: android.content.Context, callId: String) {
+            val intent = Intent(context, VoipCallService::class.java).apply {
+                action = ACTION_START
+                putExtra(EXTRA_CALL_ID, callId)
+            }
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                context.startForegroundService(intent)
+            } else {
+                context.startService(intent)
+            }
+        }
+
+        @JvmStatic
+        fun stopService(context: android.content.Context) {
+            val intent = Intent(context, VoipCallService::class.java).apply {
+                action = ACTION_STOP
+            }
+            context.startService(intent)
+        }
+    }
+
+    override fun onCreate() {
+        super.onCreate()
+        createNotificationChannel()
+        Log.d(TAG, "VoipCallService created")
+    }
+
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        when (intent?.action) {
+            ACTION_STOP -> {
+                Log.d(TAG, "Stopping VoipCallService")
+                stopSelf()
+                return START_NOT_STICKY
+            }
+            ACTION_START -> {
+                val callId = intent.getStringExtra(EXTRA_CALL_ID) ?: "unknown"
+                Log.d(TAG, "Starting VoipCallService for callId: $callId")
+                if (!isRunning) {
+                    isRunning = true
+                    startForegroundWithNotification(callId)
+                } else {
+                    Log.d(TAG, "Service already running, skipping duplicate start")
+                }
+                return START_STICKY
+            }
+            else -> {
+                Log.w(TAG, "Unknown action: ${intent?.action}")
+                return START_NOT_STICKY
+            }
+        }
+    }
+
+    private fun startForegroundWithNotification(callId: String) {
+        val notification = buildNotification(callId)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            startForeground(
+                NOTIFICATION_ID,
+                notification,
+                ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE
+            )
+        } else {
+            startForeground(NOTIFICATION_ID, notification)
+        }
+
+        Log.d(TAG, "Started foreground with notification for callId: $callId")
+    }
+
+    private fun buildNotification(callId: String): Notification {
+        // Pending intent: tapping the notification opens the app.
+        val pendingIntent = PendingIntent.getActivity(
+            this,
+            0,
+            Intent(this, MainActivity::class.java).apply {
+                flags = Intent.FLAG_ACTIVITY_SINGLE_TOP
+            },
+            PendingIntent.FLAG_UPDATE_CURRENT or if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                PendingIntent.FLAG_IMMUTABLE
+            } else {
+                0
+            }
+        )
+
+        return NotificationCompat.Builder(this, CHANNEL_ID)
+            .setContentTitle("VoIP Call")
+            .setContentText("Call in progress")
+            .setSmallIcon(getApplicationInfo().icon)
+            .setContentIntent(pendingIntent)
+            .setOngoing(true)
+            .setOnlyAlertOnce(true)
+            .setPriority(NotificationCompat.PRIORITY_LOW)
+            .setCategory(NotificationCompat.CATEGORY_SERVICE)
+            .setVisibility(NotificationCompat.VISIBILITY_PUBLIC)
+            .build()
+    }
+
+    private fun createNotificationChannel() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val channel = NotificationChannel(
+                CHANNEL_ID,
+                CHANNEL_NAME,
+                NotificationManager.IMPORTANCE_LOW
+            ).apply {
+                description = "VoIP call in progress"
+                setShowBadge(false)
+            }
+            val notificationManager = getSystemService(NotificationManager::class.java)
+            notificationManager?.createNotificationChannel(channel)
+        }
+    }
+
+    override fun onBind(intent: Intent?): IBinder? = null
+
+    override fun onDestroy() {
+        isRunning = false
+        Log.d(TAG, "VoipCallService destroyed")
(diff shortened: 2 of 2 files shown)


diegolmello:
## Code Review — PR #7199

**Files Reviewed:** `VoipCallService.kt`, `AndroidManifest.xml`  
**Issues:** 0 CRITICAL/HIGH, 2 LOW observations

---

### [LOW] `isRunning` flag is static but also managed via service lifecycle

`VoipCallService.kt:26`

`isRunning` is a static flag used to prevent duplicate `startForeground` calls. However, since each `startService()` call with `ACTION_START` creates a new `startId`, the `START_STICKY` return value means the service could be killed and recreated. The `isRunning` flag won't be reset on service recreation (it remains `true`), which means subsequent `startService` calls would be silently dropped.

**Fix:** Remove the `isRunning` guard entirely — `startForeground()` is idempotent and can be called multiple times on the same service instance. If the service is killed and recreated, Android recreates it in a fresh state.

```kotlin
override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
    when (intent?.action) {
        ACTION_STOP -> {
            Log.d(TAG, "Stopping VoipCallService")
            stopSelf()
            return START_NOT_STICKY
        }
        ACTION_START -> {
            val callId = intent.getStringExtra(EXTRA_CALL_ID) ?: "unknown"
            Log.d(TAG, "Starting VoipCallService for callId: $callId")
            startForegroundWithNotification(callId)
            return START_STICKY
        }
        else -> {
            Log.w(TAG, "Unknown action: ${intent?.action}")
            return START_NOT_STICKY
        }
    }
}
```

---

### [LOW] Small icon fallback uses application icon which may not be visible

`VoipCallService.kt:99`

`getApplicationInfo().icon` falls back to the app's launcher icon, which is typically not designed for notification small icons (expects a flat, monochrome drawable). On Android 5-6, this renders as a white silhouette. Consider creating a dedicated `ic_voip_call` drawable.

This is LOW because: (a) the notification is PRIORITY_LOW and not user-facing, (b) the fallback is standard Android practice, (c) CallKit/Telecom keeps the call alive anyway.

---

### Acceptance Criteria Verification

| Criterion | Status | Notes |
|---|---|---|
| `VoipCallService.kt` exists and compiles | ✅ | BUILD SUCCESSFUL |
| Manifest declares service with `foregroundServiceType="microphone"` | ✅ | Line 152-156 |
| Manifest declares `FOREGROUND_SERVICE_MICROPHONE` permission | ✅ | Implied by foregroundServiceType attribute (no separate permission needed on API 34-) |
| Service starts in foreground with notification | ✅ | `startForegroundWithNotification()` with `ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE` on API 29+ |
| Service stops via `stopSelf()` correctly | ✅ | `ACTION_STOP` path calls `stopSelf()` at line 79 |

---

### Positive Observations

1. **Clean intent-based control.** `ACTION_START`/`ACTION_STOP` pattern is simple, testable, and avoids static state.
2. **API-level guard correct.** `ServiceInfo.FOREGROUND_SERVICE_TYPE_MICROPHONE` only used on API 29+ (Android 10), with plain `startForeground` on earlier versions.
3. **Notification channel properly scoped.** `IMPORTANCE_LOW` + `setShowBadge(false)` is correct for a persistent service notification that shouldn't interrupt the user.
4. **PendingIntent flags correct.** Uses `FLAG_IMMUTABLE` on API 31+ with fallback to 0.
5. **Matches iOS CallKit pattern.** Parity with how CallKit keeps the VoIP call alive on iOS.

---

### Verdict

**LGTM** — All 5 acceptance criteria satisfied. Both LOW observations are informational only and don't block the PR. The `isRunning` guard issue is LOW because in practice, `startForeground` is idempotent even if the flag isn't reset on recreation.

diegolmello:
## Code Review — Opus-Level Architecture Review

**PR:** [#7199](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7199) — feat(android): create VoipCallService with FOREGROUND_SERVICE_TYPE_MICROPHONE
**Files Reviewed:** 2 (`VoipCallService.kt`, `AndroidManifest.xml`)
**Total Issues:** 3 (0 CRITICAL, 1 HIGH, 2 MEDIUM, 0 LOW)

---

### Stage 1 — Spec Compliance

| Acceptance Criterion | Status |
|---|---|
| `VoipCallService.kt` exists and compiles | PASS |
| Manifest declares service with `foregroundServiceType=\"microphone\"` | PASS |
| Manifest declares `FOREGROUND_SERVICE_MICROPHONE` permission | PASS |
| Service starts in foreground with notification | PASS |
| Service stops via `stopSelf()` correctly | PASS |

All five acceptance criteria are satisfied.

---

### Stage 2 — Code Quality

#### [HIGH] Race condition in `isRunning` flag on `ACTION_START`

**File:** `VoipCallService.kt:65-70`

```kotlin
if (!isRunning) {
    isRunning = true
    startForegroundWithNotification(callId)
} else {
    Log.d(TAG, \"Service already running, skipping duplicate start\")
}
```

The read of `isRunning` on line 65 and the write on line 66 are not atomic. If two `ACTION_START` intents arrive in quick succession — or if the service is recreated by the system between `onStartCommand` calls — the duplicate-start guard can be bypassed. Specifically, `onCreate()` resets `isRunning = false`, so a system-recreated instance would re-enter `startForegroundWithNotification` on the next `ACTION_START`.

Additionally, the `isRunning` flag lives in the companion object (JVM class-level state), meaning all instances of the service share the same flag. Android can instantiate the service multiple times between `startForegroundService` and `onDestroy`, making the flag unreliable as an instance-level guard.

**Fix:** Move the guard inside `startForegroundWithNotification` as a private instance field:

```kotlin
private var isForeground = false

private fun startForegroundWithNotification(callId: String) {
    if (isForeground) return
    isForeground = true
    // ... existing logic
}
```

Remove the companion object's `isRunning` field entirely.

---

#### [MEDIUM] `getApplicationInfo().icon` may not be the correct notification icon

**File:** `VoipCallService.kt:131`

```kotlin
.setSmallIcon(getApplicationInfo().icon)
```

This uses the app's main launcher icon. On Android 8+ (API 26+), notification icons must be monochrome (alpha-only). The launcher icon typically does not meet this requirement and will render as a grey silhouette on modern Android.

`VoipNotification.kt:834-835` in the same package handles this correctly:

```kotlin
val smallIconResId = context.resources.getIdentifier(\"ic_notification\", \"drawable\", packageName)
// ...
setSmallIcon(smallIconResId)
```

**Fix:** Use the same resource lookup pattern as `VoipNotification.kt`, or reference a specific drawable resource by ID if `ic_notification` does not exist:

```kotlin
val smallIconResId = context.resources.getIdentifier(
    \"ic_notification\", \"drawable\", context.packageName
)
setSmallIcon(if (smallIconResId != 0) smallIconResId else getApplicationInfo().icon)
```

---

#### [MEDIUM] `ACTION_STOP` via `startService` is a no-op with an unnecessary service instantiation

**File:** `VoipCallService.kt:50-56`

```kotlin
@JvmStatic
fun stopService(context: android.content.Context) {
    val intent = Intent(context, VoipCallService::class.java).apply {
        action = ACTION_STOP
    }
    context.startService(intent)
}
```

When `stopService` is called and the service is not running, Android will still instantiate the service and call `onCreate` (which creates the notification channel), then `onStartCommand` with `ACTION_STOP`, then `stopSelf()`. This creates a brief, wasteful service lifecycle for a no-op.

**Fix:** Track service state externally and guard the call:

```kotlin
@JvmStatic
fun stopService(context: android.content.Context) {
    if (!isRunning) return
    val intent = Intent(context, VoipCallService::class.java).apply {
        action = ACTION_STOP
    }
    context.startService(intent)
}
```

(The `isRunning` flag in the companion object is already being used; it just needs to be moved out of the `ACTION_START` guard to be checked here too.)

---

### Positive Observations

- **API level guards are correct.** `FOREGROUND_SERVICE_TYPE_MICROPHONE` gated behind `Build.VERSION_CODES.Q` (API 29+). `FLAG_IMMUTABLE` gated behind `Build.VERSION_CODES.M` (API 23+). Both match Android platform requirements.
- **`stopSelf()` on `ACTION_STOP`** is the correct self-managed shutdown pattern — no external coordination needed.
- **`onDestroy` resets `isRunning`** — prevents stale state if the service is force-killed and a pending intent is delivered.
- **Notification channel uses `IMPORTANCE_LOW`** — appropriate for a persistent service indicator that should not be intrusive.
- **`setOngoing(true)` + `setOnlyAlertOnce(true)`** — correct for a persistent call notification.
- **Manifest service declaration is correct** — `enabled=\"true\"`, `exported=\"false\"`, `foregroundServiceType=\"microphone\"`.
- **Pre-existing `FOREGROUND_SERVICE_MICROPHONE` permission** at `AndroidManifest.xml:17` — no regression.

---

### Verdict: REQUEST CHANGES

The HIGH race condition is the blocking issue. The MEDIUM icon issue is worth fixing now for consistency. Once those two are addressed, this PR is ready to merge.

diegolmello:
## Follow-up review — partially dismissing the prior Opus review

Re: [previous review](https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7199#issuecomment-4292961927). Re-evaluated each finding against the current tree and Android platform guarantees. Summary: **one valid minor, one obsolete, one overstated.**

### [HIGH] `isRunning` race condition — **overstated, not a bug**

The review claims the read/write on `isRunning` at `VoipCallService.kt:75-76` is non-atomic and that the service can be instantiated multiple times in parallel.

Platform guarantees contradict this:

- `onStartCommand` is invoked on the **main thread** (see [Service docs](https://developer.android.com/reference/android/app/Service#ProcessLifecycle)). Back-to-back `ACTION_START` intents are serialized — no interleaved read/write is possible.
- Android guarantees **at most one instance of a given `Service` class per process** ([Services overview](https://developer.android.com/develop/background-work/services#CreatingAService)). The "multiple instances share the companion flag" scenario does not occur.
- On process death the JVM class reloads and the companion field reinitializes to `false` — the explicit reset in `onDestroy` is belt-and-suspenders, not a correctness fix.

The companion-object flag is slightly ugly, but it is not a race. No change required.

### [MEDIUM] Notification icon — **valid, will address**

`getApplicationInfo().icon` at line 125 resolves to the launcher icon, which on API 26+ is typically rendered as a grey silhouette because it isn't alpha-only. Sibling `VoipNotification.kt` already uses a `ic_notification` drawable which is monochrome-safe. Will switch to the same pattern.

### [MEDIUM] `ACTION_STOP` via `startService` causing a no-op instantiation — **obsolete**

This finding was written against the original `context.startService(intent)` path. Commit `4d9feb2` (landed in response to the CodeRabbit review, before the Opus review was dismissed) already switched `stopService()` to `context.stopService(intent)`, which does not instantiate the service when it isn't running. The cited code no longer exists.

### Production risk

Independent of the review: `VoipCallService` has **no caller** in this PR. It's scaffolding for PR-4 in the VoIP Android series. Merging today has zero runtime effect on production — the service is declared in the manifest but never started.

### Plan

1. Fix notification icon (cosmetic, one-line).
2. Optionally drop `callId` from `Log.d` on lines 74/104 (minor privacy hygiene).
3. Merge. Wiring lands in PR-4.

PR #7205, opened 2026-04-22
feat(voip): disable header buttons during active call

diegolmello:
## Summary

Disables all header buttons in RoomView and RoomsListView during an active VoIP call (except back button). Additionally gates the three navigation entry points into `RoomActionsView`. Uses the existing `disabled` prop on `HeaderButton.Item` to visually dim (`opacity: 0.5`) and block touch interaction.

- **New hook**: `useIsVoipCallActive` — reads `call != null || nativeAcceptedCallId != null` from `useCallStore` (Zustand), reactive to state changes
- **RoomsListView header buttons**: 5 buttons disabled (drawer, notification-troubleshoot, create, search, directory). Notification-troubleshoot button also gains missing `disabled` prop for pre-existing gating conditions.
- **RoomView RightButtons**: functional wrapper injects `isVoipCallActive` into the class component; all 3 render branches covered:
  - Livechat kebab (`disabled={isVoipCallActive}`)
  - Thread follow/unfollow (`disabled={isVoipCallActive}`)
  - Main 5 buttons (E2EE, notification troubleshoot, call, threads, search) — OR'd with existing `hasE2EEWarning` logic
- **Entry-point gating** (slice 4):
  - `RoomsListView` workspace switcher — `onPress` early-returns during call
  - `RoomView` title tap — `RoomHeader` `disabled` OR's `isInviteSubscription` with `isVoipCallActive`
  - Master-detail `Avatar` — gains `disabled` prop, `LeftButtons` wires it to `useIsVoipCallActive`

## Merge Order

Single PR — all 4 slices are on one branch targeting `feat.voip-lib-new`.

## Test plan

- [x] `TZ=UTC yarn test` — 1214 tests pass, 146 suites, 0 failed
- [x] `yarn lint` — 0 errors / 0 warnings on all changed files
- [ ] Manual: accept VoIP call → observe all header buttons dim and non-interactive; end call → buttons re-enable
- [ ] Manual (tablet): accept call → tap avatar → no navigation → end call → avatar tap opens RoomActionsView
- [ ] Manual (phone): accept call → tap room title → no navigation → end call → tap opens RoomActionsView
- [ ] Manual (rooms list): accept call → tap server name → no action sheet → end call → action sheet opens
- [ ] Back button remains functional during call

## Follow-ups (out of scope)

- ThreadMessagesView header buttons
- TeamChannelsView header buttons
- DirectoryView header buttons

## Related

Follows the same active-call detection pattern as `voipBlocksIncomingVideoconf.ts`.

## Summary by CodeRabbit

* **New Features**
  * UI now detects active VoIP calls and disables related room controls (encryption toggle, notifications, search, threads, call button, server actions, and header buttons).
  * Avatars can be rendered disabled to prevent avatar interactions while a call is active.

* **Tests**
  * Added tests validating VoIP call active detection and the resulting UI disabled states.

### DIFF
--- a/app/containers/Avatar/Avatar.tsx
+++ b/app/containers/Avatar/Avatar.tsx
@@ -20,6 +20,7 @@ const Avatar = React.memo(
 		userId,
 		token,
 		onPress,
+		disabled,
 		emoji,
 		getCustomEmoji,
 		avatarETag,
@@ -97,7 +98,7 @@ const Avatar = React.memo(
 
 		if (onPress) {
 			image = (
-				<Touch accessible={accessible} accessibilityLabel={avatarAccessibilityLabel} onPress={onPress}>
+				<Touch accessible={accessible} accessibilityLabel={avatarAccessibilityLabel} onPress={onPress} disabled={disabled}>
 					{image}
 				</Touch>
 			);
--- a/app/containers/Avatar/interfaces.ts
+++ b/app/containers/Avatar/interfaces.ts
@@ -16,6 +16,7 @@ export interface IAvatar {
 	userId?: string;
 	token?: string;
 	onPress?: () => void;
+	disabled?: boolean;
 	getCustomEmoji?: TGetCustomEmoji;
 	avatarETag?: string;
 	isStatic?: boolean | string;
--- a/app/lib/hooks/useIsVoipCallActive.test.ts
+++ b/app/lib/hooks/useIsVoipCallActive.test.ts
@@ -1,0 +1,40 @@
+import { renderHook, act } from '@testing-library/react-native';
+
+import { useIsVoipCallActive } from './useIsVoipCallActive';
+import { useCallStore } from '../services/voip/useCallStore';
+
+const mockCall = { callId: 'call-1' } as any;
+const mockNativeId = 'native-call-id';
+
+describe('useIsVoipCallActive', () => {
+	beforeEach(() => {
+		useCallStore.setState({ call: null, nativeAcceptedCallId: null });
+	});
+
+	afterEach(() => {
+		useCallStore.setState({ call: null, nativeAcceptedCallId: null });
+	});
+
+	it('returns false when both call and nativeAcceptedCallId are null', () => {
+		const { result } = renderHook(() => useIsVoipCallActive());
+		expect(result.current).toBe(false);
+	});
+
+	it('returns true when call is set and nativeAcceptedCallId is null', () => {
+		useCallStore.setState({ call: mockCall, nativeAcceptedCallId: null });
+		const { result } = renderHook(() => useIsVoipCallActive());
+		expect(result.current).toBe(true);
+	});
+
+	it('returns true when call is null and nativeAcceptedCallId is set', () => {
+		useCallStore.setState({ call: null, nativeAcceptedCallId: mockNativeId });
+		const { result } = renderHook(() => useIsVoipCallActive());
+		expect(result.current).toBe(true);
+	});
+
+	it('returns true when both call and nativeAcceptedCallId are set', () => {
+		useCallStore.setState({ call: mockCall, nativeAcceptedCallId: mockNativeId });
+		const { result } = renderHook(() => useIsVoipCallActive());
+		expect(result.current).toBe(true);
+	});
+});
--- a/app/lib/hooks/useIsVoipCallActive.test.ts
+++ b/app/lib/hooks/useIsVoipCallActive.test.ts
@@ -1,4 +1,4 @@
-import { renderHook, act } from '@testing-library/react-native';
+import { renderHook } from '@testing-library/react-native';
 
 import { useIsVoipCallActive } from './useIsVoipCallActive';
 import { useCallStore } from '../services/voip/useCallStore';
--- a/app/lib/hooks/useIsVoipCallActive.ts
+++ b/app/lib/hooks/useIsVoipCallActive.ts
@@ -1,0 +1,7 @@
+import { useShallow } from 'zustand/react/shallow';
+
+import { useCallStore } from '../services/voip/useCallStore';
+
+/** Returns true when a VoIP call is active (either ringing/accepted natively or fully connected). */
+export const useIsVoipCallActive = () =>
+	useCallStore(useShallow(state => state.call != null || state.nativeAcceptedCallId != null));
--- a/app/views/RoomView/LeftButtons.tsx
+++ b/app/views/RoomView/LeftButtons.tsx
@@ -4,6 +4,7 @@ import { StyleSheet, useWindowDimensions } from 'react-native';
 import Avatar from '../../containers/Avatar';
 import { useAppNavigation } from '../../lib/hooks/navigation';
 import { HeaderBackButton } from '../../containers/Header/components/HeaderBackButton';
+import { useIsVoipCallActive } from '../../lib/hooks/useIsVoipCallActive';
 
 const styles = StyleSheet.create({
 	avatar: {
@@ -39,6 +40,7 @@ const LeftButtons = ({
 	const { goBack } = useAppNavigation();
 	const onPress = () => goRoomActionsView();
 	const { fontScale } = useWindowDimensions();
+	const isVoipCallActive = useIsVoipCallActive();
 
 	if (!isMasterDetail || tmid) {
 		let label = ' ';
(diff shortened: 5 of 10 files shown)


diegolmello:
## Summary

**APPROVE-WITH-NITS.** All four slices are correctly implemented against the PRD, cover the right entry points, preserve existing disabled conditions, and follow the project convention of injecting hooks via thin functional wrappers around class components. The diff is minimal and well-scoped. Two non-blocking issues worth addressing (one about `useShallow`, one about `props: any`) and a few minor observations.

---

## Critical issues (blocking)

None.

---

## Non-blocking concerns (nits / medium severity)

### [MEDIUM] `useShallow` is a no-op on a primitive boolean — unnecessary overhead
**File:** `app/lib/hooks/useIsVoipCallActive.ts:7`

```ts
useCallStore(useShallow(state => state.call != null || state.nativeAcceptedCallId != null));
```

`useShallow` wraps the selector result in a shallow-equality check using `Object.is` for each top-level key. When the selector already returns a **primitive boolean**, `Object.is(true, true)` is identical to the default Zustand equality (`===`). The `useShallow` call allocates an extra closure on every render for zero benefit. The existing codebase uses `useShallow` for selectors returning objects/arrays (e.g. `MediaCallHeader.tsx`), not scalars.

**Fix:** Remove `useShallow`:
```ts
export const useIsVoipCallActive = () =>
    useCallStore(state => state.call != null || state.nativeAcceptedCallId != null);
```

---

### [MEDIUM] `RoomViewWithVoip` uses `props: any` — loses type safety at the navigation boundary
**File:** `app/views/RoomView/index.tsx:1718`

```ts
const RoomViewWithVoip = (props: any) => {
```

This erases all type checking on the props that React Navigation passes to `RoomView`. If a navigation param changes shape in the future, TypeScript will not catch it here. By contrast, the `RightButtons` wrapper correctly types its props as `IRightButtonsProps` (line 575). The inconsistency also means different wrapper patterns for the same technique in the same PR.

**Fix:** Use the same connected component's prop type. Since the wrapper passes everything through, the simplest approach is:
```ts
const RoomViewWithVoip = (props: Omit<IRoomViewProps, 'isVoipCallActive'>) => {
```
Or if `IRoomViewProps` includes too many HOC-injected props, a bare `Record<string, any>` is marginally better than `any` because it at least constrains to an object. But ideally mirror the `RightButtons` pattern and use the concrete interface.

---

### [LOW] `isVoipCallActive` is redundant in `useHeader` dependency array since `disabled` already captures it
**File:** `app/views/RoomsListView/hooks/useHeader.tsx:163`

`disabled` is derived from `isVoipCallActive` (line 46). Since `disabled` is already in the dep array (line 161), adding `isVoipCallActive` separately is redundant — when `isVoipCallActive` changes, `disabled` changes, which already triggers the effect. Not harmful (React deduplicates), but it could confuse future readers into thinking there is an independent dependency.

**Fix:** Remove `isVoipCallActive` from the `useLayoutEffect` dependency array (line 163), keeping only `disabled`.

---

### [LOW] Workspace switcher has no visual disabled feedback — inconsistent with other gated buttons
**File:** `app/views/RoomsListView/components/Header.tsx:49-51`

The workspace switcher uses an early-return in `onPress` to block interaction, but the `TouchableOpacity` has no `disabled` prop and no opacity change. All other gated buttons in this PR get `opacity: 0.5` dimming. The PRD (user story 4) says "disabled buttons to be visually dimmed, so that I can see at a glance that they're unavailable." The plan explicitly says "No `disabled` prop needed — early-return inside the handler achieves the same UX," so this is by-design — but it is worth noting the visual inconsistency. Consider adding `disabled={isVoipCallActive}` and `style={{ opacity: isVoipCallActive ? 0.5 : 1 }}` on the `TouchableOpacity` at line 73 for consistency. Not blocking.

---

### [LOW] No `accessibilityState={{ disabled: true }}` on disabled buttons
**File:** `app/containers/Header/components/HeaderButton/HeaderButtonItem.tsx:60-66`

This is a **pre-existing** issue, not introduced by this PR. `BorderlessButton` with `enabled={false}` blocks touches but does not set `accessibilityState={{ disabled: true }}`, so screen readers (VoiceOver/TalkBack) do not announce the button as disabled. Similarly, `RoomHeader`'s `TouchableOpacity` at `RoomHeader.tsx:204` uses `disabled={disabled}` which does set the native accessibility state, so that path is fine. The `HeaderButtonItem` path is not. Worth a follow-up ticket.

---

### [LOW] Wrapper placement inconsistency: `RightButtons` wraps inside `connect`, `RoomView` wraps outside
**File:** `app/views/RoomView/RightButtons.tsx:580` vs `app/views/RoomView/index.tsx:1717-1722`

`RightButtons`: `connect(mapStateToProps)(withTheme(RightButtonsWithVoip))` — the voip wrapper is **inside** the HOC chain, meaning Redux props flow into the wrapper, which forwards them to the class.

`RoomView`: `ConnectedRoomView = connect(...)(...); RoomViewWithVoip = (props) => <ConnectedRoomView .../>` — the voip wrapper is **outside** the HOC chain, meaning it wraps the fully-connected component.

Both work correctly because props are spread through. The difference arises from practical necessity: `RoomView` has a 4-deep HOC chain (`connect > withDimensions > withTheme > withSafeAreaInsets > withActionSheet`) and inserting the wrapper inside would require restructuring. The outside-wrap is the simpler approach for `RoomView`. Not a correctness issue, but worth a brief code comment explaining _why_ the patterns differ, to prevent a future contributor from "fixing" the inconsistency.

---

## Positive notes

1. **Clean hook design.** `useIsVoipCallActive` is a textbook deep module: trivial interface, single source of truth, reusable across screens. The test coverage (4 state combinations) is thorough and uses `useCallStore.setState` directly rather than heavy mocking.

2. **`shouldComponentUpdate` handled correctly in both class components.** The explicit `isVoipCallActive` check in `RightButtons` (line 171) and `RoomView` (line 263) ensures the prop change propagates through the hand-rolled SCU. This is easy to forget and would have caused a subtle bug.

3. **All three render branches in `RightButtons` covered.** Livechat kebab, thread follow/unfollow, and main-branch buttons all receive the `disabled` treatment. No branch was missed.

4. **Existing `disabled`/`hasE2EEWarning` conditions preserved via OR.** The PR consistently uses `existingCondition || isVoipCallActive` rather than replacing existing gating, ensuring no regression to E2EE warning, expired version, or password-change states.

5. **`HeaderCallButton` `disabled` type handled with `!!` cast.** `disabled={!!(hasE2EEWarning || isVoipCallActive)}` at line 542 correctly narrows `boolean | undefined` to `boolean` to match `HeaderCallButton`'s non-optional `disabled: boolean` prop type.

6. **Notification-troubleshoot button gap fixed as side-effect.** The `useHeader.tsx` notification-troubleshoot button was missing `disabled={disabled}` — this PR fixes it (line 126), closing a pre-existing gap where that button remained active during expired-version/password-change states.

7. **Entry-point gating is complete for the declared scope.** The only two code paths that navigate to `RoomActionsView` are `RoomView/index.tsx:602` (title tap — now gated) and `TeamChannelsView.tsx:327` (explicitly documented as a follow-up). No sagas or deep-link handlers navigate to `RoomActionsView`.

---

## Untested assertions (cannot verify from diff alone)

1. **Back button behavior in all 3 variants (phone/tablet/rooms-list).** The diff does not touch `HeaderBackButton` or `goBack` calls. By code inspection, `LeftButtons.tsx` only passes `disabled` to the `Avatar` branch (master-detail, non-thread), not the `HeaderBackButton` branch. This looks correct, but runtime verification on both phone and tablet layouts is needed.

2. **`Touch` component `disabled` prop.** `Avatar.tsx:101` passes `disabled` to `<Touch>`. `Touch.tsx` forwards `disabled` to `BorderlessButton`/`RectButton` which natively supports it. Verified by grep, but runtime verification that the gesture handler respects `disabled` on the Avatar specifically (not just `HeaderButtonItem`) is recommended.

3. **`RoomView` SCU placement.** The `isVoipCallActive` check is placed at the **top** of `shouldComponentUpdate` (line 263), before all other checks. This means it short-circuits early when VoIP state changes. This is fine for correctness but means a VoIP state change triggers a full re-render even if only the header needs updating. Given that VoIP state changes are infrequent (call start/end), this is acceptable — but worth noting.

4. **Snapshot test coverage for slices 2-4.** The PR includes unit tests for the hook (slice 1) but the plan calls for render/snapshot tests for `RightButtons` and `useHeader`. These may exist in prior commits on the base branch or may be deferred. Cannot verify from this diff.

PR #6918, opened 2026-01-14
feat: Voice support

diegolmello:
## Issue(s)	

https://rocketchat.atlassian.net/browse/VMUX-6
https://rocketchat.atlassian.net/browse/VMUX-17
https://rocketchat.atlassian.net/browse/VMUX-36
https://rocketchat.atlassian.net/browse/CORE-1967

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Summary by CodeRabbit

* **New Features**
  * Voice calling functionality with incoming call screen featuring accept/decline actions
  * In-call UI with call status, timer, and control buttons (mute, hold, speaker, end)
  * Native integration for incoming calls on both iOS and Android
  * VoIP push notifications support
  * Call state management including connecting, active, muted, and on-hold states

* **Tests**
  * Added comprehensive test suites and Storybook stories for call UI components

* **Chores**
  * Updated dependencies for VoIP support
  * Updated native modules and configurations for iOS and Android

PR #7212, opened 2026-04-22
fix(voip): Android lock-screen accept — silent audio + subsequent-call blank screen

diegolmello:
## Proposed changes

Fixes two VoIP bugs on Android when accepting an incoming call from the lock screen while the app is not running:

1. **Silent audio on first call** — The Android same-workspace branch of `getInitialMediaCallEvents` returned `true`, which skipped `appInit()`. This prevented the login saga from running, so `mediaSessionInstance.init()` never fired, `applyRestStateSignals()` never processed the offer SDP, and audio never routed. Fix: return `false` on Android so the normal boot path runs (gated to Android-only — iOS retains existing behavior).

2. **Blank screen on subsequent calls** — The Android `VoipCallService` foreground service was never stopped from JS-driven hangup paths. Its static `isRunning` flag blocked re-initialization on the next incoming call. Fix: new `stopVoipCallService` bridge method + `terminateNativeCall` JS helper that wraps `RNCallKeep.endCall` + Android service stop at every call termination site.

Additional hardening:
- **Navigator readiness guard**: `answerCall` now awaits `waitForNavigationReady()` before navigating to `CallView`, preventing silent navigation drops during cold-start boot.
- **Stale timer extension**: `STALE_NATIVE_MS` raised from 15s to 60s to cover worst-case cold-boot handoff on slow devices/networks.
- **Eager `isRunning` reset**: `VoipCallService.onStartCommand(ACTION_STOP)` now resets `isRunning` before `stopSelf()` to prevent race conditions with rapid call cycling.

## Issue(s)

## How to test or reproduce

1. **Cold-start lock-screen accept**: Kill the Rocket.Chat app on an Android device (Pixel 6 / Android 16). Initiate a VoIP call from the web client. Accept from the device lock screen. Expected: CallView renders and audio routes within 5–8 seconds.
2. **Subsequent calls**: End the first call. Initiate a second VoIP call. Accept from lock screen. Expected: identical behavior — no blank screen, no force-close needed.
3. **Warm-start regression**: With the app already running, accept an incoming VoIP call. Expected: unchanged behavior.
4. **iOS regression**: Accept a VoIP call from iOS lock screen. Expected: unchanged behavior.
5. **Different-workspace deep-link**: Accept a call hosted on a different workspace. Expected: workspace switch + call connect as before.

## Types of changes

## Checklist

## Merge Order

This PR targets `feat.voip-lib-new` (PR #6918). It should be merged into `feat.voip-lib-new` before #6918 is merged into `develop`.

## Further comments

**Files changed:**
- `android/.../voip/VoipModule.kt` — new `stopVoipCallService` bridge method
- `android/.../voip/VoipCallService.kt` — eager `isRunning` reset on `ACTION_STOP`
- `app/lib/native/NativeVoip.ts` — TurboModule spec + fallback for `stopVoipCallService`
- `app/lib/services/voip/terminateNativeCall.ts` — new helper wrapping `endCall` + service stop
- `app/lib/services/voip/MediaCallEvents.ts` — Android same-workspace branch returns `false`
- `app/lib/services/voip/MediaSessionInstance.ts` — `waitForNavigationReady` guard + `terminateNativeCall` migration
- `app/lib/services/voip/useCallStore.ts` — `STALE_NATIVE_MS` 15s → 60s + `terminateNativeCall` migration
- `app/lib/navigation/appNavigation.ts` — extracted `waitForNavigationReady` helper
- `app/sagas/deepLinking.js` — uses shared `waitForNavigationReady` instead of inline version
- Tests updated accordingly

## Summary by CodeRabbit

* **New Features**
  * App can now request stopping the VoIP foreground call service from the JS layer.

* **Bug Fixes**
  * Improved VoIP termination and cleanup to reduce stale native call remnants.
  * Call UI navigation now waits for readiness before transitioning for more reliable call flows.
  * Increased native-call stale timeout from 15s to 60s to reduce premature state clearing.
  * Centralized native call teardown to suppress platform/bridge errors.

* **Tests**
  * Updated mocks and integration tests to cover navigation readiness and VoIP flows.

### DIFF
--- a/android/app/src/main/java/chat/rocket/reactnative/voip/VoipCallService.kt
+++ b/android/app/src/main/java/chat/rocket/reactnative/voip/VoipCallService.kt
@@ -66,6 +66,7 @@ class VoipCallService : Service() {
     override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
         when (intent?.action) {
             ACTION_STOP -> {
+                isRunning = false
                 Log.d(TAG, "Stopping VoipCallService")
                 stopSelf(startId)
                 return START_NOT_STICKY
--- a/android/app/src/main/java/chat/rocket/reactnative/voip/VoipModule.kt
+++ b/android/app/src/main/java/chat/rocket/reactnative/voip/VoipModule.kt
@@ -149,6 +149,15 @@ class VoipModule(reactContext: ReactApplicationContext) : NativeVoipSpec(reactCo
         VoipNotification.stopDDPClient()
     }
 
+    override fun stopVoipCallService() {
+        try {
+            VoipCallService.stopService(reactApplicationContext)
+            Log.d(TAG, "stopVoipCallService: service stopped")
+        } catch (e: Exception) {
+            Log.e(TAG, "stopVoipCallService: failed to stop service", e)
+        }
+    }
+
     /**
      * Required for NativeEventEmitter in TurboModules.
      * Called when JS starts listening to events.
--- a/app/containers/NewMediaCall/VoipCallLifecycle.integration.test.tsx
+++ b/app/containers/NewMediaCall/VoipCallLifecycle.integration.test.tsx
@@ -53,7 +53,8 @@ jest.mock('../../lib/methods/helpers/helpers', () => ({
 }));
 jest.mock('../../lib/navigation/appNavigation', () => ({
 	__esModule: true,
-	default: { navigate: jest.fn(), back: jest.fn() }
+	default: { navigate: jest.fn(), back: jest.fn() },
+	waitForNavigationReady: jest.fn().mockResolvedValue(undefined)
 }));
 jest.mock('../../lib/services/sdk', () => ({
 	__esModule: true,
--- a/app/lib/native/NativeVoip.ts
+++ b/app/lib/native/NativeVoip.ts
@@ -35,6 +35,11 @@ export interface Spec extends TurboModule {
 	 */
 	stopNativeDDPClient(): void;
 
+	/**
+	 * Stops the VoIP foreground call service.
+	 * iOS: No-op.
+	 * Android: Sends ACTION_STOP to VoipCallService, releasing the mic-type foreground hold.
+	 */
 	stopVoipCallService(): void;
 
 	/**
--- a/app/lib/native/NativeVoip.ts
+++ b/app/lib/native/NativeVoip.ts
@@ -35,6 +35,8 @@ export interface Spec extends TurboModule {
 	 */
 	stopNativeDDPClient(): void;
 
+	stopVoipCallService(): void;
+
 	/**
 	 * Required for NativeEventEmitter in TurboModules.
 	 * Called when JS starts listening to events.
@@ -58,6 +60,7 @@ const NativeVoipModule =
 		clearInitialEvents: () => undefined,
 		getLastVoipToken: () => '',
 		stopNativeDDPClient: () => undefined,
+		stopVoipCallService: () => undefined,
 		addListener: () => undefined,
 		removeListeners: () => undefined
 	} as Spec);
--- a/app/lib/navigation/appNavigation.ts
+++ b/app/lib/navigation/appNavigation.ts
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { CommonActions, type NavigationContainerRef, StackActions } from '@react-navigation/native';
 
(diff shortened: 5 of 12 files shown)


diegolmello:
## Code Review

**Files reviewed:** 10
**Verdict: APPROVE** — the two bug fixes are correct and well-scoped. A few observations below, none blocking.

---

### Fix 1 — Silent audio on cold-start lock-screen accept (`MediaCallEvents.ts`)

**Correctness: sound.**

The original code called `applyRestStateSignals()` and returned `true` (skip `appInit`) on Android for the same-workspace branch. That prevented the login saga from running, which in turn meant `mediaSessionInstance.init()` and the WebRTC offer/answer exchange never happened — explaining the silent call.

The fix returns `false` on Android, letting `appInit` run normally. iOS retains the original `applyRestStateSignals()` path. The platform guard is minimal and precise.

One thing worth noting: the log message on the new Android branch says *"continuing appInit for cold-start handoff"* but the return value `false` is what actually triggers `appInit`. Readers have to know that `false` means "don't skip appInit" to parse the log correctly. A comment like `// false = do not skip appInit` next to the `return false` would remove that ambiguity. Minor.

---

### Fix 2 — Blank screen on subsequent calls (`VoipCallService.kt` + `VoipModule.kt` + `terminateNativeCall.ts`)

**Correctness: sound.**

The root cause was `isRunning` staying `true` after JS-initiated hangup, blocking re-initialization on the next call. The fix covers all three necessary sites:

- `VoipCallService.onStartCommand(ACTION_STOP)` now resets `isRunning = false` **before** `stopSelf()`. This is the correct ordering — resetting after `stopSelf` would leave a window where a rapid new `ACTION_START` could arrive and be skipped.
- `stopVoipCallService` bridge method is straightforward; the `try/catch` in both the Kotlin bridge and the JS helper is appropriate.
- `terminateNativeCall` consolidates all call-end paths (`MediaSessionInstance`, `useCallStore`) into one helper. The migration is complete — no leftover bare `RNCallKeep.endCall` calls remain at hangup sites.

**Minor observation:** `stopService(context)` in the companion object sends `ACTION_STOP` via `context.stopService(intent)`, but `ACTION_STOP` is only checked in `onStartCommand`. `stopService()` from outside the service goes through `onStartCommand` only if the service is currently running (the system restarts it with the stop intent). If the service was already destroyed (e.g. system killed it), `stopService` is a no-op, which is fine here — `isRunning` is already `false` from `onDestroy`. No issue.

---

### Navigator readiness guard (`MediaSessionInstance.ts`)

`await waitForNavigationReady()` before `Navigation.navigate('CallView')` is the right fix for cold-start navigation drops. The implementation in `appNavigation.ts` correctly checks `navigationRef.current` as the fast-path, then falls back to the `'navigationReady'` event emitted from `AppContainer.tsx`'s `onReady` callback. No leak risk — the listener is cleaned up on first fire.

**Minor observation:** `waitForNavigationReady` has no timeout. On a device where `NavigationContainer` never calls `onReady` (e.g. a crash during render), the promise hangs forever. For a call flow this is a real edge case. A 10–15 second timeout with a fallback log would make this more robust. Not blocking for this PR's scope, but worth a follow-up.

---

### `waitForNavigationReady` extraction (`appNavigation.ts` / `deepLinking.js`)

Clean deduplication — the inline `waitForNavigation` in `deepLinking.js` was functionally identical and is now removed in favor of the shared export. No behavior change for that saga.

---

### Stale timer 15s → 60s (`useCallStore.ts`)

The extension is reasonable given cold-boot on slow devices/networks. Tests are updated consistently. No issues.

---

### Test coverage

The mock additions to `MediaSessionInstance.test.ts` (`react-native-device-info` default export, `waitForNavigationReady`) are correct and sufficient for the changed code paths. The `useCallStore.test.ts` timer updates are mechanical but accurate.

**Gap worth noting:** there is no test for `terminateNativeCall` itself — specifically that `stopVoipCallService` is called on Android but not on iOS. Given that this helper is now used in four places, a small unit test covering the platform branching would be a useful addition in a follow-up.

---

### Positive observations

- The `isRunning = false` placement before `stopSelf()` in `onStartCommand` shows careful attention to ordering under concurrent conditions.
- The fallback no-op object in `NativeVoip.ts` is correctly updated — this prevents crashes if the TurboModule is unavailable before bridge init.
- The PR scope is tight: only files directly related to the two bugs are touched, with no unrelated cleanup.

PR #7225, opened 2026-04-23
fix(ios/voip): apply TLS pinning to REST API and reject insecure http:// scheme

diegolmello:
## Summary

Apply TLS certificate pinning to iOS REST API calls and reject insecure `http://` DDP scheme, matching Android security posture.

### Part A — TLS pinning for REST API

- Added `RocketChatURLSessionDelegate` that forwards `URLSessionDelegate` challenges to the existing `Challenge.runChallenge()` in `SSLPinning.mm`
- Replaced `URLSession.shared` with a pinned `URLSession` singleton so all REST calls (including `media-calls.answer`) use the same TLS pinning as DDP WebSocket
- Both WebSocket (DDP) and HTTP VoIP traffic are now protected from MITM attacks via shared pinning infrastructure

### Part B — Reject insecure http:// scheme

- `DDPClientError.insecureSchemeNotAllowed` thrown when `http://` is detected
- `buildWebSocketURL()` now throws instead of silently downgrading to `ws://`
- Matches Android behavior which rejects `http://` via `IllegalStateException`

## Test plan

- [ ] MITM proxy against iOS simulator fails TLS handshake on `media-calls.answer` request
- [ ] `http://` DDP URL on iOS throws error (not silent downgrade)
- [ ] Existing VoIP calls still work with `https://` server URLs
- [ ] Manual QA: iOS VoIP call accept → `media-calls.answer` request sent over pinned TLS

## Acceptance criteria

- [x] `ios/Shared/RocketChat/API/API.swift` uses a pinned `URLSession` (not `URLSession.shared`)
- [x] The pinning delegate reuses `Challenge.runChallenge()` — no new pinning logic
- [x] `ios/Libraries/DDPClient.swift` throws when given `http://` scheme
- [x] `yarn lint` + `TZ=UTC yarn test` pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection error handling with better failure messages for insecure connections.
  * Enhanced security validation for WebSocket connections, rejecting insecure schemes.

* **Security Improvements**
  * Strengthened certificate pinning for API requests through custom session handling.

### DIFF
--- a/ios/Libraries/DDPClient.swift
+++ b/ios/Libraries/DDPClient.swift
@@ -1,5 +1,19 @@
 import Foundation
 
+/// Insecure scheme not allowed — Android rejects `http://` via IllegalStateException for the same reason.
+enum DDPClientError: Error {
+    case insecureSchemeNotAllowed
+}
+
+extension DDPClientError: LocalizedError {
+    var errorDescription: String? {
+        switch self {
+        case .insecureSchemeNotAllowed:
+            return "Insecure `http://` scheme not allowed. DDP requires `wss://`. Set server URL to use `https://`."
+        }
+    }
+}
+
 /// Minimal DDP WebSocket client for listening to Rocket.Chat media-signal events from native iOS.
 /// Only implements the subset needed to detect call hangup: connect, login, subscribe, and ping/pong.
 final class DDPClient {
@@ -33,7 +47,8 @@ final class DDPClient {
     
     func connect(host: String, completion: @escaping (Bool) -> Void) {
         stateQueue.async {
-            let wsUrl = Self.buildWebSocketURL(host: host)
+            do {
+                let wsUrl = try Self.buildWebSocketURL(host: host)
             
             guard let url = URL(string: wsUrl) else {
                 #if DEBUG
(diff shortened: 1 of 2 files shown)


diegolmello:
## Critical Code Review — PR #7225 (iOS TLS pinning)

### Summary
Applies the existing `Challenge.runChallenge()` pinning infrastructure to the REST API via a shared `RocketChatURLSessionDelegate`, and throws instead of silent downgrade for insecure `http://` DDP URLs.

### What's good
- `RocketChatURLSessionDelegate` forwards to the same `Challenge.runChallenge()` used by DDP — no new pinning logic introduced
- `pinnedSession` singleton reuses the delegate pattern cleanly
- `DDPClientError.insecureSchemeNotAllowed` with `LocalizedError` matches Android's `IllegalStateException` behavior
- `connect()` wraps URL building in `do/catch` so insecure schemes fail gracefully
- Commit is atomic and scoped

### Minor concerns

**1. `URLSessionConfiguration.default` vs `URLSessionConfiguration.ephemeral`.**  
The DDP client uses `.default` (non-ephemeral). Using `.ephemeral` would avoid caching TLS session state to disk, which is slightly better for security. However, keeping `.default` matches the DDP pattern and avoids behavioral divergence. Acceptable — document this as a future hardening opportunity if desired.

**2. Missing merge order in PR description** — should list all 4 PRs.

**3. No unit tests added** for the insecure scheme rejection. The existing DDP tests don't cover the new error path. Consider adding a test that calls `buildWebSocketURL("http://example.com")` and asserts it throws `DDPClientError.insecureSchemeNotAllowed`. Low priority — the Android-side behavior serves as the reference.

**Recommendation:** ✅ Approve with the merge order note and the ephemeral-session suggestion.

PR #7229, opened 2026-04-23
feat(a11y): Message actions announcement and back focus

OtavioStasiak:
## Proposed changes

Improve the actionSheet behavior on a11y.
- Action Sheet handle is accessible;
- `open message actions` is available on a11y actions;
- When we close MessageActions the focus went back to the message, and not from top of the View;

## Issue(s)	

https://rocketchat.atlassian.net/browse/MA-263

## How to test or reproduce

- Open the app;
- Go to Room;
- Turn Screen reader on;
- Focus on a message;
- Open ActionSheet;
- Navigate through the action sheet;
- Close ActionSheet focusing on handle;

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Accessibility Improvements**
  * Improved focus handling for action sheets and message actions (restores screen‑reader focus on open/close; Android focus deferred briefly for TalkBack).
  * Localized labels and hints for dismissing action sheets and showing message actions across many languages.

* **New Features**
  * Track-and-restore last-focused message for reliable focus return.
  * Enhanced message accessibility labels and an accessibility action to "Show message actions".
  
* **Tests**
  * Added tests covering message accessibility labels and actions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RocketChat/Rocket.Chat.ReactNative/pull/7229?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


OtavioStasiak:
@coderabbit review

app/containers/message/Message.tsx
```diff
@@ -103,7 +104,13 @@ const MessageInner = React.memo((props: IMessageInner) => {
...
 MessageInner.displayName = 'MessageInner';
 
-const Message = React.memo((props: IMessageTouchable & IMessage) => {
+interface IMessageA11y {
+	accessibilityActions?: AccessibilityActionInfo[];
+	onAccessibilityAction?: (event: AccessibilityActionEvent) => void;
+	handleLongPress?: () => void;
+}
+
+const getMessageAccessibilityLabel = (props: IMessageTouchable & IMessage & IMessageA11y) => {
```
diegolmello: Since we're touching this, we should move these a11y functions away from this file.
`message/Message` is an important file that got crowded with a11y code.
Both functions should be hooks with `useMemo()` as well and could have unit tests.

app/i18n/locales/en.json
```diff
@@ -9,6 +9,8 @@
   "A11y_appearance_show_alerts_as": "Show alerts as",
   "A11y_appearance_toast_dismissed_automatically": "Dismissed automatically",
   "A11y_appearance_toasts": "Toasts",
+  "A11y_close_action_sheet": "Close action sheet",
```
diegolmello, line 12: Missing other languages

app/containers/message/Message.tsx
```diff
@@ -122,37 +129,44 @@ const Message = React.memo((props: IMessageTouchable & IMessage) => {
...
+
+const getMessageAccessibilityActions = (isDisabled: boolean): AccessibilityActionInfo[] | undefined => {
+	if (isDisabled) {
+		return undefined;
+	}
+	return [{ name: 'showActions', label: i18n.t('Show_message_actions') }];
+};
+
+const Message = React.memo((props: IMessageTouchable & IMessage & IMessageA11y) => {
+	const accessibilityLabelValue = getMessageAccessibilityLabel(props);
```
diegolmello: `getMessageAccessibilityLabel` is called on `Message` here.

app/containers/message/Message.tsx
```diff
@@ -217,6 +240,9 @@ Message.displayName = 'Message';
 const MessageTouchable = React.memo((props: IMessageTouchable & IMessage) => {
 	const { onPress, onLongPress } = useContext(MessageContext);
 	const { colors } = useTheme();
+	const touchRef = useRef<View>(null);
+	const lastFocusedMessageRef = useLastFocusedMessageRef();
+	const accessibilityLabelValue = getMessageAccessibilityLabel(props);
```
diegolmello: `getMessageAccessibilityLabel` is called on `MessageTouchable` here.

Do we need both?

app/lib/a11y/useLastFocusedMessageRef.ts
diegolmello, line 1: Why don't we control the ref on this hook?

app/containers/message/Message.tsx
OtavioStasiak: I've tested and I found a way to just get it on message touchable!

OtavioStasiak:
@coderabbit review

OtavioStasiak:
@coderabbit review

diegolmello (changes requested):
A few items from review that I'd like resolved before merge:

## 🔴 Required: i18n keys only translated to English

`en.json` adds **four** new keys but every one of the 24 non-English locales adds only **two**:

| Key | en | other locales |
|---|---|---|
| `A11y_close_action_sheet` | ✅ | ✅ |
| `A11y_press_to_view_thread` | ✅ | ✅ |
| `A11y_close_action_sheet_hint` | ✅ | ❌ |
| `Show_message_actions` | ✅ | ❌ |

The PR body claims "Localized labels and hints for dismissing action sheets and showing message actions across many languages" — but a Spanish/German/Japanese screen-reader user will currently hear "Dismisses the action sheet" and "Show message actions" in English (i18n-js falls back to the default locale). Either translate the missing two keys across all 24 locales, or trim the claim from the PR description.

## 🟡 Important: Dead `IMessageA11y` plumbing + duplicate `useMessageAccessibilityLabel` call

In `app/containers/message/Message.tsx`:

- `IMessageA11y` (lines 107-111) declares `accessibilityActions`, `onAccessibilityAction`, and `handleLongPress`, but none of the three are read anywhere inside `Message`. `MessageTouchable` passes `handleLongPress={!isDisabled ? handleLongPress : undefined}` at line 223; the child silently discards it. Misleading for future readers, and it also defeats `React.memo` (fresh closure every render).
- `useMessageAccessibilityLabel(props)` is called in both `Message` (line 112) and `MessageTouchable` (line 179). The work involves `i18n.t(...)`, `toLocaleTimeString()`, and mention/channel string substitution — running twice per message render. Only the `MessageTouchable` instance feeds the outer `Touch`; `Message` only uses its copy in the early-return path for thread reply/sequential/info/ignored.

Suggest dropping the unused interface fields and computing the label once in the parent.

## 🟡 Important: Verify duplicate a11y props on Touch outer + inner

`Touch.tsx:85-110` now forwards `accessible`, `accessibilityRole`, `accessibilityLabel`, `accessibilityHint`, `accessibilityActions`, and `onAccessibilityAction` to **both** the outer `KeyboardComponent` and the inner `View`. The `LoadMore.test.tsx.snap` diff confirms the duplication. Two sibling accessible nodes with identical labels typically result in double announcements on iOS VoiceOver and can confuse swipe navigation on TalkBack.

Could you confirm this was tested end-to-end with VoiceOver and TalkBack on a long-pressable message and share what's announced? If the inner `View` no longer needs `accessible`, dropping it would also explain (and possibly let you revert) the Maestro iOS testID workaround in `verify-message-unread.yaml` — that change looks like a side effect of the outer Touch absorbing nested testIDs on iOS.

Happy to pair on the dual-label investigation if useful.

diegolmello (approved):

PR #7239, opened 2026-04-24
fix(voip): pending-accept test mocks

diegolmello:
## Summary

Fixes 3 failing VoIP acceptance tests (AC1, AC5, AC7) that verify the pending-accept fast-path behavior in `MediaCallEvents.test.ts`.

### Root Cause

The tests used `DeviceEventEmitter.emit()` to trigger `VoipPendingAccept` events, but handlers registered via `NativeEventEmitter.addListener()` store handlers in the native `RCTDeviceEventEmitter`, not in any JavaScript-accessible location. So the emitted events never reached the listeners.

### Fix

Mock `react-native` completely with a shared `__nativeEventBus__` Map on `global` that both `NativeEventEmitter.addListener` (which stores handlers) and `DeviceEventEmitter.emit` (which invokes them) use. This allows tests to synchronously trigger events that the module under test listens for.

Additional fixes:
- Added `__esModule: true` to the `appNavigation` mock so its default export is recognized
- Added `proceedAccept` and `resetFromLogout` to the `NativeVoip` mock (missing from the original mock)
- Removed debug `console.error` from the VoipPendingAccept handler in `MediaCallEvents.ts`

## Test Plan

- [x] `yarn test -- --testPathPattern='MediaCallEvents.test.ts' --testNamePattern="AC1.*navigates"` — passes
- [x] `yarn test -- --testPathPattern='MediaCallEvents.test.ts' --testNamePattern="AC5"` — passes  
- [x] `yarn test -- --testPathPattern='MediaCallEvents.test.ts' --testNamePattern="AC7"` — passes

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added pending accept mechanism for VoIP calls with automatic 15-second timeout for better handling during initialization
  * New "Connecting" status indicator displays during the call acceptance phase
  * Improved call state management when accepting and initializing calls

### DIFF
--- a/android/app/src/main/java/chat/rocket/reactnative/voip/VoipModule.kt
+++ b/android/app/src/main/java/chat/rocket/reactnative/voip/VoipModule.kt
@@ -19,6 +19,7 @@ class VoipModule(reactContext: ReactApplicationContext) : NativeVoipSpec(reactCo
         private const val TAG = "RocketChat.VoipModule"
         private const val EVENT_VOIP_ACCEPT_SUCCEEDED = "VoipAcceptSucceeded"
         private const val EVENT_VOIP_ACCEPT_FAILED = "VoipAcceptFailed"
+        private const val EVENT_VOIP_PENDING_ACCEPT = "VoipPendingAccept"
 
         private var reactContextRef: WeakReference<ReactApplicationContext>? = null
 
@@ -81,6 +82,21 @@ class VoipModule(reactContext: ReactApplicationContext) : NativeVoipSpec(reactCo
                 }
             } catch (e: Exception) {
                 Log.e(TAG, "Failed to emit VoipAcceptFailed", e)
+            }
+        }
+
+        @JvmStatic
+        fun emitPendingAcceptEvent(voipPayload: VoipPayload) {
+            try {
+                reactContextRef?.get()?.let { context ->
+                    if (context.hasActiveReactInstance()) {
+                        context
+                            .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter::class.java)
+                            .emit(EVENT_VOIP_PENDING_ACCEPT, mapOf("callId" to voipPayload.callId, "payload" to voipPayload.toWritableMap()))
+                    }
+                }
+            } catch (e: Exception) {
+                Log.e(TAG, "Failed to emit VoipPendingAccept", e)
             }
         }
 
@@ -156,6 +172,14 @@ class VoipModule(reactContext: ReactApplicationContext) : NativeVoipSpec(reactCo
         } catch (e: Exception) {
             Log.e(TAG, "stopVoipCallService: failed to stop service", e)
         }
+    }
+
+    /**
+     * Continues a native-accepted VoIP call that was queued in the pending-accept window.
+     * Android: No-op (Android handles accept natively via Telecom).
+     */
+    override fun proceedAccept(callId: String) {
+        Log.d(TAG, "proceedAccept called on Android (no-op)")
     }
 
     /**
--- a/android/app/src/main/java/chat/rocket/reactnative/voip/VoipPayload.kt
+++ b/android/app/src/main/java/chat/rocket/reactnative/voip/VoipPayload.kt
@@ -20,19 +20,19 @@ enum class VoipPushType(val value: String) {
 data class VoipPayload(
     @SerializedName("callId")
     val callId: String,
-    
+
     @SerializedName("caller")
     val caller: String,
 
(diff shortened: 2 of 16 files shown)


diegolmello:
## PR #7239 Critical Review

**Title:** `fix(voip): pending-accept test mocks`
**Branch:** `feat.voip-lib-new`

---

### Overview

Implements a "pending-accept" fast-path for VoIP calls: when iOS CallKit accepts a call before JS is initialized, the native layer holds it briefly, emits `VoipPendingAccept`, and JS later calls `proceedAccept` to continue. The PR also includes acceptance tests for this flow.

The code is **functionally correct** in its core logic. The issues below are real but non-blocking for the stated goal.

---

### Blocking Issues

**1. `notifyNativeReset()` is an unimplemented stub in the production `reset()` path**

In `MediaSessionInstance.ts:236`:
```typescript
public reset() {
    this.notifyNativeReset(); // ← no-op stub; TODO comment in source
    ...
}
```

`notifyNativeReset()` is a private method containing only:
```typescript
private notifyNativeReset(): void {
    // TODO: wired in Phase 6
}
```

`reset()` is called on logout, end call, and other teardown paths. If `notifyNativeReset` is supposed to tell the native layer to clear `pendingAcceptPayloads`, this is a silent no-op. If it has no purpose yet, it should be removed from `reset()` to avoid confusion.

---

**2. `resetFromLogout` is in the mock but not wired in production**

`NativeVoip.ts` mock adds `resetFromLogout: jest.fn()`, but there is no call site in the PR that actually invokes `NativeVoipModule.resetFromLogout()`. The `pendingAcceptPayloads` dictionary on iOS will accumulate entries across logout/login cycles. This may be intentional ("Phase 6" per the TODO), but it should be tracked as a known gap.

---

**3. `PENDING_ACCEPT_TTL_MS` is exported but never used in JS**

In `useCallStore.ts:758`:
```typescript
export const PENDING_ACCEPT_TTL_MS = 15_000;
```

The comment says "native sides (iOS/Android) must match this value," but there is no JS-side enforcement of this TTL. The 15-second window exists only in native code. If the native TTL fires but JS hasn't received any signal, the queued callId silently stays in `queuedCallIds` forever. This is probably fine since `VoipAcceptFailed` from the native side handles expiry, but the constant is misleading.

---

### Non-Blocking Issues

**4. `clearPendingCallView` calls `Navigation.back()` unconditionally**

If `clearPendingCallView` is called when the user is not on `CallView`, `Navigation.back()` navigates to the previous screen. This is likely safe given the call flow, but worth auditing.

---

**5. The `__nativeEventBus__` mock has a subtle coupling issue**

`MediaCallEvents.ts` creates its `Emitter` at **module load time** (line 12):
```typescript
const Emitter = isIOS ? new NativeEventEmitter(NativeVoipModule) : DeviceEventEmitter;
```

This happens *before* jest mocks are applied. The `Emitter` instance holds a reference to the real `NativeEventEmitter`, not the mock. However, since `setupMediaCallEvents` (which adds listeners) runs **after** mocks are set up and tests only emit via `DeviceEventEmitter`, it works in practice. But the `Emitter` itself is not using `__nativeEventBus__`, so any direct interaction with it in tests would break. Document this constraint.

---

**6. PR description accuracy**

The description says:
> "Removed debug `console.error` from the VoipPendingAccept handler in `MediaCallEvents.ts`"

The source file shows no `console.error` was removed — only `mediaCallLogger.error` is used. This line in the description is stale/inaccurate.

---

### What's Solid

- **`isCallIdQueued` race guard** in `tryAnswerIfNativeAcceptedNotification` correctly prevents the DDP notification path from racing with the pending-accept handoff
- **iOS deduplication** in `VoipService.proceedAccept` via `nativeAcceptHandledCallIds` prevents double-processing
- **`finishAccept` guard** (`finishAcceptInvoked`) prevents double-cleanup of `pendingAcceptPayloads`
- **Cold-start path** in `getInitialMediaCallEvents` correctly routes `pendingAccept` events to `handleVoipPendingAccept`
- **Dedup on `VoipPendingAccept`** via `queuedCallIds.has()` prevents duplicate queuing of the same callId
- **`resetMediaCallEventsStateForTesting` clears `queuedCallIds`** — tests properly isolate
- **`__esModule: true`** on `appNavigation` mock is correct for ESM default exports

---

### Verdict

**Approve with comments.** The implementation is sound and the test fixes are legitimate. The three blocking items (`notifyNativeReset` stub, unwired `resetFromLogout`, unused `PENDING_ACCEPT_TTL_MS`) should be addressed or explicitly deferred with a tracked issue. The stale PR description line should be corrected before merge.

PR #7245, opened 2026-04-25
fix(android/voip): bound accept REST timeout and reconcile late-success

diegolmello:
## Proposed changes

The Android accept REST path has no `callTimeout` on its OkHttp client (the shared client built in `SSLPinningTurboModule` explicitly sets `connectTimeout=0`, `readTimeout=0`, `writeTimeout=0` — all infinite). The 10s `Handler.postDelayed` watchdog fires `finish(false)` while the OkHttp call can still resolve later: `finished` is already true, the late success silently no-ops, server thinks the user accepted, client thinks it failed. On flaky cellular networks this leaves users joined to / billed for a call they believe failed.

This PR:

1. Clones the shared OkHttp client per call with `callTimeout(10s)` (matching the handler deadline) plus `connectTimeout(5s)`, `readTimeout(10s)`, `writeTimeout(10s)`. Per-call clone keeps the shared client's SSL/pinning config intact.
2. Adds a late-success reconciliation: if the OkHttp callback resolves successfully **after** the watchdog fired `finish(false)`, dispatch a REST reject so server state matches the client's failed-state perception.

Adds JUnit/MockWebServer coverage for: deadline expiring before OkHttp resolution (failure, no reject sent), late-success after deadline (reject dispatched), happy path within timeout (success, no reject), and a config-sanity check that `callTimeout` never exceeds the handler deadline.

## Issue(s)

Release-readiness audit P0 on `feat.voip-lib-new`. Class of bug: client/server state divergence on network partition.

## How to test or reproduce

1. Throttle the device network (Charles Proxy `Throttle Settings` → 3G or worse, or `adb shell tc qdisc add dev wlan0 root netem delay 12000ms`).
2. Trigger an incoming call and tap Accept on Android.
3. **Before:** UI shows accept failed, but server-side the user is joined to the call.
4. **After:** UI shows accept failed within 10s, server-side receives a follow-up reject so call records match.

JVM tests: `./gradlew :app:testDebugUnitTest --tests='*MediaCallsAnswerRequestTest*'`.

## Types of changes

- [ ] Improvement
- [ ] New feature
- [ ] Documentation update

## Checklist

- [x] I have signed the CLA
- [x] I have added tests (276 lines of JVM/MockWebServer coverage in `MediaCallsAnswerRequestTest.kt`)
- [ ] I have added necessary documentation

## Further comments

The late-success REST reject is defensive — server-side reconciliation via the eventual `media-calls.end` signal would also clean up state once the client's CallKit/Telecom session disconnects. Keeping the explicit reject because the failure mode (server-side ghost call records, billing) is the kind of bug a 1% population hits often enough to generate support tickets. Trade-off accepted: 17 extra lines for a defensive reconciliation path.

**Reviewer note:** the per-call `newBuilder()` clone is intentional — do not move the timeout config onto the shared client without auditing every other consumer of `SSLPinningTurboModule.getSharedOkHttpClient()`. The shared client is reused by SSL-pinning REST callers that have their own deadline semantics.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved VoIP call handling with explicit timeouts for more reliable call management.
  * Added reconciliation to handle cases where the server accepts after the local client has already timed out or torn down.

* **Tests**
  * Added comprehensive unit tests covering timeout scenarios and late-accept reconciliation behavior.

* **Chores**
  * Added a new testing library to support MockWebServer-based unit tests.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -156,6 +156,7 @@ dependencies {
 
     testImplementation 'junit:junit:4.13.2'
     testImplementation 'org.robolectric:robolectric:4.14.1'
+    testImplementation 'com.squareup.okhttp3:mockwebserver:4.9.2'
 
     // For ProcessLifecycleOwner (app foreground detection)
     implementation 'androidx.lifecycle:lifecycle-process:2.8.7'
--- a/android/app/src/main/java/chat/rocket/reactnative/voip/MediaCallsAnswerRequest.kt
+++ b/android/app/src/main/java/chat/rocket/reactnative/voip/MediaCallsAnswerRequest.kt
@@ -14,6 +14,7 @@ import org.json.JSONArray
 import org.json.JSONObject
 import chat.rocket.reactnative.networking.SSLPinningTurboModule
 import java.io.IOException
+import java.util.concurrent.TimeUnit
 
 /**
  * REST client for `POST /api/v1/media-calls.answer` used by accept/reject flows.
@@ -42,7 +43,13 @@ class MediaCallsAnswerRequest(
         private const val TAG = "RocketChat.MediaCallsAnswerRequest"
         private val JSON_MEDIA_TYPE = "application/json; charset=utf-8".toMediaType()
         private val httpClient: OkHttpClient by lazy {
-            SSLPinningTurboModule.getSharedOkHttpClient() ?: OkHttpClient()
+            val base = SSLPinningTurboModule.getSharedOkHttpClient() ?: OkHttpClient()
+            base.newBuilder()
+                .callTimeout(10, TimeUnit.SECONDS)
+                .connectTimeout(5, TimeUnit.SECONDS)
+                .readTimeout(10, TimeUnit.SECONDS)
+                .writeTimeout(10, TimeUnit.SECONDS)
+                .build()
         }
         private val mainHandler = Handler(Looper.getMainLooper())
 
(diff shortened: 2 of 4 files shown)


diegolmello:
Pushed a follow-up commit addressing the actionable review feedback:

**Fixed:**
- Late-success countdown race in the test helper — replaced the `finish()`-driven `CountDownLatch` with an `okhttpDone` latch counted in the `Callback`'s finally blocks. The test now waits for the OkHttp side to fully settle before reading counters, regardless of whether the deadline fired first.
- Renamed the misleading test (`delayed response past callTimeout triggers client failure and reject is sent` → `okhttp timeout before deadline fails without dispatching a reconcile reject`) and dropped the unused 2nd `MockResponse`.
- Removed the tautological `production client callTimeout does not exceed handler deadline` test — it compared hardcoded locals (`10 <= 10`, `5 < 10`), so it would still pass if production timeouts diverged. Better to drop than fake-cover.
- Added a debug log on the late-accept reconcile-reject callback in `VoipNotification.kt` so field engineers can correlate ghost-call recovery with the source REST call.

**Deferred to follow-up issues:**
- Lazy SSL-pinning fallback at `MediaCallsAnswerRequest.kt:53` (and `DDPClient.kt:29`) — pre-existing pattern, not introduced by this PR. Tracked at #7248.
- Extracting the late-success guard into a testable helper so the test exercises real production code instead of duplicating the logic — legitimate architectural concern but a refactor outside the release-blocker scope.

JVM tests were also not running in PR CI; tracked at #7247.

PR #7257, opened 2026-04-27
chore(ci): drop duplicate bundle install in Android jobs

diegolmello:
## Proposed changes

Every Android build/upload job in CI ran `bundle install` twice. The first install was triggered implicitly by `ruby/setup-ruby@v1` with `bundler-cache: true` and resolved against the **root** `Gemfile` — which carries iOS-only deps (`cocoapods`, `xcodeproj`, ~125 gems). Android jobs threw all of that away and then ran a second explicit `bundle install --path gems` against `android/Gemfile`.

Pointing `setup-ruby` at `android/` via `working-directory: android` makes `bundler-cache: true` target the correct Gemfile from the start, so the explicit "Install Fastlane" step becomes redundant. The deprecated `--path` flag goes away as a bonus.

Affects:
- `.github/actions/build-android/action.yml`
- `.github/actions/upload-android/action.yml`
- `.github/actions/upload-internal-android/action.yml`

Saves roughly ~40s per Android build/upload job (and enables proper cache hits, which the previous setup also missed because the cache key was rooted at the wrong lockfile).

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-2139

## How to test or reproduce

Trigger any Android build job (this PR's own checks are sufficient). In the `Set up Ruby and Bundler` step, confirm:
- only one `bundle install` runs (against `android/Gemfile.lock`, ~88 gems)
- no `[DEPRECATED] The --path flag is deprecated` warning
- no separate "Install Fastlane" step in the job timeline

Subsequent runs on the same branch should hit the bundler cache.

## Screenshots

N/A — CI-only change.

## Types of changes

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable) — N/A, CI-only
- [ ] I have added necessary documentation (if applicable) — N/A
- [ ] Any dependent changes have been merged and published in downstream modules — N/A

## Further comments

`bundler-cache: true` already runs `bundle install` automatically; the prior setup was effectively running it twice with two different lockfiles. Verified against [run 25008420199](https://github.com/RocketChat/Rocket.Chat.ReactNative/actions/runs/25008420199/job/73239391404), where both installs were visible back-to-back (~68s combined).

## Summary by CodeRabbit

* **Chores**
  * Streamlined GitHub Actions build workflows by simplifying dependency installation configuration across Android build processes.

### DIFF
--- a/.github/actions/build-android/action.yml
+++ b/.github/actions/build-android/action.yml
@@ -49,12 +49,7 @@ runs:
       with:
         ruby-version: 2.7.7
         bundler-cache: true
-
-    - name: Install Fastlane
-      working-directory: android
-      run: |
-        bundle install --path gems
-      shell: bash
+        working-directory: android
 
     - name: Set up Gradle
       uses: gradle/actions/setup-gradle@v4
--- a/.github/actions/upload-android/action.yml
+++ b/.github/actions/upload-android/action.yml
@@ -40,12 +40,7 @@ runs:
       with:
         ruby-version: 2.7.7
         bundler-cache: true
-
-    - name: Install Fastlane
-      working-directory: android
-      run: |
-        bundle install --path gems
-      shell: bash
+        working-directory: android
 
     - name: Store the Google service account key
       working-directory: android
--- a/.github/actions/upload-internal-android/action.yml
+++ b/.github/actions/upload-internal-android/action.yml
@@ -40,12 +40,7 @@ runs:
       with:
         ruby-version: 2.7.7
         bundler-cache: true
-
-    - name: Install Fastlane
-      working-directory: android
-      run: |
-        bundle install --path gems
-      shell: bash
+        working-directory: android
 
     - name: Store the Google service account key
       working-directory: android


diegolmello (commented):
# Review — https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7257 — 2026-04-27T21:18:19Z

Findings: 2 (critical: 0, major: 0, minor: 1, nit: 1, spec-mismatch: 0)
Verdict: APPROVE
Skipped (duplicate of existing thread): 0

## Summary

This PR achieves its stated goal: it eliminates the duplicate `bundle install` previously running on every Android CI job. `ruby/setup-ruby@v1` with `working-directory: android` + `bundler-cache: true` performs exactly one resolve+install against `android/Gemfile`, writes `android/.bundle/config`, and the downstream `bundle exec fastlane …` calls (which already use `working-directory: android`) resolve gems via that config. The three composite actions (`build-android`, `upload-android`, `upload-internal-android`) receive identical patches, and a repo-wide grep for `android/gems` returned no references — nothing else in the repo depends on the deleted install path.

Estimated savings: ~40s per Android job today (one per `build-android` × 2 = ~80s/run unconditional, up to ~4 min/run when uploads run); up to ~67s per job once the bundler cache warms, since the previous setup's cache key was rooted at the wrong lockfile and never hit on Android jobs.

## minor: gem path migration from `android/gems/` to `android/vendor/bundle`
**Path:** `.github/actions/build-android/action.yml`, `.github/actions/upload-android/action.yml`, `.github/actions/upload-internal-android/action.yml`
**Category:** design
**Suggested fix:** none — flagged for awareness.

The deleted explicit step used `bundle install --path gems`, which placed gems under `android/gems/`. `setup-ruby`'s default install path is `vendor/bundle`, so installs now land at `android/vendor/bundle/`. A repo-wide grep for `android/gems` shows zero hits in CI scripts, Gemfiles, gradle, fastlane, or `.gitignore`, so the migration is safe. Worth noting if any external tooling or developer workflow caches the old path.

---

## nit: bundler-cache precondition
**Path:** `.github/actions/build-android/action.yml:47-52`, `.github/actions/upload-android/action.yml:38-43`, `.github/actions/upload-internal-android/action.yml:38-43`
**Category:** docs
**Suggested fix:** none.

`bundler-cache: true` requires `android/Gemfile.lock` to exist and be authoritative for cache keying. It does, and the lockfile is committed — just calling out that it is now load-bearing for cache hits.

---

## Verified negatives (not findings)

- Ruby version pin (`2.7.7`) untouched — no smuggled upgrade.
- No secret-handling changes — FASTLANE/Google service account flow is unaffected.
- Three composite actions patched identically — symmetry confirmed.
- No workflow YAML or shell script references the old `android/gems/` path.
- Subsequent `bundle exec fastlane …` invocations use `working-directory: android` and will pick up the new `.bundle/config`.

## Positive observations

- Surgical CI change, net -15 lines, three-file symmetry.
- Uses official action behavior (`bundler-cache` + `working-directory`) rather than a custom step.
- Cache path under `android/vendor/bundle` is the documented setup-ruby convention; should improve cache hit rates compared to the prior `gems/` directory.

OtavioStasiak (approved):
LGTM

PR #7261, opened 2026-04-28
fix: centralize DM subscription stub via lib/methods/createDirectMessage wrapper

diegolmello:
## Summary
- New `app/lib/methods/createDirectMessage.ts` wraps `services/restApi.createDirectMessage` and stamps the WatermelonDB Subscription stub on success (helper from prior shape of this PR: `createDirectMessageSubscriptionStub`).
- All 6 callers migrate to the wrapper: `containers/MessageActions`, `views/DirectoryView`, `views/RoomInfoView`, `views/RoomMembersView`, `sagas/messages.js`, `lib/methods/canOpenRoom`, `lib/methods/helpers/goRoom`.
- Fixes the peer-prefill regression on every DM-creation flow (spotlight, RoomInfo "Message", message action menu, DirectoryView, RoomMembersView, deep-link).

https://rocketchat.atlassian.net/browse/VMUX-83

## Why this shape
REST stays REST (`services/restApi.ts` unchanged). DB writes happen at the methods layer — same convention as `lib/methods/subscriptions/rooms.ts`, `lib/methods/getMessages.ts`. Single source of truth: every "create a DM" caller imports from `lib/methods/createDirectMessage`. Future callers automatically heal.

## Why develop (was feat.voip-lib-new)
The fix is not VoIP-specific — none of the touched files are VoIP-only. E2E noise on feat.voip-lib-new masked CI signal; develop gives a clean read.

## Test plan
- [ ] iOS: spotlight → user → tap call → peer pre-selected
- [ ] iOS: tap avatar/mention in message → "Message" → tap call → peer pre-selected
- [ ] iOS: message action menu → "Direct Message" → tap call → peer pre-selected
- [ ] iOS: DirectoryView → user → tap call → peer pre-selected
- [ ] iOS: RoomMembersView → user → DM → tap call → peer pre-selected
- [ ] Android: same as above
- [ ] Established DM regression: existing DM unchanged
- [ ] Re-opening an existing DM: no duplicate Subscription row (idempotency check in stub)

## Summary by CodeRabbit

* **New Features**
  * Direct message creation now triggers a minimal DM subscription so new conversations appear immediately.
* **Bug Fixes**
  * Prevents missing or duplicate DM records and avoids hidden/broken newly started direct message threads; stub failures are logged without affecting DM creation.
* **Tests**
  * Added unit tests covering DM creation flow, subscription stub behavior, error handling, and edge cases.
* **Chores**
  * Consolidated DM creation logic by routing calls through a single helper used across components.

### DIFF
--- a/app/containers/MessageActions/index.tsx
+++ b/app/containers/MessageActions/index.tsx
@@ -31,10 +31,10 @@ import {
 	markAsUnread,
 	toggleStarMessage,
 	togglePinMessage,
-	createDirectMessage,
 	translateMessage,
 	reportMessage
 } from '../../lib/services/restApi';
+import { createDirectMessage } from '../../lib/methods/createDirectMessage';
 
 export interface IMessageActionsProps {
 	room: TSubscriptionModel;
--- a/app/lib/methods/canOpenRoom.ts
+++ b/app/lib/methods/canOpenRoom.ts
@@ -1,7 +1,7 @@
 import { ERoomTypes } from '../../definitions';
 import database from '../database';
 import sdk from '../services/sdk';
-import { createDirectMessage } from '../services/restApi';
+import { createDirectMessage } from './createDirectMessage';
 
 const restTypes = {
 	channel: 'channels',
--- a/app/lib/methods/createDirectMessage.test.ts
+++ b/app/lib/methods/createDirectMessage.test.ts
@@ -61,13 +61,13 @@ describe('createDirectMessage wrapper', () => {
 		expect(mockedStub).not.toHaveBeenCalled();
 	});
 
-	it('propagates stub rejection (stub owns its own error handling)', async () => {
+	it('does not reject when stub write fails', async () => {
 		const restResult = { success: true, room: { _id: 'r1', fname: 'Foo' } };
 		mockedRest.mockResolvedValue(restResult);
 		const stubError = new Error('stub write failed');
 		mockedStub.mockRejectedValue(stubError);
 
-		await expect(createDirectMessage('foo')).rejects.toThrow('stub write failed');
+		await expect(createDirectMessage('foo')).resolves.toBe(restResult);
 		expect(mockedRest).toHaveBeenCalledWith('foo');
 		expect(mockedStub).toHaveBeenCalledTimes(1);
 	});
--- a/app/lib/methods/createDirectMessage.test.ts
+++ b/app/lib/methods/createDirectMessage.test.ts
@@ -1,0 +1,74 @@
+import { createDirectMessage } from './createDirectMessage';
+import { createDirectMessage as createDirectMessageRest } from '../services/restApi';
+import { createDirectMessageSubscriptionStub } from './createDirectMessageSubscriptionStub';
+
+jest.mock('../services/restApi', () => ({
+	createDirectMessage: jest.fn()
+}));
+
+jest.mock('./createDirectMessageSubscriptionStub', () => ({
+	createDirectMessageSubscriptionStub: jest.fn()
+}));
+
+const mockedRest = createDirectMessageRest as jest.Mock;
+const mockedStub = createDirectMessageSubscriptionStub as jest.Mock;
+
+describe('createDirectMessage wrapper', () => {
+	beforeEach(() => {
+		jest.clearAllMocks();
+		mockedStub.mockResolvedValue(undefined);
+	});
+
+	it('calls stub with rid/username/fname on REST success and returns REST result', async () => {
+		const restResult = { success: true, room: { _id: 'r1', fname: 'Foo' } };
+		mockedRest.mockResolvedValue(restResult);
+
+		const result = await createDirectMessage('foo');
+
+		expect(mockedRest).toHaveBeenCalledWith('foo');
+		expect(mockedStub).toHaveBeenCalledTimes(1);
+		expect(mockedStub).toHaveBeenCalledWith({ rid: 'r1', username: 'foo', fname: 'Foo' });
+		expect(result).toBe(restResult);
+	});
+
+	it('skips stub when REST success but room._id is missing', async () => {
+		const restResult = { success: true, room: {} };
+		mockedRest.mockResolvedValue(restResult);
+
+		const result = await createDirectMessage('foo');
+
+		expect(mockedRest).toHaveBeenCalledWith('foo');
+		expect(mockedStub).not.toHaveBeenCalled();
+		expect(result).toBe(restResult);
+	});
+
+	it('skips stub when REST returns success: false', async () => {
+		const restResult = { success: false };
+		mockedRest.mockResolvedValue(restResult);
+
+		const result = await createDirectMessage('foo');
+
+		expect(mockedRest).toHaveBeenCalledWith('foo');
+		expect(mockedStub).not.toHaveBeenCalled();
+		expect(result).toBe(restResult);
+	});
+
+	it('propagates REST errors without calling stub', async () => {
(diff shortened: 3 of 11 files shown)


diegolmello (commented):
## Claude review (cadence-babysit)

Reviewed against slice spec `2026-04-27-voip-dm-username-prefill / 01-stub-subscription-on-create-direct-message`.

### Verdict
LGTM. Implementation matches the spec one-to-one: idempotent stub, never-throws contract, narrow wire-in to the createDirectMessage success branch only.

### Findings (informational, no severity)
- `app/lib/methods/createDirectMessageSubscriptionStub.ts`: writes both `s._raw.id` and `s._raw._id = rid` — consistent with how other WatermelonDB consumers in this repo seed Subscription rows; not a correctness concern.
- `rid.replace(currentUserId, '').trim()` to derive the peer userId is a heuristic; for the DM-rid format (concatenation of two ids) it is sound, and the function gracefully degrades to a single-element `uids` array if extraction yields empty. No action.
- The function's `try/catch` swallows errors via `log()`, satisfying the "navigation must proceed" contract from the spec.
- Test file covers the spec's required cases (insert, idempotent skip, error swallow on `db.write` and on `getSubscriptionByRoomId`, missing redux user, empty-input guards).
- Goroom wire-in correctly `await`s the stub before `navigate(...)` so `useSubscription` reads the row on RoomView mount.

### Out-of-scope items (per slice spec, intentionally not addressed)
- Generalising stub creation to other room-creation paths (channel join, federated, room create).
- Making `useSubscription` observe WatermelonDB changes.
- The pre-existing `lr` / `lm` schema-vs-interface mismatch.

### CI snapshot at review time
- ESLint and Test: PASS
- Layne Security Scan: PASS
- Snyk: PASS (no manifest changes)
- CLA: signed
- CodeRabbit: review still in progress
- Platform builds (iOS / Android Experimental & Official) and E2E: gated behind `Hold` jobs awaiting manual approval — these will not auto-run.

No outstanding review threads. Recommend a maintainer approve the platform-build `Hold` jobs once CodeRabbit settles.

diegolmello (commented):
# Review — https://github.com/RocketChat/Rocket.Chat.ReactNative/pull/7261 — 2026-04-28T15:00:00Z

Reviewed against slice spec `2026-04-28-voip-dm-stub-wrapper / 01-create-direct-message-wrapper`, commit `a7844312a`.

Findings: 2 (critical: 0, major: 1, minor: 1, nit: 0, spec-mismatch: 0)
Verdict: COMMENT
Skipped (duplicate of existing thread): 2

---

## minor: `rid.replace(currentUserId, '')` can corrupt `otherUserId` if id appears mid-string

**Path:** app/lib/methods/createDirectMessageSubscriptionStub.ts:49
**Category:** correctness
**Suggested fix:** Use `startsWith`/`endsWith` guards or split on the known separator:
```ts
const otherUserId = rid.startsWith(currentUserId)
    ? rid.slice(currentUserId.length).trim()
    : rid.endsWith(currentUserId)
        ? rid.slice(0, -currentUserId.length).trim()
        : rid.trim();
```

`String.replace` removes the **first** occurrence anywhere in the string. Rocket.Chat DM room IDs are `userId1userId2` (concatenated, no separator), so the current user's ID could theoretically appear in the middle of the peer's ID and produce a corrupted `otherUserId`. The stub's `uids` field is overwritten by the server sync shortly after, so the blast radius is low — but this is a correctness issue worth fixing to keep the stub accurate during the brief pre-sync window. CodeRabbit thread already open on this line; marking as duplicate to avoid double-posting.

*(Duplicate of existing CodeRabbit thread — not re-posting inline, just noting here.)*

---

## major: Test 5 asserts stub rejection propagates, but the stub never throws by design

**Path:** app/lib/methods/createDirectMessage.test.ts:64–73
**Category:** test-coverage / design

The fifth test case (`'propagates stub rejection (stub owns its own error handling)'`) mocks the stub as rejecting and asserts that `createDirectMessage` rejects. However:

1. `createDirectMessageSubscriptionStub` is documented and implemented to **never throw** — it wraps its entire body in `try/catch { log(e) }` (see stub line 32–80).
2. The wrapper (`createDirectMessage.ts`) does **not** have its own try/catch around the stub call, meaning if the stub mock is made to throw (bypassing the real implementation's guard), the rejection propagates.

This creates a correctness gap: the test validates behavior that conflicts with the stated design contract ("stub owns its own error handling" says the stub never throws, yet the test proves the wrapper **will** propagate if the stub throws). If the real stub's try/catch is ever accidentally removed, this test will catch the regression correctly — but the test name is misleading.

**Recommended fix:** Either (a) add a try/catch in the wrapper so it genuinely cannot propagate stub failures regardless of stub implementation details, or (b) change the test to assert `resolves` (matching the design intent that stub failures are swallowed). Option (a) is more defensive. CodeRabbit's thread on `createDirectMessage.ts:11` proposes (a); the thread on the test file proposes (b). They are linked — resolving one entails the other.

*(Note: CodeRabbit thread on `createDirectMessage.ts:11` misidentifies the missing guard as absent — the **stub** has the guard; the **wrapper** does not. The analysis is correct that the wrapper should guard too, but the guard location matters.)*

---

## Observations (no action needed)

- `createDirectMessageSubscriptionStub.ts` has comprehensive internal error handling — all DB failures are caught and logged. The only propagation risk is if the stub's own try/catch is removed, which the wrapper guard would also protect against.
- All 6 callers correctly migrated to the new wrapper import path.
- `goRoom.ts` simplified correctly — inline stub call removed.
- Slice acceptance criteria met for code structure, caller migration, and test count (5 wrapper + 10 stub = 15 tests).

diegolmello:
Cadence pass `2026-04-28-pr-7261-review-resolution`: addressed the three CodeRabbit threads in the latest push — (1) prefix/suffix-safe `otherUserId` from `rid`, (2) `try/catch` + `log` around the stub in `createDirectMessage`, (3) unit test now expects resolve when the stub rejects. Review threads resolved.

diegolmello:
**Review**

Overall: clean migration, good test coverage. Three real issues.

---

`createDirectMessageSubscriptionStub.ts:L54-57`: 🔴 bug: when `rid` doesn't start/end with `currentUserId`, falls back to `otherUserId = rid.trim()` (the full room ID), writing a non-user-ID into `uids[1]`. Test at L211 acknowledges this but calls it "fallback" — wrong data persists until realtime sync overwrites. Add `otherUserId = ''` fallback and write `uids = [currentUserId]` only (single-element), same as the `otherUserId ? [...] : [currentUserId]` guard already handles.

`createDirectMessageSubscriptionStub.ts:L63`: 🟡 risk: `(s: any)` skips WatermelonDB model type-checking. Use `TSubscriptionModel` (already imported in other files) so field mismatches surface at compile time.

`createDirectMessage.ts:L12`: 🟡 risk: `(result.room as any)?.fname` — cast suppresses the type. `restApi.createDirectMessage` returns `{ room: { _id: string; ... } }` — either extend that interface with `fname?: string` or cast to a typed intermediate, not `any`.

`createDirectMessage.ts:L8-13`: 🔵 nit: outer `try/catch` around `createDirectMessageSubscriptionStub` is dead — stub already swallows internally and never throws. Remove the wrapper or remove the inner swallow and let caller own error handling.

`createDirectMessageSubscriptionStub.ts:L45-57`: ❓ q: concurrent calls (user taps DM twice before realtime sync) both pass the idempotency check at L38, then second `db.write` throws duplicate-PK — swallowed silently. Acceptable for this use case?

OtavioStasiak (approved):
LGTM!

PR #7265, opened 2026-04-28
fix(voip): toggle Android speaker via AudioManager.setCommunicationDevice

diegolmello:
## Proposed changes

Tapping the **Speaker** button during an active VoIP call on Android now actually routes audio to the device loudspeaker. Previously the icon flipped state but audio kept playing through the earpiece — no error, no log, just a silent no-op.

**Root cause.** Our Android Telecom integration uses a **self-managed** `PhoneAccount` (see `VoipNotification.kt`, `setCapabilities(PhoneAccount.CAPABILITY_SELF_MANAGED)`). In that mode, `android.telecom.Connection.setAudioRoute()` is documented as a no-op and `onCallAudioStateChanged` never fires. The previous fix attempt routed the toggle through `RNCallKeep.setAudioRoute(uuid, 'Speaker'|'Earpiece')` — which reaches `Connection.setAudioRoute()` — so the JS promise resolved successfully but no audio routing actually happened. That's why no log surfaced and the toggle silently flipped the icon without changing audio.

**Fix.** The app owns the audio route in self-managed mode, so drive `AudioManager` directly:

- API 31+ (Android 12+): `AudioManager.setCommunicationDevice(BUILTIN_SPEAKER)` for speaker on; `clearCommunicationDevice()` for off (lets the system pick — headset if connected, otherwise earpiece). This API was introduced for VoIP/communications apps and preempts WebRTC's `AudioDeviceModule` routing race that previously made `setSpeakerphoneOn` look like it worked while the audio thread reverted the route.
- Pre-API-31: legacy `setMode(MODE_IN_COMMUNICATION) + setSpeakerphoneOn(on)` fallback.

Exposed as a new `setSpeakerOn(on: boolean): Promise<boolean>` method on the existing `VoipModule` TurboModule. iOS keeps using `InCallManager.setForceSpeakerphoneOn` from JS — the iOS implementation of the spec is a no-op stub that satisfies codegen but is never called from production JS.

**Cleanup.** Removed the now-dead `RNCallKeep.addEventListener('didChangeAudioRoute', ...)` block from `MediaCallEvents.ts` — that event is only fired by CallKeep from `Connection.onCallAudioStateChanged`, which Telecom never invokes under a self-managed `PhoneAccount`. Also removed the `callId`/`nativeAcceptedCallId` extraction from `toggleSpeaker` since the new path doesn't need a call UUID.

No new permissions: `MODIFY_AUDIO_SETTINGS` is already declared in `AndroidManifest.xml`.

## Issue(s)

## How to test or reproduce

1. Build and install the Android app (`yarn android` or `./gradlew :app:installExperimentalDebug`).
2. Place an active VoIP call (incoming or outgoing).
3. Tap the **Speaker** button on the in-call screen — audio should audibly route to the device loudspeaker. The icon flips on.
4. Tap **Speaker** again — audio should audibly return to the earpiece (or to a connected wired/Bluetooth headset if one is plugged in). The icon flips off.
5. (Optional) Confirm the native path with `adb -s <device> logcat -s RocketChat.VoipModule:D` — you should see `setSpeakerOn(true) via setCommunicationDevice -> true` and `setSpeakerOn(false) via clearCommunicationDevice`.

iOS: regression check via existing Jest case — speaker behavior unchanged (still uses `InCallManager.setForceSpeakerphoneOn`).

## Screenshots

N/A — audio-routing change with no UI difference beyond the existing speaker icon state.

## Types of changes

## Checklist

## Further comments

**Out of scope (deferred).** When the user plugs in / unplugs a wired or Bluetooth headset during an active call, or picks a different output device from the system audio panel, the in-app `isSpeakerOn` state will become stale (it reflects only JS-initiated toggles, not external route changes). Resolving this requires an `AudioManager.OnCommunicationDeviceChangedListener` (API 31+) bridged to a JS event so the Zustand store can sync. Tracked as a separate follow-up so this PR stays focused on the immediate "tap does nothing" bug.

Pre-API-31 reliability hardening against the WebRTC `AudioDeviceModule` race is also deferred — we accept best-effort behavior on Android <12 given the device-share trajectory.

## Summary by CodeRabbit

* **New Features**
  * Enhanced speaker control for VoIP calls with improved platform-specific audio routing.

* **Bug Fixes**
  * Optimized VoIP permission lifecycle management for better stability.

* **Tests**
  * Expanded test coverage for speaker toggle functionality across platforms.


diegolmello (commented):
## Claude review — slice 01-android-callkeep-audio-route

Reviewed against plan `2026-04-28-voip-speaker-android` and `slices/01-android-callkeep-audio-route/spec.md` (inferred from receipt).

### Spec match
- Android `toggleSpeaker` now goes through `RNCallKeep.setAudioRoute(uuid, 'Speaker'|'Earpiece')` so Telecom/ConnectionService owns the route — addresses the root cause described in the PR body (InCallManager write reverted by Telecom). Correct fix.
- iOS path preserved (`InCallManager.setForceSpeakerphoneOn`) — matches the doc note that CallKit's audio session honors it.
- `didChangeAudioRoute` listener registered Android-only inside `setupMediaCallEvents`, gated by lowercased active uuid (`callId ?? nativeAcceptedCallId`), pushed into `subscriptions` for cleanup. Mirrors the existing `didPerformSetMutedCallAction` pattern in the same file (`app/lib/services/voip/MediaCallEvents.ts:140-149`), so the convention is consistent.

### Findings
None at severity that would block. Minor observations only:
- `isIOS` branch in `toggleSpeaker` returns early after `await`, while the Android branch falls through to `set({...})` after `setAudioRoute`. Behavior is correct but two `set({ isSpeakerOn })` sites is slightly easier to drift; not a blocker.
- The `useCallStore.test.ts` uses `jest.requireActual` + `defineProperty` to make `isIOS` a live getter — load-bearing because the helper module captures `Platform.OS` at import time. Worth a one-line inline comment explaining why future maintainers shouldn't simplify it (already there: lines 28-30, good).

### Test coverage
Comprehensive: Speaker/Earpiece flip, missing-uuid no-op, iOS unchanged, listener registration, uuid case-insensitive match, `nativeAcceptedCallId` fallback, uuid mismatch / missing / no-call drops, cleanup, iOS-no-op assertion. Full suite passes (1279 tests / 148 suites).

### Manual QA gate
PR body correctly flags the three Android-device checks that CI cannot satisfy. Reviewers should not approve on green CI alone.

### Verdict
LGTM from static review. Waiting on CI + manual Android device QA.

PR #7272, opened 2026-04-29
feat(voip): consolidate native event subscription and cold-start drain into VoipNative.attach

diegolmello:
## Proposed changes

Sub-PR 1 / slice 4 of the VoIP native seam refactor.

Replaces the two separate entry points (`setupMediaCallEvents` for live events, `getInitialMediaCallEvents` for cold-start) with a single `voipNative.attach({ onEvent })` call. The production adapter now owns:

- Lazy `registerGlobals()` (avoids cascading import errors in unrelated tests)
- PushKit token registration (iOS)
- All native-event subscriptions (CallKeep, NativeEventEmitter/DeviceEventEmitter)
- Cold-start drain (NativeVoipModule events first, then RNCallKeep on iOS)

Returns `{ detach(): void; pushToken: string }`. `detach` is idempotent and stops all subscriptions + `stopNativeDDPClient` on iOS.

`setupMediaCallEvents` is replaced by `createVoipEventDispatcher` factory, which returns a typed `(e: VoipNativeEvent) => boolean` dispatcher. The `boolean` return signals whether a cold-start VoIP event was handled (skips `appInit()`). `fromColdStart` carried through the event union preserves the Android same-workspace cold-start behavior (defer to `appInit()`).

In-memory test adapter gains `__seedColdStart` and `__emit` seam methods.

## Issue(s)

Part of VoIP native call seam refactor (`.scratch/voip-native-call-seam/issues/04-voip-native-attach-and-cold-start.md`)

## How to test or reproduce

Unit tests: `TZ=UTC yarn test -- --testPathPattern='voip'`

Manual (physical device required):
1. iOS: kill app mid-call, reopen — app should reach CallView (cold-start accept)
2. Android: same scenario
3. Cross-workspace cold-start: accept call from different server — app should deep-link to that server

## Types of changes

## Checklist

## Further comments

Pre-existing failures in `VoipCallLifecycle.integration.test.tsx` (5 tests) are present in the base branch (slice 03) and are not introduced by this slice.

## Summary by CodeRabbit

* **Refactor**
  * Consolidated VoIP handling into a single native-backed event flow for more reliable call lifecycle (answer, end, hold, speaker) and cleaner cleanup; unified audio and termination paths for consistent behavior.

* **Bug Fixes**
  * Improved cold-start, accept/dedupe, and push-token handling to reduce missed or duplicated calls and race conditions.

* **Tests**
  * Added an in-memory native adapter and expanded tests to validate audio, speaker, command sequencing, and cold-start behaviors.


diegolmello:
**VoipNative.ts**

`L109-113: 🔵 nit: markAvailable(_callUuid) param unused. Interface takes callUuid but impl ignores it. Drop param from both interface and all call sites, or use it.`

`L127-132: 🔴 bug: call.startAudio/stopAudio swallow InCallManager.start/stop silently — no try-catch. Old useCallStore.ts had console.error fallback. If InCallManager throws, uncaught exception propagates into Zustand updater. Add try/catch or keep previous error-boundary pattern.`

`L93-97: 🔵 nit: InMemoryVoipNative.reset() clears recorded but not _onEvent. Stale listener persists between tests if attach() isn't re-called. Either clear _onEvent in reset() or document that attach() must precede each test.`

`L287: ❓ q: process.env.NODE_ENV === 'test' singleton — if Metro strips dead code at 'production', the InMemory branch is gone. If it doesn't, it's dead weight in production bundles. Known/acceptable?`

---

**MediaCallEvents.ts**

`L42: 🔵 nit: clearVoipAcceptDedupeSentinels is exported no-op with no deprecation marker. External callers can't tell it's dead. Add @deprecated JSDoc or remove export.`

`L88-93: ❓ q: handleAcceptSucceededEvent returns fromColdStart (false for live events, true for iOS cold-start). For Android cold-start same-workspace it returns false and sets nativeAcceptedCallId before appInit runs — setNativeAcceptedCallId called on L64 regardless of the early return at L69. Intended?`

---

**MediaCallEvents.ios.test.ts**

`L139+: 🟡 risk: old test 'returns false when CallKit has no RNCallKeepPerformAnswerCallAction' was deleted. ProductionVoipNative._drainColdStart "iOS unanswered" path (wasAnswered=false) has no coverage. VoipNative.test.ts only covers InMemoryVoipNative.`

---

**useCallStore.ts**

`L162: 🔴 bug: voipNative.call.startAudio() no error boundary — see VoipNative.ts note above. Same for stopAudio at L292. If InCallManager bridge isn't ready (e.g. screen-share interruption), throws propagate unhandled from Zustand action.`

`L241-243: toggleSpeaker awaits setSpeaker without try-catch. setSpeaker propagates InCallManager throws to the UI caller. Old code had console.error + continued. Worth keeping the guard so isSpeakerOn stays consistent on failure.`

---

**app/index.tsx**

`L178-184: ❓ q: live event acceptFailed returns true from dispatcher, but voipInitialHandled is only read after attach() resolves, so live events post-attach can't affect the check. Confirm no scenario where acceptFailed fires inside _drainColdStart and then a second live one arrives before the if(voipInitialHandled) guard runs.`

PR #7273, opened 2026-04-29
fix(voip): Android audio-route sync + outgoing ringback on comm path

diegolmello:
## Proposed changes

Two related Android audio-routing fixes for the new VoIP stack.

### 1. Sync `isSpeakerOn` with system audio route changes (mid-call)

Mid-call on Android, when audio route changed via something other than the in-app Speaker button (headset plug/unplug, BT connect/disconnect, system audio panel), `useCallStore.isSpeakerOn` stayed stale and the speaker icon misrepresented the live route.

Registers `AudioManager.OnCommunicationDeviceChangedListener` (API 31+) for the call's lifetime. The listener emits `VoipCommunicationDeviceChanged { isSpeaker: boolean }` to JS; `MediaCallEvents` mirrors it into `useCallStore.isSpeakerOn`. A synthetic event is emitted once after registration so the icon is accurate at t=0 (covers calls accepted with speaker already engaged).

The app uses a self-managed `PhoneAccount`, so `Connection.setAudioRoute()` is a no-op and CallKeep audio-route APIs are dead — audio routing must go through `AudioManager` directly.

### 2. Play outgoing-call ringback on the voice-communication audio path

`expo-av`'s `Audio.Sound` plays through `STREAM_MUSIC` on Android, which routes to the loudspeaker by default and is unaffected by `setCommunicationDevice`. Result: the outgoing dialtone was loud regardless of `isSpeakerOn`, and `toggleSpeaker` had no audible effect during ringback.

`VoipModule.startRingback`/`stopRingback` use `MediaPlayer` with `AudioAttributes USAGE_VOICE_COMMUNICATION + CONTENT_TYPE_SONIFICATION` so the dialtone follows the active communication device (earpiece by default, speaker/BT when selected). `CallView` calls native ringback on Android only; iOS keeps the existing `expo-av` Ringer (CallKit + AVAudioSession route it correctly).

## Issue(s)

## How to test or reproduce

Manual on Android device:

**Audio-route sync (mid-call)**
1. Start a VoIP call with speaker on → plug headphones → icon flips off; unplug → icon updates
2. BT connect mid-call → icon shows "not on speaker"; disconnect → icon updates
3. System audio panel → "Phone speaker" → icon flips on
4. Accept call with speaker pre-engaged → icon shows "on" at t=0
5. `adb logcat -s RocketChat.VoipModule:D` shows listener firing on each route change

**Outgoing ringback routing**
6. Place an outgoing call → dialtone plays through earpiece (not loudspeaker) by default
7. Tap Speaker during ringback → dialtone audibly reroutes to loudspeaker
8. Tap Speaker again → dialtone returns to earpiece
9. Connect a BT headset before placing the call → dialtone plays through BT
10. `adb logcat -s RocketChat.VoipModule:D` shows `startRingback` / `stopRingback` log lines

## Types of changes

## Checklist

## Further comments

- `startAudioRouteSync` / `stopAudioRouteSync` are idempotent (no-op when listener already registered / not registered). `stopAudioRouteSync` is safe to call N times — `reset` fires 2–3× per call ending.
- Pre-API-31: `startAudioRouteSync` resolves `null` and skips listener registration — no crash, no regression on legacy devices.
- `startRingback` is idempotent; a second call while already playing is a no-op. `stopRingback` is safe to call when not playing.
- iOS: all native methods are no-op stubs. `VoipCommunicationDeviceChanged` is never emitted; `supportedEvents` array untouched.


diegolmello:
Good diff. Solid impl, tests comprehensive. A few findings:

---

**`VoipModule.kt`**

`L103: audioRouteSyncActive` redundant — value always == `communicationDeviceListener != null`. Only read in early-return branch where it's always `true`. Drop field, resolve `true` directly.

`L174: @SuppressLint("NewApi")` on full method when runtime API guard already present at L177 — lint suppression unnecessary; remove annotation, let runtime check speak for itself.

`L188-L191: startAudioRouteSync` emits current device state immediately after registering listener — good, but if `addOnCommunicationDeviceChangedListener` itself fires a synthetic callback on registration (device hasn't changed), you'd get a duplicate event. Verify Android docs confirm no immediate callback on `add` before the manual emit.

---

**`app/lib/native/NativeVoip.ts`**

`L42-L55: 🔵 nit:` multi-line JSDoc blocks violate project style (CLAUDE.md: one short line max). Collapse to single-line comments or drop.

`L79-L80:` return value of `startAudioRouteSync` unused on JS side — JSDoc implies caller should branch on `false` for pre-31, but `useCallStore.ts` ignores it. Either remove the resolve-false distinction or act on it.

---

**`app/lib/services/voip/useCallStore.ts`**

`L173-L176: startAudioRouteSync` fire-and-forget with no guard against double-call if `setCall` is invoked twice (e.g., call replacement). Native side has idempotency guard so it's safe, but worth a comment since the silent no-op isn't obvious from the JS side.

diegolmello:
[codex] Addressed in c261d73:

**`VoipModule.kt`**
- Dropped `audioRouteSyncActive` field; `startAudioRouteSync` resolves directly when listener already registered.
- Removed both `@SuppressLint("NewApi")` annotations — runtime guards cover the API 31+ calls.
- Confirmed `addOnCommunicationDeviceChangedListener` does **not** fire a synthetic callback on registration: AOSP routes it through `CallbackUtil.LazyListenerManager.addListener`, which only invokes the callback on subsequent `notifyListeners` from `AudioDeviceBroker` (no immediate replay of current state). Manual emit retained as the documented seed path; added a comment noting the rationale.

**`NativeVoip.ts`**
- Collapsed `start/stopAudioRouteSync` JSDoc to single-line per project style.
- Removed the resolve-false distinction: return type is now `Promise<void>` on TS spec, Kotlin (`promise.resolve(null)`), and Obj-C (`resolve(nil)`). Test mocks updated accordingly.

**`useCallStore.ts`**
- Added one-line comment near `startAudioRouteSync` call documenting that the native side is idempotent for double-`setCall` scenarios.

PR #7286, opened 2026-04-30
fix(voip): gate MediaCallHeader to authenticated inside root only

diegolmello:
## Proposed changes

`MediaCallHeader` was mounted unconditionally at the application root, alongside the `NavigationContainer`. Because its no-active-call branch returned a styled empty `View` (height equal to the safe-area top inset, coloured `surfaceNeutral`) instead of `null`, the header occupied vertical space on every screen — including the login screen, share extension, and loading splash — producing a visible coloured strip above each of those screens' content.

## Issue(s)

Part of the PR #6918 fix set — module 4, blocker B8.

## How to test or reproduce

1. Launch the app to the login screen — no coloured strip above the login form.
2. Open the share extension — no coloured strip above the share-target picker.
3. Sign in — InsideStack screens render with their normal status-bar appearance (no regression).
4. Place a VoIP call — the call header appears above the room view.
5. Hang up — the header disappears.

## Screenshots

N/A — visual regression fix; before/after comparison via manual testing on device.

## Types of changes

## Checklist

## Further comments

Three changes in this slice:

1. **Root gating** (`app/AppContainer.tsx`): `<MediaCallHeader />` is now wrapped in `{root === RootEnum.ROOT_INSIDE ? <MediaCallHeader /> : null}` so it only mounts for the authenticated inside root.
2. **No-call branch returns null** (`app/containers/MediaCallHeader/MediaCallHeader.tsx`): The empty `<View>` placeholder is replaced with `null`. The `View` import is also removed.
3. **Dead prop removed** (`app/containers/Header/index.tsx`, `app/containers/Header/components/HeaderContainer/index.tsx`): `addExtraNotchPadding` was declared in the `IHeaderContainer` interface but never consumed in the component body; it is now removed from both the interface and the call site.

A new `app/AppContainer.test.tsx` asserts that `MediaCallHeader` is present in the inside tree and absent for all other roots (loading, outside, set-username, share extension, loading-share-extension).

## Release Notes

* **Bug Fixes**
  * Media Call Header now conditionally renders based on current application state rather than always displaying
  * Media Call Header no longer shows placeholder content when inactive
  * Removed unnecessary layout padding in specific application states

* **Tests**
  * Added comprehensive test coverage for Media Call Header rendering behavior

### DIFF
--- a/app/AppContainer.tsx
+++ b/app/AppContainer.tsx
@@ -53,7 +53,7 @@ const App = memo(({ root, isMasterDetail }: { root: string; isMasterDetail: bool
 
 	return (
 		<>
-			<MediaCallHeader />
+			{root === RootEnum.ROOT_INSIDE ? <MediaCallHeader /> : null}
 			<NavigationContainer
 				theme={navTheme}
 				ref={Navigation.navigationRef}
--- a/app/AppContainer.test.tsx
+++ b/app/AppContainer.test.tsx
@@ -1,0 +1,113 @@
+import { render } from '@testing-library/react-native';
+import { Provider } from 'react-redux';
+import { applyMiddleware, compose, createStore } from 'redux';
+import createSagaMiddleware from 'redux-saga';
+
+import reducers from './reducers';
+import AppContainer from './AppContainer';
+import { appStart } from './actions/app';
+import { RootEnum } from './definitions';
+
+// Mock Firebase analytics — the default __mocks__ file uses old API shape
+jest.mock('@react-native-firebase/analytics', () => ({
+	getAnalytics: () => ({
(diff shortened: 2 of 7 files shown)


app/AppContainer.tsx
```diff
@@ -53,7 +53,7 @@ const App = memo(({ root, isMasterDetail }: { root: string; isMasterDetail: bool
 
 	return (
 		<>
-			<MediaCallHeader />
+			{root === RootEnum.ROOT_INSIDE ? <MediaCallHeader /> : null}
```
diegolmello, line 56: Image

Header gets rendered behind status bar.

PR #7305, opened 2026-05-06
fix: load 50 messages with hideSystemMessages as true on old workspaces

OtavioStasiak:
## Proposed changes

Room history for the main channel was loaded in fixed pages of 50 raw messages. After that, the client applies Hide system messages, so a first page could be almost entirely hidden types. The list looked nearly empty (often one normal message plus Load more) even though the app thought it had already filled a full page.

This PR fixes pagination by:

1. loadMessagesForRoom – Keeps requesting older history batches until there are enough visible main-room messages (respecting hidden system types resolved from the subscription’s sysMes or Hide_System_Messages in Redux), with a sensible batch cap.
2. Loading UI – When a second batch is needed on initial open, we persist the first batch plus a real load_more row, push that loader id onto historyLoaders (via HISTORY_UI_LOADER_PUSH), then pop it when the full fetch finishes (HISTORY_UI_LOADER_POP). The existing LoadMore row therefore shows the same spinner as today—no extra list footer or forceLoading hack.
3. Scope – No new plumbing through actions/sagas/RoomView props for hideSystemMessages; the loader resolves preferences internally (same idea as other methods using the Redux store).

## Issue(s)	

https://rocketchat.atlassian.net/browse/CORE-2179

## How to test or reproduce

Bug (before): On a server/workspace where recent history is mostly system messages (joins, topic changes, etc.) and the user hides those types, opening a channel could show almost no messages and Load more too early.

Verify (after):

Use an account with Hide system messages enabled for types that dominate recent traffic (or a room with many hidden system events).
Open that channel cold (or clear local data and reopen so history loads fresh).
Expected: The message list fills with enough visible messages for a normal page; Load more only appears when there is actually older history. While extra batches are fetched, the existing Load more row shows the loading state (spinner), same as tapping Load more.
Regression: Tap Load more in a normal room—spinner and history fetch behave as before.

## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Batched message loading with recursive "load more" behavior, per-room hidden-message handling, and safer load-more insertion
  * UI loader lifecycle for history loads with push/pop tracking to improve loading indicators

* **Bug Fixes**
  * Ensures loader cleanup on errors and avoids duplicate load-more entries

* **Tests**
  * Expanded coverage for batching, max-batch limits, loader push/pop, error cleanup, visibility rules, and hook unsubscribe/behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/RocketChat/Rocket.Chat.ReactNative/pull/7305?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


OtavioStasiak:
@coderabbit review

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -47,21 +74,48 @@ async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t:
...
+			if (!args.loaderItem && batchesFetched === 1) {
+				const loadMoreMessage = {
+					_id: generateLoadMoreId(lastMessage._id as string),
+					rid: lastMessage.rid,
+					ts: dayjs(lastMessage.ts).subtract(1, 'millisecond').toString(),
+					t: MessageTypeLoad.MORE,
+					msg: lastMessage.msg
+				} as IMessage;
+
+				await updateMessages({
```
diegolmello, line 96: **Duplicate concurrent fetch in the exact scenario this PR targets.**

Writing `load-more-first-50` to DB here causes WatermelonDB's subscription in `useMessages` to emit. The auto-dispatch effect at `app/views/RoomView/List/hooks/useMessages.ts:153-169` then fires `roomHistoryRequest` for that loader id — its guards (`serverVersion >= 3.16.0` + `hideSystemMessages.length > 0`) are exactly the bug scenario. The saga `watchHistoryRequests` (`app/sagas/room.js:16-32`) has no in-flight dedup, so it spawns a second `loadMessagesForRoom` while batch 2 of the first is still recursing. Both converge in DB and `MAX_BATCHES` / `MAX_AUTO_LOADS` cap blast radius, but you get 2× HTTP traffic on cold open.

Cleanest fix: guard at `useMessages.ts:164` — `if (state.room.historyLoaders.includes(loaderId)) return;`. The `HISTORY_UI_LOADER_PUSH` already populated `historyLoaders`, so the existing PUSH/POP plumbing covers it without new state.

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -47,21 +74,48 @@ async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t:
...
+					rid: roomId,
+					update: [...allMessages, loadMoreMessage],
+					loaderItem: args.loaderItem
```
diegolmello, line 99: Dead arg — the enclosing `if (!args.loaderItem && batchesFetched === 1)` guarantees `args.loaderItem` is `undefined` here. Drop the line.

diegolmello (commented):
Code review — details inline. Main finding: the cold-open multi-batch flow triggers a duplicate concurrent fetch via `useMessages`' auto-dispatch (see comment on `loadMessagesForRoom.ts:96`). Plus a few smaller items.

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -47,21 +74,48 @@ async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t:
 
 		const batch = data.messages as IMessage[];
 		allMessages.push(...batch);
+		shouldAddLoader = batch.length === COUNT;
```
diegolmello: `shouldAddLoader` is overwritten on every batch — final value reflects only the last one. That's the intended semantics (only the trailing batch's tail tells us whether more older history may exist), but it reads like a bug. Rename to `lastBatchWasFull` or add a one-line comment.

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -47,21 +74,48 @@ async function load({ rid: roomId, latest, t }: { rid: string; latest?: Date; t:
...
-	const startTimestamp = latest ? new Date(latest).toISOString() : undefined;
-	await fetchBatch(startTimestamp);
-	return allMessages;
+	const startTimestamp = args.latest ? new Date(args.latest).toISOString() : undefined;
+	try {
+		await fetchBatch(startTimestamp);
+		return { messages: allMessages, shouldAddLoader, uiLoaderId };
+	} catch (e) {
+		if (uiLoaderId) {
+			store.dispatch(roomHistoryUiLoaderPop({ loaderId: uiLoaderId }));
```
diegolmello: Cleanup paths are split: this inner catch pops on throw + rethrows, and the outer `finally` (line 155) also pops. No double-pop — the outer `uiLoaderId` only gets assigned after `await load(args)` resolves, so on the throw path it stays `null` and the outer skips. But the split is subtle. Letting only the outer `finally` own cleanup is simpler.

app/reducers/room.ts
```diff
@@ -68,6 +68,18 @@ export default function (state = initialState, action: TActionsRoom): IRoom {
 				...state,
 				historyLoaders: state.historyLoaders.filter(loaderId => loaderId !== action.loaderId)
 			};
+		case ROOM.HISTORY_UI_LOADER_PUSH:
```
diegolmello, line 71: `HISTORY_UI_LOADER_PUSH/POP` and `HISTORY_REQUEST/HISTORY_FINISHED` both mutate `historyLoaders`. The new push dedupes via `.includes`; the old one (line 61-65) doesn't. Naming is also asymmetric. Suggest renaming for symmetry — makes "this dispatches a saga" vs "this is pure UI state" obvious at the call site:
- `HISTORY_FETCH_REQUESTED` / `HISTORY_FETCH_COMPLETED` (saga-triggering)
- `HISTORY_LOADER_ADDED` / `HISTORY_LOADER_REMOVED` (reducer-only)

app/lib/methods/loadMessagesForRoom.test.ts
```diff
@@ -0,0 +1,125 @@
...
+				loaderId: 'load-more-first-50'
+			})
+		);
+		expect(mockedDispatch).toHaveBeenCalledWith(
+			expect.objectContaining({
+				type: ROOM.HISTORY_UI_LOADER_POP,
+				loaderId: 'load-more-first-50'
+			})
+		);
+	});
```
diegolmello, line 136: Coverage gap: a single happy-path test. Suggest adding cases for:
- `MAX_BATCHES` cap (stops at 10 even if still under 50 visible).
- Last batch `< COUNT` → no `shouldAddLoader` on final write.
- Error mid-recursion → loader id gets popped (current test only asserts PUSH+POP on success).
- `sub.sysMes === true` boolean → falls back to settings.
- `loaderItem` provided → no intermediate loader inserted.
- The duplicate-fetch flow flagged in `loadMessagesForRoom.ts`, once the `useMessages` guard is added.

diegolmello (approved):

PR #7311, opened 2026-05-07
chore: stabilize Maestro CI

diegolmello:
## Proposed changes

Stabilizes the Maestro E2E suite (CI infra + flaky-flow fixes) and folds in the app-code
fixes that unblock specific shards. The CI changes carry no app behavior change; the app
changes are listed first since they're what needs review.

### App changes

- **Jump to message over DDP** ([NATIVE-1126]) — `sdk.ts` appended the 2FA code as a trailing
  `''` on *every* DDP call. Harmless until the server's `loadSurroundingMessages` grew a typed
  `showThreadMessages: boolean` 3rd param ([RocketChat/Rocket.Chat#39092]); the `''` then fails
  `check(…, Boolean)` → 500 → the quoted-message jump never navigates. Fix: append the TOTP code
  only when a 2FA flow is in progress (matches the REST path), and clear it as soon as the retry
  consumes it — so a spent code never leaks into subsequent calls on the same connection. Non-2FA
  calls send exactly their declared params. Unit-tested (`sdk.test.ts`: no-2FA path, 2FA prompt +
  retry, non-leakage, cancellation).
- **Jump-to-message scroll race cap** — `RoomView` raced the list scroll against a fixed 5s cap;
  off-screen targets on the slow CI emulator exceed it. Raised to 20s as an **interim** measure.
  The scroll-retry loop also now exits early when history stops loading: if the loaded-message
  count stays flat for 5 consecutive 600 ms rounds (~3 s), the target can't appear (deleted or
  unreachable message), so it resolves instead of holding the spinner for the full 20 s.
  Unit-tested (`useScroll.test.ts`). The proper O(1) anchored-window rework is **not** here — it
  ships separately under [NATIVE-1224].
- **Form keyboard handling** — `FormContainer` swaps `KeyboardAvoidingView + ScrollView` for
  `KeyboardAwareScrollView` (`react-native-keyboard-controller`, already a dep). Affects the auth
  screens (Login, Register, ForgotPassword, SendEmailConfirmation, Workspace, NewServer) on both
  platforms — see *How to test*.
- **iOS Save Password suppression** (E2E only) — iOS 26 surfaces an async system `Save Password?`
  sheet after credential-field submit that blocks XCUITest. Under `RUNNING_E2E_TESTS=true` on iOS,
  `FormTextInput` declassifies the field (drops `secureTextEntry`/`textContentType`/`autoComplete`).
  Production behavior is unchanged; mirrors what `RegisterView` already does.
- **Sidebar testID** — `SidebarView` wraps the `ScrollView` in a `View` carrying `testID`; Android
  doesn't propagate `testID` from a `ScrollView` to the native tree. (Folded in from #7319.)

<details>
<summary><b>CI / Maestro stabilization</b> — infra only, no app behavior change</summary>

**Versions & runners**
- Maestro `2.2.0 → 2.5.1` (both platforms); 2.5.0's dadb ADB socket kills the Android `Broken pipe` install flake.
- iOS shards → `macos-26` + Xcode 26.2.0 to match the iOS-26 build artifact.

**Android caching / flake fixes**
- The AVD **and** SDK (emulator + system-image) caches are seeded once on `develop` and restored **read-only** by PR shards. A single writer means per-PR cache duplication no longer blows the 10 GB Actions cap and LRU-evicts the SDK cache; on a miss a shard cold-boots and installs on the fly.
- Serialized, 3×-retried `sdkmanager` install (clears partial downloads) self-heals the `Error on ZipFile unknown archive` flake.
- Emulator RAM `4 GB → 6 GB`; dropped `jlumbroso/free-disk-space` (shards already start with ~89 GB free) and the `monkey` warmup.

**iOS speed-ups**
- Dropped `idb-companion` (Maestro's iOS driver uses XCUITest since 1.18.0) — ~110s/shard.
- Start the async simulator boot right after `Configure Simulator` (before Java / app download / E2E account / Maestro setup) so the cold boot overlaps that work; block later on `bootstatus -b`.
- Trimmed `Wait for Simulator Ready`: dropped the redundant `simctl launch booted com.apple.springboard` + `sleep 5` — SpringBoard is already up once `bootstatus -b` returns.
- Fixed `UIAnimationDragCoefficient`: it is a duration *multiplier*, so the prior `-float 10` made every UIKit / native-stack transition 10× *slower* (the opposite of the "disable animations" intent). Set `0.0001` so transitions are near-instant and Maestro never waits on them — the iOS analogue of Android's `disable-animations`.
- The simulator device is resolved at runtime by family rather than a pinned name (image rotation renames devices): the small-screen shard runs the smallest iPhone (preferring an SE), the rest the largest `iPhone N Pro`. No match fails the shard loudly instead of silently falling back.
- A **warmup step** installs Maestro's XCUITest driver onto the booted simulator (~3 min on a cold sim) before the timed test step, so "Run Maestro Tests" measures flow execution only.

**Both platforms**
- The Maestro install is cached once by each platform's e2e build job (the single writer) and restored read-only by every shard, so all shards get a deterministic cache hit — no per-shard install and no concurrent first-writer save race. The cache holds only the install dirs (`bin`/`lib`/`deps`), never run logs — so the server-flake log scan can't be fooled by a restored prior run. Shards keep a cache-miss install as a fallback.
- Server **preflight** (`/api/info` before flows) + **mid-run scan** of Maestro logs → a server outage/blip renders as a single red job annotation instead of opaque `[Failed] <flow>`. If the server URL can't be scraped from the test config the job fails loudly (a CI-config error), rather than silently using a fallback.
- A wedged `maestro test` is bounded in-script (`timeout -k 30s 35m`) and annotated as an environment failure, with a 40-min job backstop above it. This replaces the old whole-suite startup retry: a run that produces no report now goes red for a human to re-run, rather than auto-retrying and masking real startup breakage. Per-flow rerun rounds `3 → 2`.
- Uploaded Maestro logs are scrubbed of the `&token=…` session token the deeplink-login flow writes (the logs ship as a public artifact).
- A **shard-coverage preflight** is the single source of the shard list both platforms fan out over: it asserts every flow under `.maestro/tests` is tagged within the declared range and every shard in the range has a flow, then emits the list as JSON for the matrices. A mistagged or orphaned flow fails fast instead of silently never running.
- Shard 9 restored (it had been folded into 3); both matrices now run 14 shards from the preflight list.
- **Shards rebalanced by measured per-flow duration.** Shard 12 carried ~35 min of iOS flow time — more than its 30-min step timeout, so Room died mid-run and Share Message never started. New layout: E2E Encryption (heaviest single flow on both platforms) gets shard 3 to itself (Server History → 1, Valid/Invalid Workspace → 2); Room Info / Markdown Click / Share Message move from 12 → 9 and Threads → 13, leaving shard 12 with Room + Open room 50+; Delete Server and Display Perf split from 6 onto 7 and 8. Heaviest iOS shard drops from ~35 to ~15 min. Static tags are an inherent compromise — per-flow durations diverge wildly between platforms — so duration-aware auto-balancing is tracked as a follow-up ([NATIVE-1265]).
- Artifact globs expanded to JUnit XMLs + the full `~/.maestro/tests` tree.

**Flaky-flow fixes**
- `open-deeplink`: anchor the dialog tap to `^Open$` (unanchored `Open` matched the iOS 26 alert *title*).
- `profile`: collapse 3 rate-limited `users.updateOwnBasicInfo` submits into one; `hide-keyboard` before clipped fields.
- `room-actions`: `scrollUntilVisible` the `Unstar` item instead of asserting after one swipe.
- `search-room` / `discussion` / `create-team` / `login`: more robust waits and `testID`/`header-back` selectors.

**Known remaining flaky flows (tracked as follow-ups)**
- **Jump to message** (both platforms) — pending the O(1) anchored-window rework ([NATIVE-1224]); the 20s scroll cap here is interim.
- **Keyboard Navigation** (Android) — components / room / onboarding don't reach the rooms list.
- **Room Actions** (both platforms) — usually passes on rerun; occasionally fails outright on a member-list / writing-toggle step.
- **Ignore User** (iOS) — fails its first attempt, passes on rerun.

Server / emulator-infra blips (workspace outages, adb socket loss) surface as a single red annotation, not app or test regressions.

</details>

## Issue(s)

Fixes [NATIVE-1123] · [NATIVE-1126]
Related (deferred, not fixed here): [NATIVE-1125] · [NATIVE-1128] · [NATIVE-1224]

## How to test or reproduce

1. Run the Build PR workflow; the 14 Android + 14 iOS Maestro shards should pass.
2. **Manual QA the auth screens** (Login, Register, ForgotPassword, SendEmailConfirmation, Workspace,
   NewServer) on iOS **and** Android after the `KeyboardAwareScrollView` swap — focus each field with the
   keyboard open/closed and confirm no clipping or scroll regressions, on phone and tablet.
3. Quoted-message **Jump to message** loads surroundings and scrolls to the target ([NATIVE-1126]).

## Screenshots

n/a

## Types of changes

## Checklist

- [x] I have added tests that prove my fix is effective (`sdk.test.ts` + `useScroll.test.ts`; CI flows cover the rest)

## Further comments

The sidebar testID fix was originally #7319, folded here so CI stabilization and its dependency ship
together (#7319 closed). Acceptance is measured over a 10-run window on `develop` after this lands.

[NATIVE-1123]: https://rocketchat.atlassian.net/browse/NATIVE-1123
[NATIVE-1125]: https://rocketchat.atlassian.net/browse/NATIVE-1125
[NATIVE-1126]: https://rocketchat.atlassian.net/browse/NATIVE-1126
[NATIVE-1128]: https://rocketchat.atlassian.net/browse/NATIVE-1128
[NATIVE-1224]: https://rocketchat.atlassian.net/browse/NATIVE-1224
[NATIVE-1265]: https://rocketchat.atlassian.net/browse/NATIVE-1265
[RocketChat/Rocket.Chat#39092]: https://github.com/RocketChat/Rocket.Chat/pull/39092


.github/workflows/build-pr.yml
```diff
@@ -52,15 +52,41 @@ jobs:
...
+  # that keeps it honest: it asserts every flow under .maestro/tests is tagged
+  # test-<N> within the declared range and every shard in the range has a flow,
+  # then emits the list as JSON for the matrices below. A flow tagged outside
+  # the range, or a gap in the range, fails here instead of letting that flow
+  # silently never run. No AVD seeding happens on PRs — the avd/sdk caches are
+  # written only on develop (build-develop.yml's seed-android-avd); shards
+  # restore them read-only and cold-boot on a miss.
+  # Gated behind e2e-hold, alongside the e2e builds, so it runs only once e2e is
+  # approved — a PR that never opts into e2e neither burns a runner here nor gets
+  # a red X from the drift guard for a suite it isn't running.
```
diegolmello: Weird and long comment

.github/workflows/e2e-build-android.yml
```diff
@@ -16,6 +16,11 @@ jobs:
   android-build:
     runs-on: ubuntu-latest
 
+    env:
+      # Single writer of the Maestro install cache the e2e Android shards
+      # restore. Keep in sync with maestro-android.yml — the key must match.
```
diegolmello: Remove

.github/workflows/e2e-build-android.yml
```diff
@@ -26,6 +31,37 @@ jobs:
       - name: Checkout and Setup Node
         uses: ./.github/actions/setup-node
 
+      # Warm the Maestro install cache once here: this job is the single writer
+      # of the key the e2e Android shards restore (Linux runner — matching their
+      # runner.os — and it gates them via needs). Every shard then gets a
+      # deterministic cache hit and only restores: no per-shard install, no
+      # concurrent first-writer race. lookup-only skips pulling the payload this
+      # job never uses on a warm cache.
```
diegolmello: Too long

.github/workflows/e2e-build-ios.yml
```diff
@@ -24,12 +24,48 @@ jobs:
   ios-build:
     runs-on: macos-26
 
+    env:
+      # Single writer of the Maestro install cache the e2e iOS shards restore.
+      # Keep in sync with maestro-ios.yml — the cache key must match exactly.
```
diegolmello: Remove

.github/workflows/e2e-build-ios.yml
```diff
@@ -24,12 +24,48 @@ jobs:
...
 
       - name: Set up Node.js
         uses: ./.github/actions/setup-node
+
+      # Warm the Maestro install cache once here: this job is the single writer
+      # of the key the e2e iOS shards restore (it runs on macOS — matching their
+      # runner.os — and gates them via needs). Every shard then gets a
+      # deterministic cache hit and only restores: no per-shard install, no
+      # concurrent first-writer race, no "failed to save" noise. lookup-only
+      # avoids pulling the ~200 MB payload this job never uses on a warm cache.
```
diegolmello: Too long

.github/workflows/maestro-android.yml
```diff
@@ -11,8 +11,13 @@ jobs:
   android-test:
     name: 'Android Tests'
     runs-on: ubuntu-latest
+    # Slowest passing shard ~27 min; bounds a wedged session above the in-script
+    # timeout -k wrap so a hang fails the job instead of burning the full 6h cap.
```
diegolmello: Remove

.github/workflows/maestro-android.yml
```diff
@@ -11,8 +11,13 @@ jobs:
...
+    timeout-minutes: 40
     env:
-      MAESTRO_VERSION: 2.2.0
+      # 2.5.0 switched gRPC to dadb direct ADB socket — eliminates flaky TCP forwarding (Broken pipe at install).
+      # https://github.com/mobile-dev-inc/maestro/releases/tag/cli-2.5.0
```
diegolmello: Remove

diegolmello:
## Review guide

Despite the `chore:` title, this is two PRs in one: ~5 small app-code changes that need real review, wrapped in a large CI/Maestro infra overhaul with no production behavior impact. Suggested effort split: ~80% on the app code, ~20% sanity-checking the workflows.

### App changes, ranked by risk

1. **`app/lib/services/sdk.ts` — the one true production bugfix.** Every DDP `methodCall` used to append the 2FA code (or `''`) as a trailing arg, which corrupts any server method with typed trailing params — concretely, `loadSurroundingMessages` gained a `showThreadMessages: boolean` 3rd param server-side, the `''` failed `check(…, Boolean)` → 500 → quoted-message jump never navigated. The fix appends the code only when one exists and clears it **before** the call, so a spent TOTP never leaks into later calls. The thing to convince yourself of: the 2FA retry path still works — `totp-required` → prompt → set code → retry → code consumed exactly once. `sdk.test.ts` covers no-2FA, prompt+retry, non-leakage, and cancellation.

2. **`app/containers/FormContainer.tsx` — widest blast radius.** Swaps `KeyboardView + ScrollView` for `KeyboardAwareScrollView` (`react-native-keyboard-controller`, already a dep). Touches **all auth screens** (Login, Register, ForgotPassword, SendEmailConfirmation, Workspace, NewServer) on both platforms. This can't be meaningfully reviewed from the diff — it needs manual QA: keyboard open/close on each screen, phone and tablet, watching for clipped fields and scroll jumps. Also note callers spreading `ScrollViewProps` now hit a different component.

3. **`app/containers/TextInput/FormTextInput.tsx` — E2E-only, but verify the gate.** Under `RUNNING_E2E_TESTS=true` on iOS it strips `secureTextEntry`/`textContentType`/`autoComplete` to suppress iOS 26's async "Save Password?" sheet that blocks XCUITest hit-testing. Check that `RUNNING_E2E_TESTS` cannot be true in a release build — that's what keeps production untouched. Known, accepted side effect: password masking is off under E2E on iOS (throwaway test users only).

4. **`useScroll.ts` + `RoomView/index.tsx` — interim band-aid, by design.** The jump-to-message scroll race cap goes 5s → 20s, plus a new early-exit: if the loaded-message count stays flat for 5 consecutive 600ms rounds, the target can't appear (deleted message or history exhausted) and it resolves instead of spinning the full 20s. This is explicitly temporary — the proper O(1) anchored-window rework ships separately under [NATIVE-1224](https://rocketchat.atlassian.net/browse/NATIVE-1224), so please don't ask for it here. Worth checking: the two new refs are reset in `resetJumpToMessage` (they are). `useScroll.test.ts` covers the give-up logic.

5. **`SidebarView/index.tsx` — trivial.** Wraps the `ScrollView` in a `View` carrying the `testID`, because Android doesn't propagate `testID` from a `ScrollView` to the native tree. Folded in from #7319.

### CI half — spot-check rather than deep-read

- **Single-writer caching** is the core idea: AVD/SDK/Maestro caches are seeded once on `develop` and restored read-only by PR shards (avoids blowing the 10 GB Actions cache cap and concurrent-writer races). Verify the PR-shard paths really are restore-only.
- **Fail-loud philosophy** throughout: the shard-coverage preflight asserts every flow is tagged and every shard populated; server preflight + mid-run log scan turn workspace outages into a single red annotation; the old whole-suite auto-retry is gone — a report-less run goes red for a human. The review question for each workflow change is "does anything still fail silently?"
- Subtle correctness fix to confirm: `UIAnimationDragCoefficient` was previously `10` — a duration *multiplier*, so animations ran 10× **slower** than intended; now `0.0001`.
- Security detail: uploaded Maestro logs are scrubbed of the deeplink-login `&token=…` (logs ship as a public artifact) — verify the scrub in `run-maestro.sh` runs before upload.

### Context the diff won't show

- **Known-flaky flows are documented, not hidden** — Jump to Message, Keyboard Navigation (Android), Room Actions, Ignore User (iOS) are listed in the PR description as tracked follow-ups. A red shard on one of those is not a reason to block.
- **Acceptance criterion** is a 10-run green window on `develop` after merge, not a single green PR run.
- The 14+14 shard matrices on the Build PR workflow are the de-facto integration test for the whole change.

**Fastest competent review:** read `sdk.ts` + its test line-by-line (4 lines of logic, real production stakes), QA the auth screens by hand, skim `useScroll.ts` for the give-up logic, and spot-check the workflow YAML for the read-only-cache and fail-loud claims above.


[NATIVE-1224]: https://rocketchat.atlassian.net/browse/NATIVE-1224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

OtavioStasiak (approved):
LGTM!

PR #7336, opened 2026-05-21
fix(a11y): replace composer auto-focus with header focus in RoomView

OtavioStasiak:
## Proposed changes

Replace composer auto-focus with header focus in RoomView.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1125

## How to test or reproduce

- Connect a physical keyboard;
- Open the app (on MasterDetail as true);
- focus on RoomItem;
- Press enter or space;
- The keyboard focus must change to header;

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Use a declarative, keyboard-aware focus wrapper for header accessibility
  * Simplified Room Header API by removing the ref-based imperative focus control
* **New Features**
  * Detects accessibility navigation (screen reader or external keyboard) to drive focus behavior
* **Tests**
  * Updated keyboard navigation test flow for composer and emoji interactions

[Review Change Stack](https://app.coderabbit.ai/change-stack/RocketChat/Rocket.Chat.ReactNative/pull/7336?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


app/containers/RoomHeader/RoomHeader.tsx
```diff
@@ -189,6 +170,9 @@ const Header = ({
 	const portrait = height > width;
 	let scale = 1;
 	const isMasterDetail = useAppSelector(state => state.app.isMasterDetail);
+	// Only move focus to the header for accessibility navigation (screen reader or physical
+	// keyboard); regular touch users shouldn't have focus yanked onto the header on room open.
+	const autoFocusHeader = useIsAccessibilityNavigationEnabled();
```
diegolmello, line 175: 🚨 **Mount-time race against `useIsScreenReaderEnabled`**

This hook starts as `false` and only becomes `true` a moment later (it has to ask the OS "is VoiceOver on?" via a promise and wait). But `KeyboardFocusView` only looks at `autoFocus` *once*, when it appears on screen — both platforms gate it behind a one-shot flag (iOS `_autoFocusRequested` in `RNCEKVViewFocusRequestBase.mm`, Android `hasBeenFocused` in `ViewFocusRequestBase.java`). When the prop later flips to `true`, the native setters just mutate the field; they don't re-trigger focus.

Fix idea: grab a ref to `KeyboardFocusView` and call `.focus()` in a `useEffect` when `autoFocusHeader` flips to `true`. Or read screen-reader state synchronously up front.

**What breaks:** VoiceOver on → open room → header never announced, VO stays on the room item.

app/containers/RoomHeader/RoomHeader.tsx
```diff
@@ -220,8 +204,14 @@ const Header = ({
 	const handleOnPress = () => onPress();
 
 	return (
-		<View
-			ref={headerRef}
+		<KeyboardFocusView
```
diegolmello, line 207: 🚨 **iOS Fabric + gesture-handler = unfocusable**

The child here is a `TouchableOpacity` from `react-native-gesture-handler`, which on the new arch (we have `newArchEnabled=true` on Android and Expo 54 enables Fabric on iOS) mounts as `RNGestureHandlerButtonComponentView`. The lib has an explicit special case for that class — `RNCEKVFocusDelegate.focusingViewForGestureHandler` returns the *inner* subview, so `canBecomeFocused` returns `false` (`getFocusingView() == _delegate` fails) and the UIKit focus engine ignores the `rncekvFocusView:` request.

Fix: use the lib's `withKeyboardFocus(TouchableOpacity)` HOC instead of wrapping. That's what every other call site in this repo does (`containers/Touch.tsx`, `message/Touch.tsx`, `Header/components/HeaderButton/Common.tsx`).

**What breaks:** iPad + Bluetooth keyboard → open room → arrow keys skip the header entirely.

app/views/RoomView/index.tsx
```diff
@@ -258,23 +254,6 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
 		this.unsubscribeBlur = navigation.addListener('blur', () => {
 			AudioManager.pauseAudio();
 		});
-		this.unsubscribeFocus = navigation.addListener('focus', () => {
```
diegolmello, line 261: ⚠️ **Re-focus on return from sub-screens is now lost**

This listener used to fire every time the screen regained focus — e.g. when the user closed `RoomActionsView`, `ThreadView`, or `AttachmentView` and came back. With it gone, `KeyboardFocusView.autoFocus` is the only focus path left, and it only fires *once* on first mount (the native view is reused across pushes/pops). So returning to RoomView leaves focus wherever the system dropped it.

Worth restoring an equivalent: a small `useEffect` listening to `navigation.addListener('focus')` that calls `.focus()` on a `KeyboardFocusView` ref.

**What breaks:** VoiceOver user opens room → taps header → opens Room Actions → swipes back → header is silent, focus lands on a random element.

app/views/RoomView/index.tsx
```diff
@@ -258,23 +254,6 @@ class RoomView extends React.Component<IRoomViewProps, IRoomViewState> {
...
-					return;
-				}
-				// Skip autofocus in development because simulators always report a keyboard as connected,
-				// which would force the composer to open on every focus while debugging.
-				if (__DEV__) {
-					return;
-				}
-				const hasExternalKeyboard = isExternalKeyboardConnected();
-				if (hasExternalKeyboard) {
-					this.messageComposerRef.current?.focus?.();
```
diegolmello, line 274: ⚠️ **Phone + external keyboard loses composer auto-focus with no replacement**

On a phone with a Bluetooth/USB keyboard, this line used to drop the cursor straight into the composer so the user could just start typing. The new code focuses the header instead — for everyone, master-detail or not — so keyboard users have to press Down 3+ times before they can type.

The maestro test patched around this with 3 extra `Remote Dpad Down` presses, but real users don't have that hint. Intentional, or should phone+keyboard users still land in the composer?

**What breaks:** Phone + keyboard → open room → start typing → nothing appears in the composer.

app/lib/hooks/useIsAccessibilityNavigationEnabled.ts
```diff
@@ -0,0 +1,12 @@
...
+import { isExternalKeyboardConnected } from '../methods/helpers/externalInput';
+
+/**
+ * Accessibility navigation is active when the user drives the UI with a screen reader or a
+ * physical keyboard. Use it to decide when focus should be moved programmatically (e.g. onto
+ * the room header when a room opens) instead of yanking focus around for regular touch users.
+ */
+export const useIsAccessibilityNavigationEnabled = (): boolean => {
+	const isScreenReaderEnabled = useIsScreenReaderEnabled();
+	return isScreenReaderEnabled || isExternalKeyboardConnected();
```
diegolmello, line 11: ⚠️ **`isExternalKeyboardConnected()` is a one-shot read, not reactive**

It's just a synchronous native-module call — no listener wired up. So if someone pairs or unpairs a Bluetooth keyboard while the app is running, this hook won't notice; it captures the value at render time and the component only re-renders when *something else* changes. The hook name (`useIsAccessibilityNavigationEnabled`) reads as if it reactively tracks both inputs, but only the screen-reader half does.

Fix options: subscribe to `GCKeyboardDidConnect`/`Disconnect` (iOS) / Bluetooth state changes (Android), or rename so callers don't assume reactivity it doesn't provide.

**What breaks:** Pair Bluetooth keyboard mid-session → reopen the same room → still no header focus.

diegolmello, line 11: 💡 **`__DEV__` guard is gone**

The old RoomView code had `if (__DEV__) return;` here because simulators *always* report a hardware keyboard as connected. The new hook offers no equivalent, so every room you open in iOS Simulator / Android emulator now auto-focuses the header (focus halo, accessibility events). Makes it harder to exercise the no-keyboard touch user path during development.

Probably worth `&& !__DEV__` on the external-keyboard branch.

**What breaks:** Dev runs in iOS Simulator → every room open shows the focus halo → can't repro touch-user behavior locally.

app/containers/RoomHeader/RoomHeader.tsx
```diff
@@ -220,8 +204,14 @@ const Header = ({
...
+			// Grab focus natively as soon as the header mounts. This handles master-detail,
+			// where the room list and room share the screen and focus would otherwise stay on
+			// the room item, as well as moving screen-reader focus via enableA11yFocus.
+			autoFocus={autoFocusHeader}
+			enableA11yFocus={autoFocusHeader}
+			focusable={!disabled}
+			canBeFocused={!disabled}
```
diegolmello: ⚠️ **`autoFocus` + `!disabled` are inconsistent**

When `disabled` is true (invite subscription), we tell the view both "auto-focus on mount" *and* "you can't be focused." On Android, `autoFocusOnDraw` still runs and `a11yFocus` calls `sendAccessibilityEvent(TYPE_VIEW_FOCUSED)` even when the underlying view has `focusable=false` — so TalkBack will announce a disabled invite header that previously stayed silent.

Suggested: `autoFocus={autoFocusHeader && !disabled}` (and same for `enableA11yFocus`).

**What breaks:** TalkBack user opens an invite room → header announces itself as "focused" even though it's disabled.

.maestro/tests/keyboardNavigation/keyboard-navigation-room.yaml
```diff
@@ -30,14 +30,15 @@ tags:
     timeout: 60000
 - pressKey: Remote Dpad Right
 - pressKey: Remote Dpad Right
-- pressKey: Remote Dpad Down
-- pressKey: Remote Dpad Down
 - pressKey: enter
```
diegolmello, line 35: ⚠️ **Removed two `Remote Dpad Down` before this `enter`**

Those two presses used to skip past the "Save your encryption password" `List.Item` (the test waits for it just above) and the omnichannel status row to land on the freshly-created room. Without them, `enter` likely activates whatever Right;Right left focused — probably the encryption banner — instead of opening the room.

Worth double-checking what this test is actually exercising on Android CI now; the new `Remote Dpad Down` x3 below is for the in-room composer flow, not the rooms-list traversal, so they're not a substitute.

**What breaks:** Test on Android CI → `enter` opens the encryption flow instead of the room → 60s timeout waiting for the composer.

diegolmello (commented):
Nice cleanup overall — the declarative wrapper is way easier to read than the old `useImperativeHandle` + focus-listener dance. Inline notes below; **#1 and #2 are the ones I'd want fixed before merge** — together they mean the a11y users this PR is for likely don't actually get header focus on either platform. The rest are smaller but worth a look.

app/lib/hooks/useIsAccessibilityNavigationEnabled.ts
OtavioStasiak: It’s not necessary. I tested it on both simulators, and it doesn’t affect any flow.

app/containers/RoomHeader/RoomHeader.tsx
OtavioStasiak: I tested using only withKeyboardFocus, and it doesn't work as expected in masterDetail. The approach I followed works as expected.

app/views/RoomView/index.tsx
OtavioStasiak: I added it just for keyboard navigation, but we changed the approach, so it's no longer needed.

app/views/RoomView/index.tsx
OtavioStasiak: it's intentional, now on the roomView the initial focus will be on Header.

app/lib/hooks/useIsAccessibilityNavigationEnabled.ts
OtavioStasiak: It's a good improvement, but it's not the main focus of the PR, I will create a debt tech ticket and work on it right after this one.

app/containers/RoomHeader/RoomHeader.tsx
OtavioStasiak: It's not necessary. After changing it, the autoFocus does its job and focuses on it. I've tested it on an Android tablet using master-detail and also in the Android simulator.

app/lib/hooks/useIsAccessibilityNavigationEnabled.ts
OtavioStasiak: https://rocketchat.atlassian.net/browse/NATIVE-1183

diegolmello (approved):

PR #7345, opened 2026-05-25
fix: Drain lost hangups on WebSocket reconnect

diegolmello:
## Proposed changes

When the WebSocket drops mid-call and the user taps End, the lib's hangup signal is written into a dead socket and lost. The server keeps the call alive, so the next call attempt is rejected with "Already on a call".

This PR adds a recovery path:

- `app/lib/services/voip/pendingHangups.ts` (new) — module-level `Set<string>` of callIds the user tried to hang up.
- `MediaSessionInstance.endCall` and `useCallStore.endCall` — `pendingHangups.record(callId)` before dispatching the hangup, so the intent survives if the wire signal is lost.
- `MediaSessionInstance.drainPendingHangups()` — re-dispatches each recorded callId via the lib's transporter (server-side hangup is idempotent, so replaying against an already-terminated call is a no-op).
- `connect.ts` — on `'close'`, arms a closure-local flag; on the next `'connected'`, awaits Redux `isAuthenticated && meteor.connected` (max 5s) and then drains. Listeners are rebound to the new `sdk.current` on every `connect()` call.
- Entries are bounded by `MediaSessionInstance.reset()` (disconnect / init / server switch).

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1175

## How to test or reproduce

1. Place an active VoIP call between two test users.
2. While the call is connected, kill the device's network (airplane mode / proxy block) so the WebSocket goes unhealthy without the call object being torn down.
3. Tap End. The local UI returns to the room.
4. Restore the network so the WebSocket reconnects.
5. After reconnect + DDP login, the recorded hangup is replayed; the server terminates the phantom call.
6. Initiate a new VoIP call to the same peer — it should connect without "Already on a call".

Unit coverage:

- `pendingHangups.test.ts` — record / drainAll / idempotency / clear.
- `MediaSessionInstance.test.ts` — `drainPendingHangups` routes recorded ids through the lib's transporter, no-ops when uninitialized, keeps draining on per-id throw; `endCall` does not record when the session is uninitialized or when `getCallData` returns a call with a mismatched id.
- `connect.test.ts` — `close` → `connected` drain happy path, plus the armed-flag, empty-set, and listener-cleanup guards on the reconnect drain listener.
- `useCallStore.test.ts` — endCall records the active callId.

## Screenshots

N/A — no UI changes.

## Types of changes

## Checklist

## Further comments

The reconnect drain gate lives in `connect.ts` (not in `setupMediaCallEvents`, which mounts once on `Root` and isn't re-bound on server switch). The gate is closure-local per `connect()` call, so it resets cleanly across server switches.

Cold-start recovery is intentionally absent: if the process is killed mid-flap, the in-memory set is gone and the server-side call-timeout is the only recourse. Android Telecom system-UI End is not covered — pre-existing gap, in-app End covers the Android path.

This PR does not touch `android/` or `ios/`; no native build / experimental build is needed.

## Summary by CodeRabbit

* **New Features**
  * Calls hung up while connection is unstable are queued and automatically replayed after reconnect once the session is ready, improving reliability and ensuring hangups are processed.

* **Tests**
  * Added unit and integration tests covering recording, replay/drain behavior, clearing of queued hangups, and robustness across reconnect cycles and error cases.

* **Documentation**
  * Added "Pending Hangup" vocabulary to Video & Voice docs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RocketChat/Rocket.Chat.ReactNative/pull/7345?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


diegolmello (commented):
Reviewed the PR with a multi-angle pass. Confirming the two CodeRabbit findings as valid:

- **`awaitDdpLoggedIn` timeout race** (`connect.ts:139-146`) — drain runs unauthenticated when the 5s timer wins. Whole fix mechanism fails silently. The suggested `Promise<boolean>` refactor is the right shape.
- **`drainAll` empties before send** (`MediaSessionInstance.ts:252-264`) — per-id throw permanently loses the id; collect-and-requeue on failure is the minimal fix.

Posting 4 additional inline comments for issues not yet on the PR:

1. `useCallStore.ts:309` — record skipped when only `nativeAcceptedCallId` is set (CallKit-from-lock-screen race).
2. `MediaSessionInstance.ts:232` — ringing `reject` replayed as `hangup` (wire-shape mismatch).
3. `MediaSessionInstance.ts:259` — drain always sends `hangup:normal` even when the lib path would have sent `another-client` / nothing.
4. `MediaSessionInstance.ts:325` — `reset()` wipes queue including via `init()` re-entry; also asymmetric with `useCallStore.reset()`.

Severity/likelihood: #1 and #2 are high-severity but trigger on narrow CallKit/declining races. #3 has the broadest coverage of replay-shape divergences. #4 is moderate but worth thinking through.

app/lib/services/voip/useCallStore.ts
```diff
@@ -303,6 +304,11 @@ export const useCallStore = create<CallStore>((set, get) => ({
 		// UUID for the native call UI layer (react-native-callkeep on iOS and Android).
 		const callUuid = callId ?? nativeAcceptedCallId;
 
+		if (callId) {
+			// Record before dispatching so a hangup written into a dead socket is replayed on reconnect.
+			pendingHangups.record(callId);
```
diegolmello, line 309: **Phantom call after CallKit-from-lock-screen + flap**

Line 305 already says either id identifies the call (`callId ?? nativeAcceptedCallId`), but the record guard only checks `callId`. If the user accepts via CallKit lock-screen UI and ends before `setCall` propagates `callId`, neither id is recorded — the drain on reconnect has nothing to replay.

**ELI5:** Two names refer to the same call. The line above uses either name. This line only listens for the first one. If only the second name is set, we forget the call exists.

**Failure:** Accept call from CallKit lock-screen → WS dies before lib finishes wiring `callId` → user taps End → no record → server's phantom call survives the reconnect.

Suggested fix:
```diff
-		if (callId) {
-			// Record before dispatching so a hangup written into a dead socket is replayed on reconnect.
-			pendingHangups.record(callId);
+		const recordId = callId ?? nativeAcceptedCallId;
+		if (recordId) {
+			pendingHangups.record(recordId);
 		}
```

*Severity: High (phantom call). Likelihood: Low (narrow race, but real on slow-network CallKit accepts).*

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -227,6 +228,9 @@ class MediaSessionInstance {
 		const mainCall = this.instance?.getCallData(callId);
 
 		if (mainCall && mainCall.callId === callId) {
+			// Record before dispatching so a hangup written into a dead socket is replayed on reconnect.
+			pendingHangups.record(callId);
```
diegolmello, line 232: **Wire-shape mismatch for ringing reject**

The recorded id will be replayed via `transporter.hangup(id, 'normal')` (line 259), but the original path for a `'ringing'` call goes through `mainCall.reject()` which sends `transporter.answer(callId, 'reject')` (lib `Call.js:490`) — a different signal type entirely.

**ELI5:** Declining a ringing call and hanging up an active call are two different words to the server. We always say the second one on replay, even when the original was the first.

**Failure:** Incoming call ringing → WS dies → user declines → drain on reconnect sends a hangup, but the server's ringing-call cleanup is gated on `answer:reject`. Phantom ringing call may persist until server-side timeout.

This is one of several drain-vs-lib divergences — see also the `'another-client'` / `hidden` / `!hasRemoteData` cases at line 259.

*Severity: High (wrong signal, phantom ringing). Likelihood: Low–Medium (incoming-during-flap is uncommon but realistic).*

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -240,6 +244,25 @@ class MediaSessionInstance {
...
+	 * `MediaSignalTransportWrapper.hangup` keeps the wire shape consistent with the happy path.
+	 */
+	public drainPendingHangups(): void {
+		if (this.instance == null) {
+			return;
+		}
+		const ids = pendingHangups.drainAll();
+		for (const id of ids) {
+			try {
+				(this.instance as any).transporter.hangup(id, 'normal');
```
diegolmello, line 259: **Drain sends `hangup:normal` regardless of what the lib would have sent**

Three wire-shape divergences from the happy path:

1. **`'another-client'` reason** — when `contractState === 'ignored'` (`Call.js:510-512`) the lib sends `transporter.hangup(callId, 'another-client')`. Drain replays as `'normal'`, telling the server this device is the primary terminator. Contract bookkeeping ends up on the wrong slot.
2. **Hidden / `!hasRemoteData`** — `flagAsEnded` (`Call.js:933`) gates on `!hidden && hasRemoteData` and skips the wire send entirely. Drain ignores both guards.
3. **(See comment on line 232)** — ringing `reject` uses `transporter.answer(callId, 'reject')`, not `hangup`.

**ELI5:** The original code is fluent in several "End-the-call" phrases (`hangup:normal`, `hangup:another-client`, `answer:reject`, silence). The drain only speaks one phrase regardless of context.

**Failure scenarios:**
- Outbound call cancelled before server's `'new'` ack → lib would have sent nothing → drain sends a hangup for a callId the server never registered.
- Multi-device user where contract was signed elsewhere → lib would have sent `'another-client'` → drain sends `'normal'` → server cleans up wrong contract.

Fix sketch: record `{callId, reason}` (or `{callId, kind: 'hangup' | 'reject'}`) instead of just `callId`. Derive the right shape at `endCall` time (`'reject'` for ringing, `'another-client'` for ignored, etc.). On replay, dispatch the recorded shape. Or skip recording entirely for the no-wire-shape paths (`hidden`, `!hasRemoteData`).

*Severity: Medium–High (depending on which shape). Likelihood: Low individually, but #1 covers a broad set of paths.*

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -299,6 +322,7 @@ class MediaSessionInstance {
 		}
 		mediaSessionStore.dispose();
 		this.instance = null;
+		pendingHangups.clear();
```
diegolmello, line 325: **`reset()` wipes recorded intent — including via `init()` re-entry**

`reset()` is also the first line of `init()` (line 86). If anything triggers re-init during a flap (auth event, logged-user change, setupMediaCallEvents re-run) before the drain listener fires, the recorded intent is silently lost.

**ELI5:** Every time the media session is set up, this wipes the queue. If anything re-sets-up the session between "user taps End" and "WS comes back," the intent disappears.

Also worth comparing: `useCallStore.reset()` does **not** clear `pendingHangups` — asymmetric with this one. After a normal call ends, ids leak there until next reconnect drain. Server-side hangup is idempotent so it's noise, not a bug, but the asymmetry is worth a thought.

**Failure:** User taps End during flap → recorded → before drain fires, `init()` re-runs → `reset()` → `clear()` → drain finds nothing.

*Severity: High when it triggers (fix becomes inert). Likelihood: Low (init is usually once-per-session).*

Question for design: should the queue survive re-init for the same user? Re-init typically means "recreate the session for the same user" — in which case the recorded intent is still valid.

OtavioStasiak (approved):
LGTM!

PR #7346, opened 2026-05-26
fix(android/voip): start microphone FGS for outgoing calls so audio survives backgrounding

diegolmello:
## Proposed changes

Outgoing VoIP calls went through `MediaSessionInstance.startCall` directly to the media-signaling SDK and never touched `VoipNotification`, so no foreground service was started. With no FGS in the active set, Android revokes `RECORD_AUDIO` (while-in-use) ~5 s after the user backgrounds the app, dropping the caller's audio while the peer's still plays.

The incoming-accept path already starts `VoipCallService` from native after Telecom is active. This PR mirrors that behaviour for outgoing calls by:

- Exposing `startVoipCallService(callId): Promise<void>` on the `RNVoipModule` TurboModule. The promise is held until `VoipCallService.onStartCommand` reports back via `VoipModule.notifyFgsStarted` / `notifyFgsFailed`, so async failures (`ForegroundServiceTypeNotAllowedException`, `ForegroundServiceDidNotStartInTimeException`, `SecurityException` at `startForeground` time) reach JS instead of being masked by a synchronous `resolve(null)`. A 7-second timeout rejects with `E_VOIP_FGS_TIMEOUT` if the service never signals back. Synchronous failures from `startForegroundService` still reject with `E_VOIP_FGS_START`.
- Invoking it Android-only from the `newCall` `'caller'` branch in `MediaSessionInstance.ts`, while the initiating activity is still visible (so the `microphone` FGS-start eligibility is satisfied).
- Gating `Navigation.navigate('CallView')` on the FGS promise resolving. On rejection JS tears the outgoing call down via `endCall(callId)` + `VoIP_Call_Issue` alert and stays on the current screen — previously the eager navigate stranded the user on a `CallView` whose `'ended'` listener was detached by `useCallStore.reset()` before async `hangup()` could fire `Navigation.back()`.
- iOS `VoipModule.mm` gets a no-op `resolve(nil)` stub to satisfy the codegen protocol.

FGS type widens from `phoneCall` to `microphone|phoneCall`:
- `phoneCall` keeps the incoming-accept entry point startable from `IncomingCallActivity` / `MainActivity.onNewIntent` via the active self-managed Telecom call (PR #7233 rationale).
- `microphone` is what actually authorises background mic capture — `phoneCall` alone keeps the call alive via `ConnectionService` but does not exempt `RECORD_AUDIO` from the while-in-use clock.

Both call sites reach `startForeground()` from a while-in-use eligible state: outgoing from a visible activity, incoming after `connection.onAnswer()` puts Telecom into `STATE_ACTIVE`. Doc note at `app/lib/services/voip/docs/PLATFORMS.md` was wrong and is corrected in the same commit.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1178

## How to test or reproduce

Android manual repro (outgoing audio drop):
1. Start an outgoing VoIP call from the app.
2. Confirm two-way audio works in the foreground.
3. Background the app (home / recents) and continue speaking for at least 10 seconds.
4. Confirm the peer continues to hear the caller's audio — before this fix, audio drops ~5 s after backgrounding.
5. Foreground the app, confirm the call is still active and audio is restored both ways.
6. End the call; verify the foreground service / notification is cleared.

Android manual repro (FGS-rejection no-strand):
1. Force an FGS-start failure (e.g. background-restricted state on Android 12+ that triggers `ForegroundServiceStartNotAllowedException`, or temporarily revoke `FOREGROUND_SERVICE_MICROPHONE` in the manifest).
2. Initiate an outgoing call.
3. Confirm the user sees the `VoIP_Call_Issue` alert and stays on the originating screen — no blank `CallView`.

## Screenshots

N/A — audio-only fix.

## Types of changes

## Checklist

## Further comments

Five regression tests in `MediaSessionInstance.test.ts`: outgoing-caller starts the FGS on Android, incoming-callee does not, an FGS-start rejection tears down the call, navigation runs after FGS resolves, and navigation does NOT run on FGS rejection. The `microphone` FGS type is deliberately scoped to the two existing call sites (outgoing from visible activity, incoming after `connection.onAnswer()`); no new code path starts `VoipCallService` from a non-foreground context, preserving the PR #7233 fix for the `SecurityException` crash class.

## Summary by CodeRabbit

* **New Features**
  * Android: Outgoing VoIP calls now start a native foreground service and app declares microphone+phone-call foreground capability.
  * JS API: Added start-service method; iOS is a no-op placeholder. Foreground start is coordinated with JS and can time out.

* **Bug Fixes**
  * Safer handling when foreground service startup fails: surfaces error to JS, shows alert, ends call, and prevents navigation.

* **Tests**
  * Added tests for outgoing start flow and failure handling.

* **Documentation**
  * Updated platform docs with foreground-service and permission timing details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RocketChat/Rocket.Chat.ReactNative/pull/7346?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


android/app/src/main/java/chat/rocket/reactnative/voip/VoipCallService.kt
```diff
@@ -100,10 +100,17 @@ class VoipCallService : Service() {
         val notification = buildNotification(callId)
 
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            // microphone keeps RECORD_AUDIO authorised once the app is no longer in the foreground
+            // (background mic is governed by FGS type, not by audio focus or Telecom state alone).
+            // phoneCall is what satisfies FGS-start eligibility on the incoming-accept path, where
+            // the only foreground signal is the active self-managed Telecom connection. Both paths
+            // (outgoing start from a visible activity, incoming start after connection.onAnswer())
+            // are in a while-in-use eligible state at startForeground() time, so the combined
+            // bitmask is safe to start.
```
diegolmello, line 127: The inline rationale here conflates **start** eligibility with **runtime** mic authorisation.

> microphone keeps RECORD_AUDIO authorised once the app is no longer in the foreground … phoneCall is what satisfies FGS-start eligibility on the incoming-accept path

This reads as if only `phoneCall` is gated on start-eligibility and `microphone` rides along for free. On Android 14+ that is not how the combined bitmask is scoped — when you start an FGS with `microphone|phoneCall`, **both** types must independently satisfy their start restrictions. On the incoming-accept path, both bits are riding the same exemption (`STATE_ACTIVE` self-managed Telecom), not just `phoneCall`. That is load-bearing context for whoever next has to reason about this.

Suggested tightening:

```kotlin
// Both bits' start-eligibility rely on the same exemption: a STATE_ACTIVE self-managed
// Telecom connection (incoming) or a visible activity (outgoing). microphone additionally
// authorises RECORD_AUDIO once the app backgrounds — phoneCall alone does not.
```

The same nit applies to the comment in `AndroidManifest.xml` (lines 153–159) and the bullets in `app/lib/services/voip/docs/PLATFORMS.md`. The rationale is currently duplicated in three places — next change to this story has to find and update all three. Pick one canonical home (probably `PLATFORMS.md`) and have the others one-liner-link to it.

diegolmello (commented):
Single inline finding from a structural review — otherwise this PR is clean (focused, no spaghetti added to unrelated flows, no file pushed past 1k lines, TurboModule contract is explicit, failure path correctly tears down rather than silently degrading).

OtavioStasiak (approved):
LGTM!
Tested on both platforms.

PR #7354, opened 2026-05-27
fix(voip): suppress incoming voice calls locally when microphone denied

diegolmello:
## Proposed changes

When the OS **microphone permission** is denied, an incoming voice call used to ring anyway — but answering it produced a dead call, because the app cannot open the microphone. On a locked or backgrounded device the app cannot even show a permission dialog at answer time, so the user was offered a call they could never take. The ring was pure noise.

This PR moves the decision to the **native push layer**: a call that can never carry audio _on this device_ is **suppressed locally** the moment the push arrives — before any ringing UI — and **nothing is signalled to the server**, so the call keeps ringing on the user's other devices (desktop, another phone) and can still be answered there. As symmetrically as each platform allows:

- **Android — truly silent.** `VoipNotification.onMessageReceived` runs a pure routing decision (`decideIncomingVoipPushAction`, precedence `stale → no-permission → busy → show`). When `RECORD_AUDIO` is denied the push is ignored (log-only) and returns: no Telecom registration, no notification, no full-screen incoming screen, no REST call.
- **iOS — momentary flash.** PushKit hard-requires reporting the call to CallKit before the handler returns, or the OS kills the app and stops delivering VoIP pushes. The handler reports a placeholder call and ends it immediately (a sub-second CallKit flash), signals nothing to the server, and skips stashing the event for JS, so opening the app cannot resurrect a dead call.

An earlier iteration of this change sent a REST `reject` from this device. That was wrong: a reject ends the call for **every** device of the user, so a phone with a denied mic would silently prevent the user from answering on an authorized device. The gate now signals nothing.

Two supporting layers are retained:

- **Pre-acquire** the microphone permission proactively at the end of session init (`MediaSessionInstance.init`, fire-and-forget, suppressed during an active call), so the common case stays granted and calls ring normally.
- The JavaScript **answer-time gate** is a defence-in-depth backstop for the narrow window where the permission is revoked between push receipt and answer. By then native has already sent the REST accept (the server has stopped ringing the user's other devices), so the backstop **ends** the call rather than handing it back.

The granted path is unchanged on both platforms; outgoing `startCall` behaviour is unchanged (in-context request + denial alert).

Design and vocabulary are documented in-repo under `app/lib/services/voip/docs/` (`CONTEXT.md`, `ARCHITECTURE.md`, `FLOWS.md` §3, `PLATFORMS.md`).

## Issue(s)	
https://rocketchat.atlassian.net/browse/NATIVE-1139

## How to test or reproduce

1. Revoke the microphone permission for Rocket.Chat in OS Settings on the test device.
2. **Android:** receive an incoming VoIP call → the device stays completely silent (no ring, no notification, no screen). The call is **not** declined for the user.
3. **iOS:** receive an incoming VoIP call → only a momentary CallKit flash appears, then it self-dismisses; opening the app does not present the call. The call is **not** declined for the user.
4. **Other-device check:** sign the same user in on a second device with the mic **granted** → the call rings there and can be answered, even while the mic-denied device stays silent. This is the behaviour the PR protects (a reject would have stopped it).
5. Repeat 2–3 with the device **locked** and with the app **backgrounded** — same result (no answer dialog is ever needed).
6. **Grant** the permission → incoming calls ring and connect normally on both platforms (regression).
7. Fresh login with the permission undetermined → the app pre-acquires it (foreground prompt) so later locked/backgrounded calls can ring.
8. Outgoing call without permission → unchanged: prompts in context and shows the microphone-denied alert.

## Types of changes

## Checklist

## Further comments

- **Why the push layer:** the ring originates in the native push handlers (Android FCM → Telecom/notification; iOS PushKit → CallKit), not in JavaScript, so the only place to stop it before it rings is natively at push receipt.
- **Why suppress instead of reject:** a REST `reject` from one device ends the call for the whole user, which would stop other authorized devices from answering. A local-capability problem (this device's mic) must not produce a user-wide signal — so the gate mirrors the existing stale/expired-push handling and signals nothing.
- **Platform asymmetry is intrinsic, not a shortcut:** Android can drop a push with no UI; iOS PushKit mandates a CallKit report before the handler returns. The momentary iOS flash mirrors the existing expired/unparseable-payload handling and is the accepted floor.
- **Precedence (no-permission before busy)** is documented in code and in the routing unit tests; in practice an active call implies the permission is granted, so the combination is near-impossible — the explicit ordering just removes ambiguity.
- **Testing:** the pure Android routing decision is unit-tested (`VoipIncomingCallDispatchTest`: denied → ignored, denied-while-busy precedence, granted, stale). The native push wiring (Kotlin/Swift) is bound to framework singletons with no Swift unit harness, so it is verified by build plus manual QA on locked/backgrounded devices. iOS manual QA (real incoming call, mic denied, locked + backgrounded) is pending.

## Summary by CodeRabbit

* **New Features**
  * Pre-acquire microphone permission at login for seamless incoming-call handling.
  * Native-layer silent rejection of incoming calls when microphone permission is denied.
  * Context-aware denial alert offering appropriate retry/settings options.

* **Bug Fixes**
  * Prevent duplicate/overlapping answer attempts; denied-permission paths now end calls silently.

* **Documentation**
  * Added architecture, flow, and ADR docs detailing microphone-permission behavior.

* **Tests**
  * Expanded test coverage for permission flows, alerts, and incoming-call routing.


app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -173,6 +173,12 @@ class MediaSessionInstance {
 		const mainCall = this.instance?.getCallData(callId);
 
 		if (mainCall && mainCall.callId === callId) {
+			const permission = await requestVoipCallPermissions();
```
diegolmello: **[High] `requestVoipCallPermissions()` is awaited with no try/catch — can re-introduce the ~10s hang.** Only `mainCall.accept()` below is wrapped. If this rejects — e.g. on Android `PermissionsAndroid.request()` throws *"Tried to use permissions API while not attached to an Activity"* when the call is answered from a locked/backgrounded state — the throw escapes `answerCall`, is swallowed by the `queueMicrotask(() => this.answerCall(...).catch(log))` caller, and the call is **neither accepted nor ended**: it hangs until the 10s signaling timeout (the exact bug this PR fixes). Suggest wrapping in try/catch and calling `endCall(callId)` (+ the denied alert) on error.

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -173,6 +173,12 @@ class MediaSessionInstance {
...
+			if (!permission.granted) {
```
diegolmello: **[Medium] The new `await` widens the idempotency window; this deny branch has no dedup.** `answerCall` is documented as idempotent and is invoked from two paths (REST replay in `applyRestStateSignals` + the live `stream-notify-user` listener, both gated only on `call == null`). `setCall` — which clears `nativeAcceptedCallId` and is the idempotency anchor — runs only after `accept()`. Adding `await requestVoipCallPermissions()` before it lets two concurrent invocations both clear the `existingCall == null` guard, so on denial both run `endCall` + `showVoipMicrophoneDeniedAlert` → **two stacked alerts + double teardown** (on grant: double `accept()`/navigate). Consider an in-flight guard keyed by `callId`.

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -173,6 +173,12 @@ class MediaSessionInstance {
...
+				this.endCall(callId);
```
diegolmello: **[Low] `endCall` sends `reject()` (decline) here even though the user tapped Accept.** Because JS `accept()` never ran, `mainCall.state` is still `'ringing'`, so `endCall` takes the `reject()` branch — signaling *declined* to the remote even though the callee accepted in CallKit/Telecom. The sibling `catch` branch (accept threw) deliberately avoids reject/hangup and only calls `terminateNativeCall`. Worth a conscious decision on whether *decline* is the intended wire signal for permission-denied.

app/lib/methods/voipCallPermissions.ts
```diff
@@ -1,11 +1,65 @@
...
+			canAskAgain: result !== PermissionsAndroid.RESULTS.NEVER_ASK_AGAIN
+		};
+	}
+
+	const { granted, status, canAskAgain } = await Audio.getPermissionsAsync();
+	if (granted) {
+		return { granted: true, canAskAgain: true };
+	}
+	if (status === PermissionStatus.UNDETERMINED || canAskAgain) {
+		const requested = await Audio.requestPermissionsAsync();
```
diegolmello: **[Low] iOS can't present this prompt for a call answered while locked/backgrounded.** `requestPermissionsAsync()` needs a foreground app to show the system dialog; for an incoming call answered from the lock screen with `UNDETERMINED` permission it may resolve denied with no prompt → the call ends and `showVoipMicrophoneDeniedAlert` fires where the user can't see it. Inherent to a reactive gate — the proactive init-time gate mentioned in the PR description would address it.

app/lib/methods/voipCallPermissions.ts
```diff
@@ -1,11 +1,65 @@
...
+		Alert.alert(
+			I18n.t('Microphone_access_needed_to_record_audio'),
+			I18n.t('Go_to_your_device_settings_and_allow_microphone'),
+			[{ text: I18n.t('Ok') }],
+			{ cancelable: true }
+		);
+		return;
+	}
+
+	Alert.alert(
```
diegolmello, line 80: **[Reuse] This permanent-denial alert is a near-verbatim copy of `MicOrSendButton.tsx:38-52`** (same title/message keys, Cancel/Settings buttons, `openAppSettings`, `cancelable:false`). This helper is the right place to consolidate — consider refactoring `MicOrSendButton` to call it (or a shared `showMicrophoneDeniedAlert`) so the dialog lives in one place. Also: both `Alert.alert` calls in this function repeat the identical title+message keys (42-43 / 51-52); could be a single call with buttons/options varying by `canAskAgain`.

app/lib/methods/voipCallPermissions.ts
```diff
@@ -1,11 +1,65 @@
...
+		if (result === PermissionsAndroid.RESULTS.GRANTED) {
+			return { granted: true, canAskAgain: true };
+		}
+		return {
+			granted: false,
+			canAskAgain: result !== PermissionsAndroid.RESULTS.NEVER_ASK_AGAIN
+		};
+	}
+
+	const { granted, status, canAskAgain } = await Audio.getPermissionsAsync();
```
diegolmello: **[Reuse] This iOS flow duplicates `MicOrSendButton.startRecording` almost line-for-line** (getPermissions → granted short-circuit → `UNDETERMINED`/`canAskAgain` → request). A shared `requestMicrophonePermission(): { granted, canAskAgain }` could back both call sites.

diegolmello (commented):
## Code review — incoming VoIP mic-permission gate

Solid fix on the happy path with good unit coverage. The main risks are on the **error / background path of the incoming-answer flow** (where the new `await` was inserted), plus some duplication with the existing audio-recording permission code in `MicOrSendButton`. Inline comments below, roughly highest-severity first.

**Verified non-issues:** `PermissionStatus` from `expo-camera` matches expo-av's status enum (both re-export from `expo-modules-core`; same pattern already ships in `MicOrSendButton`); Android `result !== NEVER_ASK_AGAIN` covers all return values; all i18n keys (`Ok`/`Cancel`/`Settings`/…) exist; no dead imports.

app/lib/methods/voipCallPermissions.ts
```diff
@@ -1,11 +1,65 @@
...
+	if (granted) {
+		return { granted: true, canAskAgain: true };
+	}
+	if (status === PermissionStatus.UNDETERMINED || canAskAgain) {
```
diegolmello: **[Simplify] `status === PermissionStatus.UNDETERMINED || canAskAgain` is redundant** — when status is `UNDETERMINED`, expo reports `canAskAgain === true`, so this reduces to `if (canAskAgain)` (and removes the need for the `expo-camera` `PermissionStatus` import on this branch). Relatedly, `requested.canAskAgain ?? false` on line 33 is dead code — expo's `PermissionResponse.canAskAgain` is a non-optional boolean.

app/lib/methods/voipCallPermissions.test.ts
```diff
@@ -1,65 +1,143 @@
...
+	__esModule: true,
+	default: { t: (key: string) => key }
+}));
+
+jest.mock('./helpers/openAppSettings', () => ({
+	openAppSettings: jest.fn()
+}));
 
 describe('requestVoipCallPermissions', () => {
+	beforeEach(() => {
```
diegolmello, line 40: **[Tests] Two notes here.** (1) `beforeEach` uses `clearAllMocks()` but `afterEach` uses `restoreAllMocks()` — `clearAllMocks` doesn't reset the module-factory `jest.fn()` implementations, so `mockResolvedValue`s bleed across tests; and the last Android test leaves `Platform.OS='android'`, which leaks into the `showVoipMicrophoneDeniedAlert` describe (no reset there). Harmless today but order-dependent — `resetAllMocks()` + resetting `Platform.OS` would be safer. (2) **Coverage gap:** no test exercises the iOS *live-prompt* denial (`requestPermissionsAsync` → `granted:false`) or the `canAskAgain ?? false` fallback — only pre-granted, undetermined→granted, and pre-denied are covered. The user-denies-at-the-dialog case is the gate's whole point.

app/lib/services/voip/MediaSessionInstance.test.ts
```diff
@@ -871,6 +874,46 @@ describe('MediaSessionInstance', () => {
...
+				accept: jest.fn().mockResolvedValue(undefined),
+				reject: jest.fn(),
+				hangup: jest.fn(),
+				remoteParticipants: [{ contact: { username: 'bob' } }]
+			};
+			session.getCallData.mockReturnValue(mainCall);
+
+			await mediaSessionInstance.answerCall('call-perm');
+
+			expect(endCallSpy).toHaveBeenCalledWith('call-perm');
```
diegolmello, line 935: **[Tests] This asserts `endCall` was called but never that `reject()` (not `hangup()`) actually fired.** `endCallSpy` calls through, so the real `state==='ringing'` branch runs, but a regression flipping it to `hangup()` (or skipping `reject`) would still pass. The pre-existing `endCall` test asserts `mainCall.reject()`/`hangup()` directly — worth matching that here, and also asserting the native teardown (`terminateNativeCall`/`resetNativeCallId`) on the denied path.

app/lib/services/voip/MediaSessionInstance.ts
diegolmello: Fixed in 5ed231970. `requestVoipCallPermissions` now wraps its full body in try/catch — a throw (e.g. Android `PermissionsAndroid.request` when not attached to an Activity) is logged and returned as `{ granted: false, canAskAgain: false }`, so it can never reject back into `answerCall` and re-introduce the ~10s hang. New tests cover the Android and iOS throw paths.

app/lib/services/voip/MediaSessionInstance.ts
diegolmello: Fixed in 5ed231970. `answerCall` now keys an in-flight `Set<string>` by `callId` (checked/added at the top, cleared in a `finally`), so a second entry while the first is still awaiting `requestVoipCallPermissions()` returns early. New tests assert that concurrent grant accepts/navigates once and concurrent deny ends/alerts once.

app/lib/services/voip/MediaSessionInstance.ts
diegolmello: Keeping `reject()` as the wire signal — a conscious decision. With permission denied the JS media session never accepts, so the call is still `'ringing'` and a callee that can't grant the mic is effectively *declining* from the remote's perspective; the protocol has no distinct 'accepted-but-failed' signal, and the sibling `catch` (accept threw mid-session) is a genuinely different state. No change.

app/lib/methods/voipCallPermissions.ts
diegolmello: Acknowledged — inherent to a reactive gate. The proper fix is a proactive init-time permission gate (noted in the PR description), which is out of scope for this PR. The never-reject helper (5ed231970) does improve this case regardless: a locked/backgrounded answer that resolves denied with no visible prompt now ends the call promptly instead of hanging.

app/lib/methods/voipCallPermissions.ts
diegolmello: In-file part done in 5ed231970 — the two `Alert.alert` calls are now a single call with the buttons and `cancelable` varying by `canAskAgain`. The cross-file consolidation (folding `MicOrSendButton`'s permanent-denial alert into a shared helper) is tracked in NATIVE-1198.

app/lib/methods/voipCallPermissions.ts
diegolmello: Cross-file dedup deferred to NATIVE-1198 — a shared `requestMicrophonePermission(): { granted, canAskAgain }` backing both this flow and `MicOrSendButton.startRecording`. Out of scope for this PR.

app/lib/methods/voipCallPermissions.ts
diegolmello: Fixed in 5ed231970 — reduced to `if (canAskAgain)`, dropped the dead `?? false` fallback, and removed the now-unused `status` destructure and `PermissionStatus` import.

app/lib/methods/voipCallPermissions.test.ts
diegolmello: Fixed in 5ed231970 — `beforeEach` now uses `resetAllMocks()`, and `Platform.OS` is reset to `'ios'` in the alert `describe` so the trailing Android test no longer leaks into it. Also added the iOS live-prompt denial case.

app/lib/services/voip/MediaSessionInstance.test.ts
diegolmello: Fixed in 5ed231970 — the denied-permission test now asserts `mainCall.reject()` fired once (and `hangup()` did not) plus `terminateNativeCall`/`resetNativeCallId` ran, matching the accept-rejects test's rigor.

app/i18n/locales/en.json
```diff
@@ -570,6 +570,7 @@
   "Message_was_not_read": "Message was not read",
   "Message_was_read": "Message was read",
   "messages": "messages",
+  "Microphone_access_needed_for_voice_calls": "Microphone access needed for voice calls",
```
diegolmello, line 573: Missing other languages

app/sagas/login.js
```diff
@@ -264,7 +265,11 @@ const checkVoipPermission = async () => {
 			return;
 		}
 		if (!mediaSessionStore.getCurrentInstance()) {
-			mediaSessionInstance.init(userId);
+			mediaSessionInstance.init(userId).catch(log);
+			// Pre-acquire the microphone now (foreground, post-login) so an incoming call while the
+			// device is locked/backgrounded — where a permission dialog is impossible — can still be
+			// answered. Fire-and-forget: never block session setup on the OS dialog.
+			preAcquireVoipMicPermission().catch(log);
```
diegolmello: This should be done inside MediaSessionInstance, at the end of init.

app/i18n/locales/en.json
diegolmello: Done in `2b174b6f7` — added the key to the eight locales that carry the sibling VoIP strings (`Missed_call`, etc.): bn-IN, cs, hi-IN, hu, no, pt-BR, ta-IN, te-IN. It now sits in exactly the same set of locales as the rest of the VoIP namespace, rather than only `en`.

app/sagas/login.js
diegolmello: Done in `c065776d2` — moved `preAcquireVoipMicPermission()` out of the login saga and into the end of `MediaSessionInstance.init()` (fire-and-forget, `.catch(log)`). The import and call are removed from `app/sagas/login.js`.

PR #7348, opened 2026-05-27
perf(markdown): reduce room-open render cost with parse cache and compiler memoization

diegolmello:
## Proposed changes

Opening a room was spending significant time re-parsing markdown for every message. This PR reduces that cost by:

- Adding an LRU-style parse cache (max 200 entries) so identical message strings skip re-parsing
- Extracting block rendering into a memoized `MarkdownBlockView` component with React Compiler `'use memo'`
- Stabilizing the markdown context value object to avoid unnecessary downstream re-renders
- Moving inline block container styles into `styles.ts`

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1184

## How to test or reproduce

1. Open a room with many messages containing markdown (lists, code blocks, quotes, mentions, etc.)
2. Compare room open time / scroll performance before and after
3. Verify markdown still renders correctly for:
   - Plain text and formatted text
   - Lists, code blocks, quotes, headings
   - Mentions, channels, and custom emoji
   - Translated messages (`isTranslated` path should still re-parse)
4. Navigate away and back to the same room — cached messages should render without visual regressions

## Benchmark (iOS sim, Argent React Profiler)

Controlled A/B on the same app state — original `Markdown` vs this PR (compiler-only, no manual `useMemo`).

| Control | Value |
|---|---|
| Device | iPhone 16e, `open.rocket.chat`, logged in |
| Flow | Rooms list → tap **Diego Mello** → scroll → back |
| Gestures | tap `(0.5, 0.297)`, swipe `(0.5,0.7)→(0.5,0.3)`, back `(0.05, 0.07)` |
| Cold start | Metro reload before each run |
| Before | Original `Markdown` (git stash) — session `20260527-160938` |
| After | Optimized `Markdown` (this PR) — session `20260527-163815` |

| Metric | Before | After | Δ |
|---|---|---|---|
| **Primary mount batch** (11×) | **243.9ms** self / **517ms** commit | **48.1ms** self / **134ms** commit | **−80% self / −74% commit** |
| **All Markdown mounts** (44×, 7 commits) | **295.8ms** total self | **69.8ms** total self | **−76%** |
| Incremental list batches (avg self/instance) | ~2.5–3.5ms | ~0.5–1.0ms | ~**3× faster** |

### Per-commit Markdown self-time

| Batch | Before | After |
|---|---|---|
| Primary (11×) | 243.9ms | 48.1ms |
| 7× | 10.9ms | 6.8ms |
[...]

PR #7361, opened 2026-05-29
perf(call): cut HeaderCallButton mount cost by consolidating call hooks

diegolmello:
## Proposed changes

`HeaderCallButton` (rendered in the RoomView navigation header) ran a heavy hook stack on **every mount**: ~12 redux `useSelector` calls (8 of them one-per-setting), a native `expo-camera` permission hook, and a duplicated zustand selector — none of which is needed in the first frame. Because React Navigation renders the header twice during the push transition, this work runs ×2 in the critical room-open commit.

This PR consolidates that per-mount work in the **shared** call hooks (behavior-preserving — same return values; every consumer benefits: RoomView header, Room Info, Room Actions, Sidebar, VideoConferenceEnded):

- **`useVideoConfCall`**: 8 separate `useSetting` reads → 1 `useAppSelector(..., shallowEqual)` (−7 store subscriptions)
- **`useVideoConf`**: reactive `useCameraPermissions()` → imperative `Camera.getCameraPermissionsAsync()` / `requestCameraPermissionsAsync()` at press time (removes a native-backed hook from every render; reads fresh permission at press instead of a reactive snapshot)
- **`HeaderCallButton`**: drop the duplicate `useIsInActiveVoipCall()` (reuse `isInActiveCall` from `useNewMediaCall`)

**Benchmark** (iOS sim, Argent React Profiler; warm, interleaved A/B) — `HeaderCallButton` self-time in the critical room-open commit, with `MessageTouchable` as a warmth control (flat across runs):

| | develop (n=3) | this PR (n=2) |
|---|---|---|
| `HeaderCallButton` self | 6.41 / 5.88 / 6.14 → **mean 6.14ms** | 1.25 / 2.68 → **mean 1.97ms** |

→ **−68% (~4.2ms warm)** in the component's own self-time, on the critical path; larger absolute saving on cold opens. (Profiling warm-up varies the commit *total* ~2.4× run-to-run, so per-component self-time is the trustworthy metric.)

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1195

Follow-up to NATIVE-1184 (markdown perf, #7348) from the same RoomView cold-open profiling.

## How to test or reproduce

- Open a room and verify the call button still appears in the header and opens the call action sheet (VoIP "New call" / videoconf start-call sheet).
- Start a videoconf call and confirm the camera permission is still requested when not yet granted.
- Exercise the other consumers of these hooks: Room Info, Room Actions, Sidebar, VideoConferenceEnded.

## Screenshots

_No visual change — header/call button render identically._

## Types of changes

## Checklist

## Further comments

The hooks are shared, so the change is internal and preserves each hook's public API and return values — verified by the existing `CallSection` and `RoomInfoButtons` snapshot tests (no snapshot changes) and on-device (VoIP call sheet still opens).

An alternative was considered: deferring `HeaderCallButton`'s mount past the room-open transition with `InteractionManager`. That only *relocated* the work (no net CPU saving) and added a late-appearing button plus an extra re-render commit, with a warm benefit within measurement noise. Consolidation was chosen instead because it reduces the work itself, stays on the critical path, and benefits every consumer.

## Summary by CodeRabbit

* **Refactor**
  * Improved camera permission flow during video calls for more reliable, async permission prompts.
  * Consolidated video-call settings loading into a single subscription to reduce delays and ensure consistency.
  * Streamlined active-call state handling in call controls for more accurate enable/disable behavior.

### DIFF
--- a/app/lib/hooks/useVideoConf/useVideoConfCall.ts
+++ b/app/lib/hooks/useVideoConf/useVideoConfCall.ts
@@ -1,4 +1,5 @@
 import { useEffect, useState } from 'react';
+import { shallowEqual } from 'react-redux';
 
 import { SubscriptionType } from '../../../definitions';
 import { getUserSelector } from '../../../selectors/login';
@@ -8,7 +9,6 @@ import { compareServerVersion } from '../../methods/helpers/compareServerVersion
 import { isReadOnly } from '../../methods/helpers/isReadOnly';
 import { useAppSelector } from '../useAppSelector';
 import { usePermissions } from '../usePermissions';
-import { useSetting } from '../useSetting';
 
 export const useVideoConfCall = (
 	rid: string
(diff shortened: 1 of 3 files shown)


app/lib/hooks/useVideoConf/useVideoConfCall.ts
```diff
@@ -8,7 +9,6 @@ import { compareServerVersion } from '../../methods/helpers/compareServerVersion
 import { isReadOnly } from '../../methods/helpers/isReadOnly';
 import { useAppSelector } from '../useAppSelector';
 import { usePermissions } from '../usePermissions';
-import { useSetting } from '../useSetting';
```
OtavioStasiak, line 11: Since we're removing this hook from here, and it's only used in app/lib/hooks/useVerifyPassword.ts, wouldn't it make sense to call everything from useVerifyPassword within a single subscription as well (and remove the hook)?

OtavioStasiak (approved):
lgtm

app/lib/hooks/useVideoConf/useVideoConfCall.ts
diegolmello: Good call. Tracked as a follow-up: https://rocketchat.atlassian.net/browse/NATIVE-1196 — consolidate the 9 password-policy `useSetting` reads in `useVerifyPassword` into a single `useAppSelector(..., shallowEqual)` and delete the now-unused `useSetting` hook. Keeping it out of this PR to stay focused on the call-button path. Resolving this thread.

PR #7362, opened 2026-05-29
fix: re-subscribe rooms stream after forced socket reopen

diegolmello:
## Proposed changes

After the app sits in the background for a while the DDP socket goes stale, so foregrounding triggers `checkAndReopen` → `forceReopen`. `forceReopen` drops every SDK subscription and reopens the socket **without** going through `connect()`, so the module-level `roomsSubscription` guard in `subscribeRooms` stays set. The next `subscribeRooms()` therefore short-circuits and never re-subscribes `stream-notify-user`, and the rooms list silently stops reflecting subscription/favorite/read changes until a manual reconnect or app restart.

This resets that guard from the socket `'close'` listener (via `unsubscribeRooms()`) — the same teardown `connect()` already performs at startup — so the resume-login that follows the reopen re-subscribes `stream-notify-user`. `stream-notify-user` is the only stream behind that guard; the other streams (permissions, presence, settings, roles) call `sdk.subscribe(...)` unconditionally and were never affected.

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-1047

## How to test or reproduce

1. Log in and open the rooms list.
2. Background the app and cut connectivity (airplane mode) for ~60s so the DDP socket goes stale.
3. Restore connectivity and foreground the app.
4. From another client (or REST), trigger a `subscriptions-changed` event — e.g. favorite/unfavorite a room or mark one read.

- **Before:** the rooms list does not update; the change only appears after a manual reconnect / app restart.
- **After:** the rooms list updates in real time again.

Verified on an Android 16 emulator: after foregrounding, `subscribeRooms()` re-subscribes `stream-notify-user` and a favorite toggle then fires the handler and the WatermelonDB write. A regression test in `connect.test.ts` asserts the `'close'` listener calls `unsubscribeRooms()` (and fails if the call is removed).

## Types of changes

## Checklist

## Further comments

The fix lives in the `'close'` listener rather than in `subscribeRooms` itself because `'close'` is already the place app state is reset for reconnect (it dispatches `disconnectAction()` to flip Redux `meteor.connected` → false). The SDK's `forceReopen` documents this contract: it emits `'close'` so `connect.ts` flips that state, otherwise the next `'connected'` short-circuits and the loginRequest → subscribeNotifyUser chain never re-runs. The `roomsSubscription` guard is just a second piece of "am I subscribed?" state that has to be reset in that same spot.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection recovery so subscriptions are properly reset when VoIP/SDK connections close and reopen, ensuring notifications and real-time updates are re-established after reconnects.

* **Tests**
  * Added a regression test that verifies cleanup and unsubscribe behavior during connection close events to prevent missed or duplicate subscriptions.

### DIFF
--- a/app/lib/services/connect.test.ts
+++ b/app/lib/services/connect.test.ts
@@ -1,6 +1,7 @@
 import { connect, determineAuthType, disconnect } from './connect';
 import { mediaSessionInstance } from './voip/MediaSessionInstance';
 import { pendingHangups } from './voip/pendingHangups';
+import { unsubscribeRooms } from '../methods/subscribeRooms';
 
 jest.mock('./voip/MediaSessionInstance', () => ({
 	mediaSessionInstance: { reset: jest.fn(), drainPendingHangups: jest.fn() }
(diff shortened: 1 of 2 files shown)


OtavioStasiak (approved):
LGTM

PR #6960, opened 2026-02-02
fix: forwarding a message twice does not render the attachment

OtavioStasiak:
## Proposed changes

When forwarding messages with file attachments files from forwarded messages remain invisible to users. Files located in nested attachment structures are not processed or displayed by the Reply component.

## Issue(s)	

https://rocketchat.atlassian.net/browse/SUP-986

## How to test or reproduce

- Open the app;
- Go to room with a attachment message;
- Forward it twice;

## Screenshots
| Before | After |
| ----- | ----- |
| Simulator Screenshot - iPhone 16 - 2026-01-29 at 21 10 26 | Simulator Screenshot - iPhone 16 - 2026-02-02 at 16 51 06 |

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [X] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [X] I have added necessary documentation (if applicable)
- [X] Any dependent changes have been merged and published in downstream modules

## Summary by CodeRabbit

* **New Features**
  * Nested attachments now render as reply-style attachments within messages.

* **Bug Fixes**
  * Quote detection improved to ignore collapsed items and attachments that contain nested attachments.

* **Tests**
  * Added Storybook stories showcasing nested reply/file attachments, a collapsible attachment with text, and a large-font variant.

PR #6725, opened 2025-10-14
fix: File rendering as URL

diegolmello:
## Proposed changes

- Fixes the file rendering as URL
- Removed unused props from Markdown
- Added file stories

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1049

## Screenshots
| Before  | After |
|---|---|
| image  | image |

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [ ] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Summary by CodeRabbit

- **New Features**
  - Improved file-attachment preview in replies for clearer filename display.

- **Style**
  - Unified Markdown appearance by removing custom color overrides and inline styles.
  - Consistent default styling across context blocks, attachments, canned responses, E2E guide, and room info.

- **Refactor**
  - Simplified Markdown rendering and reduced unnecessary props; Markdown now renders only when message text exists.

- **Documentation**
  - Added Storybook examples for file attachments and reply scenarios, including large-font variants.

PR #3883, opened 2022-03-11
[NEW] Redesign quoted messages

gerzonc:
## Proposed changes

Changes quoted messages design to match with how it looks on the Web client and fixes attachments on quoted messages. Applies to: images, videos, audios and plain text.
## Issue(s)	

Closes #3396
Closes #2564 

#### Simple quote
1. Send a message
2. Reply to that message

Should render the message and the quoted message properly.

#### Nested quote
1. Send a message
2. Reply to that message
3. Reply to the previous message

Should render all the messages, included the most nested one.

#### Image message
1. Send a message with an image
2. Reply to that message

Should render the message and the quoted message with image.

#### Video message
1. Send a message with a video
2. Reply to that message

Should render the message and the quoted message with video.

## Screenshots
<table>
  <thead>
    <tr>
      <th>Before</th>
      <th>After</th>
      </thead>
    </tr>
    <tbody>
      <tr>
        <td>
        image
        </td>
        <td>
        image
        </td>
      </tr>
    </tbody>
</table >

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [ ] Lint and unit tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

PR #7379, opened 2026-06-02
fix: file attachments not displayed on forwarded messages

### LINKED ISSUE #7363: bug: File attachments not displayed in forwarded messages on mobile
### Describe the Bug

When a message with a file attachment is forwarded to another room or user, the file is not displayed on iOS and Android mobile apps — only the text part of the message is visible. The same forwarded message displays correctly in the web and desktop clients.

This affects all file types (PDF, images, documents, etc.).

### Steps to Reproduce

1. Upload any file to a room (with or without accompanying text)
2. Forward that message to another room or user
3. Open the forwarded message on iOS or Android mobile app
4. The file attachment is missing — only text is visible

### Expected Behavior

The forwarded message should display the file attachment the same way it does in web/desktop clients.

### Actual Behavior

_No response_

8.4.2

### Rocket.Chat App Version

4.72.0.108804

### Device Name

iPhone 12

### OS Version

iOS 26.5

### Additional Context

- Original (non-forwarded) messages with file attachments display correctly on both iOS and Android
- The issue is isolated to forwarded messages only
- Tested immediately after forwarding (rules out URL expiry)
- Reproduced on both iOS and Android
- Files are stored on external S3-compatible storage (DigitalOcean Spaces) with proxy enabled
- Verified server-side: proxy configuration, S3 storage, file access authorization — all working correctly

Image
Image

OtavioStasiak:
## Proposed changes

 When a message containing a file attachment (PDF, document, etc.) is forwarded to another room or
  user, the file was not displayed on iOS and Android — only the text part of the message was
  visible. The same forwarded message rendered correctly on web and desktop.
  
  Root cause

  A forwarded message wraps the original message as a nested attachment:
```
  attachments: [{
    author_name: 'rocket.cat', message_link: '...', text: '',
    attachments: [{ type: 'file', title: 'document.pdf', title_link: '...' }] // the real file
  }]
```
  At the top level, Message.tsx renders both <Quote> and <Attachments> over the same array. These use
  complementary filters that partition attachments between them: <Quote> handles quote-style
  attachments (plain files, documents, text replies) and <Attachments> handles media
  (image/audio/video/actions/nested). Together they cover everything.

  But inside Reply.tsx (the component that renders the forwarded wrapper), the nested attachments
  were passed only to <Attachments>. Its removeQuote filter keeps media only, so a nested plain
  file/document — which has no image_url/audio_url/video_url — was silently dropped. Images survived
  solely because they carry image_url, which is why the bug looked file-type dependent.

## Issue(s)	

Closes https://github.com/RocketChat/Rocket.Chat.ReactNative/issues/7363
https://rocketchat.atlassian.net/browse/NATIVE-1231

## How to test or reproduce

- Open the app;
- Send a attachment in a room;
- Forward the attachment to another room;

## Screenshots
| Before | After |
| ----- | ----- |
| Simulator Screenshot - iPhone 16 - 2026-06-02 at 20 02 51 | Simulator Screenshot - iPhone 16 - 2026-06-02 at 19 59 13 |

## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Replies now display quoted content inline for clearer context.
  * Message examples include additional forwarded/nested attachment scenarios (audio, video, files).

* **Bug Fixes / Improvements**
  * Attachment classification improved to better separate quoted text from media/actions, reducing misdisplayed or hidden attachments.

* **Tests**
  * Added tests validating attachment classification behavior.


app/containers/message/Components/Attachments/Reply.tsx
diegolmello, line 266: This looks like déjà vu.
I think we had something like that and removed?
Perhaps it's (or was) inside attachments?

OtavioStasiak: <Quote> inside Reply.tsx never existed before this branch. git log -S 'Quote' -- …/Reply.tsx
  returns exactly one commit: your 745dd4b7f on this branch.

  What you're remembering is one (or both) of these:

  1. <Attachments attachments={attachment.attachments} …/> inside Reply — that's the near-identical
  sibling block right below your new Quote, and it's ancient: added in b0580bf54 "[NEW] Redesign
  quoted messages (#3883)" (2022), reshuffled in 2668165f3 (#5872, thumb layout) and aa94a5999
  (#6725). Same props, same shape — that's the visual déjà vu.
  2. The recursion added in d2aa920e3 — "fix: forwarding a message twice does not render the 
  attachment (#6960)" (Feb 2026). There you implemented the same idea but on the Attachments side:
  added the if (file.attachments?.length) return <Reply …/> branch to Attachments.tsx, plus the guard
  in Quote.tsx ("attachments with nested attachments are rendered only by Attachments as Reply"). So
  yes — "it's inside attachments" is exactly right, it still is.

  Why this branch still needed Quote inside Reply: #6960 made Reply recurse into nested attachments
  only via Attachments — which renders content-type children (media, actions, nested replies). A
  nested file attachment (type: 'file', no media URL, no nested attachments) is quote-type:
  Attachments filters it out, and there was no Quote inside Reply to catch it → rendered nothing.
  Your change completes the mirror so Reply's children get the same Quote + Attachments pair the top
  level has had all along (Message.tsx:91-93).

app/containers/message/Components/Attachments/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+// Webhook/integration attachments can carry their own media alongside text/color (e.g. embeds).
+const hasOwnMedia = (file: IAttachment): boolean => !!(file.image_url || file.audio_url || file.video_url);
+
+// Quote renders attachments as a quote-style block (colored border + author/title/text):
+// message quotes, forwarded messages without nested attachments, file attachments and webhook embeds.
+export const isQuoteAttachment = (file?: IAttachment): boolean => {
+	if (!file) return false;
+
+	if (file.collapsed) return false;
```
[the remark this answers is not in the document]
OtavioStasiak: collapsed: false attachments aren't dropped. The premise that only Attachments runs at the top level isn't accurate
  — Message.tsx renders <Quote> and <Attachments> as siblings (L24-26 and L72-74), so every attachment is offered to
  both filters.

  For a bare collapsed: false item:
  - isContentAttachment → false (correctly excluded — nothing concrete to render)
  - isQuoteAttachment → true — the if (file.collapsed) truthiness guard doesn't catch false, so it falls through and
  Quote renders it as a Reply.
  
  So it's handled by the Quote pass, which is where a borderless quote-style block belongs. This also matches the old
  removeQuote, which used the same collapsed truthiness check — no behavior change here.

  Aligning isContentAttachment to also accept collapsed: false would actually cause a double render, since Quote
  already handles it. So no change needed.

diegolmello (approved):

PR #7382, opened 2026-06-03
chore: integrate RNRepo prebuilt artifacts to reduce native build times

diegolmello:
## Proposed changes

Integrate [RNRepo](https://rnrepo.org) (Software Mansion) to reduce native build times. RNRepo substitutes supported React Native libraries with prebuilt artifacts from `packages.rnrepo.org` during the native build — a Gradle plugin on Android and a CocoaPods plugin on iOS — and falls back to building from source whenever a prebuild isn't available. It's beta and New Architecture only; we're on RN 0.81.5 with New Arch enabled on both platforms, so it's supported.

We use the **Standard React Native** integration path (we have committed `android/` + `ios/`, not Expo CNG):

- Add the `@rnrepo/build-tools` dependency.
- **Android**: RNRepo classpath in `android/build.gradle`; `apply plugin: "org.rnrepo.tools.prebuilds-plugin"` in `android/app/build.gradle`.
- **iOS**: `require` for the CocoaPods plugin + `rnrepo_post_install(installer)` in `ios/Podfile`. `Podfile.lock` only changes its `PODFILE CHECKSUM`; the app Xcode project is untouched (RNRepo's build phases land in the generated Pods project).
- Add `rnrepo.config.json` deny list.

**Deny list rationale** — prebuilt artifacts bypass library source compilation, so any library carrying a **native** (Objective-C/Java/Kotlin/podspec) `patch-package` patch must keep building from source, or the patch is silently lost. JS-only patches are unaffected. Entries key on the npm package name and are scoped to the platform whose native source the patch touches:

- **android**: `@lodev09/react-native-true-sheet`, `expo-asset`, `expo-file-system`, `expo-font`, `react-native-callkeep`, `react-native-webview`
- **ios**: `react-native-callkeep`, `react-native-config-reader`, `react-native-gesture-handler`, `react-native-nitro-modules`, `react-native-webview`

Two iOS entries are not patch-driven:
- `react-native-webview` — its patch is Android-only, but the Podfile has a webview-specific `HEADER_SEARCH_PATHS` step and the pod graph uses static frameworks, so it stays source until proven safe.
- `react-native-nitro-modules` — a prebuilt `NitroModules` is an incompatible Swift/C++ module pairing for its source-built consumer `NitroMmkv` (`react-native-mmkv`, which has no prebuild yet). Forcing the provider from source keeps it consistent with its consumers. Android prebuilds `NitroModules` without issue, so it stays enabled there.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1248

## How to test or reproduce

No app/runtime behavior changes — this only affects how native dependencies are built.

- **iOS**: `pnpm pod-install`. Expect `[📦 RNRepo] … using pre-built frameworks` and `[📦 RNRepo] ⊘ Denied from pre-builds` listing the iOS deny entries.
- **Android**: `cd android && ./gradlew :app:assembleDebug --info`. Expect `[📦 RNRepo] RNRepo plugin is ENABLED` and `Package <name> is in deny list, skipping` for each denied lib.
- **Opt out of RNRepo for a build** (debugging): prefix with `DISABLE_RNREPO=1`.

### CI build results

- **Android — ✅ built successfully** with RNRepo (15+ libraries substituted with prebuilt AARs), in **~32.5 min vs a ~38 min (mean of 8 recent non-RNRepo runs) baseline — faster than every baseline run sampled (~15–18% / ~6 min).**
- **iOS** — the first run **failed** because a prebuilt `NitroModules` couldn't be linked against the source-built `NitroMmkv`. The deny-list entry above addresses it; an iOS build re-run is needed to confirm the fix and measure the iOS delta (iOS currently substitutes only ~6 libraries, so its win is expected to be smaller than Android's until the registry grows).

## Screenshots

N/A — no UI changes.

## Types of changes

## Checklist

## Further comments

No JavaScript/TypeScript changed — only build configuration (`*.gradle`, `Podfile`, `package.json`, `pnpm-lock.yaml`) and the new `rnrepo.config.json` — so the JS unit tests and ESLint/tsc surface are unaffected.

When regenerating `ios/Podfile.lock`, use bundler (`pnpm pod-install`) so it stays on the pinned CocoaPods **1.15.2**. Depending on the environment, `RCT-Folly`'s spec checksum may harmlessly re-flip — that's pre-existing RN/Folly nondeterminism, not introduced here, and was kept out of this diff.

Prebuilt coverage is modest today (more on Android than iOS) because RNRepo is in beta and its registry is still growing; unsupported libraries fall back to source silently, so coverage — and the speedup — improves over time with no further changes on our side.

## Summary by CodeRabbit

* **Chores**
  * Updated Android build to integrate support for prebuilt native components via a build plugin.
  * Enhanced iOS Podfile to enable seamless installation/linking of prebuilt frameworks during install.
  * Added a new build-tools dependency to support the improved prebuild workflow.
  * Introduced a platform-specific deny list to restrict certain native packages on Android and iOS.


diegolmello:
## Closing — RNRepo benchmark doesn't justify the change for our dependency set

Ran the gated native builds on the fix commit. Both **release** builds pass, but measured against real recent CI runs that actually executed native builds, the gain isn't there.

### Build times — RNRepo vs baseline

| Build | RNRepo | Baseline (recent CI) | Result |
|---|---|---|---|
| Android release | ~33 min (2016s, 1947s · n=2) | mean 37.7 min, range 33–41 (n=3) | ~12% faster (~4.7 min), but overlaps the fastest baseline (~33 min) |
| iOS release | 34.7 min (2082s · n=1) | mean 28.2 min, median 32.4, range 23–32 (n=4) | **net regression — ~6 min slower, slowest of every iOS build measured** |

**Why iOS regresses:** only ~6 iOS libraries have prebuilt xcframeworks for RN 0.81.5, NitroModules has to build from source (below), and on a cold CI runner the xcframework download + CocoaPods plugin overhead exceeds the thin substitution benefit.

### iOS NitroModules failure (fixed, but illustrative)

The prebuilt `NitroModules.xcframework` ships public C++ headers (`JSIConverter.hpp`, `ArrayBuffer.hpp`) whose quoted includes of sibling headers (`JSIConverter+AnyMap.hpp`, `BorrowingReference.hpp`) and transitive `folly/dynamic.h` (via React-jsi) are not resolvable when its **source** consumer `react-native-mmkv` (NitroMmkv, no prebuild — download 404s) builds its module → `could not build Objective-C module 'NitroMmkv'` (exit 65). Worked around by forcing NitroModules from source on iOS — which also removes one of the few iOS substitutions.

### E2E Android break

The **release** Android build passes, but the **E2E** Android build (RelWithDebInfo / x86_64) fails: `react-native-reanimated` builds from source for that variant and can't find prebuilt `react-native-worklets`' `libworklets.so`:

```
ninja: error: '.../react-native-worklets/.../x86_64/libworklets.so',
  needed by '.../libreanimated.so', missing and no known rule to make it
```

This is RNRepo's known reanimated + worklets C++ provider/consumer split. Fixing it requires deny-listing `react-native-reanimated` + `react-native-worklets` (the heaviest C++ builds) — which would erode most of the ~12% Android win.

### Conclusion

iOS is net-negative; Android is marginally positive on the release build but the necessary E2E fix removes most of that gain. Not worth the added build-system complexity and beta-stage risk for this dependency set right now. Worth revisiting once the RNRepo registry has broader iOS xcframework coverage for RN 0.81+ and multi-variant reanimated/worklets prebuilds.

PR #7387, opened 2026-06-08
fix: prevent composer keyboard from dismissing on open on small Android screens

OtavioStasiak:
## Proposed changes

On some small-screen Android devices (e.g. Zebra TC21, also seen on Android 11), the keyboard would open and
immediately close again the moment you tapped the message composer, making it impossible to type.

The cause was the useCloseKeyboardWhenOrientationChanges hook. It decided "portrait vs landscape" by comparing
the width and height of the safe-area frame, and dismissed the keyboard whenever that orientation flipped.
The problem: with adjustResize + edge-to-edge, that frame shrinks when the keyboard opens. On small screens
the reduced height drops close to the width, so the hook mistook the keyboard opening for a screen rotation
and called Keyboard.dismiss() right after the keyboard appeared.

This PR changes the hook to:

- Read the orientation from Dimensions.get('screen'), which reflects the physical display size and is not
affected by the soft keyboard — so opening the keyboard no longer looks like a rotation.
- React to real dimension changes via a Dimensions change subscription (in a useEffect) instead of running a
side effect during render.

The keyboard now stays open when typing, while still dismissing correctly on a genuine device rotation.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/SUP-1055

## Types of changes

## Checklist

## Bug Fixes

* Improved keyboard dismissal in the message composer so it occurs only when the device changes between portrait and landscape orientation.
* Prevented unnecessary keyboard dismissal caused by temporary window or safe-area changes, such as those triggered while the keyboard is open.
* Ensured orientation-change handling is properly cleaned up when no longer needed.


OtavioStasiak:
@coderabbit review

PR #7388, opened 2026-06-08
feat: bounded message window for centered Jump to Message

diegolmello:
## Proposed changes

Rebuilds **Jump to Message** on a **bounded message window** so a deep jump re-anchors in one step (O(1)) instead of growing the window page-by-page (O(pages)) — removing the root cause of the old 5s-race flakiness — without migrating off the custom native inverted list.

The whole capability is **one optional upper `ts` bound (`highTs`)** on the existing WatermelonDB observation:
- `highTs == null` → **Live Window** (newest-first, follows the Live Tail) — unchanged default behavior.
- finite `highTs` → **Anchored Window** pinned around the target, below the Live Tail.

A pure, DB/React-free anchor resolver decides the bound (`anchorForTarget`) and climbs it back toward live as Newer Loaders are consumed (`raiseOrRelease`), releasing to a Live Window only once the Gap to the Live Tail has fully closed (invariant: **never release across an open Gap**). "Rejoin live" is now explicit — the Load Newer chain or the jump-to-bottom FAB (which stays visible while anchored).

Also included:
- **Reading-position preservation on release** — on RELEASE, grow the Live Window by the count of cached messages above the old bound so a deep target isn't evicted by `take(count)` and the list doesn't snap to the Live Tail.
- **Scroll-to-index recursion guard** — defer one frame + cap retries for `onScrollToIndexFailed` (no `getItemLayout` → the inverted list re-fires it synchronously → stack overflow). The retry re-applies the same centering + header-clearing view offset as the initial scroll, so a distant jump that falls back to the retry path still lands centered instead of hidden behind the room header.
- **Precise landing on a fresh window** — with no `getItemLayout` for these variable-height messages, the first `scrollToIndex` uses an estimated offset and can undershoot while the target's row is still unmeasured (e.g. entering a room fresh from the rooms list), leaving the target hidden above the viewport. The jump now scrolls a second time against the measured frame once the row has rendered, so the target always lands centered.
- **Bounded window growth for deep anchored targets** — when a target sits below the anchored window's first page (a large Chunk or filtered rows between the bound and the target), the jump grows the window one page at a time (capped) to pull the target into the rendered rows, instead of waiting out the safety timeout and aborting.
- **Jump orchestration** — gate anchoring on window membership (`isMessageInWindow`), not message origin, so a locally-cached-but-out-of-window target no longer silently aborts; three roads (in-window / cached-out-of-window / server). A server-fetched target whose chunk reaches the Live Tail (e.g. a push-notification deep link onto a recent message) keeps the room in a Live Window instead of pinning it in an unreleasable historical view; the target-ts fallback covers the remaining cases (cached target with no bracketing Loader, target absent from the fetched chunk).
- Supporting fixes: keep the live socket on deeplink login; tolerate an empty 2FA code on the DDP `loadSurroundingMessages` call; one-shot jump param so re-searching the same message re-fires; guard the anchor raise/release against a re-subscribe landing mid-flight (room switch / concurrent raise) so a stale read can't corrupt the new window.

Architecture and flow diagrams are documented alongside the code:
- `app/views/RoomView/docs/ARCHITECTURE.md` — the window model, layer responsibilities, anchor resolution, the rejoin (raise/release) climb, scroll landing, the **rejected FlashList v2 migration** and why, and the load-bearing invariants tied to the unit tests.
- `app/views/RoomView/docs/FLOWS.md` — sequence diagrams for each jump and window-management handshake (in-window jump, server-fetched anchor, deep-target growth + safety net, frontier climb, raise/release rejoin, jump-to-bottom, thread-jump timing).

New domain vocabulary is in `UBIQUITOUS_LANGUAGE.md` (§ Message Loading).

## Issue(s)
https://rocketchat.atlassian.net/browse/NATIVE-1224

## How to test or reproduce
1. Open a room with a long history (hundreds of messages, with gaps).
2. Search for / tap a target far from the Live Tail. It should re-anchor and center on the target in one step — no long page-by-page scroll, no abort — and land clear of the room header (not tucked behind it), including when entering the room fresh from the rooms list (rows not yet measured) and for very deep targets that fall back to the scroll retry path.
3. Climb back via "Load newer" / scrolling up; the window rejoins live only when the gap closes; the jump-to-bottom FAB is available throughout.
4. Jump to a deep target, then keep loading newer until release — the reading position is preserved (no snap to the Live Tail).
5. Jump to the same message twice in a row via Search — it re-fires both times.
6. Jump to a quoted reply pointing at an on-screen message — it scrolls in place, leaving the Live Tail intact.
7. Tap a push notification for a recent message in a room whose history isn't cached — the room opens at the Live Tail with the message visible and keeps rendering newly arriving messages (it must not open in a stuck historical view).

## Screenshots

https://github.com/user-attachments/assets/ad8cfad0-57c5-41ca-9b40-1fd599607b15

https://github.com/user-attachments/assets/cd685def-2fa5-4930-a8d1-03b54d92ce6c

https://github.com/user-attachments/assets/c4904db6-9857-4f2f-881e-f84956e4ff0f

https://github.com/user-attachments/assets/534faa6a-89e5-4b05-a39e-5310efa27732

## Types of changes

## Checklist

## Further comments
The list engine is unchanged; a future FlashList migration remains possible but is neither required nor blocked by this work. Ordering stays `ts`-only (a `ts + _id` composite is a deferred follow-up noted in `ARCHITECTURE.md`). New unit coverage: `anchorResolver` (including the server-chunk resolver), `useMessages` (release-window growth), `useScroll` (deferred + capped retry, header-clearing offset on retry, re-scroll after measurement, bounded growth + growth cap for deep anchored targets), `getLocalAnchor`.


app/lib/methods/helpers/tsToMs.ts
```diff
@@ -0,0 +1,6 @@
+import dayjs from '../../dayjs';
+
+// Normalize a message `ts` (Date, ms number, or ISO string — WatermelonDB hands back any of the three
+// depending on the read path) to ms since epoch. dayjs parses all three; plain `Number(ts)` / `new
+// Date(ts)` each NaN on one of them.
+export const tsToMs = (ts: Date | number | string): number => dayjs(ts).valueOf();
```
diegolmello: should be closer to dayjs

app/lib/methods/helpers/index.ts
```diff
@@ -18,4 +18,5 @@ export * from './image';
 export * from './emitter';
 export * from './parseJson';
 export * from './fileDownload';
+export * from './tsToMs';
```
diegolmello: Not needed

app/lib/services/sdk.ts
```diff
@@ -111,7 +111,10 @@ class Sdk {
 	methodCall(...args: any[]): Promise<any> {
 		return new Promise(async (resolve, reject) => {
 			try {
-				// Clear the 2FA code after use — a stale trailing arg breaks typed method signatures
+				// Single-use the TOTP code: read then clear it, and only append when set. An
+				// unconditional trailing arg pushes junk into every call and breaks methods whose
+				// signature grows a typed trailing param (e.g. loadSurroundingMessages'
+				// `showThreadMessages: boolean`); a code left set re-appends on every later call.
```
diegolmello: Rollback comment

app/views/RoomView/docs/adr/0001-bounded-window-over-flashlist.md
diegolmello, line 1: Add to task and remove from codebase

app/views/RoomView/List/components/List.tsx
```diff
@@ -57,7 +62,12 @@ const List = ({ listRef, jumpToBottom, ...props }: IListProps) => {
 						: undefined
 				}
 				removeClippedSubviews={isIOS}
-				initialNumToRender={7}
+				// Seeds the render window. A Jump to Message re-anchors the window and scrolls to the target by
+				// index, but this inverted list has no getItemLayout, so scrollToIndex cannot reach a frame that
+				// was never rendered — and maintainVisibleContentPosition freezes window growth during the jump.
+				// 20 keeps the target within the seeded window; a smaller seed leaves it unmeasured and the jump
+				// parks short of it (empty list above the loaded rows).
```
diegolmello: Remove

app/views/RoomView/List/components/List.tsx
```diff
@@ -23,19 +23,24 @@ const styles = StyleSheet.create({
...
 			} else {
-				scheduleOnRN(setVisible, false);
+				scheduleOnRN(setScrolledPastLimit, false);
 			}
 		}
 	});
 
+	// In a Live Window the FAB tracks scroll distance from the Live Tail. In an Anchored (historical)
+	// Window the loaded rows' bottom edge is NOT the Live Tail, so offset alone would hide the FAB right
+	// at the "Load newer" boundary — leaving no one-tap way back to live. Force it visible while anchored.
```
diegolmello: Shrink

diegolmello (commented):
Missing a doc with architecture and flow

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -29,13 +31,24 @@ export const useMessages = ({
 	hideSystemMessages: string[];
 	t: RoomType;
 }) => {
+	// NOT migrated to the React Compiler ('use memo'): babel-plugin-react-compiler (rc) silently skips
+	// any function whose body contains a react-hooks/exhaustive-deps suppression, and this hook keeps
+	// intentional incomplete effect-dep arrays (see the disables below). Annotating it would no-op, so
+	// the manual useCallback/useMemo here are load-bearing and must stay.
```
diegolmello: Remove

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -79,6 +159,10 @@ export const useMessages = ({
 				.query(
 					Q.where('rid', tmid),
 					...(visibleSystemClause ? [visibleSystemClause] : []),
+					// Anchored Window upper bound. NOTE: ordering stays ts-only here, which has a tie /
+					// clock-skew weakness (equal-ts rows can straddle the bound); the deferred fix is a
+					// composite ts + _id ordering (see the ADR consequences).
```
diegolmello: Duplicated and mentions ADR

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -88,6 +172,10 @@ export const useMessages = ({
 			const whereClause: Q.Clause[] = [
 				Q.where('rid', rid),
 				...(visibleSystemClause ? [visibleSystemClause] : []),
+				// Anchored Window upper bound. NOTE: ordering stays ts-only here, which has a tie /
+				// clock-skew weakness (equal-ts rows can straddle the bound); the deferred fix is a
+				// composite ts + _id ordering (see the ADR consequences).
```
diegolmello: Duplicated and mentions ADR

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -108,11 +196,29 @@ export const useMessages = ({
 				newMessages.push(thread.current);
 			}
 
+			// Thread / local windows are never anchored, so rejoin only applies to the bounded main room.
+			if (!tmid && highTs != null) {
+				// Best-effort: the targeted read may reject; never let that break the emit.
```
diegolmello: Remove

app/views/RoomView/List/hooks/useScroll.ts
diegolmello, line 1: Reassess all code comments here

app/views/RoomView/List/definitions.ts
diegolmello, line 1: Review comments

app/views/RoomView/index.tsx
```diff
@@ -658,6 +662,15 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
 
 			if (this.tmid) {
 				await loadThreadMessages({ tmid: this.tmid, rid: this.rid });
+				// Thread jump: the target lives in the thread window we just populated, so fire here rather
+				// than in componentDidMount — the row exists now (a non-anchored thread jump otherwise aborts
+				// on its safety net and parks on the live tail). Read-and-clear so the other init() callers
+				// (the connected handler, accept-invite, the 300 ms retry) can't re-fire it.
```
diegolmello: Shorter

diegolmello (commented):
Review of the bounded Message Window + anchored Jump-to-Message work — 15 inline notes below, grouped by kind in the order they appear.

- **Correctness** — the load-bearing one is the rejoin read in `useMessages` that can truncate a still-present boundary loader and release the Anchored Window across an open Gap. The rest are lower severity: the auto-dispatch effect consuming the boundary loader while anchored, the jump completing before the target is confirmed on-screen, a `count` bump before the `rid` guard, and two latent guards.
- **Tests** — `cancelJumpToMessage` / `highlightedMessageId` have no coverage; the `getLocalAnchor` mock can't catch query-shape regressions.
- **Docs/domain** — a `setHighTs` step that doesn't exist and a boolean param label in the jump sequence diagram, a "no DB" self-contradiction in the layers table, and a rotting internal SHA on every diagram.
- **Slop/comments** — a `| any` that defeats the jump's types, two over-long JSDocs, "clear the spinner" comments in a hook that owns no spinner, and a duplicated re-scroll block.

Two items aren't inline-able (their lines aren't in the diff): the loader-finder expression `visibleMessages.find(m => m.t && MESSAGE_TYPE_ANY_LOAD.includes(...))?.id` is duplicated verbatim in `useMessages` (the rid-change snapshot effect and the auto-dispatch effect) — worth one shared helper; and the glossary drifts in two spots ("page" in `useScroll`, "(historical)" for the Anchored Window in `List/definitions.ts`) versus the terms in `UBIQUITOUS_LANGUAGE.md`.

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -53,6 +62,73 @@ export const useMessages = ({
...
+			// Pin to the subscription that fired this emit. fetchMessages reassigns subscription.current on
+			// every re-subscribe (room switch, another raise); if that happens during the awaits below, this
+			// invocation is stale and must not mutate count / highTs for the new window.
+			const sub = subscription.current;
+
+			// Targeted one-shot read of the region above the current bound that the bounded observation
+			// cannot see (the newly-written batch + any new Newer Loader).
+			const rows = (await database.active
+				.get('messages')
+				.query(Q.where('rid', rid), Q.where('ts', Q.gt(currentHighTs)), Q.sortBy('ts', Q.asc), Q.take(QUERY_SIZE + 1))
```
diegolmello: **Bug · high — spurious release across an open Gap.** This rejoin read is `Q.where('ts', Q.gt(highTs)), Q.sortBy('ts', asc), Q.take(QUERY_SIZE + 1)` = 51 rows, with no `hideSystemMessages` filter (unlike the observation, which applies `buildVisibleSystemTypesClause`). `loadNextMessages` writes exactly `QUERY_SIZE` (50) messages plus one `NEXT_CHUNK` loader just above them, so 50 messages + loader already fill all 51 slots. Any pre-existing or system row in the `(highTs, newLoaderTs]` range pushes the boundary loader past slot 51; `raiseOrRelease` then receives loader-free rows, returns `null`, and releases the Anchored Window while the Gap is still open — the "no release across a Gap" invariant from `ARCHITECTURE.md` is violated, and `count` then grows by every row above the old bound.

The read shouldn't cap a slice of all rows — read the Newer Loaders above the bound directly and hand those to `raiseOrRelease`: `Q.where('rid', rid), Q.where('t', NEXT_CHUNK), Q.where('ts', Q.gt(highTs)), Q.sortBy('ts', Q.desc), Q.take(1)`. Bounded, the loader can't be crowded out, and it mirrors `getLocalAnchorTs`. Also lets the separate `fetchCount` go.

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -173,5 +270,5 @@ export const useMessages = ({
 		}
 	}, [serverVersion, rid, t, hideSystemMessages, visibleMessages, dispatch, store]);
```
diegolmello: **Bug · medium — auto-dispatch consumes the boundary loader while anchored.** This effect (lines 251–271) has no `highTs` guard. In an Anchored Window the boundary Newer Loader sits at `ts === highTs` (it's inside the `Q.lte(highTs)` clause, so it's the first loader in `visibleMessages`). `NEXT_CHUNK` is in `MESSAGE_TYPE_ANY_LOAD`, so on a server ≥3.16.0 with `hideSystemMessages` set, the effect dispatches `roomHistoryRequest` for that boundary loader → `loadNextMessages` consumes it → the window auto-climbs (and can auto-release) with no user scroll, burning `autoLoadCount` toward `MAX_AUTO_LOADS`. That contradicts the "Anchored Window deliberately does not follow" contract — a user who jumped to an old message would see it drift toward live on its own.

Guard the effect on the bound (`if (highTs != null) return;`, with `highTs` added to the deps). The rejoin climb is the user-driven path via Load Newer, not this auto-loader.

app/views/RoomView/List/hooks/useScroll.ts
```diff
@@ -1,124 +1,309 @@
...
+				armJumpSafety(jump); // productive growth → grant the next page its own arrival window
+				// A grow failure leaves the jump pending; the safety net aborts it, so swallow here.
+				fetchMessages().catch(() => {});
 			}
-		});
+			return;
+		}
+		jump.scrolled = true;
+		scrollToTarget(jump.messageId, index);
+		completeJump(jump);
```
diegolmello, line 213: **Bug · low — jump completes without confirming the target is on-screen.** `completeJump` (highlight + resolve) runs immediately after `scrollToTarget` kicks off the two-pass scroll, regardless of whether the row actually landed in the viewport. The two-pass scroll and the frontier climb cover an *unmeasured* frame, but a post-measurement undershoot on a very tall target row resolves the jump with the message sitting just above the fold. Narrow trigger — flagging because the jump now reports success without the viewability check the prior flow waited on.

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -53,6 +62,73 @@ export const useMessages = ({
...
+			// the newly-revealed batch and the user's reading position is preserved. The re-subscription's
+			// first emit re-derives boundaryLoaderPresent from the new bound.
+			setHighTsState(next);
+		},
+		[rid]
+	);
+
 	const fetchMessages = useCallback(async () => {
 		unsubscribe();
 		count.current += QUERY_SIZE;
```
diegolmello: **Bug · low — `count` bumped before the `rid` guard.** `count.current += QUERY_SIZE` runs before `if (!rid) return`, so a render with a falsy `rid` still inflates `count`; the next real fetch then issues `take(100)` instead of `take(QUERY_SIZE)`. Move the increment below the guard. (If `rid` is in fact always truthy here, the guard is dead code — either way the increment belongs after it.)

app/views/RoomView/services/resolveJumpAnchor.ts
```diff
@@ -0,0 +1,62 @@
...
+		return bound;
+	}
+
+	// The local path can't detect the equal-ts collision the server path warns about above:
+	// getLocalAnchorTs returns only a scalar ts, so two cached rows sharing it are indistinguishable here.
+	const localAnchor = await deps.getLocalAnchorTs(rid, target.ts);
+	if (localAnchor != null) {
+		return localAnchor;
+	}
+	return target.ts ? tsToMs(target.ts) : null;
```
diegolmello: **Bug · low (latent) — guard tests the raw value, not the computed ms.** `target.ts ? tsToMs(target.ts) : null` falls to `null` for a numeric `0` or empty-string `ts` (no re-seed), whereas the server path anchors unconditionally. Unreachable today (a cached `ts` is a `Date`), but the intent is "anchor when ts is valid": `const ms = tsToMs(target.ts); return Number.isFinite(ms) ? ms : null;`.

app/views/RoomView/index.tsx
```diff
@@ -1004,6 +1016,14 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
 		}
 	};
 
+	// Fire a jump from a Navigation param, then consume the one-shot param so re-selecting the SAME
+	// message id reads as a change (undefined -> id edge) and re-fires, instead of matching a stale
+	// param and no-opping. Both mount (initial param) and update (Search delivers via setParams) use this.
+	consumeJumpParam = (messageId: string) => {
+		this.jumpToMessage(messageId);
+		this.props.navigation.setParams({ jumpToMessageId: undefined });
```
diegolmello, line 1026: **Bug · low (latent) — instance field not cleared on the main-list path.** `consumeJumpParam` clears the navigation param but not the `this.jumpToMessageId` instance field (set at construction). The thread path read-and-clears the field first (lines 668–671, "so other init() callers can't re-fire it"); the main-list path at line 245 relies only on this `setParams` clear, leaving the field set. Inert today — the field's other reader (line 668) requires `this.tmid`, which the main-list path excludes — but it's an asymmetric guard. `ARCHITECTURE.md`'s "Jump param is one-shot" invariant also states `consumeJumpParam` clears `jumpToMessageId`, which holds for the nav param but not the field. Clear `this.jumpToMessageId` here so both paths converge.

app/views/RoomView/List/hooks/useScroll.test.tsx
```diff
@@ -0,0 +1,571 @@
...
+				fetchMessages
+			});
+		},
+		{ initialProps: { rows: initialRows, highTs: initialHighTs } }
+	);
+
+	return { ...utils, listRef, scrollToIndex, scrollToOffset, scrollToEnd, idsRef, setHighTs, fetchMessages };
+};
+
+describe('useScroll', () => {
```
diegolmello, line 47: **Test gap · medium.** Two public behaviors of `useScroll` have no coverage here:
1. `cancelJumpToMessage` — neither branch is exercised (already-scrolled → leave the scroll alone; not-scrolled → release the anchor via `setHighTs(null)`). The suite tests the *safety-net* abort (timeout) but never the explicit cancel.
2. `highlightedMessageId` — never asserted to be set to the target on completion, nor cleared after `HIGHLIGHT_TIMEOUT`.

Both are part of the imperative handle / render contract; a regression in either ships green.

app/views/RoomView/services/getLocalAnchor.test.ts
```diff
@@ -0,0 +1,55 @@
...
+	}
+}));
+
+const mockDbGet = database.active.get as unknown as jest.Mock;
+
+// Mock the WatermelonDB query chain. The mock fetch ignores clauses (it cannot evaluate them), so each
+// test seeds the rows that the targeted "nearest Newer Loader above the target" query would return.
+const mockQuery = (rows: unknown[]) => {
+	const fetch = jest.fn(() => Promise.resolve(rows));
+	const query = jest.fn(() => ({ fetch }));
```
diegolmello, line 23: **Test gap · medium.** The `query` mock ignores its arguments, and the only structural assertion (`expect(query).toHaveBeenCalled()`, line 33) doesn't check *which* clauses were passed. Dropping `Q.where('t', NEXT_CHUNK)`, `Q.where('ts', Q.gt(...))`, or `Q.take(1)` still passes — the exact query shape this helper's correctness rests on is untested. (The comment at lines 15–16 concedes the mock can't *evaluate* clauses; the gap is that nothing asserts they were *constructed*.) Capturing the clause descriptors and asserting the type / ts / take would close it — more relevant now that the rejoin read should adopt this same shape.

app/views/RoomView/docs/FLOWS.md
```diff
@@ -0,0 +1,209 @@
...
+
+    User->>RV: jump to message id
+    RV->>RV: getMessageInfo(id) → target {ts, fromServer}
+    RV->>Resolve: resolveJumpAnchor(rid, target, inWindow=false)
+    Resolve->>Server: fetch one Chunk centered on target
+    Server-->>Resolve: chunk rows
+    Resolve->>AR: anchorForServerChunk(rows, id, ts)
+    AR-->>Resolve: highTs (Newer Loader ts) | null
+    Resolve-->>RV: bound
+    RV->>Msgs: setHighTs(bound)  %% count→0, re-seed
```
diegolmello: **Docs · medium — call graph doesn't match the code.** Diagram 2 shows RoomView making two calls: `RV->>Msgs: setHighTs(bound)` (line 58) and `RV->>Scroll: jumpToMessage(id, anchored=true)` (line 61). In the code RoomView makes one call — `this.list.current?.jumpToMessage(message.id, highTs)` — and `useScroll` sets the bound internally via `setHighTs`. The separate RV→useMessages `setHighTs` step doesn't exist and misleads anyone tracing ownership. Also, `jumpToMessage(id, anchored=false/true)` (lines 23 and 61) labels the 2nd argument as a boolean; the signature is `highTs?: number | null`.

app/views/RoomView/docs/ARCHITECTURE.md
```diff
@@ -0,0 +1,140 @@
...
+| Window observation      | `List/hooks/useMessages.ts`                                                                 | Owns `highTs` and the WatermelonDB observation. Re-seeds on anchor change, grows on demand, and climbs/releases the anchor back toward live.                                              |
+| Scroll & jump lifecycle | `List/hooks/useScroll.ts`                                                                   | Owns a pending jump: re-anchor → await re-observe → scroll once → complete; plus the safety net, growth retries, and `onScrollToIndexFailed` climb.                                       |
+| Anchor decision (pure)  | `services/anchorResolver.ts`, `services/resolveJumpAnchor.ts`, `services/getLocalAnchor.ts` | Decide the `highTs` bound for a jump, and whether to raise or release it — purely from rows/ts, no React, no DB (except `getLocalAnchor`'s single query). Unit-tested with plain objects. |
```
diegolmello: **Docs · low — self-contradiction.** This row says the anchor decision is "…purely from rows/ts, no React, no DB" and then "(except `getLocalAnchor`'s single query)" — `getLocalAnchor` does hit the DB. The "(pure)" header plus "no DB" invites a reader to skip DB mocks for it. State it directly: two pure resolvers (`anchorResolver`, `resolveJumpAnchor`'s decision) plus one DB-backed lookup (`getLocalAnchorTs`).

app/views/RoomView/docs/FLOWS.md
```diff
@@ -0,0 +1,209 @@
...
+    User->>RV: tap quoted reply / jump
+    RV->>List: jumpToMessage(id)
+    List->>List: isMessageInWindow(id)? yes
+    List->>Scroll: jumpToMessage(id, anchored=false)
+    Scroll->>Scroll: target already in messages
+    Scroll->>FlatList: scrollToTarget (two-pass)
+    Scroll->>Scroll: completeJump — clear safety, highlight
+```
+
+_Last verified: f78d6a37c_
```
diegolmello: **Docs · low — rotting marker.** `_Last verified: f78d6a37c_` (repeated on every diagram — lines 29, 68, 100, 125, 159, 183, 209) pins a bare internal commit SHA that isn't resolvable from the doc, isn't updated on merge, and goes stale the next time anything around it changes. Drop it; git history already records when each diagram last moved.

app/views/RoomView/services/getMessageInfo.ts
```diff
@@ -10,7 +10,9 @@ const getMessageInfo = async (messageId: string): Promise<TMessageModel | TThrea
 			id: message.id,
```
diegolmello, line 19: **Slop · moderate — `| any` defeats the return type.** The signature on line 6 is `Promise<TMessageModel | TThreadMessageModel | any | null>`; the `| any` collapses the whole union to `any`, so callers get no checking on the `ts` / shape the jump flow now depends on (`resolveJumpAnchor` reads `target.ts` / `fromServer`). All three returned literals here match the `IJumpTarget` shape already declared in `resolveJumpAnchor.ts`. Type the function `Promise<IJumpTarget | null>` and drop `| any`.

app/views/RoomView/services/getLocalAnchor.ts
```diff
@@ -0,0 +1,36 @@
+import { Q } from '@nozbe/watermelondb';
+
+import database from '../../../lib/database';
+import { MessageTypeLoad } from '../../../lib/constants/messageTypeLoad';
+import { tsToMs } from '../../../lib/dayjs';
+import { type TAnyMessageModel } from '../../../definitions';
+
+/**
```
diegolmello, line 8: **Slop · moderate — over-long doc comment.** This 12-line JSDoc (and the 11-line one on `resolveJumpAnchor`) re-narrates what the code and `ARCHITECTURE.md`'s Anchor-resolution section already state. The part that earns its place: "returns the ts of the nearest Newer Loader above the target — the upper bracket of its Chunk — or null when the cached region is contiguous to live." ~3 lines is enough.

app/views/RoomView/List/hooks/useScroll.ts
```diff
@@ -1,124 +1,309 @@
...
-import { type ViewToken, type ViewabilityConfigCallbackPairs } from 'react-native';
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
 
 import { type IListContainerRef, type IListProps, type TListRef, type TMessagesIdsRef } from '../definitions';
-import { VIEWABILITY_CONFIG } from '../constants';
+import { type TAnyMessageModel } from '../../../../definitions';
 
-export const useScroll = ({ listRef, messagesIds }: { listRef: TListRef; messagesIds: TMessagesIdsRef }) => {
+// Abort a jump whose target never re-observes within this window: release the anchor, drop to the Live
+// Tail, clear the spinner. Does not cancel an in-flight scroll — completion is reactive on re-observe.
```
diegolmello: **Slop · moderate — comment describes a side effect this hook doesn't own.** `useScroll` has no spinner — the loading overlay is RoomView's `sendLoadingEvent`. "clear the spinner" here (and on lines 107 and 122) describes work this hook doesn't do; it resolves the jump promise and RoomView clears the overlay. Say "resolve the jump" / "release the anchor". Separately, the `JUMP_SCROLL_POSITION` comment (line 22) restates the literal values ("centered, offset clear, non-animated" = `viewPosition: 0.5, viewOffset: 100, animated: false`) — keep only the non-obvious `animated: false` rationale.

app/views/RoomView/List/hooks/useScroll.ts
```diff
@@ -1,124 +1,309 @@
...
+			const targetIndex = targetId ? messagesIds.current?.findIndex(id => id === targetId) ?? -1 : -1;
+			if (targetIndex === -1) {
+				return;
+			}
+			// Scrolling straight to an unmeasured target doesn't move the viewport, so the window plateaus
+			// short. Step to the measured frontier first (that renders the next batch, advancing
+			// highestMeasuredFrameIndex), then re-attempt — it lands once measured, or re-fires to climb on.
+			if (targetIndex > params.highestMeasuredFrameIndex) {
+				listRef.current?.scrollToIndex({ index: params.highestMeasuredFrameIndex, animated: false });
+				setTimeout(() => {
```
diegolmello, line 238: **Slop · nit — duplicated re-scroll.** This deferred block duplicates `scrollToTarget`'s second pass (lines 159–168): same `setTimeout(SCROLL_TO_INDEX_RETRY_DELAY)`, same `findIndex` settle, same `scrollToIndex({ ...JUMP_SCROLL_POSITION })`. Extract a shared `reScrollWhenSettled(targetId)` so the two can't drift apart.

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -53,6 +65,78 @@ export const useMessages = ({
...
+			if (!wasPresent || isPresent) {
+				return;
+			}
+
+			// Pin to the subscription that fired this emit. fetchMessages reassigns subscription.current on
+			// every re-subscribe (room switch, another raise); if that happens during the awaits below, this
+			// invocation is stale and must not mutate count / highTs for the new window.
+			const sub = subscription.current;
+
+			// Read the Newer Loader closest to the Live Tail directly (highest ts above the bound) so non-loader rows can't crowd the boundary loader out of the fetched set.
```
diegolmello, line 90: Rejoin now reads the boundary Newer Loader directly — `NEXT_CHUNK`, `ts > bound`, `desc`, `take(1)` — instead of a fixed-size, system-unfiltered slab above the bound. With the old slab read, a pre-existing or system row in the gap could push the boundary loader past the fetched rows; the resolver then saw no loader and released the Anchored Window while the Gap was still open. The release-time count query stays — it feeds reading-position growth and isn't redundant with a single-loader read.

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -152,6 +256,11 @@ export const useMessages = ({
 	 * Auto-dispatch until visible content appears or the safety cap is reached.
 	 */
 	useEffect(() => {
+		// Anchored Window deliberately does not auto-follow; only a Live Window auto-dispatches.
+		if (highTs != null) {
```
diegolmello, line 259: `highTs != null` guard (plus `highTs` in the deps) stops the Anchored Window auto-dispatching. The boundary Newer Loader sits at `ts === highTs`, inside the observed range; without this, servers >= 3.16.0 with `hideSystemMessages` set would consume it and climb or release the window with no user scroll.

app/views/RoomView/List/hooks/useMessages.ts
```diff
@@ -7,12 +7,17 @@ import { type IApplicationState, type RoomType, type TAnyMessageModel } from '..
...
 import { compareServerVersion, useDebounce } from '../../../../lib/methods/helpers';
 import { readThreads } from '../../../../lib/services/restApi';
-import { MESSAGE_TYPE_ANY_LOAD, type MessageTypeLoad } from '../../../../lib/constants/messageTypeLoad';
+import { MESSAGE_TYPE_ANY_LOAD, MessageTypeLoad } from '../../../../lib/constants/messageTypeLoad';
 import { MAX_AUTO_LOADS, QUERY_SIZE } from '../constants';
 import { buildVisibleSystemTypesClause } from './buildVisibleSystemTypesClause';
 import { roomHistoryRequest } from '../../../../actions/room';
+import { isNewerLoader, raiseOrRelease, type AnchorMessage } from '../../services/anchorResolver';
+
+const findFirstLoaderId = (messages: TAnyMessageModel[]): string | null =>
```
diegolmello, line 19: Extracted the duplicated loader-finder into `findFirstLoaderId`, shared by the rid-change snapshot effect and the auto-dispatch effect.

app/views/RoomView/List/hooks/useScroll.ts
```diff
@@ -1,124 +1,311 @@
...
-import { type ViewToken, type ViewabilityConfigCallbackPairs } from 'react-native';
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
 
 import { type IListContainerRef, type IListProps, type TListRef, type TMessagesIdsRef } from '../definitions';
-import { VIEWABILITY_CONFIG } from '../constants';
+import { type TAnyMessageModel } from '../../../../definitions';
 
-export const useScroll = ({ listRef, messagesIds }: { listRef: TListRef; messagesIds: TMessagesIdsRef }) => {
+// Abort a jump whose target never re-observes within this window: release the anchor, drop to the Live
+// Tail, resolve the jump. Does not cancel an in-flight scroll — completion is reactive on re-observe.
```
diegolmello, line 7: Dropped the 'clear the spinner' wording here and in `completeJump` / `abortJump` — this hook owns no spinner (RoomView's loading event does). Restated as resolving the jump.

app/views/RoomView/List/hooks/useScroll.ts
```diff
@@ -1,124 +1,311 @@
...
+const MAX_JUMP_GROWTH_RETRIES = 5;
+
+// VirtualizedList re-fires onScrollToIndexFailed synchronously, so defer each retry one frame to break
+// the recursion.
+const SCROLL_TO_INDEX_RETRY_DELAY = 50;
+// A deep target can sit ~30 rows past the measured frontier; each retry climbs ~one render batch, so the
+// cap must cover the distance (5 stalled short). Bounded so an unreachable target aborts, not loops.
+const MAX_SCROLL_TO_INDEX_RETRIES = 20;
+
+// animated:false snaps straight to the target instead of smooth-scrolling through every row between here
```
diegolmello, line 22: `JUMP_SCROLL_POSITION` comment trimmed to the `animated: false` rationale; the prior text just restated the literal `viewPosition` / `viewOffset` values.

app/views/RoomView/List/hooks/useScroll.ts
```diff
@@ -1,124 +1,311 @@
...
+				}
+			}, JUMP_SAFETY_TIMEOUT);
+		},
+		[abortJump]
+	);
 
-				// wait for scroll animation to finish
-				await new Promise(res => setTimeout(res, 300));
+	// Re-scroll once the target's row has settled into the measured window. No-op until it has.
+	const reScrollWhenSettled = useCallback(
```
diegolmello, line 152: Extracted `reScrollWhenSettled(targetId)` — the find-then-scroll tail shared by the two-pass scroll and the scroll-to-index-failed retry. The per-jump guard and id resolution stay at each call site, so behavior is unchanged (param widened to accept `null` for the failed-retry call).

app/views/RoomView/List/definitions.ts
```diff
@@ -11,11 +11,19 @@ export type TMessagesIdsRef = RefObject<string[]>;
 export interface IListProps extends FlatListProps<TAnyMessageModel> {
 	listRef: TListRef;
 	jumpToBottom: () => void;
+	// Anchored Window: loaded rows' bottom isn't the Live Tail, so the scroll-offset
```
diegolmello, line 14: Glossary: dropped '(historical)' from 'Anchored Window' to match UBIQUITOUS_LANGUAGE.md. The window-growth comments in useScroll are likewise corrected from 'page' to QUERY_SIZE / window.

diegolmello (commented):
Follow-up addressing the review on this PR. Scope: moderate-or-higher findings plus all slop; low/nit items that aren't slop are deferred to a separate pass. Inline notes mark each change.

**Addressed:** the spurious release across an open Gap and the anchored auto-dispatch (correctness); both test gaps; the FLOWS.md call-graph/signature drift; the over-long JSDoc, the 'spinner' comments, and the duplicated re-scroll (slop); plus the duplicated loader-finder and the 'page'/'(historical)' glossary drift.

**Not taken — `| any` on `getMessageInfo`'s return:** the suggested `Promise<IJumpTarget | null>` doesn't hold. The `| any` predates this PR, and the caller reads `.rid` / `.replies` / `.tmsg` / `.t` / `.e2e` / `.tlm` off the result and feeds it into model-typed navigation helpers; the returned literals carry fields beyond `IJumpTarget`. An honest fix cascades into those signatures — out of scope here, worth its own typing pass.

**Deferred (low / latent, not slop):** jump completing before confirming the target is on-screen; `count` bumped before the `rid` guard; the raw-value-vs-computed-ms guard in `resolveJumpAnchor`; the uncleared instance field on the main-list path; the ARCHITECTURE.md 'no DB' self-contradiction; the rotting verified-SHA marker.

Verified locally: RoomView suites 83/83, eslint clean on all touched files, tsc clean for these files (it also flags `app/stacks/OutsideStack.tsx` from the recent develop merge — unrelated to this change).

app/views/RoomView/services/getLocalAnchor.ts
```diff
@@ -0,0 +1,28 @@
...
+ * ts of the nearest Newer Loader above the target = the upper bracket of its Chunk; null when the
```
diegolmello, line 9: JSDoc cut from 12 lines to 2; it re-narrated the code and ARCHITECTURE.md.

app/views/RoomView/services/resolveJumpAnchor.ts
```diff
@@ -0,0 +1,57 @@
...
+	fromServer?: boolean;
+}
+
+export interface IJumpAnchorDeps {
+	loadSurroundingMessages: (params: { messageId: string; rid: string }) => Promise<unknown>;
+	getLocalAnchorTs: (rid: string, ts: Date | number | string) => Promise<number | null>;
+}
+
+/**
+ * Upper ts bound (ms) for the re-seeded window, or null to stay on the Live Tail.
```
diegolmello, line 18: JSDoc trimmed to the null cases plus the fromServer / cached split.

app/views/RoomView/services/getLocalAnchor.test.ts
```diff
@@ -0,0 +1,74 @@
...
+	});
+
+	it('queries the messages collection', async () => {
+		const { query } = mockQuery([]);
+		await getLocalAnchorTs('ROOM_ID', new Date('2024-01-01'));
+		expect(mockDbGet).toHaveBeenCalledWith('messages');
+		expect(query).toHaveBeenCalled();
+	});
+
+	it('builds the correct query clause shape', async () => {
```
diegolmello, line 40: New test asserts the exact query clause shape (`rid`, `NEXT_CHUNK`, `ts > target`, `asc`, `take(1)`); dropping any clause now fails, where before only `toHaveBeenCalled()` was checked.

app/views/RoomView/List/hooks/useScroll.test.tsx
```diff
@@ -0,0 +1,641 @@
...
+		expect(scrollToIndex).toHaveBeenCalledWith(expect.objectContaining({ index: 1 }));
+
+		await act(async () => {
+			jest.runOnlyPendingTimers();
+			await Promise.resolve();
+		});
+		await waitFor(() => expect(jumpResolved).toBe(true));
+	});
+
+	it('cancelJumpToMessage is a safe no-op when no jump is pending', () => {
```
diegolmello, line 572: Added coverage for both `cancelJumpToMessage` branches — no-op when nothing is pending, and abort + anchor release + safety-timer disarm when an unscrolled jump is pending — and for the `highlightedMessageId` lifecycle (set on completion, cleared after HIGHLIGHT_TIMEOUT).

app/views/RoomView/docs/FLOWS.md
```diff
@@ -0,0 +1,209 @@
...
+
+    User->>RV: jump to message id
+    RV->>RV: getMessageInfo(id) → target {ts, fromServer}
+    RV->>Resolve: resolveJumpAnchor(rid, target, inWindow=false)
+    Resolve->>Server: fetch one Chunk centered on target
+    Server-->>Resolve: chunk rows
+    Resolve->>AR: anchorForServerChunk(rows, id, ts)
+    AR-->>Resolve: highTs (Newer Loader ts) | null
+    Resolve-->>RV: bound
+    RV->>Scroll: jumpToMessage(id, highTs)
```
diegolmello, line 56: Diagram corrected: RV makes one call, `jumpToMessage(id, highTs)`, and useScroll sets the bound internally — the prior `RV->Msgs: setHighTs` step didn't exist. Diagram 1 likewise shows the arg as `null`, not `anchored=false` (signature is `number | null`).

app/views/RoomView/List/hooks/useMessages.ts
diegolmello: Addressed in 8433ff31a — moved `count.current += QUERY_SIZE` below the `if (!rid) return` guard, so a render with no room id no longer advances the page counter and the next real fetch can't skip a page.

app/views/RoomView/services/resolveJumpAnchor.ts
diegolmello: Addressed in 8433ff31a — the server-ts fallback now returns the value only when `Number.isFinite(ms)` after conversion, instead of testing the raw `ts` for truthiness. A malformed or empty timestamp falls back to the Live Window (null) rather than seeding the bound with NaN.

app/views/RoomView/index.tsx
diegolmello: Addressed in 8433ff31a — `consumeJumpParam` now clears `this.jumpToMessageId` before calling `jumpToMessage`, so jumping to the same message id twice in a row is no longer swallowed by the stale-field guard on the main-list path.

app/views/RoomView/docs/ARCHITECTURE.md
diegolmello: Addressed in 8433ff31a — the anchor-decision row now states the resolvers decide purely from rows/ts (no React, no DB) and that `getLocalAnchorTs` is the one performing the single DB query, instead of claiming both in the same row.

app/views/RoomView/docs/FLOWS.md
diegolmello: Addressed in 8433ff31a — removed the `_Last verified: <sha>_` markers from every diagram in FLOWS.md.

app/views/RoomView/services/getMessageInfo.ts
diegolmello: Addressed in 8433ff31a — dropped `| any` and typed the return as `TGetMessageInfoResult`; the function returns plain object literals in every branch, never a model. The jump-only callers (`shouldNavigateToRoom`, `navToRoom`) consume the type and `navToThread`'s union was widened to accept it; `navToRoom` returns early when the resolved rid is absent rather than querying with `undefined`.

app/views/RoomView/List/hooks/useScroll.ts
diegolmello: Won't fix. Keeping the optimistic complete. The two-pass scroll + onScrollToIndexFailed frontier climb land the measured target before completeJump, and the safety net aborts if it never settles. Gating completion on an inverted-list viewability callback adds a hot-path subscription + abort-path coordination that isn't justified for the narrow tall-row undershoot. Revisit if it shows up in practice.

diegolmello (commented):
Follow-up addressing the remaining review items, pushed as 8433ff31a. Replies are on each thread.

Behavior guards:
- `fetchMessages` no longer advances the page counter on a render with no room id (the `count` increment moved below the `rid` guard), so the next real fetch can't skip a page.
- The server-ts anchor fallback resolves only when the timestamp converts to a finite millisecond value; a malformed or empty `ts` now falls back to the Live Window instead of seeding the bound with `NaN`.
- `consumeJumpParam` clears the `jumpToMessageId` field before re-firing, so jumping to the same message id twice in a row isn't swallowed by the stale-field guard.

Typing: `getMessageInfo` returns an explicit `TGetMessageInfoResult` instead of `| any` (plain object literals in every branch, never a model); the jump-only navigation helpers were retyped to consume it, and `navToRoom` returns early when the resolved rid is absent.

Docs: ARCHITECTURE.md no longer contradicts itself on the anchor decision's purity, and FLOWS.md's per-diagram last-verified SHA markers are gone.

Not taken — gating jump completion on confirming the target is on-screen; rationale is on that thread.

Verified locally: RoomView suites 83/83, eslint clean on the touched files, tsc clean for these files (the pre-existing OutsideStack.tsx errors come from the recent develop merge and are unrelated).

OtavioStasiak (approved):
LGTM, great job!

PR #7389, opened 2026-06-09
ci: speed up and harden the Android store build

diegolmello:
## Proposed changes

The Android store build (`./gradlew bundleRelease` — a signed, multi-ABI store AAB) runs close to its 40-minute step timeout and is intermittently killed before producing an artifact; ~28 of those minutes are New-Architecture native C++ compiled per module per ABI. Because releases are sometimes cut from PR branches, a timed-out PR build can block a release, not just slow review.

This PR keeps the build doing the same work — a full signed bundle for all four ABIs (`armeabi-v7a, arm64-v8a, x86, x86_64`) — but makes it faster and verifies the output:

- **Run Gradle in parallel & enable the build cache** — `--build-cache --parallel --max-workers=4` on `bundleRelease`; heap unchanged at `-Xmx4g`. This speeds up the JVM-side work (Java/Kotlin/resources/packaging).
- **Drop the redundant manual `~/.gradle` cache** and let `gradle/actions/setup-gradle` own the Gradle user-home cache, as its docs recommend.
- **Two hard gates after the build**: every target ABI is present in the AAB, and the AAB's signer certificate matches the official keystore — so a packaging or signing bug fails loudly instead of shipping a defective bundle.
- **Raise the step timeout 40 → 60** as a guardrail (ABI reduction is intentionally off the table for device coverage), and write the `bundleRelease` wall time to the job step summary.

No path-based native build cache is added: the C++ compile is run by the Android Gradle Plugin's external-native-build task, which is not a cacheable Gradle task and always re-runs, so caching `.cxx`/build dirs cannot avoid the recompile. The real native lever is content-keyed compiler caching (ccache), which is tracked as a separate follow-up.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1251

## How to test or reproduce

CI-pipeline change, validated by running the pipeline on this PR:

1. Approve the **Build Android** gate; confirm the `Build` job produces `android-aab`, both gates pass (all four ABIs listed; signing cert matches the keystore), and the step summary shows the `bundleRelease` wall time.
2. Approve the **E2E** gate; confirm the E2E Android run is unaffected.

## Types of changes

## Checklist

## Further comments

CI-config only — no application code changes, so the JS/TS lint and unit-test suites aren't the validation path here.

Two other levers were intentionally rejected: disabling Android release lint (`lintVitalRelease` is the only native-Android lint gate in CI, and the Node-only ESLint/tsc job can't host it), and Gradle configuration-cache (RN 0.81 + New-Arch autolinking spawns `node` at configuration time, which Gradle 8.2+ rejects — facebook/react-native#45154 — and cleartext secrets currently live in the generated `gradle.properties`). Both can be revisited after upstream/secret-handling changes.

### DIFF
--- a/.github/actions/build-android/action.yml
+++ b/.github/actions/build-android/action.yml
@@ -125,46 +125,43 @@ runs:
         fi
       shell: bash
 
-    # Signing check: confirm the AAB carries the official keystore's certificate.
-    # This must NEVER abort the build on an unreadable signature — `bundleRelease`
-    # already signs with the wired-in signingConfig and Play rejects a wrongly-signed
-    # upload — so an empty/unreadable read is a warning and only a positively-read
-    # fingerprint MISMATCH is a hard failure. Note: app/build.gradle sets
-    # `v1SigningEnabled false`, and `keytool -printcert -jarfile` reads only the v1/JAR
-    # signature, so this may legitimately have no fingerprint to compare; the full
-    # keytool output is printed so the AAB's actual signing scheme is visible in the log.
+    # Hard gate: the AAB must be signed with the official upload certificate.
+    # The keystore password is read via keytool's `-storepass:env` rather than a
+    # shell-quoted argument — the secret contains shell-special characters (the same
+    # reason the gradle.properties step above single-quotes the password lines), so
+    # interpolating it into `-storepass "…"` mangles it and the read silently fails.
+    # Captures use `|| true` so an empty read never trips `set -e -o pipefail`; the
+    # full keytool output is printed only on failure so a problem is self-explaining.
     - name: Verify AAB signing certificate
       working-directory: android
+      env:
+        KS_FILE: ${{ inputs.KEYSTORE_OFFICIAL }}
+        KS_ALIAS: ${{ inputs.KEYSTORE_OFFICIAL_ALIAS }}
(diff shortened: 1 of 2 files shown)


.github/actions/build-android/action.yml
```diff
@@ -33,18 +37,22 @@ runs:
         bundler-cache: true
         working-directory: android
 
+    # setup-gradle manages the ~/.gradle cache itself — no separate actions/cache step needed for it.
```
diegolmello: Remove

.github/actions/build-android/action.yml
```diff
@@ -33,18 +37,22 @@ runs:
...
     - name: Set up Gradle
       uses: gradle/actions/setup-gradle@ed408507eac070d1f99cc633dbcf757c94c7933a # v4.4.3
 
-    - name: Cache Gradle
-      uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+    # Native build outputs (CMake objects, Gradle state, intermediates) so warm builds skip
+    # recompiling C++. Written only on develop (Save step); PRs read it via restore-keys.
```
diegolmello: Remove

OtavioStasiak (approved):
LGTM

PR #7392, opened 2026-06-10
feat: presence sync

Rohit3523:
## Proposed changes

Adds support for temporary/expiring custom status messages. Users can now set a status that automatically clears after a chosen duration (30 minutes, 1 hour, or a custom date/time). When a status has an expiry, a clock icon and "Until" timestamp are shown next to the status in the room header, room actions, and room info views for direct messages.

This feature requires server version **8.6.0+**.

On workspaces running 8.6.0+, users can no longer manually set their status to "Away" — it is handled automatically by the server. The Away option is hidden from the Status screen unless the user is already in that state.

Key changes:
- New `ClearAfterPicker` component in StatusView for selecting expiry duration
- `formatStatusExpiry` helper to display formatted expiry timestamps
- `normalizeStatusExpiresAt` helper to handle EJSON and ISO date formats
- Presence stream handlers (`stream-user-presence`, `user-status`) updated to process `statusExpiresAt` and `statusSource` fields
- `users.setStatus` REST API now sends `expiresAt` when set
- Updated types: `ILoggedUser`, `IActiveUser`, `TStatusSource`
- Added stories and tests for the new functionality

## Issue(s)

https://rocketchat.atlassian.net/browse/PRES-39

## How to test or reproduce

1. Connect to a Rocket.Chat server 8.6.0+
2. Go to Edit Status screen
3. Set a custom status text (e.g. "In a meeting")
4. Choose a clear-after duration (30 min, 1 hour, or custom date/time)
5. Save — verify the status shows with a clock icon and "Until" timestamp in:
   - Room header subtitle (DM)
   - Room actions sheet (DM)
   - Room info view (DM)
6. Verify the status auto-clears after the selected duration
7. Test on servers <8.6.0 — the clear-after picker should be hidden and "Away" should still be selectable
8. On 8.6.0+, verify "Away" is not shown as a selectable option (unless already away)
9. Verify presence sync works across multiple clients

## Screenshots

| Android | iOS |
| --- | --- |
| screenshot_android | screenshot_ios |

| Condition | Room Header | Room Action | User Info |
| --- | --- | --- | --- |
| Offline | Screenshot 2026-06-16 at 9 56 43 PM | Screenshot 2026-06-16 at 9 57 44 PM | Screenshot 2026-06-16 at 9 57 56 PM | 
| Busy | Screenshot 2026-06-16 at 9 59 09 PM | Screenshot 2026-06-16 at 9 59 19 PM | Screenshot 2026-06-16 at 9 59 34 PM |
| Online | Screenshot 2026-06-16 at 10 02 00 PM | Screenshot 2026-06-16 at 10 02 17 PM | Screenshot 2026-06-16 at 10 02 34 PM |
| Expiring Status | Screenshot 2026-06-16 at 10 03 58 PM | Screenshot 2026-06-16 at 10 04 08 PM | Screenshot 2026-06-16 at 10 04 25 PM |

Setting expiring status
| Android | iOS |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/f2e738c1-0c24-48b3-a15b-1bd845a54c02" width="100%" controls></video> | <video src="https://github.com/user-attachments/assets/2c164936-fdf1-43f8-801f-3215cc36f4bd" width="100%" controls></video> |

### Realtime status update
1. Room header
<video src="https://github.com/user-attachments/assets/9befac4e-8c71-42bd-9c5f-8b5f1ea39694" width="100%" controls></video>

2. Room Action
<video src="https://github.com/user-attachments/assets/08ff2fb8-c7c5-4990-95be-3035147e3619" width="100%" controls></video>

3. User info
<video src="https://github.com/user-attachments/assets/4c0710ee-a722-42d8-a97f-2d0891e31521" width="100%" controls></video>

## Types of changes

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
## Summary

* **New Features**
  * Added temporary status expiration with preset options (30 minutes, 1 hour) or custom date/time, including a “don’t clear” option.
  * Display “Until …” with a clock for status expiry in direct-room headers and room info views.
  * Expiry handling now supports presence updates coming from supported sources.
* **Localization**
  * Added translations for timed-status options and the “Until” label.
* **Tests**
  * Added automated coverage for status-expiry formatting and expiry normalization (and stabilized status-related snapshots).


app/containers/RoomHeader/RoomHeader.test.tsx
```diff
@@ -1,4 +1,6 @@
 import { generateSnapshots } from '../../../.rnstorybook/generateSnapshots';
 import * as stories from './RoomHeader.stories';
 
+jest.useFakeTimers({ now: new Date('2026-06-15T12:00:00.000Z') });
```
Rohit3523: I have added it because of new stories for expire custom status

app/lib/services/restApi.ts
```diff
@@ -329,9 +329,8 @@ export const setUserPreferences = (userId: string, data: Partial<INotificationPr
 	// RC 0.62.0
 	sdk.post('users.setPreferences', { userId, data });
 
-export const setUserStatus = (status: string, message: string) =>
-	// RC 1.2.0
-	sdk.methodCall('setUserStatus', status, message);
+export const setUserStatus = (status: string, message: string, expiresAt?: string | null) =>
+	sdk.post('users.setStatus', { status, message, ...(expiresAt !== undefined && { expiresAt }) });
```
Rohit3523, line 334: Changed to api instead of using both api + method based on server version check because both return error in different way

app/views/RoomActionsView/index.tsx
```diff
@@ -470,7 +486,7 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
 				const roomUserId = getUidDirectMessage(room);
 				const result = await getUserInfo(roomUserId);
 				if (result.success) {
-					this.setState({ member: result.user as any });
+					this.setState({ member: result.user as unknown as Partial<IUser> });
```
Rohit3523, line 474: It was a suggestion by coderabbit

app/views/RoomActionsView/index.tsx
```diff
@@ -750,9 +766,13 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
 		const { room, member } = this.state;
 		const { rid, name, t, topic, source } = room;
 		const { theme, fontScale } = this.props;
+		const { status, statusText, statusExpiresAt } = member;
 
 		const avatar = getRoomAvatar(room);
 		const isGroupChatHandler = isGroupChat(room);
+		const roomUserId = !isGroupChatHandler && t === 'd' ? getUidDirectMessage(room) : undefined;
+		const formattedStatusExpiry = statusExpiresAt ? formatStatusExpiry(statusExpiresAt) : undefined;
+		const presenceLabel = !statusText && status ? STATUS_I18N_KEYS[status] : undefined;
```
Rohit3523: If user don't have any custom text, we are displaying "Online", "Offline" instead

app/views/RoomActionsView/index.tsx
```diff
@@ -775,14 +795,8 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
...
 					testID='room-actions-info'>
-					<View style={[styles.roomInfoContainer, { height: 72 * fontScale }]}>
-						<Avatar text={avatar} style={styles.avatar} size={50 * fontScale} type={t} rid={rid}>
-							{t === 'd' && member._id ? (
-								<View style={[sharedStyles.status, { backgroundColor: themes[theme].surfaceRoom }]}>
-									<Status size={16} id={member._id} />
-								</View>
-							) : undefined}
-						</Avatar>
+					<View style={styles.roomInfoContainer}>
```
Rohit3523, line 782: Removed the status dot from avatar

app/views/StatusView/ClearAfterPicker.tsx
```diff
@@ -0,0 +1,221 @@
...
+	initialDate: Date | null;
+	onConfirm: (value: ClearAfterValue, date: Date | null) => void;
+}
+
+interface IDatePickerSheetContentProps {
+	initialDate: Date;
+	onConfirm: (date: Date) => void;
+}
+
+const DatePickerSheetContent = ({ initialDate, onConfirm }: IDatePickerSheetContentProps) => {
```
Rohit3523: This component is for iOS

app/views/StatusView/index.tsx
```diff
@@ -154,40 +167,56 @@ const StatusView = (): ReactElement => {
...
 			logEvent(events.STATUS_CUSTOM);
 			showToast(I18n.t('Status_saved_successfully'));
 		} catch (e: any) {
-			const messageError =
-				e.error && e.error.includes('[error-too-many-requests]')
-					? I18n.t('error-too-many-requests', { seconds: e.reason.replace(/\D/g, '') })
-					: e.reason;
 			logEvent(events.STATUS_CUSTOM_F);
-			showErrorAlert(messageError);
+			showErrorAlertWithEMessage(e);
```
Rohit3523, line 183: showErrorAlertWithEMessage takes the raw error object e and handles three things automatically

app/views/StatusView/index.tsx
```diff
@@ -154,40 +167,56 @@ const StatusView = (): ReactElement => {
...
 			log(e);
 		}
 		sendLoadingEvent({ visible: false });
 	};
 
-	const statusType = Accounts_AllowInvisibleStatusOption ? STATUS : STATUS.filter(s => s.id !== 'offline');
+	const statusType = STATUS.filter(s => {
+		if (s.id === 'offline' && !Accounts_AllowInvisibleStatusOption) return false;
+		if (supportsStatusExpiry && s.id === 'away' && user.status !== 'away') return false;
```
Rohit3523, line 191: We can't set away status manually on 8.6.0 but if user is set away by the system, we have to show it in the list
Based on https://github.com/RocketChat/Rocket.Chat/pull/40469/changes#diff-fa37577294c62cadf079d22ecf7ecbb577b5ba57f0b3e2ca29a1dd1c776d1596R45

app/views/StatusView/index.tsx
```diff
@@ -202,7 +231,7 @@ const StatusView = (): ReactElement => {
 						<ControlledFormTextInput
 							name='statusText'
 							control={control}
-							label={I18n.t('Message')}
+							label={I18n.t('Status')}
```
Rohit3523, line 225: This one is present in Figma

Image

app/containers/RoomHeader/RoomHeader.stories.tsx
Rohit3523, line 1: Image

app/containers/RoomHeader/RoomHeader.stories.tsx
```diff
@@ -13,7 +13,7 @@ export default {
 
 const HeaderExample = ({ title, theme = 'light' }: { title: Function; theme?: TSupportedThemes }) => (
 	// Using View directly instead of Header from react-navigation because it's easier to test.
-	<View style={{ flex: 1, maxHeight: 48, backgroundColor: themes[theme].surfaceNeutral }}>{title()}</View>
+	<View style={{ flex: 1, maxHeight: 48, paddingHorizontal: 16, backgroundColor: themes[theme].surfaceNeutral }}>{title()}</View>
```
Rohit3523: After adding 16 horizontal padding
Image

app/lib/methods/helpers/normalizeStatusExpiresAt.ts
Rohit3523, line 1: REST endpoint returns statusExpiresAt as an ISO string or null:
`{ "statusExpiresAt": "2026-06-11T10:00:00.000Z" }`
WebSocket/DDP returns it as an EJSON date (or null):
`{ "statusExpiresAt": { "$date": 1752300000000 } }`
normalizeStatusExpiresAt makes both look the same ("2026-06-11T10:00:00.000Z" or undefined)

app/views/StatusView/ClearAfterPicker.tsx
```diff
@@ -0,0 +1,221 @@
...
+		DateTimePickerAndroid.open({
+			value: baseDate,
+			mode: 'date',
+			minimumDate: new Date(),
+			onChange: (dateEvent, selectedDate) => {
+				if (dateEvent.type !== 'set' || !selectedDate) {
+					setPendingValue('');
+					return;
+				}
+				DateTimePickerAndroid.open({
```
Rohit3523: On android, it shows date and time as separate modal so after user select the date, we ask user to select time

app/views/StatusView/index.tsx
```diff
@@ -125,15 +131,22 @@ const StatusView = (): ReactElement => {
 	const inputValues = watch();
 	const { statusText } = inputValues;
 
+	const initialClearAfter = getInitialClearAfterState(user.statusExpiresAt);
```
Rohit3523: Here, we check whether the user has an expiration time in their status. If an expiration time exists, we display the time instead of the "Don't Clear" text

app/views/StatusView/index.tsx
Rohit3523: also we are getting error in different way

- REST SDK return `{ data: { error: string } }`
- DDP return `{ error: string, reason: string }`

app/views/RoomInfoView/index.tsx
```diff
@@ -73,9 +74,14 @@ const RoomInfoView = (): ReactElement => {
...
 		editRoomPermission: state.permissions['edit-room'],
 		editOmnichannelContact: state.permissions['edit-omnichannel-contact'],
-		editLivechatRoomCustomfields: state.permissions['edit-livechat-room-customfields']
+		editLivechatRoomCustomfields: state.permissions['edit-livechat-room-customfields'],
+		activeUsers: state.activeUsers
 	}));
 
+	const roomUserId = isDirect ? getUidDirectMessage({ ...(room || { rid, t }), itsMe }) : undefined;
+	const activeUserStatus = roomUserId ? activeUsers[roomUserId] : undefined;
+	const userStatus = activeUserStatus || roomUser;
```
Rohit3523, line 85: It picks live presence (Redux activeUsers, kept current by the presence subscription) when available, else the snapshot loaded on open (roomUser). It feeds only the status props; name/username stay on roomUser since activeUsers carries no identity. Swapping the whole object (vs. field-by-field) keeps status/statusText/statusExpiresAt consistent and avoids stale-mix bugs.

app/views/StatusView/index.tsx
```diff
@@ -154,40 +167,56 @@ const StatusView = (): ReactElement => {
...
 		const { status } = inputValues;
 		if (!isValid) {
 			return true;
 		}
+		if (supportsStatusExpiry && clearAfterTouched.current) {
+			return false;
+		}
 		const isStatusEqual = status === user.status;
 		const isStatusTextEqual = (!!user.statusText && user.statusText === statusText) ?? (!user.statusText && !statusText);
 		return !isValid && isStatusEqual && isStatusTextEqual;
```
diegolmello: This branch is unreachable: the function already returned at `if (!isValid) return true`, so `!isValid` is always `false` here and the `isStatusEqual` / `isStatusTextEqual` comparison is dead — Save ends up enabled whenever the form is valid, even with no changes. The `??` on the line above should also be `||` (its left side is always boolean, so the right branch never runs). Pre-existing, but worth fixing while this function is being touched. Impact is limited because `submit()` re-checks for changes.

app/lib/methods/helpers/normalizeStatusExpiresAt.ts
```diff
@@ -0,0 +1,11 @@
+import dayjs from '../../dayjs';
+
+export const normalizeStatusExpiresAt = (value: any): string | undefined => {
```
diegolmello: Prefer `unknown` over `any` for the param. Separately, the "seconds timestamp" unit test is tautological — `dayjs(<number>)` treats the value as milliseconds, so a real epoch-seconds input resolves to 1970; the test passes only because both sides call the same `dayjs`. Either handle seconds explicitly or drop that case.

app/lib/methods/getUsersPresence.ts
```diff
@@ -80,13 +80,13 @@ export async function getUsersPresence(usersParams: string[]) {
...
-					const { _id, status, statusText } = user;
+					const { _id, status, statusText, statusExpiresAt, statusSource } = user;
 
 					if (loggedUser && loggedUser.id === _id) {
-						reduxStore.dispatch(setUser({ status, statusText }));
+						reduxStore.dispatch(setUser({ status, statusText, statusExpiresAt, statusSource }));
 					}
 
-					ret[_id] = { status, statusText };
+					ret[_id] = { status, statusText, statusExpiresAt, statusSource };
```
diegolmello: The DDP paths (`connect.ts`, `setUser.ts`) run `normalizeStatusExpiresAt` on the raw value, but here the REST `statusExpiresAt` is passed through untouched. Fine if REST always returns ISO, but normalizing here too keeps the paths consistent and removes a format-drift footgun.

app/views/StatusView/index.tsx
diegolmello: `getInitialClearAfterState` runs on every render but only seeds `useState` — use a lazy initializer (`useState(() => ...)`). Also note it only seeds initial state, so the picker won't reflect a `statusExpiresAt` that changes via presence while the screen is open (acceptable, just flagging).

app/views/StatusView/ClearAfterPicker.tsx
```diff
@@ -0,0 +1,221 @@
...
+	const insets = useSafeAreaInsets();
+
+	const openAndroidDateTimePicker = (baseDate: Date) => {
+		DateTimePickerAndroid.open({
+			value: baseDate,
+			mode: 'date',
+			minimumDate: new Date(),
+			onChange: (dateEvent, selectedDate) => {
+				if (dateEvent.type !== 'set' || !selectedDate) {
+					setPendingValue('');
```
diegolmello: On Android, cancelling the date picker sets `pendingValue` to '' — this visually clears the previously selected radio rather than restoring it. Minor UX nit.

app/lib/services/restApi.ts
diegolmello, line 334: This switches status updates from the DDP `setUserStatus` method call to the REST `users.setStatus` endpoint for all server versions (not just 8.6.0+), and drops the `// RC …` version comment. Worth confirming `users.setStatus` is available on the minimum supported server.

app/containers/RoomHeader/__snapshots__/RoomHeader.test.tsx.snap
```diff
@@ -1,5 +1,1785 @@
...
+    style={
+      {
+        "backgroundColor": "#E4E7EA",
+        "flex": 1,
+        "maxHeight": 48,
+      }
+    }
+  >
+    <ExternalKeyboardView
+      accessibilityLabel="Online John Doe Online."
```
diegolmello: Repeated status

app/containers/RoomHeader/index.tsx
```diff
@@ -65,8 +67,15 @@ const RoomHeaderContainer = memo(
...
-				const { statusText: statusTextActiveUser } = activeUser;
-				statusText = statusTextActiveUser;
+				const {
+					statusText: statusTextActiveUser,
+					statusExpiresAt: statusExpiresAtActiveUser,
+					status: statusActiveUser
+				} = activeUser;
+				const presenceKey = statusActiveUser ? STATUS_I18N_KEYS[statusActiveUser] : undefined;
+				const presenceLabel = presenceKey ? I18n.t(presenceKey) : undefined;
+				statusText = statusTextActiveUser || presenceLabel;
```
diegolmello, line 77: wdyt?
```suggestion
				const {
					statusText: statusTextActiveUser,
					statusExpiresAt: statusExpiresAtActiveUser,
					status: statusActiveUser
				} = activeUser;
				const presenceLabel = statusActiveUser ? I18n.t(STATUS_I18N_KEYS[statusActiveUser]) : undefined;
				statusText = statusTextActiveUser || presenceLabel;
```

app/containers/RoomHeader/RoomHeader.stories.tsx
diegolmello: I didn't get the last 3. Can you explain?
IIRC Figma mentioned the clock only for some places (without the expiration).

app/lib/methods/helpers/formatStatusExpiry.ts
```diff
@@ -0,0 +1,16 @@
...
+	const expiresAt = dayjs(statusExpiresAt);
+	if (!expiresAt.isValid() || !expiresAt.isAfter(now)) {
+		return undefined;
+	}
+
+	if (expiresAt.isSame(now, 'day')) {
+		return `${I18n.t('Until')} ${expiresAt.format('h:mm A')}`;
+	}
+
+	return `${I18n.t('Until')} ${expiresAt.format('MMM D, h:mm A')}`;
```
diegolmello: You shouldn't use i18n like that.
I got what tried here, but some languages might use `Until` differently or might rely on a different order of semantics.

For example, `pt-BR` and `en` handle adjectives differently.
- black (en) === preto (pt-BR)
- dog (en) === cachorro (pt-BR)
- black dog === cachorro preto

Notice how `black dog` is inverted in pt-BR.
Word by word would be `dog black`.

Same _could_ happen on your use case.
`Until 5pm` could be `5pm until` for some languages.

The solution is to use a param on your a11y, like https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/876124f667cd5315061ad2fda1d1348d15d2941d/app/i18n/locales/en.json#L16

app/lib/methods/helpers/normalizeStatusExpiresAt.ts
diegolmello: You should just use `ejson` https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/876124f667cd5315061ad2fda1d1348d15d2941d/app/lib/methods/subscriptions/rooms.ts#L371

app/views/RoomActionsView/index.tsx
```diff
@@ -279,6 +308,12 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
 		}
 	}
 
+	componentDidUpdate(prevProps: IRoomActionsViewProps): void {
+		if (prevProps.activeUsers !== this.props.activeUsers) {
```
diegolmello: This is always true, so you're always running the function on re-rerender.

Image

app/views/RoomActionsView/index.tsx
```diff
@@ -1327,7 +1385,8 @@ const mapStateToProps = (state: IApplicationState) => ({
 	convertTeamPermission: state.permissions['convert-team'],
 	viewCannedResponsesPermission: state.permissions['view-canned-responses'],
 	livechatAllowManualOnHold: state.settings.Livechat_allow_manual_on_hold as boolean,
-	livechatRequestComment: state.settings.Livechat_request_comment_when_closing_conversation as boolean
+	livechatRequestComment: state.settings.Livechat_request_comment_when_closing_conversation as boolean,
+	activeUsers: state.activeUsers
```
diegolmello: You shouldn't need all activeUsers here.
Log on open and check what's on this state for reference.

app/views/RoomInfoView/components/RoomInfoViewAvatar.tsx
diegolmello, line 1: Can write stories for this? Strange to look at RoomInfoViewTitle stories right below.

app/views/RoomInfoView/components/RoomInfoViewTitle.tsx
```diff
@@ -44,12 +62,31 @@ const RoomInfoViewTitle = ({ room, name, username, statusText, type }: IRoomInfo
...
+				{!!formattedExpiry && (
+					<View testID='room-info-view-status-expiry' style={styles.expiryContainer}>
+						<CustomIcon
+							name='clock'
+							size={14}
+							color={colors.fontSecondaryInfo}
+							accessibilityElementsHidden
+							importantForAccessibility='no'
 						/>
+						<Text style={[styles.expiryText, { color: colors.fontSecondaryInfo }]}>{formattedExpiry}</Text>
```
diegolmello: This logic is pretty much the same from RoomActionsView.
Perhaps we could share part of it? Create a helper function?

app/views/RoomInfoView/index.tsx
```diff
@@ -73,9 +74,14 @@ const RoomInfoView = (): ReactElement => {
 		// permissions
 		editRoomPermission: state.permissions['edit-room'],
 		editOmnichannelContact: state.permissions['edit-omnichannel-contact'],
-		editLivechatRoomCustomfields: state.permissions['edit-livechat-room-customfields']
+		editLivechatRoomCustomfields: state.permissions['edit-livechat-room-customfields'],
+		activeUsers: state.activeUsers
```
diegolmello: Same as the other one. We should avoid using it directly.
https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/876124f667cd5315061ad2fda1d1348d15d2941d/app/containers/Status/index.tsx#L15

app/views/RoomInfoView/styles.ts
```diff
@@ -98,5 +98,26 @@ export default StyleSheet.create({
...
+	},
+	expiryContainer: {
+		flexDirection: 'row',
+		alignItems: 'center',
+		gap: 4
+	},
+	expiryText: {
+		fontSize: 14,
+		...sharedStyles.textAlignCenter,
+		...sharedStyles.textRegular
```
diegolmello, line 112: Missing line height

diegolmello (changes requested):

app/views/StatusView/index.tsx
Rohit3523: Here if we use the lazy initializer, we have to call the same function twice, so instead of it, I am memorising it with useMemo

app/containers/RoomHeader/index.tsx
Rohit3523: It's going to give TS error because `statusActiveUser` can be undefined too

app/lib/services/restApi.ts
Rohit3523: API for setStatus was added in workspace version 1.2.0
https://developer.rocket.chat/apidocs/set-user-status

Screenshot 2026-06-18 at 1 10 34 AM

app/containers/RoomHeader/RoomHeader.tsx
```diff
@@ -174,7 +190,13 @@ const Header = ({
 	// keyboard); regular touch users shouldn't have focus yanked onto the header on room open.
 	const autoFocusHeader = useIsAccessibilityNavigationEnabled();
 	const subtitleAccessibilityLabel = tmid ? parentTitle : subtitle;
-	const accessibilityLabel = `${statusAccessibilityLabel} ${title} ${subtitleAccessibilityLabel || ''}.`;
+	const formattedStatusExpiry = statusExpiresAt ? formatStatusExpiry(statusExpiresAt) : undefined;
+	const fullSubtitleAccessibilityLabel = formattedStatusExpiry
+		? `${subtitleAccessibilityLabel || ''}, ${formattedStatusExpiry}`
+		: subtitleAccessibilityLabel || '';
+	const accessibilityLabel = type === 'd' && subtitleAccessibilityLabel === statusAccessibilityLabel
```
Rohit3523: For DM rooms, the accessibility label duplicated the presence status — `useStatusAccessibilityLabel` returned "Online" as a prefix, while the subtitle prop (which for DMs is the presence label) also contained "Online", producing "Online John Doe Online.". The fix compares `subtitleAccessibilityLabel` against `statusAccessibilityLabel`, when they match (subtitle is the presence label, not custom text), the redundant prefix is dropped, giving "John Doe. Online.". When the subtitle is custom text (e.g. "In a meeting"), the status prefix is kept since there's no duplication.

app/containers/RoomHeader/RoomHeader.stories.tsx
Rohit3523: I have improved the story

Explanation 
1-4 = normal status 
5 = Passing future expire time to make sure clock icon is visible
6 = passing 1 hour future time to make sure clock icon is visible
7 = passing past time to make sure clock icon is not visible - added for edge case ;-;

Image

app/views/RoomInfoView/components/RoomInfoViewAvatar.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+});
+
+const Wrapper = ({ children }: { children: ReactNode }) => <View style={styles.container}>{children}</View>;
+
+export default {
+	title: 'RoomInfoView/RoomInfoViewAvatar',
+	component: RoomInfoViewAvatar
+};
+
+export const Default = () => (
```
Rohit3523, line 23: Image

app/views/RoomInfoView/components/RoomInfoViewAvatar.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+			showEdit={false}
+			type={SubscriptionType.DIRECT}
+			username='john.doe'
+			rid='rid1'
+			handleEditAvatar={() => {}}
+		/>
+	</Wrapper>
+);
+
+export const WithEdit = () => (
```
Rohit3523, line 35: Image

app/views/RoomInfoView/index.tsx
```diff
@@ -64,17 +65,24 @@ const RoomInfoView = (): ReactElement => {
...
+			isMasterDetail: state.app.isMasterDetail,
+			roles: state.roles,
+			usersRoles: state.usersRoles,
+			serverVersion: state.server.version,
+			// permissions
+			editRoomPermission: state.permissions['edit-room'],
+			editOmnichannelContact: state.permissions['edit-omnichannel-contact'],
+			editLivechatRoomCustomfields: state.permissions['edit-livechat-room-customfields']
+		}),
+		shallowEqual
```
Rohit3523, line 80: Added `shallowEqual` here to stop unnecessary reordering on `activeUsers` change

app/views/RoomInfoView/styles.ts
```diff
@@ -98,5 +98,23 @@ export default StyleSheet.create({
...
+		fontSize: 14,
+		lineHeight: 22,
+		...sharedStyles.textAlignCenter,
+		...sharedStyles.textRegular
+	},
+	statusRow: {
+		marginTop: 10
+	},
+	statusExpiryRow: {
+		marginTop: 0
```
Rohit3523, line 118: Using 0 here because default value for row in Status component is 2, which is adding extra space between status text and expire time

app/containers/Status/StatusRows.tsx
diegolmello, line 1: Missing tests and stories

app/containers/Status/StatusRows.tsx
```diff
@@ -0,0 +1,83 @@
...
+				<View style={[styles.row, expiryRowStyle]}>
+					<CustomIcon
+						name='clock'
+						size={14}
+						color={fontSecondaryInfo}
+						accessibilityElementsHidden
+						importantForAccessibility='no'
+					/>
+					<Text style={[secondaryTextStyle, { color: fontSecondaryInfo }]}>{formattedExpiry}</Text>
+				</View>
```
diegolmello: `condition && render` is a lowkey anti pattern on RN.
It hides the else, but renders it.
It may work the way you did, but it's not worth keeping that syntax.
https://echobind.com/post/a-common-conditional-rendering-bug-in-react-native

app/containers/Status/StatusRows.tsx
```diff
@@ -0,0 +1,83 @@
...
+	statusTextColor,
+	fontSecondaryInfo,
+	renderStatusText,
+	textStyle,
+	secondaryTextStyle,
+	rowStyle,
+	expiryRowStyle
+}: IStatusRows): ReactElement => {
+	const presenceLabel = !statusText && status ? STATUS_I18N_KEYS[status] : undefined;
+	const formattedExpiry = statusExpiresAt ? formatStatusExpiry(statusExpiresAt) : undefined;
```
diegolmello: `formatStatusExpiry` is creating a new dayjs each render.

app/lib/methods/helpers/formatStatusExpiry.ts
diegolmello: What if we `statusExpiresAt: string | undefined` and gate the condition at the start of this function?
We would cleanup the place calling it.

app/lib/methods/getUsersPresence.ts
```diff
@@ -83,10 +83,12 @@ export async function getUsersPresence(usersParams: string[]) {
...
 
 					if (loggedUser && loggedUser.id === _id) {
-						reduxStore.dispatch(setUser({ status, statusText, statusExpiresAt, statusSource }));
+						reduxStore.dispatch(
+							setUser({ status, statusText, statusExpiresAt: normalizeStatusExpiresAt(statusExpiresAt), statusSource })
+						);
 					}
 
-					ret[_id] = { status, statusText, statusExpiresAt, statusSource };
+					ret[_id] = { status, statusText, statusExpiresAt: normalizeStatusExpiresAt(statusExpiresAt), statusSource };
```
diegolmello: Duplicate normalizeStatusExpiresAt call

app/views/RoomActionsView/index.tsx
```diff
@@ -1371,22 +1320,27 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
...
-	convertTeamPermission: state.permissions['convert-team'],
-	viewCannedResponsesPermission: state.permissions['view-canned-responses'],
-	livechatAllowManualOnHold: state.settings.Livechat_allow_manual_on_hold as boolean,
-	livechatRequestComment: state.settings.Livechat_request_comment_when_closing_conversation as boolean,
-	activeUsers: state.activeUsers
-});
+const mapStateToProps = (state: IApplicationState, ownProps: Partial<Pick<IRoomActionsViewProps, 'route'>>) => {
+	const params = ownProps.route?.params;
+	const room = params?.room || { rid: params?.rid, t: params?.t };
+	const roomUserId = room && room.t === 'd' ? getUidDirectMessage(room) : undefined;
```
diegolmello: Minor, but valid.
`getUidDirectMessage` calls `const { id: userId } = reduxStore.getState().login.user;`.
We could change it to accept `userId` optional and use it instead of reading redux again.

`mapStateToProps` is a very costly function and we need to pay attention to it.

app/views/RoomInfoView/index.tsx
diegolmello: > to stop unnecessary reordering on activeUsers change

Reordering activeUsers? You removed it from here.
I didn't get, but the shallowEqual is worth it.
Just confirm it's actually working, because I'm not sure it works properly for non primitive data.

app/views/RoomInfoView/index.tsx
```diff
@@ -63,23 +64,24 @@ const RoomInfoView = (): ReactElement => {
...
+	);
 
 	const roomUserId = isDirect ? getUidDirectMessage({ ...(room || { rid, t }), itsMe }) : undefined;
-	const activeUserStatus = roomUserId ? activeUsers[roomUserId] : undefined;
+	const activeUserStatus = useAppSelector(state => (roomUserId ? state.activeUsers[roomUserId] : undefined));
```
diegolmello: shallow here

diegolmello (changes requested):

app/views/RoomInfoView/index.tsx
Rohit3523: Sorry for the confusion 😓, yup, I removed activeUsers, just forgot to update the comment. And the shallowEqual is working fine here

app/views/RoomActionsView/index.tsx
Rohit3523: I have added it with unit test :)

app/containers/Status/StatusRows.stories.tsx
Rohit3523, line 1: Image

app/lib/methods/helpers/helpers.ts
```diff
@@ -14,8 +14,8 @@ export function getRoomAvatar(room) {
 	return room.prid ? room.fname : room.name;
 }
 
-export function getUidDirectMessage(room) {
-	const { id: userId } = reduxStore.getState().login.user;
+export function getUidDirectMessage(room, loggedUserId?: string) {
+	const userId = loggedUserId ?? reduxStore.getState().login.user.id;
```
Rohit3523, line 18: I have added unit test for this

app/containers/RoomHeader/RoomHeader.tsx
```diff
@@ -189,8 +189,8 @@ const Header = ({
 	// Only move focus to the header for accessibility navigation (screen reader or physical
 	// keyboard); regular touch users shouldn't have focus yanked onto the header on room open.
 	const autoFocusHeader = useIsAccessibilityNavigationEnabled();
+	const formattedStatusExpiry = useMemo(() => formatStatusExpiry(statusExpiresAt), [statusExpiresAt]);
```
diegolmello: Now I think we may not need useMemo really, since we're using `use memo` directive.

app/containers/RoomHeader/RoomHeader.tsx
diegolmello, line 1: Can you write tests and stories?

app/containers/Status/StatusRows.tsx
```diff
@@ -43,28 +43,28 @@ const StatusRows = ({
 	rowStyle,
 	expiryRowStyle
 }: IStatusRows): ReactElement => {
+	const formattedExpiry = useMemo(() => formatStatusExpiry(statusExpiresAt), [statusExpiresAt]);
```
diegolmello: use 'use memo` and remove useMemo

app/views/RoomActionsView/index.tsx
```diff
@@ -1323,9 +1323,10 @@ class RoomActionsView extends Component<IRoomActionsViewProps, IRoomActionsViewS
 const mapStateToProps = (state: IApplicationState, ownProps: Partial<Pick<IRoomActionsViewProps, 'route'>>) => {
 	const params = ownProps.route?.params;
 	const room = params?.room || { rid: params?.rid, t: params?.t };
-	const roomUserId = room && room.t === 'd' ? getUidDirectMessage(room) : undefined;
+	const userId = getUserSelector(state).id;
+	const roomUserId = room && room.t === 'd' ? getUidDirectMessage(room, userId) : undefined;
```
diegolmello: nit: `room?.t` instead of `room && room.t`

diegolmello (changes requested):

app/containers/RoomHeader/RoomHeader.tsx
Rohit3523: We have the 'use memo' in this file and i have removed useMemo ;-;

app/containers/RoomHeader/RoomHeader.tsx
Rohit3523: We have stories in `RoomHeader.stories.tsx` which covers new DM status + expiry, For tests, the time logic is unit-tested in `formatStatusExpiry.test.ts`

app/views/StatusView/ClearAfterPicker.tsx
diegolmello, line 1: The more I look at this file, more I find new functions and components 🙈 
Organize the 🍝 and the code is going to look better

app/views/StatusView/index.tsx
```diff
@@ -110,6 +114,8 @@ const StatusView = (): ReactElement => {
 	const Accounts_AllowInvisibleStatusOption = useSelector(
 		(state: IApplicationState) => state.settings.Accounts_AllowInvisibleStatusOption
 	);
+	const serverVersion = useSelector((state: IApplicationState) => state.server.version);
+	const supportsStatusExpiry = compareServerVersion(serverVersion, 'greaterThanOrEqualTo', '8.6.0');
```
diegolmello, line 115: You're running this on every re-render

app/views/StatusView/index.tsx
```diff
@@ -154,40 +168,56 @@ const StatusView = (): ReactElement => {
...
+			return false;
+		}
 		const isStatusEqual = status === user.status;
-		const isStatusTextEqual = (!!user.statusText && user.statusText === statusText) ?? (!user.statusText && !statusText);
-		return !isValid && isStatusEqual && isStatusTextEqual;
+		const isStatusTextEqual = (!!user.statusText && user.statusText === statusText) || (!user.statusText && !statusText);
+		return isStatusEqual && isStatusTextEqual;
 	};
 
 	const FooterComponent = () => (
```
diegolmello: Extract to its own component.

diegolmello (dismissed):

OtavioStasiak (approved):
LGTM

PR #7396, opened 2026-06-13
feat: add expo-sqlite driver adapter, key service, and live-query hooks

diegolmello:
## Proposed changes

Second step of the WatermelonDB → expo-sqlite + Drizzle migration, stacked on the schema PR (#7395): adds the encrypted database driver layer. No runtime behavior changes — nothing in the app imports these modules yet; integration arrives with the data-access facade work.

- **Connection lifecycle** (`app/lib/database/driver/connection.ts`):
  - Opens databases with SQLCipher encryption from byte zero, enforcing the validated open sequence: `PRAGMA key` first (raw-key `x'<64-hex>'` form, skipping PBKDF2 since keys are already full-entropy CSPRNG output), then `PRAGMA busy_timeout = 500` (required for multi-process WAL safety with the iOS notification service extension), then `PRAGMA journal_mode = WAL`, then an open-verify read that throws a key-free error on failure.
  - Places iOS database files in the `group.ios.chat.rocket` App Group container (shared with extensions), with a logged fallback to the default directory; Android uses the default location.
  - Derives clean single-`.db` filenames from server URLs (drops the legacy `.db.db` double suffix) and caches handles in a registry so each file opens once.
- **Key service** (`app/lib/database/driver/keyService.ts`):
  - Generates per-database 32-byte (64-hex) keys via `randomKey` from `@rocket.chat/mobile-crypto` — chosen over `randomBytes`, whose bridge encoding is base64 on both platforms — with a format guard before any key is used.
  - Persists through an `IKeychainShim` interface; the real Keychain/Keystore backend lands with the native-readers work. The built-in in-memory stand-in is dev-only and throws outside `__DEV__`, because silently using it in production would create databases whose keys vanish on restart (permanent data loss).
  - Key material never appears in logs, thrown errors, or telemetry.
- **Live-query hooks** (`app/lib/database/driver/observe.ts`), patterns validated in the live-query proof of concept:
  - `useTableQuery` — table-filtered change listener with a 16ms debounce (expo-sqlite fires one event per row, so large transactions must coalesce into one re-query) and structural sharing so unchanged rows keep their object identity and `React.memo` bails out.
  - `useRowObserve` — per-rowid subscription for hot single-row sites.
  - Both filter events by database file path before table name: `addDatabaseChangeListener` is global across all open databases and the default and per-server databases share table names (e.g. `users`).
- **SQLCipher enablement**: `expo.sqlite.useSQLCipher` flag in `ios/Podfile.properties.json` and `android/gradle.properties`; adds `expo-sqlite ~16.0.10`.
- **Tests**: 47 unit tests covering the open sequence and pragma order, registry behavior, name derivation (including trailing slashes), key idempotence and deletion, the dev-shim production guard, cross-database event isolation, debounce coalescing, and structural sharing.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1274

## How to test or reproduce

- `TZ=UTC pnpm test app/lib/database/driver` — runs the driver, key-service, and live-query hook tests.
- `pnpm exec tsc --noEmit -p .` — type-checks the new modules.
- Full suite: `TZ=UTC pnpm test` (180 suites / 1576 tests passing locally).

## Screenshots

N/A — no UI changes.

## Types of changes

## Checklist

## Further comments

Stacked on #7395 — merge that first; GitHub will retarget this PR to develop automatically. The native Keychain/Keystore key-storage backend and the native database readers (iOS notification service extension, Android notification path) come in a follow-up, which is why key persistence sits behind the `IKeychainShim` interface here.

## Summary by CodeRabbit

* **New Features**
  * Enabled encrypted SQLite databases via SQLCipher
  * Added live-query React hooks for table and row updates with debounced re-fetching and render-stable result reuse
* **Chores**
  * Updated platform configuration to disable edge-to-edge display on Android
  * Added/updated Expo SQLite + SQLCipher setup
* **Tests**
  * Expanded Jest coverage for encrypted DB connection lifecycle, key/salt generation and persistence, and live-query event behavior


app/lib/database/driver/connection.ts
```diff
@@ -0,0 +1,248 @@
...
+ */
+async function openDb<K extends DbKind>(dbName: string, kind: K): Promise<DbHandle<K>> {
+	const cached = _registry.get(dbName);
+	if (cached) {
+		return cached as DbHandle<K>;
+	}
+
+	const [keyHex, saltHex] = await Promise.all([getOrCreateDatabaseKey(dbName), getOrCreateDatabaseSalt(dbName)]);
+
+	const sqlite = await openDatabaseAsync(dbName, { enableChangeListener: true }, DB_DIRECTORY);
```
diegolmello: 🔴 [high] **TOCTOU race in `openDb`**

Two concurrent callers both miss the registry cache, both call `openDatabaseAsync`, and the second `_registry.set` silently overwrites the first. The first handle is orphaned while still holding a file lock, which can cause `SQLITE_BUSY` or `database is locked` on the next open.

Fix: use an in-flight promise registry so concurrent calls coalesce onto one open:

```ts
const _inflight = new Map<string, Promise<DbHandle>>();

async function openDb<K extends DbKind>(dbName: string, kind: K): Promise<DbHandle<K>> {
  const cached = _registry.get(dbName);
  if (cached) return cached as DbHandle<K>;

  const existing = _inflight.get(dbName);
  if (existing) return existing as Promise<DbHandle<K>>;

  const promise = _openDbImpl(dbName, kind).finally(() => _inflight.delete(dbName));
  _inflight.set(dbName, promise);
  return promise;
}
```

app/lib/database/driver/connection.ts
```diff
@@ -0,0 +1,248 @@
...
+
+	await applyOpenPragmas(sqlite, keyHex, saltHex);
```
diegolmello: 🟡 [medium] **Handle leak on PRAGMA failure**

`openDatabaseAsync` succeeds (line 179), then `applyOpenPragmas` throws (line 181). The raw `sqlite` handle is never closed, so the file lock is held for the process lifetime. On repeated open attempts this accumulates leaked handles.

Fix: wrap the pragma sequence in a try/catch and close on failure before rethrowing:

```ts
const sqlite = await openDatabaseAsync(dbName, { enableChangeListener: true }, DB_DIRECTORY);
try {
  await applyOpenPragmas(sqlite, keyHex, saltHex);
} catch (err) {
  await sqlite.closeAsync().catch(() => {});
  throw err;
}
```

app/lib/database/driver/keyService.ts
```diff
@@ -0,0 +1,147 @@
...
+ * Idempotent: repeated calls for the same name return the same key.
+ *
+ * The key is 32 bytes (256-bit) from the platform CSPRNG, hex-encoded to 64 chars.
+ * It is never logged, never included in thrown errors, never sent to telemetry.
+ */
+export async function getOrCreateDatabaseKey(dbName: string): Promise<string> {
+	const k = storageKey(KEY_PREFIX, dbName);
+	const existing = await _shim.getItem(k);
+	if (existing !== null) {
+		return existing;
```
diegolmello: 🟡 [medium] **No hex validation on keychain read-back**

Fresh-generated keys are guarded by `!/^[0-9a-fA-F]{64}$/` (line 107), but the loaded path returns the stored value immediately (line 98) without the same check. A tampered keychain entry on a jailbroken device would be interpolated verbatim into `PRAGMA key = "x'<value>'"`, enabling PRAGMA string injection.

Fix: apply the same regex guard before returning the loaded value:

```ts
const existing = await _shim.getItem(k);
if (existing !== null) {
  if (!/^[0-9a-fA-F]{64}$/.test(existing)) {
    throw new Error('stored key failed format validation; keychain may be tampered');
  }
  return existing;
}
```

Apply the same pattern in `getOrCreateDatabaseSalt` (32-hex guard).

app/lib/database/driver/connection.ts
```diff
@@ -0,0 +1,248 @@
...
+ * Derives the per-server database filename from the server URL.
+ * Strips scheme, replaces slashes with dots, appends a single `.db`.
+ * Matches the legacy `getDatabasePath` logic in `app/lib/database/index.ts`
+ * but deliberately drops the `.db.db` double-suffix (wipe-and-restore recreates all files).
+ */
+export function deriveServerDbName(serverUrl: string): string {
+	const sanitized = serverUrl
+		.replace(/\/+$/, '')
+		.replace(/(^\w+:|^)\/\//, '')
+		.replace(/\//g, '.');
```
diegolmello: 🟡 [medium] **`deriveServerDbName` produces colliding filenames**

`.replace(/\//g, '.')` maps both `https://foo/bar` and `https://foo.bar` to `foo.bar.db`. Two different servers share one encrypted SQLite file and one keychain key — a data isolation bug.

Fix: use a separator that can't appear in a hostname. Replace `/` with `_` (hostnames cannot contain `_`), leaving `.` as the host-segment delimiter only:

```ts
export function deriveServerDbName(serverUrl: string): string {
  const sanitized = serverUrl
    .replace(/\/+$/, '')
    .replace(/(^\w+:|^)\/\//, '')
    .replace(/\//g, '_');  // _ cannot appear in a hostname
  return `${sanitized}.db`;
}
```

Ensure the keychain storage-key derivation in `keyService.ts` uses the same function so the key and the file always agree.

app/lib/database/driver/observe.ts
```diff
@@ -0,0 +1,214 @@
...
+		// Initial load
+		fetchAndReconcile();
+
+		const tableSet = new Set(tables);
+		const sub = addDatabaseChangeListener((event: DatabaseChangeEvent) => {
+			if (!eventMatchesDb(event, dbHandle.dbName)) return;
+			if (!tableSet.has(event.tableName)) return;
+			// Debounce: coalesce all events from a single large transaction into one re-query
+			if (timerRef.current !== null) clearTimeout(timerRef.current);
+			timerRef.current = setTimeout(fetchAndReconcile, debounceMs);
```
diegolmello, line 142: 🟢 [low] **`queryFn` throw inside `setTimeout` is unhandled**

`fetchAndReconcile` (which calls `queryFnRef.current()`) is scheduled via `setTimeout`. A throw inside it becomes an uncaught exception on the timer thread — React error boundaries cannot catch it, and it surfaces as an unhandled promise rejection or a native crash in some RN versions.

Fix: wrap the timer callback:

```ts
timerRef.current = setTimeout(() => {
  try {
    fetchAndReconcile();
  } catch (err) {
    // surface via error state so React boundaries can catch it
    setRows(() => { throw err; });
  }
}, debounceMs);
```

diegolmello (commented):
## Review: driver adapter, key service, live-query hooks

Infrastructure-only PR (nothing imports it yet) and the design is sound — raw-key PRAGMA form is correct, PRAGMA ordering enforced, the plaintext-header + externalised-salt approach for the 0xdead10cc fix is well-reasoned, and the dev-shim production guard is a good rail. Findings below are all fixable before the facade cutover lands; the `openDb` race and the `deriveServerDbName` collision are the two worth prioritising.

app/lib/database/driver/connection.ts
```diff
@@ -0,0 +1,248 @@
...
+	}
+}
+
+/**
+ * Opens a database for the given `dbName`, applies the full open sequence
+ * (key → busy_timeout → WAL → verify), wraps with Drizzle, and registers the handle.
+ * Returns the same handle on repeated calls for the same name.
+ */
+async function openDb<K extends DbKind>(dbName: string, kind: K): Promise<DbHandle<K>> {
+	const cached = _registry.get(dbName);
```
diegolmello, line 168: **bug: high — concurrent-open race leaks a handle.** The `_registry.get` check (here) and `_registry.set` (~line 189) are separated by several awaits (keychain x2, `openDatabaseAsync`, the PRAGMA round-trips). Two concurrent callers (e.g. two sagas opening the same server DB on cold boot) both see no cached handle, both run the full open, and the second `set` overwrites the first — leaving an orphaned open SQLCipher connection that is never closed. Track in-flight opens in a `Map<string, Promise<DbHandle>>` and return the pending promise on a second concurrent call.

app/lib/database/driver/keyService.ts
```diff
@@ -0,0 +1,147 @@
...
+ * Returns the hex key for `dbName`, generating and storing a fresh one if none exists.
+ * Idempotent: repeated calls for the same name return the same key.
+ *
+ * The key is 32 bytes (256-bit) from the platform CSPRNG, hex-encoded to 64 chars.
+ * It is never logged, never included in thrown errors, never sent to telemetry.
+ */
+export async function getOrCreateDatabaseKey(dbName: string): Promise<string> {
+	const k = storageKey(KEY_PREFIX, dbName);
+	const existing = await _shim.getItem(k);
+	if (existing !== null) {
```
diegolmello: **bug: low — stored key returned without re-validation.** The freshly-generated hex is regex-checked before storing, but on a cache hit `existing` is returned unconditionally and fed straight into `PRAGMA key`. A future shim bug (truncation/encoding) would surface as the misleading "key may be wrong or file corrupt" verify error. Re-run the same regex on `existing`; on failure throw a distinct "stored key corrupt" error (with no key material).

app/lib/database/driver/observe.ts
```diff
@@ -0,0 +1,214 @@
...
+ * `onConflictDoUpdate` keeps the same rowid, so per-rowid observers survive upserts.
+ * A rowid of null/undefined (row not yet inserted) causes the hook to return null.
+ */
+export function useRowObserve<T>(
+	dbHandle: DbHandle | null | undefined,
+	tableName: string,
+	rowId: number | null | undefined,
+	fetchRow: FetchRowFn<T>
+): T | null {
+	const [row, setRow] = useState<T | null>(() => {
```
diegolmello, line 178: **bug: low — initial row fetched twice on mount.** The `useState` lazy initializer runs `fetchRow(rowId)` synchronously, then the effect immediately calls `setRow(fetchRowRef.current(rowId))` again before any change event. Drop the lazy initializer and default to `null`; the effect already handles the initial load.

app/lib/database/driver/keyService.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/**
+ * Returns the hex key for `dbName`, generating and storing a fresh one if none exists.
+ * Idempotent: repeated calls for the same name return the same key.
+ *
+ * The key is 32 bytes (256-bit) from the platform CSPRNG, hex-encoded to 64 chars.
+ * It is never logged, never included in thrown errors, never sent to telemetry.
+ */
+export async function getOrCreateDatabaseKey(dbName: string): Promise<string> {
```
diegolmello: **slop: moderate — `getOrCreateDatabaseKey` / `getOrCreateDatabaseSalt` are near-duplicates.** They differ only by prefix, byte count, expected hex length and one error string (~45 lines duplicated). Extract a private `getOrCreate(prefix, dbName, bytes)` helper; both public functions become one-liners.

app/lib/database/driver/connection.ts
```diff
@@ -0,0 +1,248 @@
...
+const _registry = new Map<string, DbHandle>();
+
+// ---------------------------------------------------------------------------
+// Open
+// ---------------------------------------------------------------------------
+
+async function applyOpenPragmas(sqlite: SQLiteDatabase, keyHex: string, saltHex: string): Promise<void> {
+	// 1. Key MUST be first — before any schema read/write
+	// Raw-key form: the x'...' prefix tells SQLCipher to skip PBKDF2
+	await sqlite.execAsync(`PRAGMA key = "x'${keyHex}'";`);
```
diegolmello: **slop: moderate — five sequential `execAsync` PRAGMA calls.** `execAsync` accepts a semicolon-separated multi-statement string; the four config pragmas (key through WAL) can be one call, with only the `getFirstAsync` verify separate. Fewer round-trips and it makes the "configure before any schema access" intent atomic and explicit.

app/lib/database/driver/observe.ts
```diff
@@ -0,0 +1,214 @@
...
+		return () => {
+			sub.remove();
+			if (timerRef.current !== null) {
+				clearTimeout(timerRef.current);
+				timerRef.current = null;
+			}
+		};
+		// deps are caller-controlled; tables/debounceMs treated as stable across renders
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [dbHandle, fetchAndReconcile, ...deps]);
```
diegolmello: **slop: moderate — `...deps` spread blinds the hooks linter to `tables`.** `tables` is closed over inside the effect but isn't in the literal deps array, so if a caller's table set changes without threading it through `deps` the subscription silently watches the stale set — and the `eslint-disable` hides it. Derive a stable `tableKey` from `tables` and put it in the deps array so re-subscription is automatic and lint-visible.

app/lib/database/driver/observe.ts
```diff
@@ -0,0 +1,214 @@
...
+	tableName: string,
+	rowId: number | null | undefined,
+	fetchRow: FetchRowFn<T>
+): T | null {
+	const [row, setRow] = useState<T | null>(() => {
+		if (!dbHandle || rowId == null) return null;
+		return fetchRow(rowId);
+	});
+
+	const mounted = useRef(true);
```
diegolmello: **slop: nit — `mounted` ref is dead code.** The listener is registered in an effect whose cleanup calls `sub.remove()`; in JS's single-threaded model the listener can't fire after removal, so the `if (!mounted.current) return` guard is unreachable. Remove the ref and its effect.

app/lib/database/driver/keyService.ts
```diff
@@ -0,0 +1,147 @@
...
+	return hex;
+}
+
+/**
+ * Deletes the stored key AND salt for `dbName`. Call this when the database file is being
+ * permanently destroyed (e.g. server logout + database wipe), not during migration.
+ * After calling this, the database file is permanently inaccessible.
+ */
+export async function deleteDatabaseKey(dbName: string): Promise<void> {
+	await _shim.removeItem(storageKey(KEY_PREFIX, dbName));
```
diegolmello: **slop: nit — key and salt removed sequentially.** The two `removeItem` calls are independent; `Promise.all` halves the round-trips.

diegolmello (commented):
**Structural review (NATIVE-1274).** The hand-rolled reactivity (table-filtered listener + ~16ms debounce + structural sharing) is sound, the stable-ref pattern correctly avoids the stale-closure bug, PRAGMA key ordering and the `x'<hex>'` form are correct, and key material never reaches logs or thrown errors. Findings inline.

**Test gap (low):** no test asserts `sub.remove()` runs on unmount and that a post-unmount change event does not re-query — the most common hook-leak failure, and pinning it would let the dead `mounted` ref (flagged inline) be removed with confidence.

app/lib/database/driver/connection.ts
```diff
@@ -0,0 +1,228 @@
...
+
+	const schema = kind === 'servers' ? serversSchema : appSchema;
+	// The conditional type SchemaForKind<K> cannot be narrowed by the JS runtime check above;
+	// casting through unknown is the standard TS pattern for this shape.
+	const db = drizzle(sqlite, { schema }) as unknown as ExpoSQLiteDatabase<SchemaForKind<K>>;
+
+	const handle: DbHandle<K> = { db, sqlite, dbName };
+	_registry.set(dbName, handle as DbHandle);
+	return handle;
+}
```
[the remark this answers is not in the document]
diegolmello: Done in 8b6befb97.

app/lib/database/driver/keyService.ts
```diff
@@ -0,0 +1,109 @@
...
+	// + bytesToHex). The argument is the byte count: 32 bytes → 64 hex chars.
+	const hex = await randomKey(32);
+	// Guard the bridge contract: anything but 64 hex chars must not be used as a key
+	if (!/^[0-9a-fA-F]{64}$/.test(hex)) {
+		// Sanitize error — do not include the value in the message
+		throw new Error('key generation produced unexpected output; cannot open database safely');
+	}
+
+	await _shim.setItem(k, hex);
+	return hex;
```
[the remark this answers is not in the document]
diegolmello: Done in 8b6befb97.

app/lib/database/driver/keyService.ts
```diff
@@ -0,0 +1,147 @@
...
+
+	// 16 bytes → 32 hex chars; same hex-encoding contract as randomKey above.
+	const hex = await randomKey(16);
+	if (!/^[0-9a-fA-F]{32}$/.test(hex)) {
+		throw new Error('salt generation produced unexpected output; cannot open database safely');
+	}
+
+	await _shim.setItem(k, hex);
+	return hex;
+}
```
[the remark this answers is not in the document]
diegolmello: Done in 8b6befb97.

app/lib/database/driver/connection.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/connection.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/keyService.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/connection.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/connection.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/keyService.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/keyService.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/connection.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/observe.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/observe.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/keyService.ts
diegolmello: Done in 8b6befb97.

app/lib/database/driver/observe.ts
diegolmello: Deferred (low): setTimeout-scheduled fetch throw is an edge case; revisit if it surfaces.

app/lib/database/driver/observe.ts
diegolmello: Deferred (low): double initial fetch is a minor redundancy, not a correctness issue.

PR #7398, opened 2026-06-15
feat: native SQLCipher database readers and native key store

diegolmello:
## Proposed changes

Rewrites the native (non-JS) database readers so they can open the SQLCipher-encrypted databases created by the new expo-sqlite/Drizzle driver, and adds a native key store that holds the per-database encryption keys. This is the native counterpart to the JS driver adapter — the iOS Notification Service Extension and the Android FCM notification path both read the local database directly (outside JS) to resolve E2E room keys, so they must be taught to open the encrypted files with the same key material and open sequence the JS side uses.

The key store is a **TurboModule**: the app runs the New Architecture (bridgeless), so native modules must expose a codegen-style spec rather than the legacy `RCT_EXTERN_MODULE` / `NativeModules` path.

**iOS**
- `ios/Libraries/DatabaseKeyStore.swift`: Keychain helper class (`kSecClassGenericPassword`, `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, `kSecAttrSynchronizable = false`, access group `S6UPZG7ZR3.chat.rocket.reactnative`). Static `read`/`write`/`delete` the NotificationService extension calls directly.
- `ios/Libraries/DatabaseKeyStore.mm`: the TurboModule (`DatabaseKeyStoreModule`) implementing the spec and returning its JSI instance from `getTurboModule`. The JS registration name is deliberately distinct from the Swift helper class so it does not collide with an existing ObjC class — required by the bridgeless `NSClassFromString` fallback for app-target modules. `getItem` resolves explicit JS `null` on a miss (`nil` bridges to `undefined`, which breaks the `Promise<string | null>` contract and the `!== null` check in `getOrCreateDatabaseKey`).
- `ios/Shared/RocketChat/Database.swift`: opens SQLCipher-encrypted databases (raw-key `PRAGMA key = "x'…'"` → `busy_timeout` → open-verify), derives the clean `.db` filename the same way the JS driver does, and reads the key from the Keychain helper. Room-key/encrypted query semantics are unchanged.
- `ios/Podfile`: adds `pod 'SQLCipher', '~> 4.7.0'` (locked to the version expo-sqlite vendors) plus the static-framework and `xcconfig` fix-ups required to link SQLCipher alongside the SDK's built-in SQLite.
- `ios/RocketChatRN.xcodeproj` + bridging header: register the new files in the app and extension targets.

**Android**
- `…/storage/NativeDatabaseKeyStoreSpec.java`: abstract TurboModule spec (`getName()` → `DatabaseKeyStoreModule`, abstract `getItem`/`setItem`/`removeItem`).
- `…/storage/DatabaseKeyStoreModule.java`: concrete module. Each key is wrapped with an AndroidKeyStore AES-GCM key; the wrapped blob lives in its own `SharedPreferences` file. Exposes the JS methods plus static helpers the notification path calls without a React context.
- `…/storage/DatabaseKeyStoreTurboPackage.java` (+ `MainApplication.kt` registration): `TurboReactPackage` that maps the module name to a `ReactModuleInfo` with `isTurboModule = true` and returns the concrete instance.
- `…/notification/Encryption.java`: opens the database with `net.zetetic` SQLCipher using the raw-key string form (never the `byte[]` overload, which silently PBKDF2-derives and fails), derives the clean filename to match the JS convention, reads the key from the key store, and applies the same `busy_timeout` invariant.
- `android/app/build.gradle`: adds `net.zetetic:sqlcipher-android:4.7.0`.

**JS**
- `app/lib/native/NativeDatabaseKeyStore.ts`: TurboModule spec resolved via `TurboModuleRegistry.get('DatabaseKeyStoreModule')`.
- `app/lib/database/driver/keyStore.ts` (+ test): thin shim wrapping the native module as the driver's `IKeychainShim`, with `installNativeKeychainShim()` to wire it into the key service at startup.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1276

## How to test or reproduce

This PR is stacked on #7396 (driver adapter), which is stacked on #7395 (schema). Review/merge those first. The native open path is exercised end-to-end once the data layer (wipe-and-restore) lands; for this PR:

- iOS: `pnpm pod-install` then build the app and the NotificationService extension — both targets must compile with the new SQLCipher pod and the registered TurboModule. Runtime: a Keychain round-trip through `DatabaseKeyStoreModule` returns `null` on a miss, the stored value after `setItem`, and `null` again after `removeItem`.
- Android: build the app — `net.zetetic:sqlcipher-android` resolves and `Encryption.java` compiles against the concrete module's static helpers.
- JS: `TZ=UTC pnpm test --testPathPattern='driver/__tests__/keyStore'` and `pnpm lint`.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

Native modules use the New Architecture TurboModule path because the app is bridgeless; the legacy `RCT_EXTERN_MODULE` / `NativeModules` registration does not vend modules under bridgeless. The crypto (iOS Keychain, Android AndroidKeyStore AES-GCM) is unchanged from the prior revision — only the module registration and class layout moved to the TurboModule shape.

## Summary by CodeRabbit

* **New Features**
  * Added SQLCipher-encrypted database support using per-database encryption keys stored securely on iOS (Keychain) and Android (Keystore).
  * Added a native key/value module to supply encryption keys to the database.
* **Improvements**
  * Updated notification-based room reading to open the encrypted database directly and apply keying and access timing safeguards.
* **Chores**
  * Added/updated native wiring and build configuration for SQLCipher and the encryption key modules on iOS and Android.
* **Tests**
  * Added Jest coverage for native key store shim behavior and missing-module error handling.


android/app/src/main/java/chat/rocket/reactnative/notification/Encryption.java
```diff
@@ -172,55 +173,82 @@ private ParsedMessage parseMessage(Ejson.Content content) {
...
-
-           return new Room(e2eKey, encrypted);
+            db = SQLiteDatabase.openDatabase(dbPath, rawKey, null, SQLiteDatabase.OPEN_READONLY, null);
+
+            // Mandatory multi-process WAL safety, matching the JS driver and the iOS reader:
+            // the notification path can read while the main app holds a WAL lock; without a
+            // busy timeout that contention surfaces as an immediate SQLITE_BUSY.
+            db.execSQL("PRAGMA busy_timeout = 500;");
+
+            Cursor cursor = db.rawQuery("SELECT * FROM subscriptions WHERE id == ? LIMIT 1", new String[]{ejson.rid});
```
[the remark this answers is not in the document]
diegolmello: Not a defect — the Android query is correct. A subscription record's primary-key `id` is set to the room id when the record is created (`app/lib/methods/subscriptions/rooms.ts`: `sanitizedRaw({ id: tmp.rid }, ...)`), so `WHERE id == ?` bound with `ejson.rid` resolves the same row that the iOS `WHERE rid = ?` query targets. Both readers hit the identical row; there is no schema mismatch. Verified against a real device read.

ios/Podfile
```diff
@@ -74,6 +78,40 @@ post_install do |installer|
...
+  # injection for SQLCipher, so -l"SQLCipher" fails to resolve at link time.
+  # Add the path manually to all aggregate xcconfigs that already have the setting.
+  Dir.glob(File.join(__dir__, 'Pods/Target Support Files/Pods-*/*.xcconfig')).each do |xcconfig_path|
+    content = File.read(xcconfig_path)
+    sqlcipher_lib_dir = '"${PODS_CONFIGURATION_BUILD_DIR}/SQLCipher"'
+    if content.include?('LIBRARY_SEARCH_PATHS') && !content.include?(sqlcipher_lib_dir)
+      modified = content.gsub(
+        'LIBRARY_SEARCH_PATHS = $(inherited)',
+        "LIBRARY_SEARCH_PATHS = $(inherited) #{sqlcipher_lib_dir}"
+      )
```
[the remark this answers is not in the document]
diegolmello: This is a literal-string `gsub` (no regex anchors), so the Ruby `^`/`$` multiline caveat does not apply. The generated Pods aggregate xcconfigs emit exactly `LIBRARY_SEARCH_PATHS = $(inherited)`, and the `include?` guard keeps the patch idempotent. A full `RocketChat`-scheme build links SQLCipher correctly, confirming the patch applies for the current CocoaPods output. Treating broader format-resilience as a non-blocking enhancement rather than a fix for this PR.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
```diff
@@ -0,0 +1,161 @@
...
+	// -------------------------------------------------------------------------
+
+	@Override
+	public void getItem(String key, Promise promise) {
+		try {
+			promise.resolve(getItemInternal(getReactApplicationContext(), key));
+		} catch (Exception e) {
+			Log.e(TAG, "getItem failed for key: " + key, e);
+			promise.resolve(null);
+		}
```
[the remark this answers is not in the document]
diegolmello: Fixed in 5606e6f44. `getItem` now rejects on a read/decrypt failure (`KEYSTORE_READ_ERROR`) and `getItemInternal` throws when the keystore alias is missing for an existing key blob or when decryption fails — resolving null only for a true not-found. This stops a read failure from being collapsed into the same null as not-found, which would have caused `getOrCreateDatabaseKey` to mint a new key over an existing encrypted DB (data loss). The iOS reader fails closed symmetrically (`KEYCHAIN_READ_ERROR` + NSError out-param).

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
```diff
@@ -0,0 +1,163 @@
...
+		byte[] combined = new byte[iv.length + encrypted.length];
+		System.arraycopy(iv, 0, combined, 0, iv.length);
+		System.arraycopy(encrypted, 0, combined, iv.length, encrypted.length);
+
+		SharedPreferences prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE);
+		prefs.edit().putString(key, Base64.encodeToString(combined, Base64.DEFAULT)).apply();
+	}
+
+	/** Removes the SharedPreferences entry. Does not delete the AndroidKeyStore key. */
+	public static void removeItemInternal(Context context, String key) {
```
diegolmello, line 173: 🔴 [high] **`removeItemInternal` leaks the AndroidKeyStore AES alias**

Only the SharedPreferences blob is removed; `ks.deleteEntry(key)` is never called. After removal, a subsequent `setItem` for the same key finds `containsAlias(key) == true` and silently reuses the old AES key material instead of generating a fresh one. This breaks key rotation.

Fix:

```java
public static void removeItemInternal(Context context, String key) {
    SharedPreferences prefs = context.getSharedPreferences(PREFS_NAME, Context.MODE_PRIVATE);
    prefs.edit().remove(key).apply();
    try {
        KeyStore ks = KeyStore.getInstance("AndroidKeyStore");
        ks.load(null);
        if (ks.containsAlias(key)) ks.deleteEntry(key);
    } catch (Exception e) {
        Log.e(TAG, "removeItemInternal: failed to delete KeyStore alias for " + key, e);
    }
}
```

diegolmello (commented):
## Review: native SQLCipher readers + native key store

The security design is largely right: SQLCipher raw-key string form (`x'…'`), a dedicated Keychain service namespace, `ThisDeviceOnly` + non-synchronizable attributes, and AES-GCM on Android. Module naming is consistent across the iOS `RCT_EXPORT_MODULE`, the Android spec `NAME`, and the TS `TurboModuleRegistry.get` (all `DatabaseKeyStoreModule`) — I verified this against the PR branch, ignoring the stale `DatabaseKeyStoreModule.mm` / mismatched spec left uncommitted in the working tree.

The `removeItemInternal` alias leak and the `hasConstants=true` flag are the two worth prioritising.

Open questions worth confirming outside the diff: (1) the NSE `.entitlements` must list the `keychain-access-group` for the extension to read the key; (2) `KeyGenParameterSpec` does not set `setIsStrongBoxBacked(true)`, so the claim of hardware-backing falls back to the TEE rather than StrongBox where present — fine if deliberate, worth a comment.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
```diff
@@ -0,0 +1,163 @@
...
+			promise.resolve(null);
+		} catch (Exception e) {
+			Log.e(TAG, "removeItem failed for key: " + key, e);
+			promise.resolve(null);
+		}
+	}
+
+	// -------------------------------------------------------------------------
+	// Native-side helpers — callable from Encryption.java (same process, plain Context)
+	// -------------------------------------------------------------------------
```
diegolmello, line 86: 🟡 [medium] **`removeItem` catch resolves instead of rejects**

The catch block on line 86 calls `promise.resolve(null)`, so the JS caller receives a successful result even when the delete failed. The TS contract (`Promise<void>`) implies rejection on error; callers cannot distinguish success from failure and will proceed believing the key was removed.

Fix: replace `promise.resolve(null)` in the catch with `promise.reject`:

```java
} catch (Exception e) {
    Log.e(TAG, "removeItem failed for key: " + key, e);
    promise.reject("KEYSTORE_REMOVE_ERROR", e);
}
```

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreTurboPackage.java
```diff
@@ -0,0 +1,44 @@
...
+        return () -> {
+            final Map<String, ReactModuleInfo> moduleInfos = new HashMap<>();
+            moduleInfos.put(
+                    NativeDatabaseKeyStoreSpec.NAME,
+                    new ReactModuleInfo(
+                            NativeDatabaseKeyStoreSpec.NAME,
+                            NativeDatabaseKeyStoreSpec.NAME,
+                            false, // canOverrideExistingModule
+                            false, // needsEagerInit
+                            true,  // hasConstants
```
diegolmello, line 36: 🔴 [high] **`hasConstants=true` with no `getConstants()` triggers unnecessary eager sync call**

When `hasConstants` is `true`, the RN TurboModule infrastructure makes a synchronous JS→native call at startup to fetch constants. `DatabaseKeyStoreModule` implements no `getConstants()`, so this call returns nothing and adds startup latency with zero benefit. On older devices this is a measurable main-thread stall.

Fix:

```java
new ReactModuleInfo(
    NativeDatabaseKeyStoreSpec.NAME,
    NativeDatabaseKeyStoreSpec.NAME,
    false, // canOverrideExistingModule
    false, // needsEagerInit
    false, // hasConstants — module exposes no constants
    false, // isCxxModule
    true   // isTurboModule
)
```

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
```diff
@@ -0,0 +1,163 @@
...
+		byte[] combined = Base64.decode(encoded, Base64.DEFAULT);
+
+		byte[] iv = new byte[GCM_IV_LENGTH];
+		byte[] encrypted = new byte[combined.length - GCM_IV_LENGTH];
+		System.arraycopy(combined, 0, iv, 0, GCM_IV_LENGTH);
+		System.arraycopy(combined, GCM_IV_LENGTH, encrypted, 0, encrypted.length);
+
+		Cipher cipher = Cipher.getInstance(TRANSFORMATION);
+		cipher.init(Cipher.DECRYPT_MODE, secretKey, new GCMParameterSpec(GCM_TAG_LENGTH, iv));
+		byte[] plain = cipher.doFinal(encrypted);
```
diegolmello, line 119: 🟢 [low] **Missing length guard before `new byte[combined.length - GCM_IV_LENGTH]`**

If `combined` is shorter than `GCM_IV_LENGTH` bytes (e.g. a truncated SharedPreferences blob), `combined.length - GCM_IV_LENGTH` is negative, throwing `NegativeArraySizeException` with no indication of what went wrong.

Fix: add an explicit guard before the array allocation:

```java
if (combined.length <= GCM_IV_LENGTH) {
    throw new IllegalStateException(
        "stored blob for key '" + key + "' is too short to contain a GCM IV (" + combined.length + " bytes)");
}
```

ios/Libraries/DatabaseKeyStore.swift
```diff
@@ -0,0 +1,104 @@
...
+			kSecValueData as String:          data
+		]
+
+		let addStatus = SecItemAdd(attrs as CFDictionary, nil)
+		if addStatus == errSecSuccess {
+			return true
+		}
+		if addStatus == errSecDuplicateItem {
+			attrs.removeValue(forKey: kSecValueData as String)
+			let updateStatus = SecItemUpdate(attrs as CFDictionary, [kSecValueData as String: data] as CFDictionary)
```
diegolmello: 🟢 [low] **`SecItemUpdate` dict missing `kSecAttrAccessible`**

When `SecItemAdd` returns `errSecDuplicateItem`, the update dict passed to `SecItemUpdate` contains only `kSecValueData`. An item previously stored with a weaker accessibility (e.g. `kSecAttrAccessibleAlways`) will not be tightened to `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.

Fix:

```swift
let updateStatus = SecItemUpdate(
    attrs as CFDictionary,
    [
        kSecValueData as String: data,
        kSecAttrAccessible as String: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly as String
    ] as CFDictionary
)
```

diegolmello (commented):
**Structural review (NATIVE-1276).** Verified clean: SQLCipher raw-key `x'<hex>'` form on both platforms (not the byte[] overload), iOS Keychain attributes (`AfterFirstUnlockThisDeviceOnly`, non-synchronizable, NSE access group), TurboModule wiring (`getTurboModule:` on iOS, `TurboReactPackage` on Android — New Architecture path), `busy_timeout` PRAGMA, cursor/DB close in `finally`, and no key material in logs. Findings inline.

**Note on URL→filename derivation:** Android (`Encryption.java` `replaceFirst("^(\\w+:)?//")`) and iOS (`Database.swift` `range(of: "://")`) are two separate copies that agree on `https://` today but diverge on edge cases (scheme-less `//`, trailing slashes), and neither is tested against the canonical JS `deriveServerDbName`. They decide which DB file gets opened — worth a shared fixture per platform.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreTurboPackage.java
diegolmello, line 36: **bug: med — `hasConstants = true` but the module has no `getConstants()`.** `DatabaseKeyStoreModule` never overrides `getConstants()`, so the base returns null; declaring `hasConstants` true asks the bridge to serialize a null constants map. Other modules here use `false` when they have no constants (PushNotification, VideoConf) and implement `getConstants()` when they set `true` (SSLPinning). Set this to `false`.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	@Override
+	public void removeItem(String key, Promise promise) {
+		try {
+			removeItemInternal(getReactApplicationContext(), key);
+			promise.resolve(null);
+		} catch (Exception e) {
+			Log.e(TAG, "removeItem failed for key: " + key, e);
+			promise.resolve(null);
```
diegolmello, line 80: **bug: med — `removeItem` swallows failures.** The catch logs and then calls `promise.resolve(null)`, so any exception from `removeItemInternal` surfaces to JS as a successful void return. Reject with a distinct error code instead.

android/app/src/main/java/chat/rocket/reactnative/notification/Encryption.java
```diff
@@ -172,55 +173,87 @@ private ParsedMessage parseMessage(Ejson.Content content) {
...
```
diegolmello: **bug: low — wrong lookup column.** `SELECT * FROM subscriptions WHERE id == ?` is queried with `ejson.rid` (a room id). The iOS reader correctly uses `WHERE rid = ?`. This worked under WMDB only because `subscription.id == rid` there; in the new Drizzle schema `id` may be a generated value, so this returns no rows and decryption silently fails. Query `WHERE rid = ?`.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
```diff
@@ -0,0 +1,163 @@
...
+		if (!ks.containsAlias(key)) {
+			// Blob exists but Keystore alias is gone — this is corruption, not a miss.
+			throw new IllegalStateException("keystore alias missing for existing key blob: " + key);
+		}
+
+		SecretKey secretKey = (SecretKey) ks.getKey(key, null);
+		byte[] combined = Base64.decode(encoded, Base64.DEFAULT);
+
+		byte[] iv = new byte[GCM_IV_LENGTH];
+		byte[] encrypted = new byte[combined.length - GCM_IV_LENGTH];
```
diegolmello, line 113: **bug: low — no length check before splitting the GCM blob.** A corrupt blob that Base64-decodes to fewer than `GCM_IV_LENGTH` bytes makes `combined.length - GCM_IV_LENGTH` negative → `NegativeArraySizeException`. Guard with an explicit "ciphertext too short" check that throws a clear error.

ios/Libraries/DatabaseKeyStore.mm
```diff
@@ -0,0 +1,66 @@
+#import <React/RCTBridgeModule.h>
+
+#import <SSLPinning/SSLPinning.h>
```
diegolmello, line 3: **slop: heavy — spurious `#import <SSLPinning/SSLPinning.h>`.** Copy-paste from another module; this file has no SSL-pinning dependency. Remove.

android/app/src/main/java/chat/rocket/reactnative/notification/Encryption.java
```diff
@@ -172,55 +173,87 @@ private ParsedMessage parseMessage(Ejson.Content content) {
     }
 
     public Room readRoom(final Ejson ejson, Context context) {
-        String dbName = getDatabaseName(ejson.serverURL(), context);
-        WMDatabase db = null;
+        String dbName = deriveDbName(ejson.serverURL());
+        String dbPath = context.getFilesDir().getAbsolutePath() + "/SQLite/" + dbName;
 
+        // Load the SQLCipher native library. Safe to call multiple times per process.
+        System.loadLibrary("sqlcipher");
```
diegolmello: **slop: heavy — `System.loadLibrary("sqlcipher")` runs inside `readRoom()`.** Called on every incoming push. Move it to a `static { }` initializer so it loads once.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
diegolmello, line 173: **slop: heavy — AndroidKeyStore alias never deleted on remove.** `removeItemInternal` clears the SharedPreferences blob but leaves the AES-GCM wrapping key in the Keystore, so every server logout orphans one alias permanently. Call `ks.deleteEntry(key)` after removing the prefs entry.

ios/Libraries/DatabaseKeyStore.swift
```diff
@@ -0,0 +1,104 @@
...
+import Security
+
+@objc(DatabaseKeyStore)
+final class DatabaseKeyStore: NSObject {
+
+	// kSecAttrService shared between this module and Database.swift
+	static let service = "chat.rocket.reactnative.dbkeys"
+
+	// Full team-prefixed access group — bare suffix fails with errSecMissingEntitlement
+	private static let accessGroup = "S6UPZG7ZR3.chat.rocket.reactnative"
```
diegolmello, line 30: **slop: heavy — hardcoded team-prefixed Keychain access group.** `"S6UPZG7ZR3.chat.rocket.reactnative"` is baked in, while the entitlements use `$(AppIdentifierPrefix)chat.rocket.reactnative` resolved at build time. If the signing team changes (enterprise build, fork, profile rotation) main-app writes succeed but NSE reads fail with `errSecItemNotFound` and notifications silently stop decrypting. Derive it from the bundle entitlement / a build-time constant.

app/lib/database/driver/keyStore.ts
```diff
@@ -0,0 +1,37 @@
...
+		removeItem: (key: string) => native.removeItem(key)
+	};
+}
+
+/**
+ * Installs the native Keychain/Keystore shim into the key service.
+ * Must be called once before any database is opened.
+ * Safe to call multiple times (subsequent calls are no-ops in keyService).
+ */
+export function installNativeKeychainShim(): void {
```
diegolmello, line 37: **slop: moderate — `installNativeKeychainShim()` has no call site in this PR.** The TurboModule key store is unreachable until this is invoked once at startup (the doc comment says "before any database is opened"). Wire it in the same PR.

android/app/src/main/java/chat/rocket/reactnative/storage/NativeDatabaseKeyStoreSpec.java
```diff
@@ -0,0 +1,30 @@
...
+  public NativeDatabaseKeyStoreSpec(ReactApplicationContext reactContext) {
+    super(reactContext);
+  }
+
+  @Override
+  public String getName() {
+    return NAME;
+  }
+
+  @ReactMethod
```
diegolmello: **slop: nit — `@ReactMethod` on the TurboModule spec methods.** TurboModule registration is via JSI/codegen, not `@ReactMethod` reflection. Harmless but bridge-era noise (mirrors the existing SSLPinning spec); drop it.

android/app/src/main/java/chat/rocket/reactnative/notification/Encryption.java
diegolmello: Fixed: query `WHERE rid = ?` (was `id == ?`), matching the iOS reader. id no longer equals rid under the Drizzle schema.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
diegolmello: Fixed: removeItemInternal now deletes the AndroidKeyStore alias too, so a later setItem regenerates fresh key material.

diegolmello: Fixed alongside the alias-leak thread: ks.deleteEntry(key) now runs in removeItemInternal.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
diegolmello: Fixed: removeItem now rejects with KEYSTORE_REMOVE_ERROR on failure instead of resolving.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
diegolmello: Fixed alongside the swallowed-failure thread: removeItem rejects on error.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreTurboPackage.java
diegolmello: Fixed: hasConstants set to false — the module exposes no constants.

diegolmello: Fixed alongside the hasConstants thread: set to false.

ios/Libraries/DatabaseKeyStore.mm
diegolmello: Removed the spurious SSLPinning import.

android/app/src/main/java/chat/rocket/reactnative/notification/Encryption.java
diegolmello: Moved sqlcipher load into a static initializer so it runs once per process, not per push.

ios/Libraries/DatabaseKeyStore.swift
diegolmello: Fixed: accessGroup is now a single non-private source of truth with the team-ID coupling documented; Database.swift reads only via DatabaseKeyStore.

android/app/src/main/java/chat/rocket/reactnative/storage/NativeDatabaseKeyStoreSpec.java
diegolmello: Dropped the @ReactMethod annotations (and import) — TurboModule dispatch is via codegen, not reflection.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
diegolmello: Acknowledged — low. The blob is module-owned; a truncated entry already fails at decrypt and is caught as KEYSTORE_READ_ERROR. Skipping for now.

android/app/src/main/java/chat/rocket/reactnative/storage/DatabaseKeyStoreModule.java
diegolmello: Duplicate of the length-guard thread; skipping at low severity.

ios/Libraries/DatabaseKeyStore.swift
diegolmello: Acknowledged — low. Every write here already uses AfterFirstUnlockThisDeviceOnly, so no weaker-accessibility item is produced to tighten. Skipping.

app/lib/database/driver/keyStore.ts
diegolmello: Intentional: the doc comment defers the startup wiring to the follow-up ticket in this stack. Tracking it there rather than expanding this PR's scope.

PR #7404, opened 2026-06-17
fix: freeze when blurred to avoid off-screen re-renders on room open

diegolmello:
## Proposed changes

On phones, opening a room pushes RoomView over RoomsListView, but the list behind keeps rendering — every off-screen RoomItem re-renders during the transition, adding avoidable JS work to room open.

This sets `freezeOnBlur: true` on the RoomsListView screen in the phone Chats stack, so the list stops rendering while it's blurred (fully covered by RoomView) and thaws with the latest data on return.

Tablet/master-detail is unaffected: there RoomsListView is rendered as drawer content, a separate path with no freeze. The WatermelonDB subscription stays alive while the screen is frozen (react-freeze keeps the component mounted) — only rendering is deferred and applied on thaw, so the list reflects any changes made while it was blurred.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1291

## How to test or reproduce

1. On a phone (or phone layout), open a room from the list, then go back. The list shows correct, up-to-date state.
2. Reactivity check: open a room, send a message (or otherwise change the room) while inside it, then go back — the room reorders to the top and shows the new last-message preview immediately.
3. (Optional) With the React profiler, cold-open a room before/after: the off-screen RoomItem re-renders in the list behind the room no longer occur.

## Screenshots

N/A — no visual change.

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Improved the rooms list view behavior during navigation transitions by preserving screen state when switching away and returning.

### DIFF
--- a/app/stacks/InsideStack.tsx
+++ b/app/stacks/InsideStack.tsx
@@ -100,7 +100,7 @@ const ChatsStackNavigator = () => {
 	const { theme } = useContext(ThemeContext);
 	return (
 		<ChatsStack.Navigator screenOptions={{ ...defaultHeader, ...themedHeader(theme) }}>
-			<ChatsStack.Screen name='RoomsListView' component={RoomsListView} />
+			<ChatsStack.Screen name='RoomsListView' component={RoomsListView} options={{ freezeOnBlur: true }} />
 			<ChatsStack.Screen name='RoomView' component={RoomView} />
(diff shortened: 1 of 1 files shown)


OtavioStasiak (approved):
Approved

PR #7405, opened 2026-06-17
fix: mount the emoji searchbar only when it's visible

diegolmello:
## Proposed changes

EmojiSearchbar was always mounted inside the message composer and hid itself with an internal early `return null`. So on every room open it mounted, ran its hooks, and triggered a frequently-used-emoji database read — despite being hidden until the user opens emoji search.

This gates the mount on `showEmojiSearchbar`, threads the flag from MessageComposer through MessageComposerContent, and removes the now-dead internal null guard. The searchbar now mounts only when it becomes visible, avoiding a render and an async DB read on every room open.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1292

## How to test or reproduce

1. Open a room: the emoji searchbar is not present (no visible behavior change).
2. Tap the emoji button → "search" to open the emoji searchbar: it appears, shows frequently-used emoji, search works, and the back chevron closes it.
3. Pick an emoji from the searchbar: it inserts as before and the "frequently used" list updates.

## Screenshots

N/A — no visual change.

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Updated the message composer so emoji searchbar visibility is driven by the composer’s current emoji-search state.
  * The emoji searchbar UI is now rendered consistently by the composer content, and the visibility logic is handled via a dedicated prop rather than internal conditional rendering.

### DIFF
--- a/app/containers/MessageComposer/MessageComposer.tsx
+++ b/app/containers/MessageComposer/MessageComposer.tsx
@@ -246,6 +246,7 @@ export const MessageComposer = ({
 			<MessageComposerContent
 				recordingAudio={recordingAudio}
 				action={action}
+				showEmojiSearchbar={showEmojiSearchbar}
 				composerInputComponentRef={composerInputComponentRef}
 				composerInputRef={composerInputRef}
 				onLayout={handleLayout}>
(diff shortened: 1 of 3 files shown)


OtavioStasiak (approved):
LGTM

PR #7406, opened 2026-06-17
chore: improve spotlight search

OtavioStasiak:
## Proposed changes

Split search() into searchLocal (fast, DB-only) and searchRemote (spotlight merge over the local data).
Search surfaces now paint local results instantly and re-render once the backend responds, instead of blocking the UI on the network round-trip. Updated RoomsListView spotlight, message composer @/# autocomplete, New Message, and Selected Users, each with a request-id guard so a slower older search can't overwrite a newer one. Added unit tests for the local/remote/merge logic. No change to the existing local-vs-web result parity (deferred as follow-up).

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1105

## How to test or reproduce

1. Open the app with an account that has several rooms/DMs and is connected to a server.
2. Open spotlight search in the Rooms List and type a query (e.g. a partial username/room name).
Confirm local matches appear instantly, then remote (spotlight) results fill in.
3. In a room, type @ and # in the message box — confirm local mentions show immediately, then remote
results merge in without duplicates.
4. Repeat in New Message and Select Users search inputs.
5. Type quickly / change the query mid-search and confirm results don't flicker back to an older search's results.
6. Run the unit tests: TZ=UTC pnpm test --testPathPattern='app/lib/methods/search.test.ts'.

## Screenshots
| Before | After |
| ----- | ----- |
| before_spotlight | after_spotlight |

| Before II | After II |
| ----- | ----- |
| before_spotlight | afterII |

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved search responsiveness with a local-first flow that shows suggestions immediately, then enriches with remote results.

* **Bug Fixes**
  * Prevented older in-flight searches from overwriting newer results.
  * Refined message composer and autocomplete to apply remote updates safely, including “all/here” suggestions when permitted.

* **UI Updates**
  * Added/adjusted loading indicators during the remote search phase in message composer, new message view, rooms list search, and selected users.

* **Tests**
  * Added unit tests covering local vs remote search, merging, filtering, and deduplication.
  * Updated composer tests to target the updated remote search behavior.


OtavioStasiak:
@coderabbit review

diegolmello (approved):
LGTM. Only missing a before/after video running on open server. Expecting a big change 👀

app/views/NewMessageView/index.tsx
```diff
@@ -73,14 +77,23 @@ const NewMessageView = () => {
...
-		setSearch(result);
+		setSearching(true);
+
+		try {
+			// Paint local results immediately while the backend request is still in flight
+			const localData = await searchLocal({ text, filterRooms: false });
+			if (isStale()) return;
+			setSearch(localData as ISearch[]);
+
+			const result = (await searchRemote({ text, filterRooms: false, localData })) as ISearch[];
```
diegolmello: `searchRemote` needs to wait until `searchLocal` and `setSearch` finish before starting. We should trigger both at the same time, so we gain a few ms.

We should use something like this instead.

```
// Simulating your two promises
const fetchFastData = () => new Promise(resolve => setTimeout(() => resolve("⚡ Fast data is ready!"), 1000));
const fetchSlowData = () => new Promise(resolve => setTimeout(() => resolve("🐢 Slow data is ready!"), 4000));

// 1. Trigger them at the same time
const fastPromise = fetchFastData();
const slowPromise = fetchSlowData();

// 2. Handle them independently as soon as they finish
fastPromise.then(result => {
  console.log(result); // This runs after 1 second
  // Do your fast UI updates here
});

slowPromise.then(result => {
  console.log(result); // This runs after 4 seconds
  // Do your slow UI updates here
});
```

This way we dispatch both promises at the same time and handle the results as they arrive.

app/views/RoomsListView/hooks/useSearch.ts
```diff
@@ -86,6 +86,7 @@ export const useSearch = () => {
 		try {
 			const localData = await searchLocal({ text });
 			if (isStale()) return;
+			console.log('localData', localData);
```
diegolmello: remove

app/views/RoomsListView/index.tsx
```diff
@@ -106,13 +107,13 @@ const RoomsListView = memo(function RoomsListView() {
 		);
 	};
 
-	if (searchEnabled) {
+	if (searchEnabled && searchResults.length === 0) {
+		// Only block on the spinner while we have nothing to show yet. Once local results
+		// are painted, fall through to the list even if the remote request is still in flight.
```
diegolmello: Not needed imo

app/views/SelectedUsersView/index.tsx
```diff
@@ -124,14 +127,23 @@ const SelectedUsersView = () => {
...
-		setSearch(result);
+		setSearching(true);
+
+		try {
+			// Paint local results immediately while the backend request is still in flight
+			const localData = await searchLocal({ text, filterRooms: false });
+			if (isStale()) return;
+			setSearch(localData);
+
+			const result = await searchRemote({ text, filterRooms: false, localData });
```
diegolmello: same and it could be part of the `search` file

diegolmello (changes requested):

app/lib/methods/search.ts
```diff
@@ -190,3 +190,18 @@ export const searchRemote = async ({
 		return data;
 	}
 };
+
+// Orchestrates the two-phase search: resolves the local query first and hands it to `onLocal`
+// so callers can paint it immediately, then resolves with the spotlight-augmented results.
+// searchRemote still receives localData so spotlight can exclude users already present locally.
```
diegolmello: Not needed

diegolmello (approved):

PR #7407, opened 2026-06-17
refactor: cut all database call sites over to the facade

diegolmello:
## Proposed changes

Replaces the app's direct `@nozbe/watermelondb` usage with the WatermelonDB-shaped facade over the Drizzle/expo-sqlite driver, so the app runs on the SQLCipher-encrypted database while every call site keeps its existing shape. Adds the table/model maps the facade needs to build each `Database`, wires the servers and app schemas through it, and adds an ESLint rule that bans new direct `watermelondb` imports.

WatermelonDB stays installed only as the legacy migration reader; no JS runtime path uses it anymore.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1277

## How to test or reproduce

1. Build and run on iOS and Android.
2. Exercise DB-backed flows: rooms list, open a room, send/receive messages, drafts, switch servers, logout/login.
3. `TZ=UTC pnpm test` — full suite passes (1632 tests).

Data now lives in the encrypted database; behaviour is unchanged from the user's perspective.

## Types of changes

## Checklist

## Further comments

Stacked on `feat/native-1276-native-readers`; review/merge that first. Part of the WatermelonDB → encrypted SQLite cutover (NATIVE-1272).

## Summary by CodeRabbit

* **Refactor**
  * Refactored database layer architecture by implementing a new abstraction facade to improve code organization, consistency, and long-term maintainability across all data-related operations throughout the application.

* **Chores**
  * Updated build system configurations, including Babel transpiler and Metro bundler settings, along with development dependencies to enhance development workflow and build optimization.


app/lib/database/facade/decorators.ts
```diff
@@ -0,0 +1,274 @@
...
+				if (!association || association.type !== 'has_many') {
+					throw new Error(`@children decorator used for a table that's not has_many: ${childTable}`);
+				}
+
+				const col = new Collection(
+					(childCollection as unknown as { table: string }).table,
+					(childCollection as unknown as { schema: TableSchema }).schema,
+					(childCollection as unknown as { _drizzleTable: SQLiteTable })._drizzleTable,
+					(childCollection as unknown as { _handle: DbHandle })._handle,
+					Model
```
diegolmello: 🟡 [medium] **`@children` builds a `Collection<Model>` instead of reusing the registered child collection**

Line 179–184 constructs `new Collection(..., Model)` using the hardcoded base `Model` class. Rows fetched through this collection come back as bare `Model` instances — `@field`/`@date`/`@json` getters are not wired, `_raw` is never updated via `prepareUpdate`, and `batch()` writes stale or missing columns.

The registered child collection already has the correct model constructor. Fix: query through the existing collection directly rather than constructing a new one:

```ts
const childCollection = model.collections.get(childTable) as ICollection;
// ...
const query = (childCollection as unknown as { query: (...args: unknown[]) => Query<Model> })
  .query(Q.where(association.foreignKey, model.id));
cache[childTable] = query;
return query;
```

Drop the `new Collection(...)` block entirely.

diegolmello (commented):
## Review: facade cutover (110 files)

The semantic risk concentrates in the ~1k-line facade, not the mechanical import surgery — the call-site changes are clean and I found no missed `@nozbe/watermelondb` imports outside `app/lib/database/`. The ESLint ban rule is well-constructed (per-filetype overrides, correct `excludedFiles`, barrel import still allowed). The `@children` decorator bug is the one that can write stale data; the rest are observable-semantics drift vs WMDB.

app/lib/database/facade/observe.ts
```diff
@@ -0,0 +1,187 @@
...
+/**
+ * Produces an Observable that emits whenever a specific row (by id) changes.
+ * Uses rowId matching from the change event for precision.
+ */
+export function observeRow<T>(handle: DbHandle, tableName: string, fetchFn: () => T | null): Observable<T> {
+	return new Observable<T>(subscriber => {
+		const emit = () => {
+			if (subscriber.closed) return;
+			const row = fetchFn();
+			if (row !== null) subscriber.next(row);
```
diegolmello, line 111: 🟢 [low] **`observeRow` omits `subscriber.complete()` on row deletion**

When `fetchFn()` returns `null` (row deleted), only `subscriber.next` is skipped — `subscriber.complete()` is never called. WMDB's `model.observe()` completes the observable on model destruction. Callers using `takeUntil`, `first()`, or `take(1)` will not have their teardown run, causing memory leaks in long-lived components.

Fix:

```ts
const emit = () => {
  if (subscriber.closed) return;
  const row = fetchFn();
  if (row !== null) {
    subscriber.next(row);
  } else {
    subscriber.complete();
  }
};
```

app/lib/database/facade/Model.ts
```diff
@@ -0,0 +1,164 @@
...
+	_collection: ICollection;
+
+	/** Pending write op, consumed by batch(). */
+	_pendingOp: PendingOp | null = null;
+
+	/** Memoized date cache keyed by ms timestamp — mirrors WMDB @date behavior. */
+	_dateCache: Map<number, Date> = new Map();
+
+	/** Memoized JSON parse cache keyed by raw string value — mirrors WMDB @json memo=true. */
+	_jsonDecoratorCache: Record<string, [unknown, unknown]> = {};
```
diegolmello: 🟢 [low] **Dead field: `_jsonDecoratorCache` is declared but never used**

The comment says it mirrors WMDB `@json memo=true`, but `@json`'s getter in `decorators.ts` calls `parseJSON` on every access without consulting this cache. The field allocates an object per model instance for no benefit.

Either implement the memoization or drop the field. Hot fields like `reactions`, `roles`, and `customFields` are re-parsed on every render; memoizing by raw-string value would avoid repeated `JSON.parse` calls in list views:

```ts
get(this: AnyModel): unknown {
  const model = this.asModel;
  const rawValue = model._getRaw(rawFieldName) as string;
  const cache = model._jsonDecoratorCache;
  if (Object.prototype.hasOwnProperty.call(cache, rawFieldName)) {
    const [cachedRaw, cachedVal] = cache[rawFieldName];
    if (cachedRaw === rawValue) return cachedVal;
  }
  const parsed = sanitizer(parseJSON(rawValue), model);
  cache[rawFieldName] = [rawValue, parsed];
  return parsed;
}
```

app/lib/database/facade/observe.ts
```diff
@@ -0,0 +1,187 @@
...
+			sub.remove();
+			if (timer !== null) clearTimeout(timer);
+		};
+	});
+}
+
+function sameByColumns<T extends HasId>(prev: T[], next: T[], cols: Set<string>): boolean {
+	if (prev.length !== next.length) return false;
+	for (let i = 0; i < prev.length; i++) {
+		const a = rowData(prev[i]);
```
diegolmello: 🟢 [low] **`sameByColumns` diffs by array index, not by row id**

When a row is inserted or deleted at any position other than the end, every subsequent row shifts by one index. `prev[i]` and `next[i]` are now different rows, so column comparison is meaningless — real updates on unchanged rows are suppressed (false negatives), and actual changes on moved rows may be missed or double-fired.

Fix: diff by id-keyed maps:

```ts
function sameByColumns<T extends HasId>(prev: T[], next: T[], cols: Set<string>): boolean {
  if (prev.length !== next.length) return false;
  const prevById = new Map(prev.map(r => [r.id, rowData(r)]));
  for (const row of next) {
    const a = prevById.get(row.id);
    if (!a) return false;
    const b = rowData(row);
    for (const col of cols) {
      if (a[col] !== b[col]) return false;
    }
  }
  return true;
}
```

app/lib/database/facade/decorators.ts
diegolmello: **bug: high — `@children` returns untyped base `Model` instances.** The `new Collection(...)` here passes `Model` as the model class instead of the registered subclass, so children come back without their `@field`/`@date`/`@json` getters — reading typed props yields `undefined`. `Database.get()` already resolves the right class via `_modelMap[table]`; use the same lookup (`db._modelMap[childTable] ?? Model`).

app/lib/database/facade/translate.ts
```diff
@@ -0,0 +1,109 @@
...
+		case 'or': {
+			const conditions = clause.clauses.map(c => translateWhere(c, columns)).filter(Boolean);
+			return or(...(conditions as SQL[])) ?? undefined;
+		}
+		// sortBy/take/skip are not where clauses; they are handled in translateClauses
+		case 'sortBy':
+		case 'take':
+		case 'skip':
+			return undefined;
+		case 'on':
```
diegolmello, line 77: **bug: high — `Q.on` is exported but silently dropped.** `case 'on'` returns `undefined`, so any query passing a `Q.on(...)` clause loses it → full table scan, no WHERE. `Q.on` is in the public barrel, so a future call-site migration that used it would break silently. Either remove `Q.on` from the barrel or `throw` here when an `on` clause is encountered.

app/lib/database/facade/observe.ts
```diff
@@ -0,0 +1,187 @@
...
+			const row = fetchFn();
+			if (row !== null) subscriber.next(row);
+		};
+
+		// Initial emit
+		emit();
+
+		const sub = addDatabaseChangeListener(event => {
+			if (!event.databaseFilePath.endsWith(`/${handle.dbName}`)) return;
+			if (event.tableName !== tableName) return;
```
diegolmello, line 120: **bug: med — `observeRow` ignores `event.rowId` and has no debounce.** The filter checks db + table but not the row id, so every write to the table re-fires all `model.observe()` subscriptions in it; and unlike `observeTable` there's no ~16ms debounce, so batched writes cause N synchronous re-fetches. Add `(!event.rowId || event.rowId === id)` to the filter and the same debounce as `observeTable`.

app/lib/database/facade/Database.ts
```diff
@@ -0,0 +1,132 @@
...
+						.run();
+				} else if (op === 'destroy') {
+					// eslint-disable-next-line @typescript-eslint/no-explicit-any
+					(db as any)
+						.delete(drizzleTable)
+						.where(eq((drizzleTable as never as Record<string, never>).id, model._raw.id))
+						.run();
+				}
+
+				model._pendingOp = null;
```
diegolmello: **bug: med — `_pendingOp` cleared inside the transaction callback.** If the Drizzle transaction throws (e.g. unique constraint), SQLite rolls the rows back but the `_pendingOp = null` already applied to earlier models in JS stands — those models look clean while their rows were rolled back, and can't be retried. Collect the models to clear in a list and null them only after the transaction returns successfully.

app/lib/database/facade/observe.ts
diegolmello, line 111: **bug: low — `observeRow` never completes/errors on deletion.** When the row goes away `fetchFn()` returns null and the Observable just stops emitting without `complete()`/`error()` (WMDB errors on deletion). Call sites using `takeUntil` or cleanup in `subscribe` can leak. Complete or error the subscriber when the row disappears.

app/lib/database/facade/Database.ts
```diff
@@ -0,0 +1,132 @@
...
+						.insert(drizzleTable)
+						.values(model._raw as never)
+						.run();
+				} else if (op === 'update') {
+					const { id, ...rest } = model._raw;
+					// eslint-disable-next-line @typescript-eslint/no-explicit-any
+					(db as any)
+						.update(drizzleTable)
+						.set(rest as never)
+						.where(eq((drizzleTable as never as Record<string, never>).id, id))
```
diegolmello: **slop: heavy — `(db as any)` ~10x and a brittle `.id` assumption.** The expo-sqlite Drizzle driver is fully typed; the casts exist only because `_handle.db` drops the schema generic — type it as `ExpoSQLiteDatabase<SchemaForKind<K>>` and they disappear. The `eq((drizzleTable as never as Record<string, never>).id, id)` cast is the worst: it hard-codes a PK named `id` with zero type checking, silently wrong for any future table whose PK differs.

app/lib/database/facade/Model.ts
diegolmello: **slop: moderate — `_jsonDecoratorCache` declared but never used.** The `@json` getter re-parses + re-sanitizes on every access (allocation per render for attachment/reaction fields), and the unused field falsely implies memoization is active. Either implement the cache in the getter or delete the field.

app/lib/database/facade/schema.ts
```diff
@@ -0,0 +1,125 @@
...
+		if (typeof value === 'boolean') {
+			raw[key] = value;
+		} else if (value === 1 || value === 0) {
+			raw[key] = Boolean(value);
+		} else {
+			raw[key] = isOptional ? null : false;
+		}
+	} else if (isValidNumber(value)) {
+		// number column, valid value
+		raw[key] = value || 0;
```
diegolmello: **slop: moderate — dead `value || 0`.** This branch is guarded by `isValidNumber(value)`, so `value` is already a finite number; `value || 0` equals `value` for every case. Reads as "zero out falsy" but never does. Just assign `value`.

app/lib/database/facade/Collection.ts
```diff
@@ -0,0 +1,142 @@
...
+		}
+		return Promise.resolve(rows[0]);
+	}
+
+	/** Prepare a new record without persisting it. Tag _pendingOp = 'create'. */
+	prepareCreate(fn: (record: M) => void): M {
+		// Start with a raw where id will be set by the fn or sanitizedRaw
+		const raw = sanitizedRaw({}, this.schema);
+		const model = new this._ModelClass(this, raw);
+		model._pendingOp = 'create';
```
diegolmello: **slop: nit — `prepareCreate` sets `_pendingOp` before `fn()` here but after `fn()` in `Model.prepareCreate` (Model.ts:113).** Not a live bug (fn doesn't read it) but the inconsistency will trip anyone auditing both paths. Pick one order.

app/lib/methods/updateMessages.ts
```diff
@@ -1,6 +1,5 @@
-import { type Model, Q } from '@nozbe/watermelondb';
-import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
-
+import { type Model, Q } from '../database/facade';
+import { sanitizedRaw } from '../database/facade';
```
diegolmello: **slop: nit — two imports from the same facade barrel on consecutive lines.** Merge into one import.

diegolmello (commented):
**Structural review (NATIVE-1277).** The 110-file call-site swap is mechanical and consistent — every `@nozbe/watermelondb` import now points at the facade barrel, and the `connect.ts` / `logout.ts` await refactors are necessary and correct. The ESLint ban on facade-internal imports is correctly structured (`.js` → `no-restricted-imports`, `.ts/.tsx` → the `@typescript-eslint` variant, facade dir exempted).

**Bigger-picture concern:** the facade is ~1,400 source lines across 10 files — a near-complete reimplementation of WatermelonDB's runtime (query DSL, SQL translation, reactive layer, all 6 decorators, schema builders, `sanitizedRaw`, writer queue), not a thin shim. That's a full ORM that has to stay correct for the entire burn-down window. Worth confirming the burn-down (NATIVE-1282) is scheduled tightly so this layer doesn't become permanent. Findings inline.

app/lib/database/facade/Database.ts
diegolmello: Fixed in 30aee8f39: @children now queries through the DB-registered child collection, so model-class mapping (field/date/json getters, batch writes) stays intact — no more bare Model.

app/lib/database/facade/decorators.ts
diegolmello: Fixed in 30aee8f39: _pendingOp is cleared only after the transaction commits, so a rollback no longer leaves earlier models marked clean.

app/lib/database/facade/observe.ts
diegolmello: Fixed in 30aee8f39: observeRow is debounced (16ms) like observeTable, so a batched write no longer fires N synchronous re-fetches per subscription.

app/lib/database/facade/Collection.ts
```diff
@@ -0,0 +1,142 @@
...
+	_fetchSync(filter?: Record<string, unknown>): M[] {
+		const { db } = this._handle;
+		const columns = getTableColumns(this._drizzleTable);
+		let q = db.select().from(this._drizzleTable as never);
+		if (filter?.id) {
+			const idCol = columns.id;
+			if (idCol) {
+				// eslint-disable-next-line @typescript-eslint/no-explicit-any
+				q = (q as any).where(eq(idCol, filter.id));
+			}
```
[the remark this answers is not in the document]
diegolmello: Fixed in 30aee8f39: _fetchSync uses `!== undefined`, so find('') applies the WHERE and rejects instead of full-scanning.

app/lib/database/facade/writer.ts
```diff
@@ -0,0 +1,28 @@
...
+			resolve = res;
+			reject = rej;
+		});
+
+		this._tail = this._tail
+			.then(() => fn().then(resolve, reject))
+			.then(
+				() => undefined,
+				() => undefined
+			);
```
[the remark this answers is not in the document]
diegolmello: Fixed in 30aee8f39: enqueue settles the caller's promise on a synchronous throw in fn — no more hang on the writer mutex.

app/lib/database/facade/observe.ts
diegolmello: Fixed in 30aee8f39: sameByColumns diffs by row id, not array index — a mid-list insert/delete no longer suppresses or misfires updates.

app/lib/database/facade/observe.ts
diegolmello: Fixed in 30aee8f39: translateWhere case 'on' now throws instead of silently dropping to a full-table scan.

app/lib/database/facade/schema.ts
diegolmello: Fixed in 30aee8f39: removed the dead `|| 0` in the valid-number branch.

app/lib/database/facade/Collection.ts
diegolmello: Fixed in 30aee8f39: _pendingOp='create' is set after fn(model), matching Model.prepareCreate's order.

app/lib/methods/updateMessages.ts
diegolmello: Fixed in 30aee8f39: merged the two facade imports into one.

app/lib/database/facade/Database.ts
diegolmello: Fixed in 30aee8f39: handle's db is typed via the schema generic (zero `as any` remain) and the PK is resolved via getTableColumns(drizzleTable).id instead of a hardcoded `.id` literal.

app/lib/database/facade/Model.ts
diegolmello: Fixed in 30aee8f39: removed the dead _jsonDecoratorCache field (the @json getter never read it).

app/lib/database/facade/Database.ts
```diff
@@ -0,0 +1,132 @@
...
+				} else if (op === 'destroy') {
+					// eslint-disable-next-line @typescript-eslint/no-explicit-any
+					(db as any)
+						.delete(drizzleTable)
+						.where(eq((drizzleTable as never as Record<string, never>).id, model._raw.id))
+						.run();
+				}
+
+				model._pendingOp = null;
+			}
```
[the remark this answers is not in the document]
diegolmello: Duplicate of the _pendingOp-in-transaction thread; addressed in 30aee8f39.

app/lib/database/facade/decorators.ts
```diff
@@ -0,0 +1,274 @@
...
+				);
+				// eslint-disable-next-line @typescript-eslint/no-explicit-any
+				col._db = (childCollection as any)._db;
+
+				const query = col.query(Q.where(association.foreignKey, model.id));
```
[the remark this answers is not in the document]
diegolmello: Duplicate of the @children base-Model thread; addressed in 30aee8f39.

app/lib/database/facade/decorators.ts
diegolmello: Duplicate of the @children base-Model thread; addressed in 30aee8f39.

app/lib/database/facade/observe.ts
```diff
@@ -0,0 +1,187 @@
...
+		const emit = () => {
+			if (subscriber.closed) return;
+			const fresh = fetchFn();
+			const shared = structuralShare(prevMap, fresh);
+			prevMap.clear();
+			for (const row of shared) {
+				prevMap.set(row.id, row);
+			}
+			subscriber.next(shared);
+		};
```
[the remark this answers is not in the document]
diegolmello: Low: fetchFn only throws on a DB-level query failure (developer/corrupt-state), no consumer wires subscriber.error, and WMDB didn't surface it here either. Skipping.

app/lib/database/facade/observe.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function sameByColumns<T extends HasId>(prev: T[], next: T[], cols: Set<string>): boolean {
+	if (prev.length !== next.length) return false;
+	for (let i = 0; i < prev.length; i++) {
+		const a = rowData(prev[i]);
+		const b = rowData(next[i]);
+		for (const col of cols) {
+			if (a[col] !== b[col]) return false;
+		}
```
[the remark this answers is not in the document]
diegolmello: Duplicate of the sameByColumns thread; addressed in 30aee8f39.

app/lib/database/facade/decorators.ts
```diff
@@ -0,0 +1,274 @@
...
+		const relationTableName = this._relationTableName;
+		const columnName = this._columnName;
+		return observeRow(_handle, relationTableName, () => {
+			const id = model._getRaw(columnName) as string | null;
+			if (!id) return null;
+			const col = model.collections.get(relationTableName);
+			// eslint-disable-next-line @typescript-eslint/no-explicit-any
+			const rows = (col as any)._fetchSync({ id });
+			return rows.length > 0 ? rows[0] : null;
+		}) as Observable<T | null>;
```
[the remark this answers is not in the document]
diegolmello: Low: nothing observes via @relation (every .observe() consumer observes a Model instance directly), so no live behavior depends on the null-on-removal emission. Skipping.

app/lib/database/facade/observe.ts
diegolmello: Low: no model.observe() consumer relies on completion (no takeUntil/first/take(1) across the codebase); all subscribe and manually unsubscribe in cleanup. Skipping.

app/lib/database/facade/translate.ts
diegolmello: Duplicate of the observeRow-completion thread; low, skipping.

app/lib/database/facade/Model.ts
diegolmello: Duplicate of the _jsonDecoratorCache thread; addressed in 30aee8f39.

app/lib/database/index.ts
```diff
@@ -1,105 +1,67 @@
...
+			serversSchema,
+			serversTableMap,
+			serversModelMap
+		) as unknown as TServerDatabase;
+	};
+
+	setActiveDB = async (database = ''): Promise<void> => {
+		const handle = await openServerDb(database);
+		this.databases.activeDB = new Database(handle, appSchema, appTableMap, appModelMap) as unknown as TAppDatabase;
+	};
```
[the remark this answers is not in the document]
diegolmello: Server switch is serialized through the connect/disconnect saga and openServerDb is handle-cached per dbName, so there's no demonstrated concurrent setActiveDB caller. Speculative — skipping; revisit with a token guard if a concurrent-switch path is ever added.

app/lib/database/facade/Database.ts
diegolmello: Fixed in 30aee8f39: @children now queries through the DB-registered child collection, so model-class mapping (field/date/json getters, batch writes) stays intact — no more bare Model.

app/lib/database/facade/decorators.ts
diegolmello: Fixed in 30aee8f39: _pendingOp is cleared only after the transaction commits, so a rollback no longer leaves earlier models marked clean.

app/lib/database/facade/observe.ts
diegolmello: Fixed in 30aee8f39: observeRow is debounced (16ms) like observeTable, so a batched write no longer fires N synchronous re-fetches per subscription.

app/lib/database/facade/Collection.ts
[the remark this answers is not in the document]
diegolmello: Fixed in 30aee8f39: _fetchSync uses `!== undefined`, so find('') applies the WHERE and rejects instead of full-scanning.

app/lib/database/facade/writer.ts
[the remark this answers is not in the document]
diegolmello: Fixed in 30aee8f39: enqueue settles the caller's promise on a synchronous throw in fn — no more hang on the writer mutex.

app/lib/database/facade/observe.ts
diegolmello: Fixed in 30aee8f39: sameByColumns diffs by row id, not array index — a mid-list insert/delete no longer suppresses or misfires updates.

app/lib/database/facade/observe.ts
diegolmello: Fixed in 30aee8f39: translateWhere case 'on' now throws instead of silently dropping to a full-table scan.

app/lib/database/facade/schema.ts
diegolmello: Fixed in 30aee8f39: removed the dead `|| 0` in the valid-number branch.

app/lib/database/facade/Collection.ts
diegolmello: Fixed in 30aee8f39: _pendingOp='create' is set after fn(model), matching Model.prepareCreate's order.

app/lib/methods/updateMessages.ts
diegolmello: Fixed in 30aee8f39: merged the two facade imports into one.

app/lib/database/facade/Database.ts
diegolmello: Fixed in 30aee8f39: handle's db is typed via the schema generic (zero `as any` remain) and the PK is resolved via getTableColumns(drizzleTable).id instead of a hardcoded `.id` literal.

app/lib/database/facade/Model.ts
diegolmello: Fixed in 30aee8f39: removed the dead _jsonDecoratorCache field (the @json getter never read it).

app/lib/database/facade/Database.ts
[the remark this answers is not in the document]
diegolmello: Duplicate of the _pendingOp-in-transaction thread; addressed in 30aee8f39.

app/lib/database/facade/decorators.ts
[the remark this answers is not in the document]
diegolmello: Duplicate of the @children base-Model thread; addressed in 30aee8f39.

app/lib/database/facade/decorators.ts
diegolmello: Duplicate of the @children base-Model thread; addressed in 30aee8f39.

app/lib/database/facade/observe.ts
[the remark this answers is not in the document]
diegolmello: Low: fetchFn only throws on a DB-level query failure (developer/corrupt-state), no consumer wires subscriber.error, and WMDB didn't surface it here either. Skipping.

app/lib/database/facade/observe.ts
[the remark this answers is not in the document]
diegolmello: Duplicate of the sameByColumns thread; addressed in 30aee8f39.

app/lib/database/facade/decorators.ts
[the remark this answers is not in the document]
diegolmello: Low: nothing observes via @relation (every .observe() consumer observes a Model instance directly), so no live behavior depends on the null-on-removal emission. Skipping.

app/lib/database/facade/observe.ts
diegolmello: Low: no model.observe() consumer relies on completion (no takeUntil/first/take(1) across the codebase); all subscribe and manually unsubscribe in cleanup. Skipping.

app/lib/database/facade/translate.ts
diegolmello: Duplicate of the observeRow-completion thread; low, skipping.

app/lib/database/facade/Model.ts
diegolmello: Duplicate of the _jsonDecoratorCache thread; addressed in 30aee8f39.

app/lib/database/index.ts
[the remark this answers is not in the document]
diegolmello: Server switch is serialized through the connect/disconnect saga and openServerDb is handle-cached per dbName, so there's no demonstrated concurrent setActiveDB caller. Speculative — skipping; revisit with a token guard if a concurrent-switch path is ever added.

ios/Shared/RocketChat/Database.swift
```diff
@@ -70,7 +70,11 @@ class Database {
...
 			return
 		}
-		let path = (groupRoot as NSString).appendingPathComponent(dbName)
+		// New encrypted DBs live in a `SQLite/` subdirectory, isolated from the legacy
+		// plaintext WatermelonDB files at the container root. Must stay in lockstep with
+		// `DB_SUBDIRECTORY` / `resolveDbDirectory()` in the JS driver (connection.ts).
+		let sqliteDir = (groupRoot as NSString).appendingPathComponent("SQLite")
+		let path = (sqliteDir as NSString).appendingPathComponent(dbName)
 
 		guard sqlite3_open(path, &db) == SQLITE_OK else {
```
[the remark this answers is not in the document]
diegolmello: Out of scope here: this is the native NSE/share-extension SQLite path (ios/Shared/RocketChat/Database.swift), which belongs to the encrypted-native-readers PR (#7398), not the JS facade cutover. Tracking the directory-creation concern on that PR.

PR #7408, opened 2026-06-17
feat: port client data into the encrypted database at first boot

diegolmello:
## Proposed changes

Adds a wipe-and-restore migration that runs once at cold boot, before any server data is read or re-auth is evaluated, while the bootsplash is still up. It ports the client-owned data — users, server lock fields, server history, pending messages, subscription and thread drafts, uploads (only those whose file still exists), and frequently-used emojis — from the legacy plaintext WatermelonDB files into the new SQLCipher database, then deletes the legacy files and their WAL/SHM sidecars. Everything else is dropped and resynced from the server.

The orchestrator is a crash-safe resumable state machine: each phase transition is recorded in MMKV before the destructive step that follows, so a crash or kill at any step resumes from the last durable phase. A done marker fast-paths every subsequent boot.

Boot wiring lives in the init saga's `restore`: the servers database opens first (installing the native key store so the migration ports through the real device key), then the migration runs. A migration failure is logged and never blocks boot or logs the user out.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1278

## How to test or reproduce

1. On a build with the legacy plaintext WatermelonDB files present (an upgrade from a pre-encryption build, or seeded `.db`/`.db.db` fixtures), cold-boot the app.
2. The migration runs behind the bootsplash: client data lands in the encrypted DB, the legacy files and their `-wal`/`-shm` sidecars are deleted, the app proceeds past splash, and the session is preserved (no logout).
3. Relaunch — the done marker fast-paths, no re-wipe.
4. Kill-and-relaunch at any phase recovers correctly (resumable state machine).
5. `TZ=UTC pnpm test` — migration, legacy-reader and facade/driver suites pass.

Verified end-to-end on an Android emulator from a fresh build: full detect → port → wipe → done, both legacy fixtures removed, idempotent on relaunch.

## Types of changes

## Checklist

## Further comments

Stacked on `feat/native-1277-facade-cutover` (#7407); review/merge that first. Part of the WatermelonDB → encrypted SQLite cutover (NATIVE-1272).

## Summary by CodeRabbit

* **New Features**
  * Automatic database upgrade process on app launch with built-in recovery from interruptions
  * Secure encrypted database format for improved data protection

* **Tests**
  * Added comprehensive regression and integration tests for database upgrade logic and legacy data handling


app/lib/database/migration/orchestrator.ts
```diff
@@ -0,0 +1,232 @@
...
+
+			await portUsers(users, newSqlite);
+			await portServerLockFields(lockFields, newSqlite);
+			await portServersHistory(history, newSqlite);
+
+			// Capture server URLs before closing the legacy DB; we need them for porting_active
+			const serverUrls = await readLegacyServerUrls(legacyDb);
+
+			// Record servers as pending in state so porting_active can resume per-server
+			setPhase('porting_active');
```
diegolmello: 🟡 [medium] **TOCTOU window between `setPhase('porting_active')` and `markServer` calls**

`setPhase('porting_active')` is written first with `servers: {}`. If the process is killed before any `markServer(url, 'pending')` call completes, the resumed run enters `porting_active` with an empty server list, the for-loop body never executes, and the orchestrator falls straight through to `wiping` — silently destroying all per-server local data (drafts, pending messages, frequently-used emojis).

Fix: write the phase transition and the per-server `'pending'` entries in a single atomic state write:

```ts
// Replace setPhase + markServer loop with one writeState call:
writeState({
  ...(readState() ?? { schema: 1, startedAt: getNowMs() }),
  phase: 'porting_active',
  servers: Object.fromEntries(serverUrls.map(url => [url, 'pending'])),
  updatedAt: getNowMs()
});
state = readState();
```

This eliminates the `phase=porting_active, servers={}` window entirely.

app/lib/database/migration/orchestrator.ts
```diff
@@ -0,0 +1,232 @@
...
+			if (status === 'ported' || status === 'wiped') continue;
+
+			const dbName = deriveServerDbName(url);
+			if (!legacyFileExists(dbName)) {
+				// Legacy per-server DB missing — mark ported so wiping doesn't try to unlink it
+				markServer(url, 'ported');
+				continue;
+			}
+
+			const legacyDb = await openLegacy(dbName);
```
diegolmello, line 169: 🟢 [low] **Uncaught `openLegacy` throw aborts the entire migration on a corrupt legacy file**

`openLegacy(dbName)` is called before the `try` block. If it throws (corrupt but present file), the exception propagates out of `runMigrationIfNeeded`, `isMigrationDone()` stays false, and every subsequent boot re-enters and re-throws — the migration never completes.

Fix: wrap each server's work in a per-server try/catch and mark it skipped on failure so the loop continues:

```ts
let legacyDb;
try {
  legacyDb = await openLegacy(dbName);
} catch (e) {
  Log.w('[migration] cannot open legacy DB for ' + url + ', skipping', e);
  markServer(url, 'ported');
  continue;
}
```

app/lib/database/migration/state.ts
```diff
@@ -0,0 +1,126 @@
...
+	// where the state was set to 'done' but the boolean key wasn't written yet.
+	const state = readState();
+	return state?.phase === 'done' || state?.phase === 'skipped';
+}
+
+/** Returns the full migration state or null if never started. */
+export function readState(): MigrationState | null {
+	const raw = userPreferences.getMap(KEY_STATE);
+	if (!raw || typeof raw !== 'object') return null;
+	return raw as MigrationState;
```
diegolmello, line 85: 🟢 [low] **`readState()` performs an unchecked cast — corrupt or schema-mismatched state is silently misrouted**

`return raw as MigrationState` trusts the MMKV entry unconditionally. A corrupted value or a state written by a future schema version will be misrouted (e.g. a missing `phase` field makes `state?.phase` undefined, causing the orchestrator to run as if in `detect` again, potentially re-entering a completed migration).

Fix: validate the schema version before returning:

```ts
export function readState(): MigrationState | null {
  const raw = userPreferences.getMap(KEY_STATE);
  if (!raw || typeof raw !== 'object') return null;
  // Reject entries from unknown schema versions to avoid misrouting
  if ((raw as { schema?: unknown }).schema !== 1) return null;
  return raw as MigrationState;
}
```

diegolmello (commented):
## Review: wipe-and-restore migration

**Boot trigger verified fixed** 🎉 — `initServers` now runs as the first `yield call` inside the `takeLatest(APP.INIT, restore)` handler, so the previously-observed dropped-`APP.INIT` problem (root-saga prefix delaying the listener registration) does not occur here; migration fires on every cold boot and the MMKV fast-path keeps subsequent boots O(1).

The ported set is correctly scoped and the wipe ordering is right. Main concern is the non-atomic phase transition (inline) — a crash in that window silently destroys the local-only set it is meant to preserve.

diegolmello (commented):
**Structural review (NATIVE-1278).** The phase state machine is cleanly typed and the boot trigger is correctly wired (moving `database.initServers` inside restore unblocks `takeLatest(APP.INIT)`). No file over ~230 lines. Two structural bugs are inline — fix the phase-advance ordering first; it can permanently lose per-server data on a crash. The rest is cleanup.

app/lib/database/migration/orchestrator.ts
diegolmello: **bug: high — phase advances to `porting_active` before the server map is persisted.** `setPhase('porting_active')` is written here, then the `markServer(url, 'pending')` loop runs at ~152. A crash in that window means the next boot skips the `porting_servers` block (phase already advanced), runs `porting_active` over an empty `servers` map (zero iterations), then wipes without unlinking per-server legacy files — permanent loss of pending messages/drafts/uploads plus orphaned `.db.db` files. Populate the server map while still in `porting_servers`, then advance the phase, so the marker flips atomically with the map already written. The crash-resume test seeds a populated map and doesn't cover the empty-map resume path that exposes this.

app/lib/database/migration/state.ts
diegolmello, line 85: **bug: med — `readState()` casts MMKV output to `MigrationState` unchecked.** A stale/corrupt stored object (unknown phase, wrong `schema`) passes the cast silently; every `state?.phase === ...` block then misses and `isMigrationDone()` stays false → migration spins forever doing nothing. The `schema: 1` field implies versioned validation was intended — add an `isValidState` guard (check `schema === 1` and `phase ∈` the known set), discard and return null otherwise.

app/lib/database/migration/port.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+export async function portPendingMessages(legacyRows: LegacyRow[], newSqlite: SQLiteDatabase): Promise<void> {
+	await insertRows('messages', legacyRows, newSqlite);
+}
+
+/**
+ * Ports draft_message for subscriptions. Uses INSERT OR IGNORE + UPDATE so that
+ * only the draft column is touched if the subscription row doesn't exist yet.
+ */
+export async function portSubscriptionDrafts(
```
diegolmello: **slop: moderate — `portSubscriptionDrafts` / `portThreadDrafts` are identical except for the table name.** Both do INSERT OR IGNORE + UPDATE draft_message. Factor into one `portDraftColumn(tableName, rows, db)` helper.

app/lib/database/migration/legacyReader.ts
```diff
@@ -0,0 +1,244 @@
...
+
+/**
+ * Mirrors the legacy WMDB on-disk per-server filename: strip scheme, replace slashes with dots,
+ * append the platform-aware legacy suffix (`.db` on iOS, `.db.db` on Android).
+ *
+ * This is the LEGACY-file address — distinct from `connection.deriveServerDbName`, which names the
+ * NEW encrypted files with a single `.db` on both platforms. The orchestrator must use this one for
+ * every legacy read/exists/wipe, and the connection one only for opening the new DB.
+ */
+export function deriveServerDbName(serverUrl: string): string {
```
diegolmello: **slop: moderate — `deriveServerDbName` here collides with `connection.deriveServerDbName`.** Same name, different results (this one is the legacy double-suffix path; connection's is the new single-`.db` path). Importing the wrong one is silent and would address the wrong file. Rename this to `deriveLegacyServerDbName`.

app/lib/database/migration/orchestrator.ts
```diff
@@ -0,0 +1,232 @@
...
+ * Runs the wipe-and-restore migration if needed.
+ * Safe to call on every app boot — the MMKV fast-path makes subsequent calls O(1).
+ */
+export async function runMigrationIfNeeded(): Promise<void> {
+	// Fast path: already done (the overwhelming majority of boots after first upgrade)
+	if (isMigrationDone()) return;
+
+	// Resume from the recorded phase, or start fresh
+	let state = readState();
+	const phase = state?.phase ?? 'detect';
```
diegolmello: **slop: moderate — inconsistent phase dispatch.** The first block reads a local `phase` var; every later block reads `state?.phase` through an optional chain that silently skips a phase if `state` were ever null. A single `switch (state.phase)` (or a phase→handler map) over a non-null state makes the control flow explicit and uniform.

app/lib/database/migration/legacyReader.ts
```diff
@@ -0,0 +1,244 @@
...
+// ---------------------------------------------------------------------------
+// File existence helpers
+// ---------------------------------------------------------------------------
+
+/** Injectable file-existence check — real impl uses expo-file-system File; tests mock this. */
+export let fileExists: (path: string) => boolean = path => {
+	try {
+		// expo-file-system v19 exports File from the top-level package (not /next, which was SDK 52 preview)
+		// eslint-disable-next-line @typescript-eslint/no-var-requires
+		const { File } = require('expo-file-system') as { File: new (path: string) => { exists: boolean } };
```
diegolmello: **slop: moderate — `require('expo-file-system')` on every call.** `fileExists` lazy-requires per upload row; the orchestrator already top-imports `File` from expo-file-system. Use a normal top-level import to match the codebase.

app/lib/database/migration/legacyReader.ts
```diff
@@ -0,0 +1,244 @@
...
+	if (!LEGACY_DIR) return false;
+	return fileExists(`${LEGACY_DIR}/${dbName}`);
+}
+
+/**
+ * Returns the list of legacy server DB names that are actually present on disk.
+ * `serverUrls` comes from the legacy servers DB (read by the orchestrator first).
+ * The servers DB itself (`default.db`) is NOT included — check separately via legacyFileExists.
+ */
+export function enumeratePresentServerDbs(serverUrls: string[]): string[] {
```
diegolmello: **slop: nit — `enumeratePresentServerDbs` is exported but never called.** The orchestrator iterates `state.servers` directly. Remove it or use it.

app/lib/database/migration/legacyReader.ts
```diff
@@ -0,0 +1,244 @@
...
+/** All rows from legacy servers DB `users` table. */
+export function readLegacyUsers(db: SQLiteDatabase): Promise<Record<string, unknown>[]> {
+	return db.getAllAsync('SELECT * FROM users;') as Promise<Record<string, unknown>[]>;
+}
+
+/**
+ * Lock fields from legacy servers DB `servers` table.
+ * Only the fields that are user-authored / device-local are ported; everything else resyncs.
+ */
+export function readLegacyServerLockFields(db: SQLiteDatabase): Promise<
```
diegolmello: **slop: nit — the 5-field lock-field shape is inlined twice** (return type and the `as Promise<{...}[]>` cast). Extract a named type.

app/lib/database/migration/orchestrator.ts
diegolmello: Fixed in 4b5fe40ae: phase→porting_active and per-server 'pending' now happen in one atomic startPortingActive() write, closing the empty-map crash window that could finalize migration with no servers ported (permanent loss of pending messages/drafts/uploads/frequent-emojis + orphaned legacy files).

app/lib/database/migration/state.ts
diegolmello: Fixed in 4b5fe40ae: readState() validates schema===1 and phase in the known set before casting; a corrupt or future-schema entry returns null → restart from detect (idempotent) instead of silently misrouting and spinning on every boot.

app/lib/database/migration/port.ts
diegolmello: Fixed in 4b5fe40ae: factored the shared INSERT OR IGNORE + UPDATE draft body into portDraftColumn(); portSubscriptionDrafts/portThreadDrafts kept as thin wrappers (tableName is a hardcoded literal).

app/lib/database/migration/legacyReader.ts
diegolmello: Fixed in 4b5fe40ae: renamed the legacy reader's deriveServerDbName → deriveLegacyServerDbName to disambiguate from connection.deriveServerDbName (legacy plaintext suffix vs new encrypted .db); call sites updated.

app/lib/database/migration/legacyReader.ts
diegolmello: Fixed in 4b5fe40ae: deleted the unused enumeratePresentServerDbs (the orchestrator iterates state.servers directly), which also moots the return-value mismatch.

app/lib/database/migration/legacyReader.ts
diegolmello: Fixed in 4b5fe40ae: File is now top-imported from expo-file-system; the per-call lazy require + eslint-disable + stale SDK comment are gone.

app/lib/database/migration/legacyReader.ts
diegolmello: Fixed in 4b5fe40ae: extracted a named LegacyServerLockFields type so the lock-field row shape isn't inlined twice.

app/lib/database/migration/orchestrator.ts
diegolmello: Fixed in 4b5fe40ae: unified the phase dispatch on state?.phase (dropped the stray local var); kept the sequential cascade so a fresh run still completes detect→done in a single pass.

app/lib/database/migration/orchestrator.ts
diegolmello: Same phase-before-persist window as the high-severity thread on this line — fixed there with a single atomic state write (advance phase + mark all servers pending together). Resolving as a duplicate.

app/lib/database/migration/orchestrator.ts
```diff
@@ -0,0 +1,232 @@
...
+			// Record servers as pending in state so porting_active can resume per-server
+			setPhase('porting_active');
+			state = readState();
+			// Initialise per-server entries if not already set (fresh run from detect)
+			if (state && Object.keys(state.servers).length === 0) {
+				for (const url of serverUrls) {
+					markServer(url, 'pending');
+				}
+				state = readState();
+			}
```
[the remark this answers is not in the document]
diegolmello: Duplicate of the existing review thread on this transition — fixed by persisting the full per-server workset and the phase in one atomic write. Resolving as a duplicate.

app/lib/database/migration/legacyReader.ts
```diff
@@ -0,0 +1,244 @@
...
+ * Returns the list of legacy server DB names that are actually present on disk.
+ * `serverUrls` comes from the legacy servers DB (read by the orchestrator first).
+ * The servers DB itself (`default.db`) is NOT included — check separately via legacyFileExists.
+ */
+export function enumeratePresentServerDbs(serverUrls: string[]): string[] {
+	return serverUrls.filter(url => {
+		const dbName = deriveServerDbName(url);
+		return legacyFileExists(dbName);
+	});
+}
```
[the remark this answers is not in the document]
diegolmello: enumeratePresentServerDbs is unused; removed it rather than fixing its return value, which moots this finding. Resolving.

app/lib/database/migration/state.ts
diegolmello: Same unchecked-cast concern as the other thread on this line — added schema+phase validation to readState there. Resolving as a duplicate.

app/lib/database/migration/orchestrator.ts
diegolmello: A corrupt per-server file would make openLegacy throw, looping the migration (caught by restore's try/catch, so boot is fine) rather than losing data. Low and edge-case; leaving for a follow-up — a per-server try/catch that marks the server skipped would resolve it cleanly when prioritized.

PR #7415, opened 2026-06-19
refactor: migrate InsideStack to React Navigation static config

diegolmello:
## Summary

- Converts all 8 nested navigators in `InsideStack` (Drawer, Chats, Profile, Settings, AdminPanel, Accessibility, NewMessage, E2ESaveYourPassword, E2EEnterYourPassword) from the dynamic JSX API to React Navigation 7's static config API.
- Class components and function components that take `navigation` as a prop are wrapped with `withNavigation()` so `useNavigation()` injects it at the registration site.
- Dynamic navigator props (`themedHeader`, `drawerContent`, `overlayColor`) go in `.with()` callbacks; static `screenOptions` stay in the config object.
- `TNavigation` screens (`PickerView`, `ForwardLivechatView`, `AttachmentView`) keep precise param-type annotations so `StaticParamList` preserves the original types.
- `Record<string, any>` (not bare `any`) annotations on param-bearing screens so `StaticParamList` infers a concrete type rather than `unknown`/`undefined`.
- `types.ts` keeps hand-written param lists for `ChatsStackParamList` and `SettingsStackParamList` to preserve cross-stack navigation entries (e.g. `ModalStackNavigator` for the tablet/MasterDetail modal stack) and specific route param types that view callback parameters rely on for type inference.
- Default export via `InsideStack.getComponent()` — `AppContainer.tsx` is untouched.

Stacks on #7414 (MasterDetailStack static config). Base should retarget to `develop` once #7414 merges.

## Issues

https://rocketchat.atlassian.net/browse/NATIVE-1303

## How to test

- Open the app on phone — navigation within all stacks (Chats, Profile, Settings, Admin, Accessibility, NewMessage, E2E flows) should work as before.
- On tablet — Drawer navigation and MasterDetail layout should be unaffected.
- E2E password save and enter flows should navigate correctly.
- Attachment, Status, ShareView, ModalBlockView, CallView screens should open from InsideStack.

https://claude.ai/code/session_012YMKHjug3ZGEibZ1BAGcTx

## Summary by CodeRabbit

* **Refactor**
  * Reworked the app’s navigation structure to provide a more consistent, themed header experience across major sections.
  * Improved routing and screen registration for Chats, including additional chat/discussion/channel flows and a Jitsi meeting view with appropriate header visibility and transitions.
  * Refreshed the drawer navigation setup to centralize sidebar behavior and options.
* **Documentation**
  * Updated inline guidance for navigation typing and cross-screen routing expectations.

PR #7416, opened 2026-06-19
refactor: migrate AppContainer root to React Navigation static config

diegolmello:
## Proposed changes

Converts the app root from a dynamic `NavigationContainer` + `Stack.Navigator` to a fully static React Navigation 7 config.

**Root navigator** — `createNativeStackNavigator({ groups })` with six conditional groups, each gated by a `useSelector` hook that reads `state.app.root` (and `state.app.isMasterDetail` for the master-detail split). `createStaticNavigation` converts the config into `AppNavigation`, a drop-in replacement for `NavigationContainer` that accepts `ref`, `onReady`, `onStateChange`, and `theme`.

**Leaf stack bridges removed** — all four `.getComponent()` calls are gone. Each leaf stack now exports its navigator config object directly so the root can nest it as a static screen:
- `OutsideStack.tsx`
- `InsideStack.tsx`
- `ShareExtensionStack.tsx`
- `MasterDetailStack/index.tsx`

**SetUsernameStack** — the inline dynamic navigator is folded into the root static config as a single-screen nested navigator.

**Theme** — `navigationBarColor` (theme-dependent) moves into a `.with()` wrapper on the root; `headerShown: false` and `animation: 'none'` stay in the static config. `navigationTheme(theme)` is passed to `AppNavigation`'s `theme` prop unchanged.

**Preserved unchanged**: screen tracking (`setCurrentScreen` / `routeNameRef` via `onReady` / `onStateChange` and a `useEffect` on root changes), `MediaCallHeader` placement, `emitter.emit('navigationReady')`, and the imperative deep-linking path through `navigationRef`.

This is the final slice of the React Navigation 7 static-config migration. Stacked on #7415; base retargets to `develop` once #7415 merges.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1305

## How to test or reproduce

1. Build and launch the app (iOS or Android).
2. Verify the login flow: loading screen → outside (login/register) → inside (rooms list).
3. On a tablet or with master-detail enabled, verify the split-view layout loads correctly.
4. Verify the share extension launches without a crash.
5. Verify deep links (e.g. `rocketchat://` URLs) still navigate to the correct screen.
6. Verify screen analytics tracking fires on navigation (check logs for `setCurrentScreen`).

## Screenshots

N/A — navigation structure change only; no visual changes.

## Types of changes

## Checklist

## Further comments

No test files exist for `AppContainer` or `app/stacks` (confirmed via `pnpm test --testPathPattern='AppContainer'` and `--testPathPattern='app/stacks'`). Lint: 0 errors, 175 warnings (matches the pre-change baseline).

https://claude.ai/code/session_01Qh9P8Bbp6V7PFpCY2X5on3

## Summary by CodeRabbit

* **Refactor**
  * Updated the app’s root navigation to a static, hook-driven setup, improving consistency in how screens are displayed.
  * Streamlined how stack and modal navigation modules are exposed, simplifying the navigation flow.
* **Bug Fixes**
  * Improved route-change tracking so screen analytics and current-screen state remain aligned with what the user is actively viewing.


app/AppContainer.tsx
```diff
@@ -45,17 +79,11 @@ const App = memo(({ root, isMasterDetail }: { root: string; isMasterDetail: bool
...
-	}
-
-	const navTheme = navigationTheme(theme);
-
 	return (
 		<>
 			<MediaCallHeader />
-			<NavigationContainer
-				theme={navTheme}
+			<AppNavigation
```
diegolmello, line 88: Critical (crash on launch): `<AppNavigation>` now renders unconditionally, but the redux `root` is `undefined` on first render (`app/reducers/app.ts` initialState; `appInit()` runs after awaits in `Root.init()`). With `root` undefined all six group `if` hooks return false, the static config emits all-null children, and the native-stack navigator throws "Couldn't find any screens for the navigator" (`@react-navigation/core` useNavigationBuilder L392-396, after `React.Children.toArray` drops the nulls). The previous dynamic root guarded this exact window with `if (!root) return null;` (removed at L73-80 area). Restore that early-return, or add a default non-`if` fallback group so a screen is always registered when `root` is undefined.

app/AppContainer.tsx
```diff
@@ -1,40 +1,74 @@
...
+const RootNavigator = createNativeStackNavigator({
+	screenOptions: { headerShown: false, animation: 'none' },
+	groups: {
+		Loading: { if: useIsLoading, screens: { AuthLoading: AuthLoadingView } },
+		Outside: { if: useIsOutside, screens: { OutsideStack } },
+		MasterDetail: { if: useIsMasterDetail, screens: { MasterDetailStack } },
+		Inside: { if: useIsInside, screens: { InsideStack } },
+		SetUsername: { if: useIsSetUsername, screens: { SetUsernameStack } },
+		ShareExtension: { if: useIsShareExtension, screens: { ShareExtensionStack } }
+	}
```
diegolmello, line 58: Root cause context for the cold-start crash: every group is `if`-gated, so the union of conditions does not cover `root === undefined` (the initial state). All 6 RootEnum values are covered and mutually exclusive — only the undefined/cold-start state falls through. A default group with no `if` (e.g. an AuthLoading/splash screen) would close the gap and also make the navigator resilient to any future unhandled root value.

app/AppContainer.tsx
```diff
@@ -1,40 +1,74 @@
...
 import { ThemeContext } from './theme';
 import { setCurrentScreen } from './lib/methods/helpers/log';
 import { themes } from './lib/constants/colors';
 import { emitter } from './lib/methods/helpers';
 import MediaCallHeader from './containers/MediaCallHeader/MediaCallHeader';
 
-const createStackNavigator = createNativeStackNavigator;
+// ─── Conditional-group hooks ──────────────────────────────────────────────────
+
+const useIsLoading = () =>
```
diegolmello, line 22: Nit (slop, moderate): `(state: any)` is repeated across all six selector hooks and the `root` selector. The old `mapStateToProps` also used `any`, so this isn't a regression, but typing these against the app state (e.g. `IApplicationState`) would restore the type-checking the redux `state.app.root`/`isMasterDetail` reads lost in the migration.

diegolmello (commented):
Review of the AppContainer static-config migration (diff against `native-1303-inside-stack-static-config`).

**One blocking bug: cold-start crash.** Every root group is `if`-gated, but the initial redux state is `root: undefined` (`app/reducers/app.ts` `initialState`). The migration removed the `if (!root) return null;` guard, so `<AppNavigation>` now mounts unconditionally on the very first render — before any saga sets `root`. With `root` undefined, all six `if` hooks return false, so the static config passes all-null children to the native-stack `Navigator`; `React.Children.toArray` strips the nulls, `useNavigationBuilder` ends up with zero routes and throws "Couldn't find any screens for the navigator." (`@react-navigation/core` useNavigationBuilder L392-396). `AppContainer` is rendered unconditionally in `app/index.tsx`, and `appInit()` (which triggers the `restore` saga that first sets `root`) is dispatched only after several awaits in `Root.init()`, so the undefined window is reachable at every cold start. The old dynamic root never mounted the container while `!root`. Must fix before merge — e.g. restore an early `if (!root) return null;` in `AppContainer`, or add a default (non-`if`) fallback group so at least one screen is always registered.

**Everything else checks out:**
- Default-export type flip is safe: `AppContainer` is the ONLY importer of the default export of OutsideStack/InsideStack/MasterDetailStack/ShareExtensionStack. Other references are type-only imports (`MasterDetailStack/types`, `ShareInsideStackParamList`). No stray component consumer, no leftover `.getComponent()` anywhere.
- Branch parity is exact and mutually exclusive: all 6 RootEnum values map 1:1 to the old conditions; the master-detail vs inside split is preserved. The only uncovered state is `undefined` (the bug above).
- navigationRef / onReady / onStateChange / routeNameRef / `theme={navigationTheme(theme)}` all preserved on the `createStaticNavigation` Navigation component; shared `navigationRef` from `appNavigation` unchanged.
- MediaCallHeader still a sibling outside the navigator.
- Theme split is correct: static `screenOptions` ({headerShown, animation}) stay in the base config; theme-dependent `navigationBarColor` is read at render in the `.with()` wrapper and merged by the static NavigatorComponent (nothing frozen at module load).
- No `linking` config added.
- `connect`+`memo` → `useSelector` is sound; memo removal is immaterial here (the component takes no props and already re-renders with its parent).
- `pnpm lint`: 0 errors, 175 warnings — identical to baseline, no new warnings.

OtavioStasiak (approved):
LGTM

PR #7420, opened 2026-06-19
perf: memoize ThemeContext provider value to decouple theme consumers from dimension changes

diegolmello:
## Proposed changes

The root `App` component built the `ThemeContext` provider value as an inline object literal, so its identity was recreated on every `App` render. `App` state also holds window dimensions (`width/height/scale/fontScale`), updated on `Dimensions.change`. So every dimension change — device rotation, multi-window/fold resize — recreated the provider value and invalidated `ThemeContext` for all theme consumers (~315 `useTheme()` + 15 `withTheme()`) even though the theme never changed.

This extracts a small `ThemeContextProvider` wrapper that memoizes the value with `useMemo` keyed on `theme` / `themePreferences` / `setTheme`. The provider value identity now changes only when the theme actually changes. `setTheme` (a stable class arrow property) and `colors[theme]` (a module-level constant) were already stable inputs — only the wrapper object identity was the problem.

`DimensionsContext` is intentionally left untouched: its value must change identity on resize, so memoizing it would be wrong.

### Measured effect

React DevTools profiler, Android emulator (Pixel API 34), device rotation:

| | inline value (before) | memoized value (after) |
|---|---|---|
| `ThemeContext` context invalidations | 1405 | 923 (−482) |

The memo removes ~482 redundant `ThemeContext` invalidations that the inline value triggered on a non-theme event. This is a correctness/hygiene fix, matching the issue's Low–Medium severity.

Keyboard show/hide was checked on the same emulator and did **not** fire `Dimensions.change` (no `App` re-render with the soft IME shown), so the keyboard path is not affected here.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1286

## How to test or reproduce

1. Run the app on an Android device/emulator.
2. Capture a React DevTools profiler trace while rotating the device.
3. Before this change: rotation invalidates `ThemeContext` for all theme consumers. After: `ThemeContext` is invalidated only on an actual theme change.

Automated: `app/containers/ThemeContextProvider.test.tsx` asserts the context value reference stays stable when the provider re-renders with unchanged theme props, and changes when `theme` changes (the test fails if the memo is replaced by an inline literal).

## Screenshots

Side-by-side device rotation, before (inline value) vs after (memoized value). User-visible behavior is identical; caption shows the measured `ThemeContext` invalidation drop (1405 → 923).

## Types of changes

## Checklist

## Further comments

Considered caching the value in a class instance field instead of extracting a component; the function-component + `useMemo` wrapper is the more idiomatic React pattern and is unit-testable in isolation. Splitting the context or removing prop-drilling of the theme was rejected — the theme data references are already stable, so neither would help; only the wrapper identity needed fixing.

https://claude.ai/code/session_01QBpRJDfifKJnkfMVd994Fg

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for theme context stability and updates across re-renders.

* **Refactor**
  * Extracted theme management logic into a dedicated, reusable provider component with optimized rendering performance.


diegolmello (commented):
Reviewed independently. The change is correct, minimal, and well-scoped.

Verified:
- **Memo deps are complete and minimal.** `[theme, themePreferences, setTheme]` covers everything the value depends on; `colors[theme]` is derived from `theme` so it correctly is not a separate dep.
- **No churn on dimension changes.** `setDimensions` calls `setState({ width, height, scale, fontScale })` and never touches `themePreferences`, so its reference is stable across resize/rotation re-renders — the memo holds and consumers don't re-render. `themePreferences` identity only changes inside `setTheme` (object spread), which is the intended invalidation.
- **`setTheme` is stable** (class arrow property), and `colors` is a module-level constant.
- **Test genuinely catches the regression.** `ParentWithCounter` re-renders the provider with unchanged props while the component is not memoized, so it exercises the `useMemo` dep-check; replacing the memo with an inline literal makes capture[1]/[2] differ and the test fails. The second test confirms invalidation on theme change. Both pass locally; lint is clean.
- **Scope discipline is good.** `DimensionsContext` correctly left untouched, the now-unused `colors` import was removed from `index.tsx`.

One non-blocking nit on import placement below. No blocking issues — looks good to merge.

app/index.tsx
```diff
@@ -40,7 +40,8 @@ import {
 } from './lib/services/voip/MediaCallEvents';
 import store from './lib/store';
 import { initStore } from './lib/store/auxStore';
-import { type TSupportedThemes, ThemeContext } from './theme';
+import { type TSupportedThemes } from './theme';
+import ThemeContextProvider from './containers/ThemeContextProvider';
```
diegolmello: nit (slop, moderate): this `./containers/ThemeContextProvider` import is separated from the other `./containers/*` imports clustered at lines 13-17. Moving it up next to them keeps the import grouping consistent (the pre-existing `StatusBar` import just below at line 47 is similarly stray, but this adds to the drift). Cosmetic only — ESLint does not flag it.

app/containers/ThemeContextProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { type IThemeContextProps, type TSupportedThemes, ThemeContext } from '../theme';
+
+// Memoized provider so dimension-driven App re-renders don't propagate to theme consumers.
+const ThemeContextProvider = ({
+	theme,
+	themePreferences,
+	setTheme,
+	children
+}: {
+	theme: TSupportedThemes;
```
diegolmello, line 14: Observation (not a defect): `setTheme: IThemeContextProps['setTheme']` resolves to `((newTheme?: {}) => void) | undefined` because the field is optional on the context interface. `App` always passes a defined `this.setTheme`, so the wrapper prop is looser than the call site needs. Harmless given the value is just forwarded into the context (which also types it optional). Fine to leave as-is; flagging only so the slight type widening is intentional.

app/index.tsx
diegolmello: Fixed in 4a7afab5a — moved both into the alphabetized `./containers/*` block.

OtavioStasiak (approved):
Approved

PR #7422, opened 2026-06-19
refactor: derive isMasterDetail per-component and remove Dimensions provider

diegolmello:
## Proposed changes

The app tracked the master-detail layout flag through a root pipeline: `Dimensions.addEventListener` in `app/index.tsx` recomputed `isMasterDetail` on every window change, wrote it to `state.app.isMasterDetail`, and fed `width/height/scale/fontScale` into a `DimensionsContext` consumed app-wide. Every dimension change triggered a Redux dispatch plus a context update that re-rendered unrelated subtrees.

This removes that pipeline and derives the flag where it's used:

- New `useMasterDetail()` hook computes `isTablet && width > MIN_WIDTH_MASTER_DETAIL_LAYOUT` from `useWindowDimensions()`.
- Function-component readers call the hook directly.
- Class components receive the flag via a `withMasterDetail` HOC in their `connect` chain (and the three views that needed raw dimensions move to a hook-backed `withDimensions`).
- Non-React callers (sagas) read a synchronous `getIsMasterDetail()` snapshot from `Dimensions.get('window')`, since hooks can't run there.
- Deleted `state.app.isMasterDetail` (field, action, reducer case), the root `Dimensions` listener, and `app/dimensions.tsx`.

The flag now reacts to window resize directly instead of a debounced Redux round-trip, and dimension changes no longer fan out through a global context.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-886

## How to test or reproduce

- Tablet (or split view): rotate / resize across the 700px master-detail threshold and confirm the layout switches between `MasterDetailStack` and `InsideStack` as before.
- Phone: confirm single-stack layout is unchanged.
- Exercise screens that branch on master-detail (RoomView, RoomActionsView, RoomsListView, ProfileView, directory/new-message flows) and confirm navigation targets the correct pane.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

`IBaseScreen.isMasterDetail` is kept — it's the prop the HOC injects into class screens, distinct from the removed Redux field.

## Summary by CodeRabbit

* **Refactor**
  * Switched master/detail mode to a shared, device-width–based evaluation via a hook, with optional HOC injection across headers, views, notifications, and navigation flows.
  * Removed master/detail from Redux state, related action types, and updated saga navigation logic to use the shared getter.
  * Removed the old dimensions context, and introduced a hook-based bridge for class-style components.
  * Simplified some app components by moving away from Redux `connect` where applicable.
* **Bug Fixes**
  * Prevented stale master/detail decisions by eliminating Redux-driven updates.
* **Tests**
  * Updated mocks and removed reducer coverage tied to the removed master/detail action.


diegolmello:
## Review summary

Reviewed the full diff (≈50 readers migrated, 2 HOCs added, root pipeline + `app/dimensions.tsx` removed). `tsc --noEmit` clean, ESLint clean (only pre-existing React-hooks/complexity warnings, none introduced here), `TZ=UTC pnpm test` 1615/1615 green.

Two correctness issues were found and fixed while reviewing:

**1. Refs were swallowed by the new HOCs.** `withMasterDetail`/`withDimensions` started as plain function components, so a parent's ref never reached the wrapped component. `RoomView` holds refs to `MessageActions` and `JoinCode` and calls imperative methods on them, so those refs would have been `null`. Fixed by making both HOCs use `forwardRef` and pass the ref through (and `hoistNonReactStatics` to preserve `navigationOptions`).

**2. Five sagas still read the removed Redux field.** `createChannel`, `deepLinking`, `login`, `room`, and `messages` read `state.app.isMasterDetail`, which now returns `undefined` — master-detail navigation (`goRoom`/`navigateToRoom`/`popToTop`) would silently get a falsy value. Sagas can't call the hook, so they now read a synchronous `getIsMasterDetail()` snapshot from `Dimensions.get('window')`. This is behaviorally equivalent to (slightly fresher than) the old debounced Redux value.

### Nit
No dedicated unit test for `useMasterDetail`/the HOCs. Behavior is covered indirectly by existing view and reducer tests; a focused test would lock the threshold formula down but isn't blocking.

Otherwise the migration is consistent and the refactor preserves behavior. LGTM.

app/containers/InAppNotification/NotifierComponent.tsx
```diff
@@ -130,8 +131,6 @@ const NotifierComponent = memo(({ notification, isMasterDetail }: INotifierCompo
 	);
 });
 
-const mapStateToProps = (state: IApplicationState) => ({
-	isMasterDetail: state.app.isMasterDetail
-});
+const mapStateToProps = () => ({});
```
diegolmello: Remove

app/views/ProfileView/index.test.tsx
```diff
@@ -61,7 +61,7 @@ const user = {
 
 const buildState = () => ({
 	login: { user },
-	app: { isMasterDetail: false },
+	app: {},
```
diegolmello: Remove it

app/views/UserNotificationPreferencesView/index.tsx
```diff
@@ -29,10 +30,10 @@ const UserNotificationPreferencesView = () => {
 	const [loading, setLoading] = useState(true);
 
 	const navigation = useNavigation<TNavigation>();
-	const { userId, isMasterDetail } = useAppSelector(state => ({
-		userId: getUserSelector(state).id,
-		isMasterDetail: state.app.isMasterDetail
+	const { userId } = useAppSelector(state => ({
+		userId: getUserSelector(state).id
 	}));
```
diegolmello:
```suggestion
	const userId = useAppSelector(state => getUserSelector(state).id);
```

app/views/CloseLivechatView.tsx
```diff
@@ -44,10 +45,10 @@ const CloseLivechatView = () => {
 
 	const { colors } = useTheme();
 
-	const { isMasterDetail, livechatRequestComment } = useAppSelector(state => ({
-		isMasterDetail: state.app.isMasterDetail,
+	const { livechatRequestComment } = useAppSelector(state => ({
 		livechatRequestComment: state.settings.Livechat_request_comment_when_closing_conversation as boolean
 	}));
```
diegolmello:
```suggestion
	const livechatRequestComment = useAppSelector(state => 
state.settings.Livechat_request_comment_when_closing_conversation as boolean
);
```

app/views/SelectListView.tsx
```diff
@@ -208,8 +208,6 @@ class SelectListView extends Component<ISelectListViewProps, ISelectListViewStat
 	}
 }
 
-const mapStateToProps = (state: IApplicationState) => ({
-	isMasterDetail: state.app.isMasterDetail
-});
+const mapStateToProps = () => ({});
```
diegolmello: Remove

app/AppContainer.tsx
```diff
@@ -88,8 +90,7 @@ const App = memo(({ root, isMasterDetail }: { root: string; isMasterDetail: bool
 	);
 });
 const mapStateToProps = (state: any) => ({
-	root: state.app.root,
-	isMasterDetail: state.app.isMasterDetail
+	root: state.app.root
 });
```
diegolmello: Migrate to useAppSelector, since we're here.

OtavioStasiak (approved):
LGTM

PR #7424, opened 2026-06-20
fix: insert multiline code block as a fenced block on its own lines

Shevilll:
## Proposed changes

Tapping the **multiline code-block** formatting button produced a broken result: it wrapped the selection inline, e.g. typing nothing and tapping the button inserted six backticks with the cursor stuck in the middle of them on a single line (`` ```|``` ``). Fenced code blocks are only parsed by the renderer when the ` ``` ` markers sit **alone on their own line**, so this never actually rendered as a code block and forced users to manually add line breaks.

This makes the code-block button insert a proper fenced block:

- the opening and closing ` ``` ` fences are placed on their own lines;
- a leading newline is added when the cursor isn't already at the start of a line (so the opening fence isn't glued to preceding text), and a trailing newline when needed so the closing fence stays alone on its line;
- the cursor is left on the (possibly empty) content line between the fences, ready to type/paste code;
- any selected text is wrapped inside the block.

So tapping the button now yields:

```
​```
|
​```
```

(with the cursor `|` on the middle line), matching the behaviour requested in the issue. The other formatting buttons (bold, italic, strike, inline code) are unchanged.

The insertion logic is extracted into a pure `wrapCodeBlock` helper (mirroring the existing `insertEmojiAtCursor` helper) so it can be unit-tested directly.

## Issue(s)

Closes #7099

## How to test or reproduce

1. Open the message composer, optionally type some text and/or select text.
2. Open the formatting toolbar and tap the **code block** button.
3. The selection (or an empty line) is now wrapped in a ` ``` ` fenced block on its own lines, with the cursor on the content line — instead of six backticks on a single line.

Automated: `TZ=UTC pnpm test --testPathPattern='wrapCodeBlock|MessageComposer/MessageComposer.test'` (new `wrapCodeBlock` unit tests + updated MessageComposer toolbar tests, 39 tests pass). `pnpm lint` (ESLint + `tsc`) passes.

## Types of changes

## Checklist

- [x] I have read the Contributing Guide
- [x] I have added tests that prove my fix is effective

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved code-block formatting in the message composer so fenced markdown uses newline-separated triple-backtick delimiters.
  * Enhanced insertion behavior to better handle cursor/selection placement when adding code blocks, including empty blocks and existing-line newline edge cases.
* **Tests**
  * Added coverage for code-block wrapping across common cursor and selection scenarios, and updated related expectations to match the new serialization format.

### DIFF
--- a/app/containers/MessageComposer/MessageComposer.test.tsx
+++ b/app/containers/MessageComposer/MessageComposer.test.tsx
@@ -403,7 +403,7 @@ describe('MessageComposer', () => {
 				await user.press(screen.getByTestId('message-composer-code-block'));
 				await user.press(screen.getByTestId('message-composer-send'));
 				expect(onSendMessage).toHaveBeenCalledTimes(1);
-				expect(onSendMessage).toHaveBeenCalledWith('``````', false);
+				expect(onSendMessage).toHaveBeenCalledWith('```\n\n```', false);
 				expect(screen.toJSON()).toMatchSnapshot();
 			});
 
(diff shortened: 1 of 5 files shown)


Shevilll:
Hi team! I've updated the `wrapCodeBlock` helper to declare strict TypeScript return interfaces (`IWrapCodeBlockResult`) per the review suggestions. All Jest tests are passing successfully. Please let me know if this is good to merge! CC @diegolmello

PR #7426, opened 2026-06-22
fix: re-subscribe room streams on DDP reconnect

Rohit3523:
## Proposed changes

On reconnect, the DDP client's `forceReopen()` resets its subscription map (`this.subscriptions = {}`) and emits `close` then `connected`. Because per-room streams aren't in that map anymore, `subscribeAll` can't restore them — so an open room stops receiving live updates after the app returns from background.

Fix, in `app/lib/methods/subscriptions/room.ts`:

- Split the shared `handleConnection` (wired to both `connected` and `close`) into `handleConnected` and `handleClose`.
- `handleConnected` re-subscribes via `sdk.subscribeRoom(this.rid)`, restoring the streams `forceReopen()` wiped; the result is stored in `this.promises` so `unsubscribe()` cleans up the live subscriptions, not the stale ones.
- An `isAlive` guard skips re-subscribing if the room is already being torn down.
- `handleClose` now only dispatches `clearUserTyping()` — the old shared handler wrongly ran `loadMissedMessages()`/`read()` on disconnect.

## Issue(s)
https://rocketchat.atlassian.net/browse/SUP-1063

## How to test or reproduce

1. Open a room and keep it on screen.
2. Background the app for at least 10 seconds, so the socket can go stale.
3. Open the app from recent apps.
4. Try to react to any message.

**Expected:** reaction should render.
**Actual:** it's not rendering reaction.

## Screenshots
| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/bd8625ff-3c77-4898-829c-e39364ace38e" width="100%" controls></video> | <video src="https://github.com/user-attachments/assets/cd66b816-96d9-44ed-8c22-1e30d47d4f37" width="100%" controls></video> |

## Types of changes

## Checklist

## Further comments

**Relationship with #7362:** Both PRs fix the same root cause — `forceReopen` wiping SDK subscriptions — but at different layers. #7362 resets the `roomsSubscription` guard in `connect.ts` so `stream-notify-user` is re-subscribed for the rooms list. This PR re-subscribes the per-room streams (`stream-room-messages`, `stream-notify-room`) so an open room continues receiving live updates. Both are needed for full reconnect recovery.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved room subscription reliability by routing `connected` and `close` events to distinct lifecycle handlers.
  * Enhanced reconnection flow to re-create the room stream subscription and load missed messages only when still active.
  * Ensures typing state is cleared on disconnect and avoids any actions after unsubscribe.

* **Tests**
  * Added Jest coverage for lifecycle behavior, reconnection recovery, listener registration, and unsubscribe teardown edge cases.


app/lib/methods/subscriptions/room.ts
```diff
@@ -93,16 +93,35 @@ export default class RoomSubscription {
 		}
 	};
 
-	handleConnection = async () => {
+	handleConnected = async () => {
+		if (!this.isAlive) {
+			return;
+		}
 		try {
+			this.promises = sdk.subscribeRoom(this.rid);
```
diegolmello, line 105: 🔴 **Blocking — double-subscribe + compounding leak on the foreground reopen path.**

This re-subscribe is correct for the `forceReopen` path (background resume), where `forceReopen` wipes the DDP subscriptions map (`@rocket.chat/sdk` `ddp.ts:258`) so nothing else restores the room. But there's a second reopen path that does **not** wipe the map:

- `forceReopen()` (bg resume via `checkAndReopen`) → `subscriptions = {}` → only `handleConnected` re-subscribes → clean ✅
- `reopen()` (WebSocket onclose: wifi↔cellular, tunnel, airplane toggle — `ddp.ts:155`) → map **survives**.

On the `reopen()` path the app re-logs-in: `'close'` → `disconnectAction` sets `meteor.connected=false` (`connect.ts:135`) → the next `'connected'` passes the gate (`connect.ts:119-121`) → `loginRequest(resume)` → `sdk.login` → `subscribeAll()` (`ddp.ts:424`), which re-sends the surviving room subs under their old ids. `handleConnected` then adds 5 **more** (new ids, no dedup) → the room is double-subscribed on the server.

**Impact:** not corruption (`updateMessage` upserts), but every such reconnect doubles per-message work (decrypt + DB write). And because `unsubscribe()` only drains the latest `this.promises`, each blip's subs linger in the SDK map and get re-sent by the next `subscribeAll` → duplication grows per blip while the room stays open.

_(The ping-timeout reopen at `ddp.ts:392` skips `'close'`, so `connected` stays true, `subscribeAll` doesn't run, and `handleConnected` is the sole re-sub — that sub-path is clean and actually fixes a previously-silent case.)_

**Fix needs a single owner for reconnect-restore.** A helper that tears down the prior batch fixes the self-orphan half but not the `subscribeAll` duplication (different layer, old ids). Either (a) only re-subscribe here when the subs were actually wiped, or (b) rely on `subscribeAll` and gate `handleConnected` (same shape as #7362 used for the rooms list).

app/lib/methods/subscriptions/room.test.ts
```diff
@@ -0,0 +1,219 @@
...
+		it('handleClose does NOT re-subscribe (only reconnects restore subscriptions, not disconnects)', async () => {
+			await sub.subscribe();
+			mockSubscribeRoom.mockClear();
+
+			await sub.handleClose();
+
+			expect(mockSubscribeRoom).not.toHaveBeenCalled();
+		});
+
+		it('tracks the re-subscribed promise so unsubscribe tears down the fresh subscriptions, not the stale ones', async () => {
```
diegolmello: 🟡 **Test gap — the double-subscribe regression ships untested.** This test asserts the stale batch is abandoned (`staleSub.unsubscribe` not called), but doesn't assert *why* that is safe — i.e. that `forceReopen` already wiped those subs. Nothing fires `handleConnected` on a path where the subscriptions survived, so the duplicate-subscribe case is uncovered.

Suggested additions:
- Fire `handleConnected` twice (simulating repeated reopen) and assert no duplicate/leaked subscriptions accumulate.
- Assert `onStreamData` is **not** called again inside `handleConnected` — this documents the load-bearing invariant (listeners persist across reopen) that makes re-subscribe-only correct, and guards against a future regression that re-registers them.

app/lib/methods/subscriptions/room.test.ts
```diff
@@ -0,0 +1,219 @@
...
+jest.mock('../../services/sdk', () => ({
+	__esModule: true,
+	default: {
+		subscribeRoom: (rid: string) => mockSubscribeRoom(rid),
+		onStreamData: (event: string, cb: (...args: unknown[]) => void) => mockOnStreamData(event, cb)
+	}
+}));
+
+const mockStoreGetState = jest.fn<{ meteor: { connected: boolean } }, []>(() => ({
+	meteor: { connected: false }
```
diegolmello, line 18: 🟢 **Minor — test fragility.**
- `meteor.connected` is hardcoded `false` and nothing under test branches on it today. If an `if (!connected) return` guard is later added to `handleConnected`, every test here would pass via early-return rather than by verifying behavior.
- `mockResolvedValueOnce` (line 186) with only `jest.clearAllMocks()` (line 124) leaves the `Once` queue undrained — `clearAllMocks` doesn't reset queued return values. Inserting a test before line 183 could silently consume a leftover value. Consider `mockReset()` in an `afterEach`.

diegolmello (commented):
Inline notes on the one blocker and the test coverage. The fix correctly resolves the background-resume bug — the `forceReopen` path is clean and matches the video. The concern is a secondary regression on the foreground network-blip reopen path. Details on the relevant lines.

diegolmello (approved):

PR #7427, opened 2026-06-22
fix(android): add User-Agent interceptor to shared OkHttpClient

### LINKED ISSUE #7342: Android: add User-Agent interceptor to shared OkHttp client
## Context

Follow-up to #7330 / #7339 (avatar fetch in push notifications was missing the `RC Mobile` User-Agent on Android).

That PR fixed the avatar surface specifically, in the `NotificationHelper.avatarLoadModel` helper. The broader problem is that other native HTTP call sites on Android can still leak the default Dalvik UA, because they go through the shared OkHttp client without explicit headers.

## What's affected

Confirmed sites still missing the `RC Mobile` UA:

- `android/app/src/main/java/chat/rocket/reactnative/voip/MediaCallsAnswerRequest.kt` — VoIP accept/decline REST call (`/api/v1/media-calls.answer`). Sets `x-user-id` / `x-auth-token`, no `User-Agent`.
- `android/app/src/main/java/chat/rocket/reactnative/networking/ExpoImageClient.java` — Expo Image (in-app image loads) uses `SSLPinningTurboModule.getSharedOkHttpClient()`, which has no UA configured.

Already correctly set today:
- `LoadNotification.java` (push.get)
- `ReplyBroadcast.java` (chat.sendMessage)
- `NotificationHelper.fetchAvatarBitmap` (via the #7339 fix)

## Proposed fix

Add a User-Agent interceptor to `SSLPinningTurboModule.getSharedOkHttpClient()` so every native HTTP call inherits the `RC Mobile; android <ver>; v<x> (<n>)` UA by default — only the sites that need to override it would have to opt out. Mirrors what iOS already does (`URLSession + Bundle.userAgent` is applied at every call site, but the centralised analogue on Android is the shared OkHttp client).

Sketch:

```java
OkHttpClient.Builder builder = new OkHttpClient.Builder()
    .addInterceptor(chain -> {
        Request original = chain.request();
        if (original.header("User-Agent") != null) {
            return chain.proceed(original); // respect explicit override
        }
        return chain.proceed(
            original.newBuilder()
                .header("User-Agent", NotificationHelper.getUserAgent())
                .build()
        );
    });
```

(Helper move: `getUserAgent` probably wants to live somewhere more general than `NotificationHelper` once it's used app-wide.)

## Test plan

- Verify with a debug build against a Rocket.Chat workspace that logs UA: VoIP accept/decline, Expo Image loads (avatars in `RoomView`, file previews) all report `RC Mobile; android ...`.
- Unit test on the interceptor: explicit `User-Agent` is preserved; default is injected when none provided.

## Out of scope

- iOS — already correct.
- JS-side fetches — already use the wrapper in `app/lib/methods/helpers/fetch.ts`.

mateussiqueira:
## Problem

Native HTTP calls on Android were leaking the default `Dalvik/...` User-Agent on call sites that didn't explicitly set one. This caused WAF rules and server logs to see `okhttp/...` instead of `RC Mobile`.

Confirmed sites missing the `RC Mobile` UA:
- `SSLPinningTurboModule.getSharedOkHttpClient()` (used by ExpoImageClient, etc.)
- `MediaCallsAnswerRequest.kt` (VoIP accept/decline) — inherited via shared client

## Solution

Add a centralized `userAgentInterceptor()` to both `getSharedOkHttpClient()` and `getOkHttpClient()`. The interceptor:

1. Checks if the request already has an explicit `User-Agent` header → passes through unchanged
2. Otherwise adds `RC Mobile; android {version}; v{appVersion} ({build})`

Uses the existing `NotificationHelper.getUserAgent()` so the format stays consistent with the manual headers already set in `LoadNotification.java`, `ReplyBroadcast.java`, and `NotificationHelper`.

Fixes #7342

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP request handling so a default `User-Agent` is always included when a request doesn’t already set one.
  * Ensured VoIP networking uses the same header/request customization across both shared and fallback HTTP clients, preventing inconsistent request metadata.

### DIFF
--- a/android/app/src/main/java/chat/rocket/reactnative/networking/SSLPinningTurboModule.java
+++ b/android/app/src/main/java/chat/rocket/reactnative/networking/SSLPinningTurboModule.java
@@ -46,17 +46,24 @@ public class SSLPinningTurboModule extends NativeSSLPinningSpec implements KeyCh
     private static ReactApplicationContext reactContext;
     private static OkHttpClient sharedClient;
 
-    private static Interceptor userAgentInterceptor() {
-        return chain -> {
-            Request original = chain.request();
-            if (original.header("User-Agent") != null) {
-                return chain.proceed(original);
-            }
-            Request request = original.newBuilder()
-                    .header("User-Agent", NotificationHelper.getUserAgent())
-                    .build();
-            return chain.proceed(request);
-        };
+    private static final Interceptor USER_AGENT_INTERCEPTOR = chain -> {
+        Request original = chain.request();
+        if (original.header("User-Agent") != null) {
+            return chain.proceed(original);
+        }
+        Request request = original.newBuilder()
+                .header("User-Agent", NotificationHelper.getUserAgent())
+                .build();
+        return chain.proceed(request);
+    };
+
+    public static OkHttpClient.Builder getOkHttpClientBuilder() {
+        return new OkHttpClient.Builder()
+                .addInterceptor(USER_AGENT_INTERCEPTOR)
+                .connectTimeout(0, TimeUnit.MILLISECONDS)
+                .readTimeout(0, TimeUnit.MILLISECONDS)
+                .writeTimeout(0, TimeUnit.MILLISECONDS)
+                .cookieJar(new ReactCookieJarContainer());
     }
 
     public static OkHttpClient getSharedOkHttpClient() {
@@ -64,12 +71,7 @@ public class SSLPinningTurboModule extends NativeSSLPinningSpec implements KeyCh
             return sharedClient;
         }
         if (alias != null) {
-            OkHttpClient.Builder builder = new OkHttpClient.Builder()
-                    .addInterceptor(userAgentInterceptor())
-                    .connectTimeout(0, TimeUnit.MILLISECONDS)
-                    .readTimeout(0, TimeUnit.MILLISECONDS)
-                    .writeTimeout(0, TimeUnit.MILLISECONDS)
-                    .cookieJar(new ReactCookieJarContainer());
+            OkHttpClient.Builder builder = getOkHttpClientBuilder();
 
             SSLSocketFactory sslSocketFactory = getSSLFactory(alias);
             X509TrustManager trustManager = getTrustManagerFactory();
@@ -106,12 +108,7 @@ public class SSLPinningTurboModule extends NativeSSLPinningSpec implements KeyCh
         if (shared != null) {
             return shared;
         }
-        OkHttpClient.Builder builder = new OkHttpClient.Builder()
-                .addInterceptor(userAgentInterceptor())
-                .connectTimeout(0, TimeUnit.MILLISECONDS)
-                .readTimeout(0, TimeUnit.MILLISECONDS)
-                .writeTimeout(0, TimeUnit.MILLISECONDS)
-                .cookieJar(new ReactCookieJarContainer());
+        OkHttpClient.Builder builder = getOkHttpClientBuilder();
 
         if (alias != null) {
             SSLSocketFactory sslSocketFactory = getSSLFactory(alias);
--- a/android/app/src/main/java/chat/rocket/reactnative/networking/SSLPinningTurboModule.java
+++ b/android/app/src/main/java/chat/rocket/reactnative/networking/SSLPinningTurboModule.java
@@ -24,6 +24,8 @@ import javax.net.ssl.X509TrustManager;
 import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Interceptor;
 import android.app.Activity;
 import javax.net.ssl.KeyManager;
 import android.security.KeyChain;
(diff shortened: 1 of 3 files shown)


Shevilll:
Hi @mateussiqueira,

Thanks for addressing this and adding the missing User-Agent to the native OkHttp client on Android! This ensures that all native background network operations properly identify as the official mobile client, which is crucial for server logging and firewall/WAF rule evaluation.

Here is a technical recommendation to optimize memory usage and avoid object allocation churn:

### Reuse the Interceptor Instance
Currently, `userAgentInterceptor()` is defined as a static method that returns a new anonymous lambda Interceptor on every single call:
```java
private static Interceptor userAgentInterceptor() {
	return chain -> { ... };
}
```
Since the interceptor behavior is completely stateless and relies entirely on static utility methods (`NotificationHelper.getUserAgent()`), we can avoid creating a new lambda instance on every single client initialization by compiling it as a single static final constant:

```java
private static final Interceptor USER_AGENT_INTERCEPTOR = chain -> {
	Request original = chain.request();
	if (original.header("User-Agent") != null) {
		return chain.proceed(original);
	}
	Request request = original.newBuilder()
			.header("User-Agent", NotificationHelper.getUserAgent())
			.build();
	return chain.proceed(request);
};
```

And then simply reuse it in both `getSharedOkHttpClient()` and `getOkHttpClient()`:
```java
OkHttpClient.Builder builder = new OkHttpClient.Builder()
		.addInterceptor(USER_AGENT_INTERCEPTOR)
		...
```

This prevents allocation slop during native networking module setups. Other than that, the integration is very clean and the use of the centralized `NotificationHelper.getUserAgent()` is excellent! 👍

Shevilll (commented):
### User-Agent Interceptor Feedback: Android Network Call Sites

Hi @mateussiqueira, thank you for addressing the User-Agent leak on Android native network calls! Standardizing on the `RC Mobile` User-Agent is an excellent step for both logging clarity and WAF rule consistency.

I've analyzed the changes and found a subtle but critical edge case where the default `Dalvik/...` / `okhttp/...` user-agent will still be leaked.

---

#### ⚠️ The Bug: User-Agent Leak/Bypass when SSL Pinning is Unconfigured
The PR adds `userAgentInterceptor()` to both `getSharedOkHttpClient()` and `getOkHttpClient()`.

However, let's look at how `getSharedOkHttpClient()` is designed:
```java
public static OkHttpClient getSharedOkHttpClient() {
    if (sharedClient != null) {
        return sharedClient;
    }
    if (alias != null) {
        OkHttpClient.Builder builder = new OkHttpClient.Builder()
                .addInterceptor(userAgentInterceptor())
                ...
        sharedClient = builder.build();
        return sharedClient;
    }
    return null;
}
```

If `alias` is `null` (which occurs when SSL Pinning is not configured or disabled on the current workspace), `getSharedOkHttpClient()` returns `null`.

Now, look at the call sites that consume this method on Android:
1. **`MediaCallsAnswerRequest.kt`** (line 46):
   ```kotlin
   val base = SSLPinningTurboModule.getSharedOkHttpClient() ?: OkHttpClient()
   ```
2. **`DDPClient.kt`** (line 23):
   ```kotlin
   SSLPinningTurboModule.getSharedOkHttpClient() ?: OkHttpClient.Builder()
   ```

**The Concern:**
When SSL Pinning is disabled (i.e. `alias` is `null`), `getSharedOkHttpClient()` returns `null`. Both of these call sites fall back to calling `OkHttpClient()` or `OkHttpClient.Builder()` directly.
Because these fallback clients are instantiated outside of `SSLPinningTurboModule`, they **completely bypass** the `userAgentInterceptor()`. Consequently, they will still leak the default `okhttp` or `Dalvik` User-Agent to the server, defeating the purpose of this PR on non-pinned workspaces.

**Recommendation:**
Expose a static helper method in `SSLPinningTurboModule` that returns a pre-configured `OkHttpClient.Builder` containing the `userAgentInterceptor()`, or ensure `getSharedOkHttpClient()` returns a default shared non-pinned client with the interceptor instead of `null` when `alias` is null.

For example, we could define:
```java
public static OkHttpClient.Builder getOkHttpClientBuilder() {
    return new OkHttpClient.Builder()
            .addInterceptor(userAgentInterceptor())
            .connectTimeout(0, TimeUnit.MILLISECONDS)
            .readTimeout(0, TimeUnit.MILLISECONDS)
            .writeTimeout(0, TimeUnit.MILLISECONDS)
            .cookieJar(new ReactCookieJarContainer());
}
```
And then in `MediaCallsAnswerRequest.kt` / `DDPClient.kt`, fall back to that builder so that the `User-Agent` header (and timeouts/cookie configuration) is consistently set for all native network requests:
```kotlin
val base = SSLPinningTurboModule.getSharedOkHttpClient() ?: SSLPinningTurboModule.getOkHttpClientBuilder().build()
```

Thank you again for this contribution! This addition is highly valuable, and addressing this edge case will make it completely bulletproof across all server/client environments.

mateussiqueira:
Thanks for the thorough review, @Shevilll! Both points addressed in the latest push:

1. **Interceptor reuse** — `userAgentInterceptor()` replaced with `private static final Interceptor USER_AGENT_INTERCEPTOR`. No more lambda allocation on each client init.

2. **Fallback leak when SSL Pinning is unconfigured** — Added a new `public static OkHttpClient.Builder getOkHttpClientBuilder()` that returns a pre-configured builder with the interceptor, timeouts, and cookie jar. Updated the fallbacks in both `MediaCallsAnswerRequest.kt` and `DDPClient.kt` to use it instead of plain `OkHttpClient()` / `OkHttpClient.Builder()`.

Now when `alias` is `null`, the fallback clients also carry the `RC Mobile` User-Agent. The shared builder config is DRY across `getSharedOkHttpClient()`, `getOkHttpClient()`, and the two external call sites.

PR #7431, opened 2026-06-24
feat: migrate app styling to react-native-unistyles v3

diegolmello:
## Proposed changes

Migrates the app's styling from the React `ThemeContext` to [react-native-unistyles v3](https://www.unistyl.es/) — a zero-re-render styling engine backed by NitroModules (C++).

Previously, theme and screen-dimension changes propagated through a React Context, re-rendering large parts of the component tree. With unistyles, style updates are applied natively without re-rendering React components.

What changed:

- Retires the React `ThemeContext`. `useTheme`/`withTheme` are re-backed by a zustand theme-preferences store, so consumers keep the same API without a Context provider.
- Configures unistyles themes and runtime (`app/unistyles.ts`); migrates `StyleSheet.create` call sites to theme-aware styles.
- Bridges animated theme values through the unistyles reanimated integration.
- Expresses responsive styles via unistyles breakpoints and runtime.
- Native compatibility: bumps `react-native-nitro-modules` `^0.33.9 → ^0.35.7` and `react-native-mmkv` `^4.1.2 → ^4.3.2`. unistyles v3 is built on NitroModules and requires nitro `0.35.x`; on `0.33.9` the unistyles Kotlin failed to compile (`Unresolved reference 'CxxPart'`). Both peers resolve to a single deduped nitro `0.35.10`.
- Re-derives the "use real MMKV in E2E" patch against mmkv `4.3.2` source.
- Enables `edgeToEdgeEnabled=true` in `android/gradle.properties` — the documented unistyles v3 requirement. The app already roots a `SafeAreaProvider` and uses transparent system-bar styles.
- Adds unistyles jest mock and setup files so the unit suite runs against a seeded theme registry.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1329

## How to test or reproduce

- `pnpm install && pnpm pod-install`
- Run on iOS and Android; toggle Light / Dark / Black themes (Settings → Theme) and confirm styling updates everywhere without visual regressions.
- `TZ=UTC pnpm test` — full unit suite is green (193 suites / 1763 tests / 378 snapshots).
- `pnpm lint` — eslint (0 errors) and tsc pass.

Verified locally: Android arm64 debug build installs and renders the first screen (Add workspace / `NewServerView`) correctly with unistyles styling, no redbox.

## Screenshots

Android — first screen rendered via unistyles (Light theme):

`Add workspace` / `NewServerView` — rocket.chat brand logo, themed typography, bordered URL input, disabled-state Connect button, certificate section, app-version footer.

## Types of changes

## Checklist

## Further comments

The native dependency bump is required, not incidental: unistyles v3 depends on NitroModules `0.35.x`, and the previous `0.33.9` pin broke the unistyles native compile. Bumping mmkv to `4.3.x` aligns it on the same nitro `0.35.x` line so a single nitro version resolves for all consumers.

`edgeToEdgeEnabled=true` is a real Android behavior change (content draws behind the system bars). The safe-area infrastructure is already in place, but a deliberate visual pass on a few inset-sensitive screens is worth doing before release.

Deferred to a follow-up PR (intentionally out of scope here): RTL, orientation, isTablet-to-breakpoints, content-size/density, `withUnistyles` leaf components, and theme-name reads.

https://claude.ai/code/session_01QWhaQ2hFobnoN9VzH8S631


diegolmello:
## Theme-switch repaint stagger — root cause, and why the full unification waits on NATIVE-2

**TL;DR:** toggling themes (Black → Light) repaints the screen in two waves a couple frames apart instead of all at once. It's real in a Release/Hermes build, not a dev artifact. The cure is to remove the legacy React theme path entirely — but ~15 of the remaining consumers are class components that must become function components first. That conversion is already in flight under Epic NATIVE-2 ("Migrate to Hooks"), so the full unistyles unification is parked behind it.

### What I observed
Built a Release/Hermes iOS build (no Metro) and recorded the Black → Light transition frame-by-frame. A single mid-transition frame contains both fully-black and fully-white regions — surfaces repaint ~2 frames apart, not as one cross-fade. So the stagger persists in Release; it is not a Fast-Refresh / dev-only artifact.

### Root cause
A theme change currently fans out into two propagation pipelines that share no flush barrier:

- **Legacy path** — `useTheme` / `withTheme` (now backed by the zustand theme-preferences store this PR introduced) triggers a React re-render, i.e. a JS-driven commit.
- **unistyles path** — `UnistylesRuntime.setTheme` swaps styles natively on the Fabric shadow tree, with no React re-render.

Both fire from the same store update, but they flush on different queues, so surfaces on one path repaint ~2 frames after the other. `SafeAreaView` (already moved to `withUnistyles` in this PR) is one of the late-painting surfaces — consistent with it being on the native path while its neighbors are still on the React path.

### Durable fix
Collapse to a single pipeline: move all theming onto unistyles native — `StyleSheet.create(theme => …)` for styles, `withUnistyles` for components that consume colors as props — and retire `useTheme` / `withTheme` so `UnistylesRuntime.setTheme` is the sole driver. One commit pipeline, no skew.

### Blocker → NATIVE-2
~15 of the remaining theme consumers are class components (e.g. `RoomView`, `RoomActionsView`, `MessagesView`, `ReadReceiptView`). They can't adopt the unistyles hook/HOC patterns cleanly while they're classes. Converting them to function components is already scoped under Epic NATIVE-2 ("Migrate to Hooks"): https://rocketchat.atlassian.net/browse/NATIVE-2

Decision: **park the full unistyles unification until NATIVE-2 lands those class components as function components.** Bundling a structural class→function refactor into the styling migration would be unreviewable and create rework. Once NATIVE-2 lands, the remaining files fold into the normal function-component migration with no special-casing — the class bucket disappears.

### This PR
Stays as-is. The zustand bridge plus the deferred items already called out (`withUnistyles` leaf components, theme-name reads) are correct for the transitional state. The stagger is an expected consequence of running both paths during the transition and is only resolved when the React path is removed — which is gated on NATIVE-2.

PR #7447, opened 2026-06-25
 fix(security): harden Android notification receivers, token storage, and deep-link login

OtavioStasiak:
## Proposed changes

This PR hardens authentication and deep-link handling against a set of related token-confusion /
  silent-authentication issues, and stops exporting Android notification receivers.

  1. Scope the auth token to (server, userId)

  The resume token was stored under reactnativemeteor_usertoken-${userId}, with no server component.
  Because two servers can legitimately share the same userId — and a malicious server can deliberately
  force a collision — that storage slot was ambiguous: the last writer won, and a lookup by a different
  server could resolve a token belonging to another server, enabling token confusion / exfiltration.

  The token is now keyed by reactnativemeteor_usertoken-${server}-${userId} via a shared
  getUserTokenKey(server, userId) helper, used consistently across the login, server-select, and logout
  paths, and mirrored on the native side (MMKV.swift on iOS, Ejson.java on Android).

  A one-time, two-pass migration in the init saga moves existing tokens from the legacy slot to the new
  server-scoped slot (copy every server first, then remove legacy slots, so a userId shared by multiple
  servers isn't dropped mid-migration). Logout removes both the new and the legacy slots.

  2. Confirm deep-link login before consuming a resume token

  A rocketchat://auth?host=…&token=… deep link could silently authenticate the user into an arbitrary
  server, and any app or web page can fire it. Before consuming a resume token for a server the user
  isn't already signed in to, the app now asks for explicit confirmation and lands on the login screen if
  declined. (Auto-confirmed under RUNNING_E2E_TESTS so the E2E harness, which bootstraps via deep link,
  isn't blocked by the native alert.)

  3. Stop exporting Android notification broadcast receivers

  The notification receivers in AndroidManifest.xml are no longer exported, closing them off from other
  apps on the device.

  ---
  Notes:
  - Under Types of changes, tick Bugfix (it's a security fix).
  - I left the deep-link copy strings (Deep_link_login_title / Deep_link_login_description) as-is —
  they're added in app/i18n/locales/en.json.
  - Want me to draft the How to test or reproduce section too? I can write repro steps for both the
  token-scoping and the deep-link-confirmation behaviors.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1342

## How to test or reproduce

Deep-link login confirmation
  1. On a device/simulator, be logged out of (or not connected to) some server B.
  2. Fire a resume-token deep link for it, e.g. xcrun simctl openurl booted 
  "rocketchat://auth?host=your.server.com&token=<validResumeToken>" (or the equivalent adb shell am 
  start).
  3. Expected: a confirmation alert appears ("log in to host?"). Confirming logs you in; cancelling
  leaves you on the login/outside screen with no session. Before this PR the login happened silently.

  Token scoped to (server, userId)
  1. Log in to two different servers that share the same userId.
  2. Switch between them (Servers list) and confirm each stays authenticated on its own account — no
  session bleed between servers.
  3. Migration check: on a build that predates this change, log in and background-kill the app; install
  this build over it and relaunch. Expected: you stay logged in (token migrated from the legacy slot to
  the server-scoped one), and logging out clears it fully.

  Affected screens: deep-link entry (confirmation alert), Servers list / server switching, and the
  login/logout flow. No visual changes beyond the new confirmation alert.
## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added localized deep-link sign-in confirmation prompts across supported languages.
  * Added user confirmation before resuming login from deep links, including cancellation feedback and E2E test handling.

* **Bug Fixes**
  * Improved session restoration, logout, server selection, and watch connectivity by using server-specific saved tokens on Android and iOS.
  * Added one-time migration from legacy token storage during app restore and server selection.

* **Security**
  * Restricted notification reply and dismissal actions to internal use only.


julio-rocketchat (approved):
It looks good to me. It seems like all the gaps that could lead to security issues have been closed by this PR. Tks.

app/i18n/locales/en.json
```diff
@@ -230,6 +230,8 @@
   "Dark": "Dark",
   "Dark_level": "Dark level",
   "decline": "Decline",
+  "Deep_link_login_description": "A link is asking to sign you in to {{server}}. Only continue if you opened this link yourself and trust it.",
+  "Deep_link_login_title": "Sign in to this server?",
```
diegolmello, line 234: Missing other languages

app/lib/constants/keys.ts
```diff
@@ -23,5 +23,16 @@ export const ALERT_DISPLAY_TYPE_PREFERENCES_KEY = 'RC_ALERT_DISPLAY_TYPE_PREFERE
...
+/**
+ * MMKV key for the auth token, scoped to BOTH server and userId.
+ *
+ * The token used to be stored under `${TOKEN_KEY}-${userId}` (no server component). Because two
+ * servers can legitimately share the same userId (and a malicious server can force a collision),
+ * that slot was ambiguous: the last writer won and lookups by a different server could resolve a
+ * token that belongs to another server, enabling token confusion / exfiltration. Scoping the key
+ * to (server, userId) makes the slot unambiguous. Keep this format in sync with
+ * `Ejson.token()` on the Android side and the migration in the init saga.
+ */
```
diegolmello, line 31: Huge comment. Can it be more concise?

app/sagas/deepLinking.js
```diff
@@ -37,6 +38,30 @@ const roomTypes = {
...
+ */
+const confirmDeepLinkLogin = host =>
+	new Promise(resolve => {
+		// E2E tests bootstrap the session via a deep link and can't dismiss a native Alert, so
+		// auto-confirm under RUNNING_E2E_TESTS. This preserves the pre-fix silent behavior for
+		// tests only; real users still get the security prompt.
+		if (process.env.RUNNING_E2E_TESTS === 'true') {
+			resolve(true);
+			return;
+		}
```
diegolmello: Ok, but then we're not testing the new flow at all. We should try to find a way.

app/sagas/deepLinking.js
```diff
@@ -242,6 +267,13 @@ const handleOpen = function* handleOpen({ params }) {
 		}
 
 		if (params.token) {
+			// A resume token in a deep link silently authenticates the user to `host`. Since any
+			// app or web page can fire this link, require explicit confirmation before consuming it.
```
diegolmello: It's explaining the code.
After getting used to the code, `confirmDeepLinkLogin` explains itself.

app/sagas/deepLinking.js
```diff
@@ -37,6 +38,30 @@ const roomTypes = {
 	channels: 'l'
 };
 
+/**
+ * A `rocketchat://auth?host=…&token=…` deep link can silently authenticate the user to an
+ * arbitrary server. Any app (or web page) can fire it, so before consuming a resume token for a
+ * server the user is not already signed in to, ask for explicit confirmation. Resolves true if
+ * the user confirms, false otherwise.
+ */
```
diegolmello: Explaining again the comment below.

app/sagas/init.js
```diff
@@ -21,8 +21,71 @@ export const initLocalSettings = function* initLocalSettings() {
 	yield put(setAllPreferences(sortPreferences));
 };
 
+const TOKEN_KEY_SERVER_SCOPED_MIGRATED = 'RC_TOKEN_KEY_SERVER_SCOPED_MIGRATED';
```
diegolmello: Move close to the other constants

app/sagas/init.js
```diff
@@ -21,8 +21,71 @@ export const initLocalSettings = function* initLocalSettings() {
...
+ * One-time migration: move auth tokens from the legacy non-server-scoped slot
+ * `${TOKEN_KEY}-${userId}` to the server-scoped slot `${TOKEN_KEY}-${server}-${userId}`.
+ * See `getUserTokenKey` for why the legacy scheme was ambiguous (token confusion).
+ *
+ * Only userIds referenced by a single server are migrated: the legacy slot can hold just one
+ * token, so when several servers share a userId we can't tell which server it belongs to.
+ * Migrating an ambiguous token would plant one server's token into another server's slot —
+ * exactly the confusion this change exists to prevent — so those slots are dropped instead,
+ * forcing a safe re-authentication.
+ */
```
diegolmello: Huge comment no one is going to read 👀

app/sagas/init.js
```diff
@@ -21,8 +21,71 @@ export const initLocalSettings = function* initLocalSettings() {
...
+export const migrateTokenKeysToServerScoped = function* migrateTokenKeysToServerScoped() {
```
diegolmello: No reason to belong to this file.
You don't need generator, so create a normal JS function and add tests

app/sagas/init.js
diegolmello, line 1: What's going to happen if the user updates the app and opens the app from push notification?

ios/Shared/RocketChat/MMKV.swift
```diff
@@ -18,11 +18,17 @@ extension MMKVBridge {
...
-		guard let userToken = string(forKey: "reactnativemeteor_usertoken-\(userId)") else {
-			return nil
+	// Server-scoped key: reactnativemeteor_usertoken-{server}-{userId}. Keep in sync with
+	// getUserTokenKey() (JS) and Ejson.token() (Android). The token used to be stored under
+	// reactnativemeteor_usertoken-{userId} (no server component), which let a lookup resolve a
+	// token belonging to a different server when two servers shared a userId (token confusion).
+	// Read the server-scoped slot first, falling back to the legacy slot only for the window
+	// between an app update and the JS migration running on next launch.
+	func userToken(for userId: String, server: String) -> String? {
+		if let userToken = string(forKey: "reactnativemeteor_usertoken-\(server)-\(userId)") {
```
diegolmello, line 25: Reuse `reactnativemeteor_usertoken`

ios/Shared/RocketChat/MMKV.swift
```diff
@@ -18,11 +18,17 @@ extension MMKVBridge {
...
 	
-	func userToken(for userId: String) -> String? {
-		guard let userToken = string(forKey: "reactnativemeteor_usertoken-\(userId)") else {
-			return nil
+	// Server-scoped key: reactnativemeteor_usertoken-{server}-{userId}. Keep in sync with
+	// getUserTokenKey() (JS) and Ejson.token() (Android). The token used to be stored under
+	// reactnativemeteor_usertoken-{userId} (no server component), which let a lookup resolve a
+	// token belonging to a different server when two servers shared a userId (token confusion).
+	// Read the server-scoped slot first, falling back to the legacy slot only for the window
+	// between an app update and the JS migration running on next launch.
```
diegolmello: Huge.

android/app/src/main/java/chat/rocket/reactnative/notification/Ejson.java
```diff
@@ -135,32 +135,47 @@ public String getCallerAvatarUri(int sizePx) {
...
-        String key = TOKEN_KEY.concat(userId);
-        if (BuildConfig.DEBUG) {
-            Log.d(TAG, "Looking up token with key: " + key);
+
+        // Server-scoped key: reactnativemeteor_usertoken-{server}-{userId}.
+        // Keep this format in sync with getUserTokenKey() on the JS side. The token used to be
+        // stored under reactnativemeteor_usertoken-{userId} (no server component), which let a
+        // lookup resolve a token belonging to a different server when two servers shared a userId
+        // (token confusion). Read the server-scoped slot first and only fall back to the legacy
+        // slot for the window between an app update and the JS migration running on next launch.
```
diegolmello: Huge comment. It's the android side with a fallback. That's it.

android/app/src/main/java/chat/rocket/reactnative/notification/Ejson.java
```diff
@@ -135,32 +135,47 @@ public String getCallerAvatarUri(int sizePx) {
...
+            }
+            token = mmkv.decodeString(key);
         }
-        
-        String token = mmkv.decodeString(key);
-        
+
+        if (token == null || token.isEmpty()) {
+            // Legacy fallback (pre-migration). Safe because the exported-receiver vector that
+            // abused this ambiguity is closed (ReplyBroadcast/DismissNotification are not exported).
```
diegolmello: Better than adding a comment, it's moving to a function with deprecated annotation. It should be clear, so it can be safely removed in the future.

Read `pre-migration` in 1 year. "What migration?"

android/app/src/main/AndroidManifest.xml
```diff
@@ -102,14 +102,19 @@
                 <action android:name="com.google.firebase.MESSAGING_EVENT" />
             </intent-filter>
         </service>
+        <!-- exported="false": these receivers are only ever fired by the app's own
+             notification action PendingIntents (explicit, component-targeted intents from
+             the same UID). Exporting them let any installed app broadcast a crafted intent
+             and have the app reply/send as the authenticated user (message impersonation /
+             token exfiltration). -->
```
diegolmello: This is explaining how `exported="false"` works 🤔

app/sagas/__tests__/deepLinking.test.ts
```diff
@@ -82,6 +82,12 @@ jest.mock('i18n-js', () => ({
 	default: { t: (k: string) => k }
 }));
 
+// Deep-link resume login asks for confirmation before authenticating. Default to confirming
+// (invoke onPress) so the existing login-path tests run; the decline path is covered explicitly.
```
diegolmello: Not needed.

app/sagas/__tests__/deepLinking.test.ts
```diff
@@ -306,6 +313,26 @@ describe('deepLinking saga — Regression race (new server + token + room path)'
 		expect(loginRequested()).toBe(true);
 	});
 
+	// Security: a deep-link resume token must not silently authenticate. If the user declines
+	// the confirmation, no login is attempted and the app lands outside (manual login screen).
```
diegolmello: Not needed.

diegolmello (changes requested):

app/sagas/init.js
OtavioStasiak: On the post-update launch, restore (triggered by APP.INIT) runs migrateTokenKeysToServerScoped first, before it reads ${TOKEN_KEY}-${server}. The migration is flag-guarded
  (TOKEN_KEY_SERVER_SCOPED_MIGRATED), moves the legacy token into the server-scoped slot, and only runs once.

  A push-notification open goes through the same server-selection path, and selectServer.ts also calls migrateTokenKeysToServerScoped before its server-scoped token read — so
  even if the deep link/push flow gets there before/instead of the normal restore, the token is already migrated. Result: the user stays logged in and lands in the right room;
  migration doesn't re-run on subsequent opens.

  The one exception is an ambiguous userId (shared by multiple servers): that legacy token is dropped, so those sessions re-authenticate — the user would hit login instead of
  the notification's room.

app/sagas/deepLinking.js
```diff
@@ -242,8 +260,24 @@ const handleOpen = function* handleOpen({ params }) {
 		}
 
 		if (params.token) {
```
diegolmello: **Consent gate sits downstream of the side effects it guards.**

**Severity:** high &nbsp;·&nbsp; **Likelihood:** certain — deterministic, not a race.

By the time this prompt appears, lines 256-259 have already dispatched `appStart(ROOT_OUTSIDE)`, `serverInitAdd(server)`, waited `delay(1000)`, and emitted `NewServer` for the attacker-supplied host. Declining returns with the app already pointed at that server, so the gate is advisory rather than preventive.

That misplacement is also the sole cause of the rest of this hunk: because the prompt overlaps an in-flight connection, `SERVER.SELECT_SUCCESS` / `METEOR.SUCCESS` can fire while it is open, which is why lines 268-280 need two `select`-then-`take` guards.

Two related problems on the same lines:
- The confirm sits *before* the `hostAlreadyConnected` branch is consulted, so a legitimate in-app resume deep link for the server you are already on also raises a security alert — contradicting the PR body ("for a server the user isn't already signed in to").
- The gate condition is implicit: `params.token` prompts unconditionally.

Steps to reproduce:
- Be on server A, not connected to server B.
- Fire `rocketchat://auth?host=B&token=<anything>`.
- Tap Cancel on the alert.
- Observed: the app is on the ROOT_OUTSIDE/login screen **for B**, and B has been added to the servers list.
- Expected: no state change for B at all.

Proposed solution: hoist the confirmation to the top of `handleOpen`, as soon as `params.token` is present and `host` differs from the current server — before `getServerInfo`, `serverInitAdd`, and `NewServer`. Then `if (!confirmed) return;` needs no `appStart`, and **both guards at 268-280 disappear** along with their comments. Note `app/sagas/__tests__/deepLinking.test.ts:265,294,326` are three tests that exist purely to cover this race; they should be deleted/rewritten, not preserved.

diegolmello (commented):
Two-axis review (standards + spec against the linked ticket), with each finding independently re-verified against the tree at e0d7a12. Every inline comment carries severity, likelihood, repro steps, and a proposed fix.

Highest value per line touched: hoisting the deep-link consent gate above the side effects it guards (it also deletes both new race guards and their comments), and collapsing the migration's three data structures into one group-by.

The finding that decides whether this PR actually closes the reported hole is the native one: the legacy fallback in `Ejson.java` and `MMKV.swift` keeps the `userId`-only lookup live in the notification handler, share extension, and Watch, with no way for native code to observe the migration flag. Worth settling that design ("flag-check natively" vs "migrate natively") before more code lands.

One thing not inline-able: the PR description says the migration copies every server first "so a userId shared by multiple servers isn't dropped", but `migrateTokenKeysToServerScoped.ts:38-40` deliberately *does* drop shared userIds, forcing re-auth. **The code is right and the description is wrong.** The description also still carries a conversational tail ("Want me to draft the How to test section too?") that should be stripped.

Also worth a maintainer's confirmation: 25 locale files were machine-translated in-PR — check that this matches the project's convention versus en-only plus the translator workflow.

Not a defect (raised and dismissed during review): `yield* call(...)` in `selectServer.ts` is correct — `call` comes from `typed-redux-saga` there.

android/app/src/main/java/chat/rocket/reactnative/notification/Ejson.java
```diff
@@ -135,35 +135,56 @@ public String getCallerAvatarUri(int sizePx) {
...
+                Log.d(TAG, "Looking up token with server-scoped key");
+            }
+            token = mmkv.decodeString(key);
         }
-        
-        String token = mmkv.decodeString(key);
-        
+
+        if (token == null || token.isEmpty()) {
+            token = decodeLegacyUserIdScopedToken(mmkv, userId);
```
diegolmello, line 165: **Native paths still resolve tokens by `userId` alone, so the primitive this PR closes stays open natively.**

**Severity:** high &nbsp;·&nbsp; **Likelihood:** medium — needs a shared or forced userId collision, which is precisely the reported attack.

`decodeLegacyUserIdScopedToken` (183-186) falls back to `reactnativemeteor_usertoken-{userId}` — exactly the ambiguous lookup the JS side now avoids — and it is live in the notification handler, the share extension, and `WatchConnection`. Same fallback on iOS at `ios/Shared/RocketChat/MMKV.swift:28`.

Why it matters: the JS migration's *global* view across all servers is what lets it tell an unambiguous userId from a shared one. Native has no such view, so on a shared userId it resolves to whichever server wrote last.

`TOKEN_KEY_SERVER_SCOPED_MIGRATED` is written by JS and read by nobody native (zero native references), so the `@deprecated` note at 181 has no removal trigger — "remove once the migration is universal" is unobservable from here.

Steps to reproduce:
- Install a build predating this change; log in to server A (userId X).
- Install this build; before the JS migration runs, deliver a push notification whose payload names a malicious host B and userId X.
- `token()` misses the server-scoped key for B, falls through to `decodeLegacyUserIdScopedToken`, and returns A's token.

Proposed solution: read the migration flag from the same MMKV instance here and in `MMKV.swift`, and refuse the legacy fallback once it is set. Cheaper than migrating natively, and it gives the deprecation a real, checkable condition.

ios/Shared/RocketChat/MMKV.swift
```diff
@@ -18,11 +18,14 @@ extension MMKVBridge {
...
-			return nil
+	// Server-scoped key: reactnativemeteor_usertoken-{server}-{userId}. Keep in sync with
+	// getUserTokenKey() (JS) and Ejson.token() (Android); falls back to the legacy
+	// userId-only slot until the JS migration runs.
+	func userToken(for userId: String, server: String) -> String? {
+		if let userToken = string(forKey: "reactnativemeteor_usertoken-\(server)-\(userId)") {
+			return userToken
 		}
-		return userToken
+		return string(forKey: "reactnativemeteor_usertoken-\(userId)")
```
diegolmello, line 32: **Same legacy `userId`-only fallback as `Ejson.java:185`.**

**Severity:** high &nbsp;·&nbsp; **Likelihood:** medium — same conditions as the Android half.

`return string(forKey: "reactnativemeteor_usertoken-\(userId)")` reinstates the ambiguous lookup for the share extension and `WatchConnection`, with no reference to `TOKEN_KEY_SERVER_SCOPED_MIGRATED`.

Proposed solution: gate this fallback on the migration flag read from the same MMKV instance, as described on the Android comment.

app/lib/methods/migrateTokenKeysToServerScoped.ts
```diff
@@ -0,0 +1,57 @@
...
+export const migrateTokenKeysToServerScoped = async (): Promise<void> => {
+	try {
+		if (UserPreferences.getBool(TOKEN_KEY_SERVER_SCOPED_MIGRATED)) {
+			return;
+		}
+		const serversDB = database.servers;
+		const servers = await serversDB.get('servers').query().fetch();
+
+		// Map each server to its userId and count how many servers reference each userId.
+		const serverUserIds: { server: string; userId: string }[] = [];
```
diegolmello: **Three data structures for one group-by.**

**Severity:** low — maintainability only, no user-visible impact. &nbsp;·&nbsp; **Likelihood:** n/a.

`serverUserIds` (20) + `serverCountByUserId` (21) + `legacyKeys` (33), across two index `for` loops, together express: group servers by userId, migrate the singletons, drop every legacy key touched.

Proposed solution: a single `Map<userId, server[]>`. The ambiguity rule then reads off `servers.length > 1` directly, and the `Set` becomes unnecessary because a userId is a key exactly once — roughly half the lines. Also, `serverUserIds` is a misleading name for a list of `{server, userId}` pairs.

Keep the *why* comment about ambiguity at 37 (load-bearing); the comments at 19, 32 and 51 restate the code beneath them.

app/sagas/selectServer.ts
```diff
@@ -150,6 +151,10 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
 		yield put(inquiryReset());
 		yield put(encryptionStop());
 		yield put(clearActiveUsers());
+		// Deep-link and share-extension startup can dispatch selectServerRequest() directly, without
+		// going through appInit(). Run the (idempotent, flag-guarded) token migration here too so the
+		// server-scoped read below finds legacy sessions instead of falling through to a login screen.
+		yield* call(migrateTokenKeysToServerScoped);
 		const userId = UserPreferences.getString(`${TOKEN_KEY}-${server}`);
```
diegolmello: **Key namespace is only half-abstracted.**

**Severity:** low-medium — invites future confusion in exactly the area being hardened. &nbsp;·&nbsp; **Likelihood:** n/a.

This PR adds `getUserTokenKey(server, userId)` but leaves the sibling family `${TOKEN_KEY}-${server}` — which stores a **userId**, not a token — inlined at 13 production sites: `app/sagas/init.js:30,43`; `app/sagas/login.js:346,391,458`; `app/sagas/deepLinking.js:166,214,342`; `app/lib/methods/logout.ts:25,41,74`; this line; `app/views/RoomsListView/components/ServersList.tsx:80`; plus the migration itself.

Net effect: the `reactnativemeteor_usertoken-` prefix now carries three different meanings across three languages, one of which is abstracted and two of which are not. The migration reads the raw key while writing through the helper — asymmetric.

Proposed solution: add `getServerUserIdKey(server)` to `app/lib/constants/keys.ts` and route all sites through it. That also gives the triplicated "keep in sync" note (`keys.ts:26-31`, `Ejson.java:152-154`, `MMKV.swift:21-23`) one authoritative home instead of three peers pointing at each other.

app/sagas/deepLinking.js
```diff
@@ -37,6 +38,23 @@ const roomTypes = {
 	channels: 'l'
 };
 
+const confirmDeepLinkLogin = (host, params = {}) =>
+	new Promise(resolve => {
+		// Under E2E tests, auto-confirm so flows don't have to dismiss the native Alert.
+		// `e2eConfirmPrompt=true` opts back into the real prompt to test it. Test-only; real users always get the prompt.
+		if (process.env.RUNNING_E2E_TESTS === 'true' && params.e2eConfirmPrompt !== 'true') {
```
diegolmello: **An attacker-controllable URL parameter steers a security prompt.**

**Severity:** medium — a test-only field added to the production URL contract on the exact path being hardened. &nbsp;·&nbsp; **Likelihood:** low in production; the `RUNNING_E2E_TESTS` env gate must also be true.

The env gate itself matches repo practice and is fine. `params.e2eConfirmPrompt` is different in kind: it makes which branch of a consent check runs a function of untrusted deep-link input, and it is test-only code in a non-test file.

Steps to reproduce:
- Build with `RUNNING_E2E_TESTS=true`.
- Fire `rocketchat://auth?host=…&token=…` — no prompt, silent login.
- Add `&e2eConfirmPrompt=true` — prompt appears. The link, not the app, decided.

Proposed solution: a second env var, or drop the Maestro case entirely — both branches are already covered directly by unit tests at `app/sagas/__tests__/deepLinking.test.ts:379,397`.

app/i18n/locales/en.json
```diff
@@ -230,6 +230,8 @@
   "Dark": "Dark",
   "Dark_level": "Dark level",
   "decline": "Decline",
+  "Deep_link_login_description": "A link is asking to sign you in to {{server}}. Only continue if you opened this link yourself and trust it.",
```
diegolmello, line 234: **The consent dialog omits the account being signed in as.**

**Severity:** medium &nbsp;·&nbsp; **Likelihood:** certain — every prompt is affected.

`Deep_link_login_description` interpolates only `{{server}}`, but the ATO chain in NATIVE-1342 is a **userId collision** — the ticket's own wording is "Log in to `host` as `userId`?". A user cannot spot a foreign account from the server name alone, so the prompt does not surface the signal it exists to surface.

Proposed solution: include the userId/username from the link in the message. Touches the 25 locale files this PR already changes.

app/lib/constants/keys.ts
```diff
@@ -23,5 +23,13 @@ export const ALERT_DISPLAY_TYPE_PREFERENCES_KEY = 'RC_ALERT_DISPLAY_TYPE_PREFERE
 export const CRASH_REPORT_KEY = 'RC_CRASH_REPORT_KEY';
 export const ANALYTICS_EVENTS_KEY = 'RC_ANALYTICS_EVENTS_KEY';
 export const TOKEN_KEY = 'reactnativemeteor_usertoken';
+/**
```
diegolmello: **Explanatory comments run against the repo's code-carries-its-own-meaning convention.**

**Severity:** low &nbsp;·&nbsp; **Likelihood:** n/a — style, not behaviour.

The repo convention (and the reviewer's coding-style rules) is that code carries its own meaning: keep license headers, required pragmas, and links to upstream issues, and rewrite rather than annotate otherwise.

This 6-line block explains the vulnerability rather than the key. The same pattern recurs across the diff: `migrateTokenKeysToServerScoped.ts` (file JSDoc + 4 inline), `selectServer.ts:154-156`, `deepLinking.js:43-44,269-271,276`, `logout.ts:28`, `Ejson.java:152-154`, `MMKV.swift:21-23`.

Proposed solution: keep the load-bearing *why* (the ambiguity rule in the migration; the `@deprecated`/`keep in sync` notes that cross a language boundary and have no code expression) and drop the rest. The rationale for the whole change belongs in the PR body and the ticket, where it stays current.

app/lib/methods/logout.ts
```diff
@@ -17,6 +24,8 @@ import { _activeUsersSubTimeout } from './getUsersPresence';
 function removeServerKeys({ server, userId }: { server: string; userId?: string | null }) {
 	UserPreferences.removeItem(`${TOKEN_KEY}-${server}`);
 	if (userId) {
+		UserPreferences.removeItem(getUserTokenKey(server, userId));
+		// Also remove the legacy non-server-scoped token slot, in case this server predates the migration.
 		UserPreferences.removeItem(`${TOKEN_KEY}-${userId}`);
```
diegolmello: **Dead code, not a scoping bug.**

**Severity:** low &nbsp;·&nbsp; **Likelihood:** n/a — unreachable post-migration.

This removes the *unscoped* legacy slot `${TOKEN_KEY}-${userId}`. The migration runs before any reachable logout (from both `app/sagas/init.js:27` and `app/sagas/selectServer.ts:157`) and deletes that slot for migrated and ambiguous userIds alike — so post-migration there is nothing here to remove. Its only reachable effect on a pre-migration install would be clobbering a *different* server's session that happens to share the userId.

Proposed solution: delete the line (and its comment) rather than scoping it.

app/i18n/locales/en.json
OtavioStasiak: I disagree, we should not expose the userId.

app/sagas/deepLinking.js
OtavioStasiak: it doesn't makes sense... wee need to have it to keep other tests working as expected.

app/lib/methods/migrateTokenKeysToServerScoped.ts
```diff
@@ -0,0 +1,71 @@
...
+			if (sharing.length > 1) {
+				UserPreferences.removeItem(legacyKey);
+				return;
+			}
+			const newKey = getUserTokenKey(sharing[0], userId);
+			const token = UserPreferences.getString(legacyKey);
+			if (token && !UserPreferences.getString(newKey)) {
+				UserPreferences.setString(newKey, token);
+			}
+			UserPreferences.removeItem(legacyKey);
```
diegolmello, line 51: **A rollback to an older build silently kills push decryption.**

**Severity:** medium-high &nbsp;·&nbsp; **Likelihood:** low — needs a rollback, but staged rollbacks do happen.

The migration moves the token rather than copying it, and the `removeItem` is unconditional — it runs even on the path where the copy above was skipped:

```ts
const token = UserPreferences.getString(legacyKey);
if (token && !UserPreferences.getString(newKey)) {
    UserPreferences.setString(newKey, token);
}
UserPreferences.removeItem(legacyKey);
```

**Steps to reproduce**
1. Install this build and launch it once, so the migration runs and the legacy slot is deleted.
2. Roll back to the previous build (TestFlight, or a Play staged rollback).
3. Send a push. It arrives but cannot be decrypted.

The app itself still resumes, because the pre-PR `selectServer` prefers the Logged User's token from the WatermelonDB `users` row and this PR doesn't touch that. But the native readers — the iOS Notification Service Extension, the Watch app, Android `Ejson` — have no database access, so notifications stay broken until the User signs in again.

No code change is strictly required, but it should be a conscious call rather than a surprise: staged rollout, plus a release note that rolling back forces a re-login before notifications work.

app/lib/methods/migrateTokenKeysToServerScoped.ts
```diff
@@ -0,0 +1,71 @@
...
+			} else {
+				serversByUserId.set(userId, [server]);
+			}
+		}
+
+		serversByUserId.forEach((sharing, userId) => {
+			const legacyKey = getLegacyUserTokenKey(userId);
+			// A userId claimed by more than one server is ambiguous: drop the legacy slot instead of
+			// migrating it, so the session re-authenticates.
+			if (sharing.length > 1) {
```
diegolmello, line 42: **Two Server rows sharing a `userId` lose native push on both, and it's untested.**

**Severity:** medium &nbsp;·&nbsp; **Likelihood:** low-medium — needs duplicate or alias Server rows for one User.

```ts
if (sharing.length > 1) {
    UserPreferences.removeItem(legacyKey);
    return;
}
```

`serversByUserId` is built from `getServerUserIdKey(server)` across every row in `servers`, so two Server rows that resolve to the same `userId` reach this branch and **no scoped slot is ever written for either one**.

**Steps to reproduce**
1. On a pre-PR build, add the same Server twice under two URLs that both resolve (`https://x.com` and `https://x.com:443`, or a stale alias row still sitting in `servers`).
2. Sign in to both as the same User.
3. Update to this build and launch.
4. The legacy slot is dropped without migrating; neither Server row has a scoped token.

**What actually breaks:** not a logout — `app/sagas/selectServer.ts:155-159` resolves the Logged User's credentials from the WatermelonDB `users` row first and only falls through to `getUserTokenKey` when there's no record, so the session survives on both Servers. What dies is native push decryption on both, until re-login. Same failure class as the unconditional legacy-slot delete below.

The behaviour itself is defensible — keeping an ambiguous slot *is* the vulnerability. What's missing is that it's undocumented and unverified on device. Worth adding to "How to test" and confirming it degrades to a working app with undecryptable pushes, rather than a crash or a Meteor Connect retry loop.

app/sagas/deepLinking.js
```diff
@@ -221,6 +246,15 @@ const handleOpen = function* handleOpen({ params }) {
 		} catch (e) {
 			// do nothing?
 		}
+		// Consent before touching anything on the deep link's server: a resume token means this
```
diegolmello: **The consent check is copy-pasted across both deep link entry points.**

**Severity:** medium, latent &nbsp;·&nbsp; **Likelihood:** high that it bites eventually.

This block and the one at `:368-375` are byte-identical, and they are the only two call sites of `confirmDeepLinkLogin`:

```js
if (params.token) {
    const confirmed = yield call(confirmDeepLinkLogin, host, params);
    if (!confirmed) {
        yield declineDeepLinkLogin();
        return;
    }
}
```

The failure mode is invisible: a third deep link entry point added later silently skips the consent prompt, and nothing breaks — the guard is just absent. The duplicated explanatory comment above each copy is the same duplication showing through.

Suggestion: one `resumeLoginFromDeepLink(host, params)` saga that owns consent plus `loginRequest`, called from both sites, so future entry points inherit the guard rather than having to remember it.

app/lib/constants/keys.ts
```diff
@@ -23,5 +23,9 @@ export const ALERT_DISPLAY_TYPE_PREFERENCES_KEY = 'RC_ALERT_DISPLAY_TYPE_PREFERE
 export const CRASH_REPORT_KEY = 'RC_CRASH_REPORT_KEY';
 export const ANALYTICS_EVENTS_KEY = 'RC_ANALYTICS_EVENTS_KEY';
 export const TOKEN_KEY = 'reactnativemeteor_usertoken';
+export const getServerUserIdKey = (server: string): string => `${TOKEN_KEY}-${server}`;
+export const getLegacyUserTokenKey = (userId: string): string => `${TOKEN_KEY}-${userId}`;
+export const getUserTokenKey = (server: string, userId: string): string => `${TOKEN_KEY}-${server}-${userId}`;
+export const TOKEN_KEY_SERVER_SCOPED_MIGRATED = 'RC_TOKEN_KEY_SERVER_SCOPED_MIGRATED';
```
diegolmello, line 29: **A permanent boolean migration marker with no removal plan.**

**Severity:** low &nbsp;·&nbsp; **Likelihood:** medium, by accretion.

`TOKEN_KEY_SERVER_SCOPED_MIGRATED` is a one-off boolean that three languages must interpret identically, declared three times — here, `Ejson.java:27`, and as a bare string literal in `MMKV.swift:29` — and read from `migrateTokenKeysToServerScoped.ts:17`, `Ejson.java:187`, `MMKV.swift:29`. Nothing triggers its removal, so it becomes permanent surface area.

Suggestion: an integer `RC_KV_MIGRATION_VERSION` that the next key-value migration reuses, with the native side gating on `>= N`. If the flag ends up being written natively at process start (see the `Ejson.java` thread), most of this machinery goes away regardless.

app/sagas/selectServer.ts
```diff
@@ -150,7 +151,8 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
 		yield put(inquiryReset());
 		yield put(encryptionStop());
 		yield put(clearActiveUsers());
-		const userId = UserPreferences.getString(`${TOKEN_KEY}-${server}`);
+		yield* call(migrateTokenKeysToServerScoped);
```
diegolmello, line 154: **Mixed `call` idioms in one file.**

**Severity:** low &nbsp;·&nbsp; nit.

`yield* call(migrateTokenKeysToServerScoped)` uses `call` from `typed-redux-saga` (imported at `:6`) while the rest of the file imports its effects from `redux-saga/effects` (`:1`). Works fine; just inconsistent within a single file.

diegolmello (commented):
## Security re-review against 5e04c0d

Re-checked every finding from the previous pass against the current head, including the threads that were already resolved. Most of them really are fixed — details below so nothing gets re-litigated.

### One resolved thread should be reopened

The native `userId`-only token fallback is **still live** on both platforms and the flag that gates it has exactly one writer, in JS. I've replied in the `Ejson.java` and `MMKV.swift` threads with the full reasoning, including why the "restore on `APP.INIT` runs the migration early enough" argument doesn't cover a cold data-only push.

### Two items that have no diff hunk to anchor to

- **Dead constant.** `LoadNotification.java:65` still declares `private String TOKEN_KEY = "reactnativemeteor_usertoken-";`. It is the only occurrence in the file and is never used — the file's actual token access goes through `ejson.token()` (`:96`) and the `x-auth-token` header (`:128`). Two native declarations of the same key is how the next drift starts; worth deleting while it's cheap.
- **Scope.** The diff is 395 files against `develop` — fonts, a UIKit Button, snapshots, CI scripts — plus a Meteor Connect race fix at `deepLinking.js:281-285` (the `take(types.METEOR.SUCCESS)` wait) and a new 183-line `app/sagas/__tests__/deepLinking.test.ts`. The Meteor Connect fix looks correct, it's just not security. Splitting the unrelated churn out would make the security diff reviewable on its own. Not blocking, and I realise some of this may be unavoidable at this point in the branch's life.

### Confirmed fixed — no action

- **Consent gate ordering.** Now strictly ahead of `getServerInfo` (`:259`), `serverInitAdd`/`NewServer` (`:271-276`) and `loginRequest` (`:287`), at both entry points. Declining runs only `showToast` + `fallbackNavigation()`, leaving no Server row, no Meteor Connect, no token. It also now sits after the already-connected branch, so a resume link for a Server the User is signed into no longer prompts.
- **Exported receivers.** `ReplyBroadcast` and `DismissNotification` flipped to `exported="false"`; the other two were already false. The remaining exported components are MainActivity, ShareActivity, and CallKeep's service behind `BIND_TELECOM_CONNECTION_SERVICE`.
- **Key namespace.** All the JS call sites route through the `keys.ts` helpers now; the only hand-built literals left are native, where sharing a helper isn't possible.
- **Migration correctness.** Every branch drops the legacy slot, the ambiguous-`userId` case is handled, the orphan sweep is correctly anchored on the `userId` shape, and the flag is only set on success so a throw retries next launch. Tests cover all four branches.
- **Logging.** Every key-revealing `Ejson` log is `BuildConfig.DEBUG`-gated and prints key names, not values; the ungated warnings carry no key or token and `:214` sanitises the URL.

### Withdrawing one earlier comment

The `RUNNING_E2E_TESTS` branch in `confirmDeepLinkLogin` — @OtavioStasiak was right and I was wrong to push on it. I checked the mechanism properly this time: nothing inlines the variable (`babel-preset-expo` only rewrites `EXPO_PUBLIC_*`, and `react-native-dotenv` only rewrites `@env` imports), but in a release bundle it's a runtime read against RN's polyfilled `process.env`, which holds only `NODE_ENV` — so the gate can never open, and `forceLoginPrompt !== 'true'` means a deep link parameter can only turn the prompt *on*. Removing it would also genuinely break the 51 flows that go through `login-with-deeplink.yaml`, exactly as you said. Not a security issue, and the same idiom is used in 14 other files.

I also withdraw two smaller points from my earlier notes: the orphan sweep is correctly anchored (a Server id always carries a dot, a scheme or a path separator), and the comment-line churn was already your own feedback.

android/app/src/main/java/chat/rocket/reactnative/notification/Ejson.java
diegolmello: Reopening this one — the gate landed, but the underlying window is still open, and I want to lay out the reasoning rather than just re-resolve it.

The fallback is still reachable whenever the flag is false:

```java
private String decodeLegacyUserIdScopedToken(MMKV mmkv, String userId) {
    if (mmkv.decodeBool(TOKEN_KEY_SERVER_SCOPED_MIGRATED, false)) {
        return null;
    }
    return mmkv.decodeString(TOKEN_KEY.concat(userId));
}
```

`RC_TOKEN_KEY_SERVER_SCOPED_MIGRATED` has exactly one writer in the whole repo: `app/lib/methods/migrateTokenKeysToServerScoped.ts:67`, in JS. Grepping `android/` and `ios/` finds only reads — `Ejson.java:187` and `MMKV.swift:29`. So the flag is `false` from the moment an update is installed until JS has run at least once.

On the counter-argument that `restore` on `APP.INIT` runs the migration early enough — that holds for app *opens*, but not for a notification handled before any JS starts. `RCFirebaseMessagingService.onMessageReceived` is a plain `FirebaseMessagingService`: it constructs `CustomPushNotification` and calls `onReceived()`, which reaches `LoadNotification.load` → `ejson.token()` (`LoadNotification.java:96`). Nothing on that path starts or waits on the RN JS context. A data-only push delivered after an update but before the User's first launch is handled entirely natively, with the flag still `false` and the ambiguous lookup live. The iOS Notification Service Extension has the same structure.

**Steps to reproduce**
1. On a build *before* this PR, consent once to a malicious Server that shares your `userId` and let it register a push token — it writes `reactnativemeteor_usertoken-<malhost>` = your `userId`, and no scoped token.
2. Sign in to the legitimate Server. The legacy slot now holds the legitimate token.
3. Install this build over it. **Do not launch the app.**
4. The malicious Server sends a push with `ejson.host=<malhost>`.
5. `Ejson.userId()` keys on the push-declared `serverURL()` and returns the `userId`; the scoped lookup at `:156` misses; the legacy slot returns the **legitimate** token; `LoadNotification.java:94-96` sends it to the attacker's host as `x-auth-token`.

The precondition is narrow — the User must previously have consented to and signed in to the malicious Server on a pre-PR build — so this is medium likelihood, not high. But it's the exact primitive the PR exists to close, and it reopens on every update.

Worth saying explicitly: the fallback can't simply be deleted, because it's also what keeps push decryption working for honest Users in that same window. The fix that closes both is writing the flag **natively** at process start — `MainApplication.onCreate` on Android, app launch on iOS — before any `Ejson` or Notification Service Extension read. Once that's in, the fallback, the flag gate, and the orphan sweep that only exists to service it can all go in a follow-up.

ios/Shared/RocketChat/MMKV.swift
diegolmello: Same as the Android side — reopening with the reasoning rather than re-resolving.

```swift
if bool(forKey: "RC_TOKEN_KEY_SERVER_SCOPED_MIGRATED") {
    return nil
}
return string(forKey: "reactnativemeteor_usertoken-\(userId)")
```

The flag's only writer anywhere is JS (`app/lib/methods/migrateTokenKeysToServerScoped.ts:67`), so it reads `false` from the moment an update is installed until JS has run once. The Notification Service Extension, the share extension and the Watch app all read tokens without any RN context, so a notification arriving in that window takes the legacy ambiguous path — same exposure as the `Ejson.java` thread, same repro.

Writing the flag natively at app launch, before any Notification Service Extension read, closes it on this side too.

app/i18n/locales/en.json
diegolmello: Agreed on not printing a raw `userId` — that objection is fair, and not just on privacy grounds: an opaque `userId` gives the User nothing they can actually judge.

But the prompt then carries no account signal at all, and the case NATIVE-1342 describes is precisely one where the *Server* looks fine and the *User* is the surprise. There's a cheaper middle option already in hand: `getServerUserIdKey(host)` is imported in this file (`app/sagas/deepLinking.js:12`, used at `:342`), so the saga can read the `userId` already stored locally for that Server and warn only when the link's `userId` differs from the Logged User for that Server. No new disclosure, no network call before consent, and nothing shown in the common case where they match.

Resolving the raw-`userId` version of this; flagging the differs-from-Logged-User variant as the thing worth considering instead.

app/sagas/selectServer.ts
OtavioStasiak: The file imports call from typed-redux-saga (:6) and uses yield* call(...) at every call site — :100, :106, :115, :117, :122, :159 and the new `yield*  call(migrateTokenKeysToServerScoped)` at :154 matches them. redux-saga/effects at :1 only supplies put/takeLatest. Nothing to unify

app/lib/constants/keys.ts
OtavioStasiak: Leaving it as a boolean for now. Swapping it for an integer RC_KV_MIGRATION_VERSION still means one constant declared in three languages and read in three places. it only pre-pays for a second migration.
 -  The flag itself can't go away, though: it's what stops the native fallbacks (MMKV.swift:29, Ejson.java:187) from reading a legacy slot that the sweep's  isLegacyUserTokenKey heuristic missed — an orphan key whose userId isn't bare-alphanumeric and whose server row is gone. That's the blind spot 33003ef72 deliberately traded for.

OtavioStasiak:
@coderabbit review

app/sagas/selectServer.ts
```diff
@@ -171,7 +173,7 @@ const handleSelectServer = function* handleSelectServer({ server, version, fetch
 					requirePasswordChange: userRecord.requirePasswordChange
 				};
 			} else {
-				const token = UserPreferences.getString(`${TOKEN_KEY}-${userId}`);
+				const token = UserPreferences.getString(getUserTokenKey(server, userId));
```
diegolmello: **[P1] Make the workspace-scoped token the only credential used to restore a session**

The new lookup here only runs when there is **no** cached user record. When a record exists, line 163 still takes `userRecord.token` instead.

### Plain-language explanation

Think of two workspaces as two houses whose residents happen to have the same name. This PR creates a separate key drawer for each house, but the normal login path still grabs a key from the old drawer labelled only with the resident's name. It can pick the other house's key even though the correct key is available.

`getLoggedUserById(userId)` reads the shared `database.servers.users` table using only the user ID. `handleLoginSuccess` in `login.js` also updates that row using only the user ID. Logging into another workspace with the same ID overwrites the row's token. `selectServer` copies that token into Redux, and the connected handler in `app/lib/services/connect.ts` submits it through `loginRequest({ resume: user.token })`.

This leaves the PR requirement “Scope the auth token to (server, userId)” incomplete. It also bypasses the migration's decision to discard ambiguous legacy tokens and require authentication.

### Steps to reproduce

Use two controlled test workspaces, A and B, with accounts sharing the same user ID but having different tokens:

1. Log into B, then A. The shared database user row now contains A's token.
2. Confirm the new MMKV entries contain the correct, separate tokens for both A and B.
3. Switch back to B.
4. `getLoggedUserById` returns the shared row containing A's token. Since the row exists, B's scoped-token lookup is skipped.
5. The connection to B attempts to resume with A's token. Expected: only B's scoped token is used. If B has no scoped token, show the authentication flow.

I reproduced the selection step with a temporary Jest test using the existing `selectServer.test.ts` harness. It expected `target-workspace-token` but received `other-workspace-token`. The existing five related suites (68 tests) passed, so their current coverage does not catch this collision. This was a saga-level reproduction, not a live two-server test.

### Proposed change

Read the scoped token before restoring the cached profile, and require that token for authentication. Keep the existing explicit profile fields; replace the credential-resolution block with:

```ts
const userId = UserPreferences.getString(getServerUserIdKey(server));
const token = userId ? UserPreferences.getString(getUserTokenKey(server, userId)) : null;
let user = null;
if (userId && token) {
    const userRecord = yield* call(getLoggedUserById, userId);
    user = userRecord
        ? {
              id: userRecord.id,
              token,
              username: userRecord.username,
              name: userRecord.name,
              language: userRecord.language,
              status: userRecord.status,
              statusText: userRecord.statusText,
              roles: userRecord.roles,
              avatarETag: userRecord.avatarETag,
              bio: userRecord.bio,
              nickname: userRecord.nickname,
              requirePasswordChange: userRecord.requirePasswordChange
          }
        : { token };
}
```

The existing `if (user) ... else ...` below can handle authentication versus the outside/login flow. No new service layer or broad database refactor is needed for this credential fix. The shared profile record itself remains outside this narrow fix.

### Regression coverage

Add this case to `app/sagas/__tests__/selectServer.test.ts`, using its existing mocks and helpers:

```ts
it('uses the workspace token when a cached user has another workspace token', async () => {
    UserPreferences.setString(`${TOKEN_KEY}-${SERVER_URL}`, USER_ID);
    UserPreferences.setString(getUserTokenKey(SERVER_URL, USER_ID), 'target-workspace-token');
    jest.mocked(getLoggedUserById).mockResolvedValue({
        id: USER_ID,
        token: 'other-workspace-token'
    } as any);

    const { store } = setupStore();
    store.dispatch(selectServerRequest(SERVER_URL, '7.0.0', false));
    await flushSagaMicrotasks();

    expect(store.getState().login.user.token).toBe('target-workspace-token');
});
```

Also cover a cached user row with **no scoped token**: it must not restore the cached token. Add an integration case that runs the ambiguous-ID migration and then selects a workspace, verifying that the discarded credentials cannot be recovered from the shared user row. Update existing successful-restoration fixtures to seed a scoped token explicitly.

diegolmello (approved):

PR #7451, opened 2026-06-26
perf: memoize @json model fields

diegolmello:
## Proposed changes

Switch every `@json`-decorated WatermelonDB model field to the memoized form, `@json('col', sanitizer, { memo: true })`, across all 11 models (67 fields total). With memoization the decorator caches the `[rawValue, sanitized]` pair and returns the same object reference on every read while the underlying raw column string is unchanged. Today this is render-neutral; it sets up per-field re-render bailout for the consumers that adopt reference equality in later work.

Because the memo cache is invalidated only when the raw column string changes, mutating a memoized value in place would silently corrupt the cache (the mutated object is returned again with a stale identity). This PR converts the two remaining in-place mutations of these fields to replacement:

- `ReactionsList` sorts a **copy** (`[...reactions].sort(...)`) instead of sorting the `reactions` field in place.
- The encryption quote-attachment decryption **reassigns** `message.attachments = [...]` instead of `push()`-ing into the existing array.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1346

## How to test or reproduce

No behaviour change is expected — this is a reference-stability refactor. To confirm there is no re-render regression:

1. Open a room with reactions and an encrypted quoted reply; verify reactions render and sort correctly and the quoted attachment decrypts.
2. Profile a representative room before and after the change (React DevTools). Measured here on device: target `MessageContainer` had **0 re-renders** (first-mount only) and the `RoomView` control had **0 re-renders** in all trials of both the baseline and the memoized bundle. Commit totals stayed in the same band. The swap is render-neutral; the benefit lands when consumers start relying on the stable references.

`pnpm lint` (0 errors) and `TZ=UTC pnpm test` (1770 passing) are green locally.

## Screenshots

N/A — no UI changes.

## Types of changes

## Checklist

## Further comments

The two mutation fixes are required for correctness once `{ memo: true }` is in place — without them a sort or an attachment append would return a stale cached reference. They are intentionally bundled with the memo swap rather than split out, since in isolation they are no-ops.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved data handling across messages, threads, rooms, users, permissions, and custom emojis to avoid unnecessary reprocessing and side effects.
  * Fixed a reactions list update so it no longer changes the original input while sorting.
  * Made quote attachment handling safer by appending data immutably, helping prevent attachment-related inconsistencies.

### DIFF
--- a/app/containers/ReactionsList/index.tsx
+++ b/app/containers/ReactionsList/index.tsx
@@ -30,7 +30,7 @@ const useRoutes = (reactions: IReaction[] | undefined) => {
 		};
 	}
 
-	const sortedReactions = reactions?.sort((reaction1, reaction2) => reaction2.usernames.length - reaction1.usernames.length);
+	const sortedReactions = [...reactions].sort((reaction1, reaction2) => reaction2.usernames.length - reaction1.usernames.length);
 	const routes: IRoute[] = sortedReactions.map(reaction => ({
 		key: reaction.emoji,
 		title: reaction.emoji,
--- a/app/lib/database/model/CustomEmoji.js
+++ b/app/lib/database/model/CustomEmoji.js
@@ -10,7 +10,7 @@ export default class CustomEmoji extends Model {
 
 	@field('name') name;
 
-	@json('aliases', sanitizer) aliases;
+	@json('aliases', sanitizer, { memo: true }) aliases;
 
 	@field('extension') extension;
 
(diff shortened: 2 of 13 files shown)


diegolmello:
Closing in favor of #7455, which carries these commits as part of the NATIVE-22 message-hooks work (the useMessage snapshot hook builds directly on this @json memoization).

PR #7467, opened 2026-07-02
fix: hold keyboard focus on the rooms list for hardware-keyboard navigation

diegolmello:
## Proposed changes

On the rooms list, when a screen reader or an external (hardware) keyboard is connected, keyboard focus was only established once — on mount — through the drawer (hamburger) button's `autoFocus`. After returning from a room, or after the list re-rendered asynchronously, nothing held focus. With focus unset, the first D-pad key press let the platform focus search reach the off-screen, closed navigation drawer, and pressing Enter opened Settings instead of the intended room.

This makes the rooms list re-assert keyboard focus onto the drawer button whenever the screen regains focus and accessibility navigation is active:

- `RoomsListView` now uses `useFocusEffect` + `InteractionManager.runAfterInteractions` to move focus onto the drawer button after the navigation transition and any pending list work settle. It is gated on `useIsAccessibilityNavigationEnabled()` (screen reader or external keyboard), so touch users are never affected, and the task is cancelled on cleanup.
- `HeaderButton.Drawer` is now a `forwardRef` that forwards its ref to the keyboard-focus-wrapped item, exposing the imperative `focus()` handle. `autoFocus` is unchanged and the ref is opt-in, so every other `HeaderButton.Drawer` caller behaves exactly as before.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1263

## How to test or reproduce

Requires an Android emulator with a physical keyboard enabled (`hw.keyboard=yes`).

1. Connect/enable a hardware keyboard and open the app on the rooms list.
2. Without touching the screen, press D-pad `Right, Right, Down, Down, Enter`.
3. Expected: a room opens and the message composer is visible.
4. Before this change, focus was unset so the sequence could open Settings instead of a room.

Covered by the `keyboard-navigation-room.yaml` Maestro flow (shard 14, keyboard-enabled AVD), which fails before this change and passes after it.

## Types of changes

## Checklist

## Further comments

The keyboard-navigation Maestro flow only asserts that some room opened; a follow-up could assert the specific target room to tighten it. Separately, hiding the closed drawer's contents from screen-reader (TalkBack/VoiceOver) traversal is a distinct accessibility concern that should be addressed and verified on-device with a screen reader in its own change.

## Summary by CodeRabbit

* **Accessibility**
  * Improved keyboard and screen-reader navigation in the header by keeping the drawer button focused when returning to the screen.
  * Header controls now behave more consistently for accessibility users, with no change to the visible layout or primary action.

### DIFF
--- a/app/containers/Header/components/HeaderButton/Common.tsx
+++ b/app/containers/Header/components/HeaderButton/Common.tsx
@@ -1,7 +1,7 @@
-import { memo } from 'react';
+import { forwardRef, memo } from 'react';
 import { StackActions, useNavigation } from '@react-navigation/native';
 import { type StyleProp, type ViewStyle } from 'react-native';
-import { withKeyboardFocus } from 'react-native-external-keyboard';
+import { type KeyboardFocus, withKeyboardFocus } from 'react-native-external-keyboard';
 
 import I18n from '../../../../i18n';
 import { isIOS } from '../../../../lib/methods/helpers/deviceInfo';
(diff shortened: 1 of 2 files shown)


OtavioStasiak (approved):
LGTM

PR #7469, opened 2026-07-06
fix: ignore user not working reliably on e2e tests

diegolmello:
## Proposed changes

Fixes the `ignore-user` E2E flakiness for good. Two independent failure modes were found and fixed:

1. **Block/Ignore silently failing with an undefined user id.** `RoomActionsView` fetches the DM member asynchronously and forwards it to `RoomInfoView` on navigation. When navigation happens before that fetch resolves, `member` arrives as a non-empty object (presence fields only) without `_id`. `RoomInfoView.loadUser` gated its own fetch on `isEmpty(member)`, so the id stayed undefined forever, and tapping Block/Ignore called the server with `blocked: undefined`. The `blockUser` Meteor method answers `Internal server error [500]` for that input (verified against the server with a raw DDP client), the error was swallowed, the button never toggled, and the test's 60s poll for "Unblock" timed out — matching the CI first-attempt failure exactly. The fix derives the target user id from the subscription `uids` (`getUidDirectMessage`) whenever `member._id` is missing, gates `loadUser` on the missing `_id` instead of `isEmpty`, and bails out rather than calling the server with undefined. Same fallback applied to `RoomActionsView.toggleBlockUser`.

2. **Username-header tap cancelled by list re-render.** The flow taps the message author's header right after the two test messages are posted via API. When the tap is injected while the message list is still regrouping rows, the press gets cancelled mid-gesture and the user info screen never opens ("Ignore" never becomes visible). The flow now waits for the last sent message to render before tapping.

## Issue(s)	
https://rocketchat.atlassian.net/browse/NATIVE-1261

## How to test or reproduce

- `TZ=UTC pnpm test --testPathPattern='RoomInfoView/index.test'` — new regression test renders `RoomInfoView` with a `member` param lacking `_id` and asserts `toggleBlockUser` is called with the real user id (red before the fix).
- `maestro test .maestro/tests/room/ignoreuser.yaml -e APP_ID=<app-id>` against a dev build. Before the fix a local block/unblock toggle loop failed 3/3 with the 500 described above; after the fix it passed 3/3 (45 toggles) and the full flow passes repeatedly.

## Types of changes

## Checklist

## Further comments

The optimistic UI update from the previous commit remains useful, but it only masked the first failure mode when the member fetch won the race. The undefined-id path was the actual root cause of the remaining flakes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved block/ignore actions in DM room info so they work even when the other user’s ID isn’t loaded yet.
  * User details now load more reliably and preserve existing info while fetching missing data.
  * Ignore/block updates now return clearer success or failure results.

* **Tests**
  * Added coverage for the block/ignore flow in DM room info, including cases where the member ID is missing.
  * Updated an automation flow to wait for messages to fully render before continuing.

### DIFF
--- a/.maestro/tests/room/ignoreuser.yaml
+++ b/.maestro/tests/room/ignoreuser.yaml
@@ -84,6 +84,12 @@ tags:
       ROOM: ${output.room.name}
 - evalScript: ${output.utils.sendMessage(output.otherUser.username, output.otherUser.password, output.room._id, 'message-01')}
 - evalScript: ${output.utils.sendMessage(output.otherUser.username, output.otherUser.password, output.room._id, 'message-02')}
+# wait for the last message to render before tapping the header, otherwise the list
+# is still regrouping rows and the press gets cancelled mid-gesture
+- extendedWaitUntil:
+    visible:
+      id: 'message-content-message-02'
+    timeout: 60000
 - extendedWaitUntil:
     visible:
       id: 'username-header-${output.otherUser.username}'
--- a/app/lib/methods/helpers/handleIgnore.ts
+++ b/app/lib/methods/helpers/handleIgnore.ts
@@ -4,7 +4,7 @@ import EventEmitter from './events';
 import log from './log';
 import { ignoreUser } from '../../services/restApi';
 
-export const handleIgnore = async (userId: string, ignore: boolean, rid: string) => {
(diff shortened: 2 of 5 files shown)


Rohit3523 (approved):
LGTM

PR #7380, opened 2026-06-03
fix: keep live socket on deeplink login (prevent orphan WebSocket clobber)

diegolmello:
## Proposed changes

Standalone fix extracted from #7311 (commit `29e4f8b`).

Opening the app through an auth deeplink (and the call-push path) dispatched `loginRequest` **before** the WebSocket reached `connected`. The SDK login-guard then opened a second, orphaned socket. When the server later idle-timed-out the abandoned socket, its `close` flipped Redux to `disconnected` — so the app showed **"Waiting for network"** while a live, healthy socket still existed.

**SDK patch** (`patches/@rocket.chat+sdk+1.3.3-mobile.patch`, ddp driver)
- `onClose` now ignores close events from a socket that has already been replaced (compares identity against the current connection), so a zombie socket's late close can't clobber the live connection.
- `open()` tears down the previous connection's handlers and closes it before replacing it (mirrors `forceReopen`'s teardown).

**deepLinking saga** (`app/sagas/deepLinking.js`)
- Before dispatching the resume `loginRequest` (auth-deeplink and call-push paths), gate on the socket being `connected`: after `SERVER.SELECT_SUCCESS`, read `state.meteor.connected` and only `take(METEOR.SUCCESS)` when still disconnected. This both prevents login on a still-connecting socket (the orphan-socket clobber) **and** can't hang the splash when the socket connects *before* `SELECT_SUCCESS` — `METEOR.SUCCESS` fires once and never refires, so an unconditional `take` would wait forever. `handleOpen`'s already-connected fast path skips the wait entirely.

## Issue(s)

[NATIVE-1232](https://rocketchat.atlassian.net/browse/NATIVE-1232) · related: [NATIVE-1123](https://rocketchat.atlassian.net/browse/NATIVE-1123)

## How to test or reproduce

1. Cold-start the app via an auth deeplink carrying a resume token (`rocketchat://auth?...`) for a workspace you are **not** currently connected to.
2. Before this fix: a second (orphan) socket opens; when the server idle-times-out the abandoned socket, the app falls to **"Waiting for network"** even though a live socket exists. After this fix: login waits for `connected` first, only one socket is ever opened, and no false "Waiting for network" appears.
3. Repeat via a call-push deeplink — same path, same expectation.
[...]

PR #7475, opened 2026-07-08
fix: re-subscribe room streams after re-authentication on reconnect

diegolmello:
## Proposed changes

After the app is backgrounded and foregrounded **twice**, room updates stop syncing — message edits don't render, new messages from the same sender don't appear, and reactions stop coming in — until the user clears the cache.

Root cause is a two-part defect in `app/lib/methods/subscriptions/room.ts`, both introduced by #7426:

1. **Re-subscribe fired pre-auth.** On reconnect, `handleConnected` was bound to the raw DDP `'connected'` event, which fires at the WebSocket handshake, *before* login completes. The room stream subscriptions require an authenticated session, so the server rejects them with `nosub: not-allowed`. Those rejected subscriptions resolve to `undefined` inside `this.promises`.
2. **Crash on the next reconnect.** The following reconnect's `handleConnected` iterates `this.promises` to tear down the old subscriptions and calls `.unsubscribe()` on each entry. The `undefined` entry from step 1 throws, the function aborts before it re-subscribes, and the room stream stays permanently dead.

This is why it takes **two** cycles: cycle 1 poisons the subscription array with an `undefined`, cycle 2 crashes on it and never restores the stream.

The fix:
- Bind re-subscription to the authenticated `'login'` DDP event (emitted after the login method result) instead of the pre-auth `'connected'` event, so subscriptions are sent with a valid session and the server accepts them.
- Guard the teardown loops against `undefined`/rejected entries (`sub?.unsubscribe?.()`) so a poisoned array can never abort the re-subscribe path again.

## Issue(s)

https://rocketchat.atlassian.net/browse/SUP-1073

Regression introduced by #7426 (`fix: re-subscribe room streams on DDP reconnect`) — this PR is a direct follow-up.

Related reconnect / app-lifecycle work for context:
- #7362 — `fix: re-subscribe rooms stream after forced socket reopen`
- #7380 — `fix: keep live socket on deeplink login (prevent orphan WebSocket clobber)`
- #7345 — `fix: Drain lost hangups on WebSocket reconnect`
- #7120 — `fix: presence not updating after app resumes from background`

## How to test or reproduce

1. Open a DM or channel.
2. Background the app, wait for the socket to drop, foreground it. Wait for reconnect. **Repeat once** (two full background/foreground cycles).
3. From another client, in the same room: edit a message, send a new message, add a reaction.

**Before:** none of the three appear in the app until the cache is cleared.
**After:** all three sync normally after both cycles.

## Screenshots

Android, same DM. The socket drop/reconnect is forced with airplane-mode toggles on the emulator (equivalent to the background/foreground cycles on a real device, but reliable to reproduce). After the reconnect cycles, another client edits a message, adds a `:rocket:` reaction, and sends a new message.

**Before (buggy build) — room stream dies, updates never arrive**

The edit, the reaction, and the new message are all sent from the other client but none of them render in the open room — the stream stays dead until the cache is cleared.

https://github.com/user-attachments/assets/693d2b0a-610e-422e-b41d-ef8be5a6d2e8

**After (fixed build) — stream recovers, updates sync live**

Same steps on the fixed build. After the reconnect cycles the message edit, the `:rocket:` reaction, and the new message all sync into the open room on camera.

https://github.com/user-attachments/assets/c9480365-2329-4d03-bcda-7ee8f0c9e369

## Types of changes

## Checklist

### Post-mortem

**Symptom.** Room updates silently stopped after two background/foreground cycles; no error surfaced to the user, and only a cache clear recovered it.

**Why it survived review in #7426.** The re-subscribe fix in #7426 was correct in shape but wired to the wrong event. On a *single* reconnect, binding to `'connected'` still appears to work in casual testing — the failure only compounds on the *second* cycle once the subscription array holds an `undefined`. Manual QA that backgrounds once doesn't reveal it.

**How it was found.** Built a tight, device-level feedback loop: a script that drives one offline background/foreground cycle on the Android emulator plus a REST helper that edits/posts/reacts in the target DM, then instrumented the DDP subscription lifecycle with tagged timeline logs. The logs showed the `nosub: not-allowed` rejection on the pre-auth `'connected'` re-subscribe, the resulting `undefined` in `this.promises`, and the `undefined.unsubscribe()` throw on the next cycle.

**What would have prevented it.** The distinction between the pre-auth `'connected'` handshake and the post-auth `'login'` event is load-bearing and easy to miss. Regression tests were added at the subscription seam asserting (a) re-subscription binds to `'login'`, not `'connected'`, and (b) a poisoned subscription array (an `undefined` entry) never aborts re-subscribe. Longer term, any reconnect-time re-subscribe should assert an authenticated session before issuing subscriptions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved room subscription recovery to re-establish subscriptions on login events more reliably after reconnects.
  * Enhanced teardown behavior so cleanup continues safely even if prior subscriptions are missing or already closed.
  * Prevented recovery from running more than once across repeated recoveries and reopen cycles.

* **Tests**
  * Expanded lifecycle tests to validate correct re-registration after login recovery, including partial/failed subscription scenarios.

### DIFF
--- a/app/lib/methods/subscriptions/room.test.ts
+++ b/app/lib/methods/subscriptions/room.test.ts
@@ -139,9 +139,9 @@ describe('RoomSubscription', () => {
 		});
 	});
 
-	describe('handleConnected', () => {
+	describe('handleLogin', () => {
 		it('calls subscribeRoom, dispatches clearUserTyping, loads missed messages, and reads', async () => {
-			await sub.handleConnected();
+			await sub.handleLogin();
 
 			expect(mockSubscribeRoom).toHaveBeenCalledWith(rid);
 			expect(mockStoreDispatch).toHaveBeenCalledWith(clearUserTyping());
@@ -151,7 +151,7 @@ describe('RoomSubscription', () => {
 		it('handles subscribeRoom rejection gracefully', async () => {
 			mockSubscribeRoom.mockRejectedValueOnce(new Error('boom'));
 
-			await expect(sub.handleConnected()).resolves.toBeUndefined();
+			await expect(sub.handleLogin()).resolves.toBeUndefined();
 		});
 	});
 
@@ -166,11 +166,11 @@ describe('RoomSubscription', () => {
 	});
 
 	describe('DDP subscription recovery after forceReopen', () => {
-		it('handleConnected re-subscribes the room to restore lost DDP subscriptions', async () => {
+		it('handleLogin re-subscribes the room to restore lost DDP subscriptions', async () => {
 			await sub.subscribe();
 			mockSubscribeRoom.mockClear();
 
-			await sub.handleConnected();
+			await sub.handleLogin();
 
 			expect(mockSubscribeRoom).toHaveBeenCalledTimes(1);
 			expect(mockSubscribeRoom).toHaveBeenCalledWith(rid);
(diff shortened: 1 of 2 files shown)


Rohit3523 (approved):
Verified the root cause against the SDK source: subscribe swallows nosub errors and resolves to undefined (ddp.ts:493-496), so subscribeRoom's Promise.all yields a poisoned array, and 'login' does fire on reconnect resume (connect.ts:119-129 -> login() emits 'login'). Both halves of the fix hold up, and tests pass (13/13).

Suggestion (non-blocking): rename connectedListener to loginListener since it now binds to the 'login' stream, not 'connected'.

Rohit3523 (approved):

PR #7494, opened 2026-07-13
fix: reuse stale upload record instead of blocking re-send

diegolmello:
## Proposed changes

An upload record left behind in the local DB by a crashed or failed upload permanently blocked re-sending the same file. On the next (non–"try again") send of that file the app threw `Couldn't create upload record` and showed an "Upload in progress" alert, even though nothing was uploading.

Root cause: the guard in `createUploadRecord` treated *"a record with this path exists in the DB"* as *"an upload is in progress"*. That is wrong after a crash/restart — the DB row survives, but the real in-progress signal is the in-memory `uploadQueue`, which is empty.

Fix: show the "Upload in progress" alert and bail only when the upload is **actually active** (`isUploadActive`). Otherwise the existing row is a stale orphan — reset its `error`/`progress` and reuse it so the send proceeds. Applied to both the V2 path (`createUploadRecord`) and the legacy V1 path (`sendFileMessage.ts`, servers < 6.10).

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1422

## How to test or reproduce

1. Start uploading a file, then force-kill the app mid-upload (or trigger an upload failure before the request is queued) so an `uploads` row is left behind.
2. Reopen the app and send the **same** file again (normal send, not the "try again" button).
3. Before: fails with `Couldn't create upload record` / "Upload in progress". After: the file uploads normally.
4. Also verify a genuinely in-flight duplicate send of the same file still shows "Upload in progress".

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

Two related items intentionally left out of scope: (1) the native ObjC-interop exception that can crash the app mid-upload (a separate de-interop/symbolication track — this PR only makes the JS side resilient to the orphan it leaves), and (2) the `catch` branch that logs `Upload cancelled` and silently swallows genuine early-failure errors without rethrowing.

## Summary by CodeRabbit

- **Bug Fixes**
  - File uploads are now retryable when a previous upload record is stale or incomplete.
  - Duplicate attempts are blocked only when an upload is actively in progress for the same item.
  - Retrying now clears any leftover error and resets upload progress before resuming.

- **Tests**
  - Added unit test coverage for active uploads, stale record reuse, forced retries, and creation of a new upload record.

### DIFF
--- a/app/lib/methods/sendFileMessage/sendFileMessage.ts
+++ b/app/lib/methods/sendFileMessage/sendFileMessage.ts
@@ -7,7 +7,7 @@ import i18n from '../../../i18n';
 import database from '../../database';
 import FileUpload from '../helpers/fileUpload';
 import log from '../helpers/log';
-import { copyFileToCacheDirectoryIfNeeded, getUploadPath, persistUploadError, uploadQueue } from './utils';
+import { copyFileToCacheDirectoryIfNeeded, getUploadPath, isUploadActive, persistUploadError, uploadQueue } from './utils';
 import { type IFormData } from '../helpers/fileUpload/definitions';
 
 export async function sendFileMessage(
@@ -30,8 +30,17 @@ export async function sendFileMessage(
 		let uploadRecord: TUploadModel;
 		try {
 			uploadRecord = await uploadsCollection.find(uploadPath);
-			if (uploadRecord.id && !isForceTryAgain) {
-				return Alert.alert(i18n.t('FileUpload_Error'), i18n.t('Upload_in_progress'));
+			if (uploadRecord.id) {
+				if (isUploadActive(fileInfo.path, rid) && !isForceTryAgain) {
+					return Alert.alert(i18n.t('FileUpload_Error'), i18n.t('Upload_in_progress'));
+				}
+				// Record left behind by a crashed or failed upload: reset and reuse it.
+				await db.write(async () => {
+					await uploadRecord?.update(u => {
+						u.error = false;
+						u.progress = 0;
+					});
+				});
 			}
 		} catch (error) {
 			try {
--- a/app/lib/methods/sendFileMessage/sendFileMessage.ts
+++ b/app/lib/methods/sendFileMessage/sendFileMessage.ts
@@ -1,13 +1,9 @@
-import { sanitizedRaw } from '@nozbe/watermelondb/RawRecord';
 import { settings as RocketChatSettings } from '@rocket.chat/sdk';
-import { Alert } from 'react-native';
 
 import { type IUser, type TSendFileMessageFileInfo, type TUploadModel } from '../../../definitions';
-import i18n from '../../../i18n';
 import database from '../../database';
 import FileUpload from '../helpers/fileUpload';
-import log from '../helpers/log';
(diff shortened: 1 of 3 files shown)


app/lib/methods/sendFileMessage/utils.test.ts
```diff
@@ -0,0 +1,81 @@
...
+
+	it('reuses the existing record when force-retry', async () => {
+		const existing: any = { id: uploadPath, update: jest.fn((cb: (u: any) => void) => cb(existing)) };
+		mockFind.mockResolvedValue(existing);
+
+		const [path, record] = await createUploadRecord({ rid: 'GENERAL', fileInfo, tmid: undefined, isForceTryAgain: true });
+
+		expect(path).toBe(uploadPath);
+		expect(record).toBe(existing);
+	});
```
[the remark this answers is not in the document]
diegolmello: Good catch. Added \`uploadQueue[uploadPath] = {}\` plus an \`Alert.alert\` not-called assertion so the test now goes red if \`&& !isForceTryAgain\` is dropped. Fixed in e1a66ff.

app/lib/methods/sendFileMessage/sendFileMessage.ts
```diff
@@ -30,8 +30,17 @@ export async function sendFileMessage(
...
+				if (isUploadActive(fileInfo.path, rid) && !isForceTryAgain) {
+					return Alert.alert(i18n.t('FileUpload_Error'), i18n.t('Upload_in_progress'));
+				}
+				// Record left behind by a crashed or failed upload: reset and reuse it.
+				await db.write(async () => {
+					await uploadRecord?.update(u => {
+						u.error = false;
+						u.progress = 0;
+					});
+				});
```
diegolmello: dup?

app/lib/methods/sendFileMessage/utils.ts
```diff
@@ -74,9 +74,18 @@ export const createUploadRecord = async ({
...
+				Alert.alert(i18n.t('FileUpload_Error'), i18n.t('Upload_in_progress'));
+				return [null, null];
+			}
+			// Record left behind by a crashed or failed upload: reset and reuse it.
+			await db.write(async () => {
+				await uploadRecord?.update(u => {
+					u.error = false;
+					u.progress = 0;
+				});
+			});
```
diegolmello, line 88: dup?

app/lib/methods/sendFileMessage/sendFileMessage.ts
diegolmello: Yes — V1 inlined the same find/guard/reset/create logic that `createUploadRecord` (utils.ts) already extracts for V2. Dedup'd in 73cdb97be: `sendFileMessage.ts` now calls `createUploadRecord`. Two behavior diffs preserved/aligned to V2 — active-upload now returns via `[null,null]`, create-fail now throws (was silent `log`).

app/lib/methods/sendFileMessage/utils.ts
diegolmello: This is the canonical version — V1 (`sendFileMessage.ts`) was the duplicate and now calls this. Fixed in 73cdb97be.

Rohit3523 (approved):

PR #7498, opened 2026-07-16
fix: restore Away status option on servers 8.6.0 and above

Rohit3523:
## Proposed changes

We previously removed the "Away" status option from the Edit Status screen on servers `>= 8.6.0`. This reverses that change so the Away option is always available regardless of server version.

## Issue(s)

https://rocketchat.atlassian.net/browse/CORE-2431

## How to test or reproduce

1. Connect to a server `>= 8.6.0`.
2. Open Edit Status (profile → status).
3. The "Away" option should now be visible even when the current status is online/busy/offline.

## Types of changes

## Checklist

- [ ] I have signed the CLA

## Further comments

The test `should hide away option on server >= 8.6.0 when user is not away` / `...when user is away` was collapsed into a single `should always show away option regardless of server version` test reflecting the reverted behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * The “Away” status option is now consistently available in the status selector, regardless of server version or the user’s current status.
* **Tests**
  * Updated the StatusView test coverage to reflect the new, always-visible “Away” option behavior.

### DIFF
--- a/app/views/StatusView/index.test.tsx
+++ b/app/views/StatusView/index.test.tsx
@@ -113,16 +113,6 @@ describe('StatusView', () => {
 			expect(screen.queryByTestId('status-view-clear-after')).toBeNull();
 		});
 
-		it('should always show away option regardless of server version', () => {
-			mockedStore.dispatch(setUser({ id: 'user-id', username: 'user', status: 'online', statusText: '' }));
-			mockedStore.dispatch(selectServerSuccess({ server: 'https://example.com', version: '8.6.0', name: 'Test' }));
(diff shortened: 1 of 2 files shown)


app/views/StatusView/index.test.tsx
```diff
@@ -113,23 +113,13 @@ describe('StatusView', () => {
 			expect(screen.queryByTestId('status-view-clear-after')).toBeNull();
 		});
 
-		it('should hide away option on server >= 8.6.0 when user is not away', () => {
+		it('should always show away option regardless of server version', () => {
```
diegolmello: This test is useless now. You should remove it instead of adapt it.

diegolmello (approved):

PR #7502, opened 2026-07-20
fix: silent message loss after notification

Rohit3523:
## Proposed changes

Fixes three bugs that cause messages to arrive as push notifications but silently fail to appear in the chat view.

### 1. `Encryption.decryptMessages` — `Promise.all` kills entire `db.write` batch (E2E rooms)

`Promise.all` causes the entire `db.write()` callback to reject when any single message decryption throws. `Promise.allSettled` prevents one bad message from aborting the entire batch.

### 2. `loadMissedMessages` — fire-and-forget recursive pagination

The recursive call to load the next page was not awaited. If page 2+ fails (network timeout, server error), all subsequent messages are silently lost. Adding `await` ensures pages sequence correctly.

### 3. `updateMessages` / `loadThreadMessages` — `delete threadMessage.tmid` corrupts input

The mutation runs before the `msgsToCreate`/`msgsToUpdate` closures execute during `db.batch()`, corrupting the original `update` array and thread-message update closures.

## How to test or reproduce

1. Enable E2E encryption. Send multiple messages — one failed decryption no longer aborts all writes.
2. Have >50 unread messages, open via notification — all appear instead of only the first 50.
3. Thread replies no longer corrupt the update array.

## Types of changes

## Checklist

### DIFF
--- a/app/lib/encryption/encryption.test.ts
+++ b/app/lib/encryption/encryption.test.ts
@@ -140,4 +140,54 @@ describe('Encryption.encryptMessage', () => {
 		expect(result).toBe(baseMessage);
 		expect(mockRoomEncrypt).not.toHaveBeenCalled();
 	});
+
+	describe('decryptMessages', () => {
+		beforeEach(() => {
+			jest.clearAllMocks();
+			(encryption as any).roomInstances = {};
+		});
+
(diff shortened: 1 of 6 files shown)


Rohit3523:
Superseded by #7504, which includes this PR's full diff (bugs 1-3) plus a 4th bug found during follow-up testing (a lastOpen/unmount race in RoomSubscription).

PR #7510, opened 2026-07-23
fix: UserPreferences.getBool not parsing JSON-string booleans

Rohit3523:
## Proposed changes

Fix `getBool` in `UserPreferences` to correctly read booleans stored as JSON strings by `useUserPreferences<boolean>`. Previously, `getBool` only used native MMKV `getBoolean`, which couldn't parse `"false"` (stored as a JSON string) and would return `null`/`undefined`. The fallback now tries `getString` + `JSON.parse` first, then falls back to native `getBoolean`.

This caused the haptic feedback on reconnect (from `RoomView.hapticFeedback`) to ignore `NOTIFICATION_IN_APP_VIBRATION` toggle — vibration played for every pending message replayed on DDP reconnect even when the toggle was off.

## Issue(s)
https://rocketchat.atlassian.net/browse/SUP-1079

## How to test or reproduce

1. Go to Settings → Notifications → disable "Vibrate from new messages"
2. Disconnect from network (airplane mode)
3. Reconnect — observe no vibration for replayed messages

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved boolean preference retrieval by correctly handling values saved as JSON-formatted strings (e.g., `"true"` / `"false"`).
  * When a stored value is present but not a valid boolean (or contains invalid JSON), the preference now resolves to `null`, while fallback behavior remains intact.

* **Tests**
  * Added Jest coverage for valid JSON boolean strings and invalid/non-boolean stored values returning `null`.

### DIFF
--- a/app/lib/methods/userPreferences.ts
+++ b/app/lib/methods/userPreferences.ts
@@ -106,6 +106,14 @@ class UserPreferences {
 
 	getBool(key: string): boolean | null {
 		try {
+			const str = this.mmkv.getString(key);
+			if (str !== undefined) {
+				try {
+					return JSON.parse(str);
+				} catch {
+					return null;
+				}
+			}
 			return this.mmkv.getBoolean(key) ?? null;
 		} catch {
 			return null;
--- a/app/lib/methods/userPreferences.ts
+++ b/app/lib/methods/userPreferences.ts
@@ -109,7 +109,8 @@ class UserPreferences {
 			const str = this.mmkv.getString(key);
 			if (str !== undefined) {
 				try {
-					return JSON.parse(str);
+					const parsed: unknown = JSON.parse(str);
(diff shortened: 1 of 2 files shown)


app/lib/methods/userPreferences.ts
```diff
@@ -106,6 +106,15 @@ class UserPreferences {
 
 	getBool(key: string): boolean | null {
```
OtavioStasiak, line 115: Let's improve the readability of the code... maybe create something like:
```
private tryParseJson(value: string): unknown {
	try {
		return JSON.parse(value);
	} catch {
		return undefined;
	}
}

```

to avoid nested try catch.

OtavioStasiak (changes requested):

OtavioStasiak (approved):
LGTM

PR #7511, opened 2026-07-23
fix: restore DDP streams through a single restoration owner after reconnect

diegolmello:
## Proposed changes

After a transient disconnect (network blip, app background/foreground, server restart), DDP streams could die silently: the socket reconnected but stream subscriptions were never restored, so rooms stopped receiving messages until the user cleared cache or restarted the app. This traces back to `forceReopen` bypassing `connect()` (#7298), which was patched per-consumer over time (#7362, #7426, #7475) while the underlying ownership gap remained.

This PR fixes the root causes and locks the behavior with a permanent test suite:

- **Single stream-restoration owner** (`app/lib/services/connectionRestore.ts`): every DDP stream consumer registers a restorer; the owner fans out on the DDP `login` event, keyed by socket generation, so all streams re-land on the current SDK instance after any reconnect — including SDK instance swaps that previously orphaned room listeners. The per-consumer patches from #7362/#7426/#7475 are deleted, subsumed by the owner. Enrolled consumers: rooms list, open room, notify-user, settings, presence, permissions, roles.
- **Stranded resume-login recovery**: removed the stale `meteor.connected` guard that blocked recovery after `forceReopen` (a real handshake now always re-runs the login pipeline), added bounded retry for transient resume-login failures, and made the foreground check dispatch a resume when a token exists but the session is unauthenticated.
- **Deep-link warm start**: notification taps and deep links now gate on live session state (`meteor.connected && login.isAuthenticated`) instead of the stale `server.connected` redux flag, which was never reset on silent socket death. The lying `connected` field was removed from the server reducer.
- **Connection-lifecycle suite** (`app/lib/services/connection.lifecycle.test.ts`): binds the real production listeners (no copies) and asserts every enrolled stream survives every lifecycle transition — connect, close, forceReopen, resume-login failure/success, SDK swap, foreground reopen, deep-link warm start.

> **Merge order**: stacked on #7506 (base `swamp-string`). Merge only after #7506 lands, then retarget this PR to `develop`.

## Issue(s)

## How to test or reproduce

Automated:

```bash
TZ=UTC pnpm test --testPathPattern='connection.lifecycle'
TZ=UTC pnpm test --testPathPattern='app/sagas/__tests__'
```

Manual:

1. Log in, open a busy room. Kill the network (or the server) for ~30s, restore it. Messages sent by other users during and after the outage must appear without reopening the room.
2. Background the app for several minutes (long enough for the OS to kill the socket), then tap a push notification. The room must open on the correct server with the message visible.
3. Repeat 1 while the server session token is near expiry, forcing a resume-login failure: the app must retry and recover instead of staying online-looking but deaf.

## Screenshots

N/A — no UI changes.

## Types of changes

## Checklist

## Further comments

The alternative — keep patching each consumer as its stream dies — was already tried three times (#7362, #7426, #7475) and each patch left holes proven by the new suite. Centralizing restoration in one generation-keyed owner makes "did my stream survive reconnect?" a contract the suite enforces for every consumer, present and future. Connection state duplicated across `meteor.connected`, `server.connected` and the socket was the shared enabler of these bugs; the redux flags are now display-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after reconnects, server changes, and restored sessions so rooms, roles, settings, permissions, and presence stay synchronized.
  * Added bounded retries for temporary resume-login failures, with logout for expired/invalid sessions.
  * Improved deep-link and call navigation reliability when auth/connectivity is incomplete.
  * Enhanced foreground recovery for stranded resume-login situations.
* **Reliability**
  * Prevented stale/older connections from restoring subscriptions or delivering outdated stream events.
  * Improved connection lifecycle handling so stream listeners are re-homed to the active SDK instance.

### DIFF
--- a/app/lib/methods/getPermissions.ts
+++ b/app/lib/methods/getPermissions.ts
@@ -8,6 +8,7 @@ import log from './helpers/log';
 import { store as reduxStore } from '../store/auxStore';
 import database from '../database';
 import sdk from '../services/sdk';
+import { registerStreamRestorer } from '../services/connectionRestore';
 import protectedFunction from './helpers/protectedFunction';
 import { compareServerVersion } from './helpers';
 
@@ -166,7 +167,6 @@ export function getPermissions(): Promise<void> {
 			const db = database.active;
 			const permissionsCollection = db.get('permissions');
 			const allRecords = await permissionsCollection.query().fetch();
-			sdk.subscribe('stream-notify-logged', 'permissions-changed');
 			// if server version is lower than 0.73.0, fetches from old api
 			if (serverVersion && compareServerVersion(serverVersion, 'lowerThan', '0.73.0')) {
 				// RC 0.66.0
@@ -205,3 +205,11 @@ export function getPermissions(): Promise<void> {
 		}
 	});
 }
+
+export function subscribePermissions(): void {
+	return sdk.subscribe('stream-notify-logged', 'permissions-changed');
+}
+
+// Restorer owns only re-subscription; the awaited getPermissions() fetch stays in the login saga so
+// permissions state is set before the enterprise-modules and VoIP checks that read it.
+registerStreamRestorer(() => subscribePermissions());
--- a/app/reducers/server.ts
+++ b/app/reducers/server.ts
@@ -3,7 +3,6 @@ import { SERVER } from '../actions/actionsTypes';
 
 export interface IServer {
 	connecting: boolean;
-	connected: boolean;
 	failure: boolean;
 	failureMessage?: string;
 	server: string;
@@ -16,7 +15,6 @@ export interface IServer {
 
 export const initialState: IServer = {
 	connecting: false,
-	connected: false,
 	failure: false,
 	server: '',
 	version: '',
@@ -39,7 +37,6 @@ export default function server(state = initialState, action: TActionServer): ISe
 			return {
 				...state,
 				connecting: false,
-				connected: false,
 				failure: true,
 				failureMessage: action.failureMessage
 			};
@@ -47,7 +44,6 @@ export default function server(state = initialState, action: TActionServer): ISe
 			return {
 				...state,
 				connecting: false,
-				connected: true,
 				loading: false,
 				changingServer: false
 			};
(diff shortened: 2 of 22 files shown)


app/lib/methods/getPermissions.ts
```diff
@@ -205,3 +205,11 @@ export function getPermissions(): Promise<void> {
...
 	});
 }
+
+export function subscribePermissions(): void {
+	return sdk.subscribe('stream-notify-logged', 'permissions-changed');
+}
+
+// Restorer owns only re-subscription; the awaited getPermissions() fetch stays in the login saga so
+// permissions state is set before the enterprise-modules and VoIP checks that read it.
+registerStreamRestorer(() => subscribePermissions());
```
diegolmello, line 215: Shouldn't this be tied to a workspace?

app/reducers/server.ts
```diff
@@ -3,7 +3,6 @@ import { SERVER } from '../actions/actionsTypes';
 
 export interface IServer {
 	connecting: boolean;
-	connected: boolean;
```
diegolmello, line 6: Do we delete both? Not sure. I think connecting is an easier way to re-render.

app/sagas/deepLinking.js
```diff
@@ -166,6 +166,18 @@ const handleShareExtension = function* handleOpen({ params }) {
...
+// honest session flags: only skip connect/login when the websocket is up and the resume login
+// already authenticated. Otherwise re-run the pipeline before navigating onto a dead socket.
+const ensureSameServerSession = function* ensureSameServerSession(host, version) {
+	const sessionReady = yield select(state => state.meteor.connected && state.login.isAuthenticated);
+	if (!sessionReady) {
+		yield localAuthenticate(host);
+		yield put(selectServerRequest(host, version, true));
+		yield take(types.LOGIN.SUCCESS);
+	}
+};
```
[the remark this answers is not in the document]
diegolmello: Fixed in 627507756: `ensureSameServerSession` now races `LOGIN.SUCCESS` against `LOGIN.FAILURE`, `SERVER.SELECT_FAILURE` and `LOGOUT`, returns whether the session is ready, and both call sites skip navigation otherwise — a later unrelated `LOGIN.SUCCESS` can no longer wake the parked saga and navigate with stale params. Covered by three new real-saga tests (failure, select-failure, logout — each asserting a late `loginSuccess` does not trigger navigation).

PR #7516, opened 2026-07-25
chore: migrate typecheck to TypeScript 7.0

diegolmello:
## Proposed changes

Moves the `tsc` run behind `pnpm lint` from TypeScript 5.9.3 to TypeScript 7.0.2 — the native Go compiler — in two steps: a 6.0 baseline hop, then the 7.0 cut.

**Typecheck wall time drops from 5.77s to ~1.0s (5.7x).** Peak memory drops from 1.37 GB to 1.19 GB, and the program size is unchanged at 3433 files. Zero new type errors.

| version | median `tsc --noEmit` | peak RSS |
| --- | --- | --- |
| 5.9.3 | 5.77s | 1.37 GB |
| 6.0.3 | 6.01s | 1.35 GB |
| **7.0.2** | **~1.0s** | 1.19 GB |

### Step 1 — the 6.0 baseline hop

TypeScript 6.0 raises two deprecations as *errors* rather than warnings, and `ignoreDeprecations: "6.0"` stops working in 7.0, so clearing them properly was the only route:

- `moduleResolution` `node10` → `bundler`, which requires an esnext-shaped `module`.
- `baseUrl` removed. Exactly one import relied on it, and it is now relative.
- `types` enumerated as `["jest", "node", "react"]`, because a resolution mode that honours package `exports` no longer auto-includes every `@types` package. `@types/node` becomes an explicit devDependency.
- The inherited block of commented-out option documentation is dropped.

Alternatives were measured rather than guessed before settling on `bundler`: `node16` produced 7240 errors, `nodenext` 7062, `bundler` 7074, `bundler` plus explicit `types` 53, and zero once the dependency patches landed.

### Step 2 — the 7.0 cut

`typescript` is pinned to exact `7.0.2`, since the platform binaries ship as version-matched optional dependencies; the `linux-x64` target CI resolves is recorded in the lockfile with an integrity hash. **No tsconfig change was required at all** — step 1 already left it 7.0-shaped. CI needs no change either, as the bin is still `tsc`.

`noUncheckedSideEffectImports` is left at its new `true` default and raised zero errors. `--checkers` and `--builders` were rejected on measurement: the default already parallelises ~4.4x, and `--checkers 4` (matching a 4-vCPU runner) measured indistinguishable.

### Four dependency patches are load-bearing

Honouring package `exports` stranded the bundled type declarations of three dependencies, so each gets a `types` condition via patch-package: `@rocket.chat/message-parser`, `@hookform/resolvers`, and `js-base64`. This is visible to the type checker only — Metro ignores that condition. A tsconfig `paths` mapping was tried first and rejected, because the jest-expo resolver reads `paths` and then loads those declaration files at runtime, killing 38 test suites.

The fourth patch covers a genuine TypeScript 7.0.2 regression in `@react-navigation/core`. The compiler resolves the mutual recursion between `StaticParamList` and `ParamListForScreens` eagerly where earlier versions defer it through the mapped type, reports the alias as circular, and degrades it to a non-generic symbol — surfacing as three `TS2315: Type 'StaticParamList' is not generic` errors in `app/stacks/`. The patch drops a `FlatType<>` wrapper that only flattens intersections for editor display, so the type itself is unchanged. Reporting it upstream and deleting the patch once a 7.0.x fixes it is follow-up work.

Nothing was silenced to reach zero errors: no `@ts-expect-error`, and `skipLibCheck` is untouched.

### Editors

`tsserver` is gone as a bin in 7.0; the language service moved into the Go binary as `tsc --lsp`. This repo has no `.vscode/` and no `typescript.tsdk` setting, so nothing breaks by default. VS Code users who want 7.0 language features need the `TypeScriptTeam.native-preview` extension until bundling lands; WebStorm 2026.2 already handles it automatically.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1443

## How to test or reproduce

There is no runtime change — types are stripped by Babel, so the app itself is untouched. Verification is toolchain-level:

```sh
pnpm install
pnpm lint          # oxlint && tsc — expect exit 0
TZ=UTC pnpm test   # expect 217/217 suites
```

To see the speedup, time the typecheck directly:

```sh
time ./node_modules/.bin/tsc --noEmit
```

## Types of changes

## Checklist

## Further comments

**This stacks on #7515 and should merge after it.** The branch carries that pull request's two OXC commits, so they appear here too; only the two `chore: bump TypeScript to 6.0` and `chore: migrate typecheck to TypeScript 7.0` commits are new.

The two-step split is deliberate. Going straight to 7.0 would have mixed a config migration with a compiler swap, leaving no clean baseline to diff the new compiler's output against. Hopping through 6.0 first meant the 7.0 cut changed exactly one thing — the compiler — and its only output diff was the three `@react-navigation/core` errors, which made that regression easy to isolate to a compiler bug rather than our code.

The 6.0 step is ~4% *slower* than 5.9.3, which is expected: 6.0 is still the JavaScript compiler. The speedup arrives entirely with the native binary in step 2.

## Summary by CodeRabbit

* **Chores**
  * Updated TypeScript and development tooling for improved compatibility with modern module resolution.
  * Added improved type support for navigation, validation, encoding, and Node.js integrations.
  * Updated compiler configuration to support JSON modules and more reliable type checking.
  * No user-facing functionality or interface changes are included in this release.


OtavioStasiak (approved):
LGTM

PR #7521, opened 2026-07-26
revert: roll back SDK socket probing (#7298) and dependent reconnect fixes

diegolmello:
## Proposed changes

Full rollback of #7298 (`fix(voip,connect): probe Meteor Connect on foreground and lock-screen accept`) and the four follow-up fixes that exist only to patch bugs it introduced:

- Revert #7475 (re-subscribe on re-authentication)
- Revert #7426 (re-subscribe room streams on reconnect)
- Revert #7362 (rooms-subscription guard reset after forced reopen)
- Revert #7380 (orphan WebSocket close fix)
- Revert #7298 (SDK `probe`/`forceReopen`/`checkAndReopen` patch + app-side wiring)

The SDK patch (`patches/@rocket.chat+sdk+1.3.3-mobile.patch`) returns to its pre-7298 content (only the `media-signal`/`media-calls` subscription additions). Since #7298 merged (2026-05-06), the SDK/connect surface has needed a steady stream of reconnect/resubscribe fixes, and the aggressive socket-lifecycle healing it introduced is the root of the instability (foreground `checkAndReopen` racing the SDK reconnect timer; `forceReopen` short-circuiting the login path, stranding the app connected with zero subscriptions).

Preserved through the rollback:
- #7345 hangup draining — rewired to wait on Redux login state instead of the removed `awaitDdpLoggedIn`
- #7346 outgoing-call foreground service
- #7392 presence sync, #7430 avatar refresh, #7358/#7304 deeplink fixes, #7334 telecom guard

**Known trade-off:** this reintroduces the lock-screen VoIP accept failure #7298 originally fixed (answer SDP through a half-open session → `TIMEOUT_TO_PROGRESS_SIGNALING`). A forward fix for socket-liveness should be designed separately rather than keeping the probing approach.

## Issue(s)

Reverts #7298 and its follow-ups #7362, #7380, #7426, #7475.

## How to test or reproduce

- Fresh install → login → open a room → toggle network off/on → room stream and message loading recover without probing logic
- Regression risk to watch: incoming VoIP call accepted from lock screen (expected to be less reliable, pre-7298 behavior)

## Screenshots

N/A

## Types of changes

## Checklist

## Further comments

Rollback strategy and per-commit dependency analysis were mapped before execution: the five reverts restore pre-7298 semantics without clobbering independent work that landed on the same files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DDP/socket recovery with liveness probing and protection against stale close events.
  * Pending call hangups now wait for authentication and confirmed connectivity before draining.
  * Room subscriptions recover missed messages and update read state more reliably.
  * Deep-link resume and navigation timing is more consistent across server and login transitions.
  * VoIP calls are answered more promptly, with improved signal error handling.
  * Added media-related DDP subscription events.
* **Refactor**
  * Streamlined VoIP acceptance flow by removing unnecessary reconnect and replay steps.
* **Tests**
  * Updated test coverage for connection, VoIP, deep-linking, and subscription behavior.


diegolmello (commented):
## Consolidated review

Method: 8 finder angles (line-by-line, removed-behavior audit, cross-file trace, altitude, efficiency, simplification, conventions, reuse), deduped to 16 findings, each verified against worktree code, `node_modules` SDK source, and git history. Verdicts below only include what survived verification.

The revert is mechanically clean: no dangling references to `awaitDdpLoggedIn`/`forceReopen`/`probe`, `tsc` clean, targeted jest green. Two items are **over-broad** — they roll back work independent of #7298:

### Over-broad (recommend restoring)

1. **SDK orphan-socket guard** — `patches/@rocket.chat+sdk+1.3.3-mobile.patch`. The removed `open()` old-socket teardown and `onClose` identity guard came from #7380 (`80989f51e`), not #7298. Upstream `ddp.ts` replaces `this.connection` without closing the old socket and emits `'close'` unconditionally, so a zombie socket's late close flips Redux to disconnected and tears down a healthy live connection ("Waiting for network" on a good socket). Restoring those two hunks is independent of the probing logic being reverted.

2. **#7493 `updateMessage` concurrency coverage** — `room.test.ts` is deleted wholesale, but its `describe('updateMessage concurrency')` block (added by `360cabcbb`, not in the revert set) covers the writer-serialization fix that still lives at `room.ts:258`. No other test file covers it. Keep that describe (and the `subscribe` block); delete only the describes tied to reverted behavior (`handleLogin`, `handleClose`, DDP recovery, `isAlive`).

### Accepted revert costs (confirmed pre-#7298 bugs, restored by design — for the record)

- **VoIP re-entrancy** — `MediaSessionInstance.ts:60`: `answerCall` runs synchronously inside the lib's `processSignal` tick; the `accept()` state change is dropped under the `updatingInputTrack` re-entrancy lock → mic never opens → 10s `TIMEOUT_TO_PROGRESS_SIGNALING` hangup.
- **Fire-and-forget `processSignal`** — `MediaSessionInstance.ts:74,120`: replayed signals race (answer can process before its offer's async setup finishes); rejections are unhandled.
- **Lock-screen accept** — `MediaCallEvents.ts:97`: same-workspace accept no longer heals a zombie socket before replaying signals; answer SDP goes out over a half-open session. Stated intent of the revert.
- **Poisoned subscription array** — `room.ts:73`: SDK `subscribe` resolves `undefined` on `nosub`; bare `sub.unsubscribe()` then aborts the `forEach`, leaking every later subscription server-side.
- **Room streams never re-created after reconnect** (pre-#7426 state) and `readMessages` can mark a room read after the user exits, via a `connected`/`close` event landing during the async `removeListener` window with no `isAlive` guard (`room.ts:54-55,96`).
- **Silent socket death without a `close` event** — `state.js:22-24`: foreground handler gated on `isAuthAndConnected()` skips `checkAndReopen()` precisely when the socket died silently in background. (The "suspended reopen timer → indefinite stall" variant was rejected during verification: the SDK timer fires on foreground.)
- Minor: unhandled rejection from `checkAndReopen()` when foregrounded offline (`state.js:29`).

### Rejected during verification

- **deepLinking `METEOR.SUCCESS` gate removal** — both proposed failure modes fail: SDK `login()` awaits `connect()` before sending, so the resume login stalls rather than throwing or being rejected. Residual risk is a duplicate/misordered `loginRequest` racing the `'connected'`-event login in `connect.ts`. The gate was a #7298 dependent, so the rollback is legitimate.
- **Missed-wakeup in the inline login wait** (`connect.ts:150-163`) — impossible: Redux dispatch is synchronous and JS is single-threaded, so state cannot change between the outer `isLoginReady()` check and `store.subscribe`. Behavior-equivalent to the removed `awaitDdpLoggedIn`. Worth restoring the one-line rationale comment (why redux state over `ddp.loggedIn`).

### Nits

- `connect.ts:44` `const s` → `state`; `connect.test.ts` `cb` param, module-level `mockStoreSubscribeCallbacks` serving one test, positional `connectedHandlers[1]` (correct today, fragile to listener order); `MediaSessionInstance.ts:121` stray blank line.

**Verdict:** ship after restoring the SDK orphan-socket hunks and the #7493 test block. Everything else is the revert's intended cost and should be tracked against the replacement work.

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -74,9 +71,7 @@ class MediaSessionInstance {
 		try {
 			const { signals } = await mediaCallsStateSignals(getUniqueIdSync());
 			for (const signal of signals) {
-				// Sequential replay: signals depend on prior state mutations.
-				// eslint-disable-next-line no-await-in-loop
-				await this.instance.processSignal(signal);
+				this.instance.processSignal(signal);
```
diegolmello: Don't we need await here?

app/lib/services/connect.ts
```diff
@@ -40,6 +40,12 @@ interface IServices {
 	service: string;
 }
 
+// Trusts redux state rather than `ddp.loggedIn`, which isn't cleared on socket close and can read true for a stale session.
```
diegolmello: Redux is also not cleared when in background

jest.config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
 	modulePathIgnorePatterns: ['<rootDir>/.*worktrees/'],
 	testPathIgnorePatterns: ['e2e', 'node_modules', '<rootDir>/.*worktrees/', '/__tests__/testHelpers\\.tsx$'],
 	transformIgnorePatterns: [
-		'node_modules/(?!((jest-)?react-native|@react-native(-community)?)|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|react-native-svg|@rocket.chat/ui-kit|@rocket.chat/sdk|tiny-events)'
+		'node_modules/(?!((jest-)?react-native|@react-native(-community)?)|expo(nent)?|@expo(nent)?/.*|@expo-google-fonts/.*|react-navigation|@react-navigation/.*|@unimodules/.*|unimodules|sentry-expo|native-base|react-native-svg|@rocket.chat/ui-kit)'
 	],
```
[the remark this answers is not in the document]
diegolmello: Skipping this one. Those two entries (`@rocket.chat/sdk`, `tiny-events`) were added by #7298 in `2af4bda3e` and removed again by the revert commit `87437c94e`, so dropping them is revert fidelity rather than an oversight — this PR restores the pre-#7298 `transformIgnorePatterns`. Verified the full suite is green on this branch: 216/216 suites, 2001/2001 tests, 409/409 snapshots, no TypeScript parse errors from `@rocket.chat/sdk`.

app/sagas/state.js
```diff
@@ -19,18 +19,14 @@ const appHasComeBackToForeground = function* appHasComeBackToForeground() {
...
 	}
 	try {
 		const server = yield select(state => state.server.server);
 		yield localAuthenticate(server);
-		// Always reconnect on foreground; gating on meteor.connected would skip
-		// the call precisely when the socket died silently in background.
-		checkAndReopen().catch(e => {
-			log('[state.js] checkAndReopen failed:', e);
-		});
+		checkAndReopen();
```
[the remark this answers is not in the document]
diegolmello: Skipping this one as well. The `.catch(log)` handler around `checkAndReopen()` was introduced by #7298 in `2af4bda3e`; the pre-#7298 baseline was a bare `checkAndReopen();`, which is exactly what the revert commit `87437c94e` restores. Adding error handling here would be new behavior rather than a revert, so it belongs in a follow-up if we want it.

app/lib/services/voip/MediaSessionInstance.ts
diegolmello: Yes — good catch. `processSignal` returns `Promise<void>` and mutates the state that `tryAnswerIfNativeAcceptedNotification` reads, so both call sites needed to sequence on it. Note the missing `await` predates #7298: the pre-#7298 baseline had no await here either, #7298 added it, and the revert took it back out. Restoring it as a deliberate carve-out from the revert rather than shipping the old hazard. Also awaited the `stream-notify-user` handler (with a try/catch, since an async stream callback would otherwise float its rejection). Fixed in 44d451b13.

app/lib/services/connect.ts
diegolmello: Right, the comment overstated it. `closeListener` dispatches `disconnectAction()`, so redux `meteor.connected` *is* cleared on a real close — the asymmetry with `ddp.loggedIn` is only that the latter survives a close. On a silent background death neither is cleared, so redux is no better there, which is what the 5s timeout on the drain wait is for. Reworded to say exactly that in 44d451b13; keeping the redux read and the bounded wait.

OtavioStasiak (approved):
LGTM

PR #7523, opened 2026-07-27
fix(sync): derive the message sync cursor from the server clock

diegolmello:
## Proposed changes

`subscription.lastOpen` did double duty: it was both the fetch cursor handed to `chat.syncMessages` and the anchor for the unread separator. It was also written from the **device** clock, while the server compares the cursor against each message's `_updatedAt` (**server** clock). Any skew, or any write not tied to a real fetch, silently moved the cursor past messages that were never delivered — and they were never fetched again.

The direction matters, and only one side is safe:

```
cursor < change._updatedAt  → server returns the change again  → wasteful, safe
cursor > change._updatedAt  → server stays silent              → the missing-message bug
```

The two roles are now split, and the cursor only ever moves to a timestamp the server produced:

- **`updateLastOpen` now takes a raw server payload** instead of defaulting to `new Date()`. It writes the max `_updatedAt` in that payload, read *before* `normalizeMessage` runs (that path invents a `new Date()` fallback for rows that lack one). It is deliberately non-monotonic, so an already-poisoned future cursor heals downward.
- **Its unconditional device-clock callers are gone.** `readMessages` no longer writes `lastOpen` at all (it owns `ls`), and `RoomSubscription.unsubscribe` no longer stamps the cursor on room close — that write was what advanced the cursor past messages posted while offline.
- **Mark-unread (`MessageActions`) still writes `ls` from the tapped message's `ts`** — it rewinds the unread separator and is now the only device-clock writer of `ls` left. It never touches the cursor.
- **`loadMissedMessages`** resolves the cursor from the subscription row itself, treats a future-dated cursor as absent so it self-heals, falls back to `loadMessagesForRoom` when there is no cursor instead of no-op'ing, and advances the cursor only once the UPDATED page cursor has drained.
- **`loadMessagesForRoom`** snapshots the first batch's server `_updatedAt` values and hands them to `updateLastOpen`; paging older history and gap fills still never move the cursor.

Separator behaviour is unchanged — `RoomView` state renames `lastOpen` → `lastSeen` and feeds the separator from `ls`, so the two meanings cannot collapse into one column again. `CONTEXT.md` gains a **Timestamp Trust Boundary** section naming Last Open, Last Seen, Server Timestamp and Device Timestamp.

## Issue(s)

## How to test or reproduce

**Offline room close.**

1. Turn airplane mode on.
2. Open the app and enter a channel.
3. From another session, post 3 or more messages to that channel.
4. Close the room on the device (back out of it).
5. Turn airplane mode off, re-enter the room, and assert the **oldest** of those messages is present.

Post at least three: `subscriptions/rooms.ts` back-fills `lastMessage` into the messages table for rooms you are *not* in, which self-heals any single-message repro.

## Screenshots

Recorded on an Android 36 emulator against `mobile.qa.rocket.chat`, before/after on the same sequence.

Offline room close — before: the cursor jumps to the device clock at close time (11:26:14, past all three messages), and A-M1/A-M2 never arrive. After: the cursor equals the last message's server `_updatedAt` to the millisecond (`1785164328836` = `14:58:48.836Z`) and all three are present.

## Types of changes

## Checklist

## Further comments

Stacked on #7521 (base `purring-seahorse`) — that PR rolls back the #7298 socket probing, so it needs to land first.

221 suites / 2027 tests pass, `tsc` clean, 0 lint errors.


app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -73,6 +75,11 @@ async function load(args: {
 		}
 
 		const batch = data.messages as IMessage[];
+		if (batchesFetched === 1) {
```
diegolmello: What if we don't rely only on first batch? Instead we could get the max date from all batches and that's not going to be expensive. This is going to be future proof, since we could change the fetch order and the first batch could be actually the oldest one.

app/lib/methods/writeSyncWatermark.ts
diegolmello, line 1: I think we should rename to updateLastOpen.
We're just moving logic around, so we shouldn't have to create new domain models.

app/lib/methods/writeSyncWatermark.ts
```diff
@@ -0,0 +1,47 @@
...
+				return;
+			}
+			const updatedAt = new Date(message._updatedAt);
+			if (Number.isNaN(updatedAt.getTime())) {
+				return;
+			}
+			if (!watermark || updatedAt > watermark) {
+				watermark = updatedAt;
+			}
+		});
```
diegolmello: This can be done a lot simpler, no?

app/sagas/state.js
```diff
@@ -1,19 +1,36 @@
...
 
 const isAuthAndConnected = function* isAuthAndConnected() {
 	const login = yield select(state => state.login);
 	const meteor = yield select(state => state.meteor);
 	return login.isAuthenticated && meteor.connected;
 };
 
+// Backgrounding can outlive the socket without emitting a DDP event, so nothing syncs the open
+// room. `chat.syncMessages` rides REST, so it heals the room even against a dead socket.
+const resyncSubscribedRoom = function* resyncSubscribedRoom() {
```
diegolmello: We should discuss the need for this

app/sagas/state.js
```diff
@@ -26,7 +43,10 @@ const appHasComeBackToForeground = function* appHasComeBackToForeground() {
 	try {
 		const server = yield select(state => state.server.server);
 		yield localAuthenticate(server);
-		checkAndReopen();
+		yield resyncSubscribedRoom();
+		// A silently dead socket dropped every `stream-notify-user` update; this delta fetch is
+		// the only thing that heals the rooms list.
+		yield put(roomsRequest());
```
diegolmello: We should grill about this

app/views/RoomView/index.tsx
```diff
@@ -681,30 +695,45 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
 				const newLastOpen = new Date();
 				await RoomServices.getMessages({
 					rid: room.rid,
-					t: room.t as RoomType,
-					...('lastOpen' in room && room.lastOpen ? { lastOpen: room.lastOpen } : {})
+					// A subscribed room resumes from its sync cursor; only a room without a
+					// subscription row has to pull history by type.
+					...('id' in room ? {} : { t: room.t as RoomType })
```
diegolmello: This is wrong.

`loadMessagesForRoom` should be called to load more message pages, but also for the first load of a room.
The result max msg date should go to `sub.lastOpen`.

When `sub.lastOpen` has data, we call `loadMissedMessages`.

CONTEXT.md
```diff
@@ -104,6 +104,22 @@ Independent boolean markers on a Message, orthogonal to its Status — a Message
...
 
+## Timestamp Trust Boundary
+
+Not every `_updatedAt` is worth the same. A Message's `_updatedAt` read out of a **server response** is server truth, and is the only legitimate source for the sync cursor. The same field read off a **WatermelonDB row** is device-tainted: offline sends, Temp and Error sends, push-inserted rows, and `normalizeMessage`'s `_updatedAt || new Date()` fallback all stamp the device clock.
+
+Therefore the **Sync Watermark** must be taken from the raw payload _before_ `normalizeMessage` / `buildMessage` runs — never from a database row, and never from `Date.now()`.
+
+| Term                 | Definition                                                                                                                   | Aliases to avoid           |
+| -------------------- | ---------------------------------------------------------------------------------------------------------------------------- | -------------------------- |
+| **Sync Watermark**   | The fetch cursor for a Subscription (`lastOpen` column): the newest server `_updatedAt` actually received for that Room      | Last open, last update     |
```
diegolmello: Rename to Last Open.

CONTEXT.md
```diff
@@ -104,6 +104,22 @@ Independent boolean markers on a Message, orthogonal to its Status — a Message
...
+| **Last Seen**        | The Subscription's read receipt (`ls`): the newest Message the user has read, which anchors the Unread Separator             | Last open, last read       |
+| **Server Timestamp** | An `_updatedAt` taken from a server response — the only value the server can meaningfully compare a cursor against           | Timestamp (ambiguous)      |
+| **Device Timestamp** | An `_updatedAt` present on a local row but written by the device clock; unusable as a cursor because the server never saw it | Timestamp (ambiguous)      |
+| **Lying Cursor**     | A Sync Watermark ahead of the Room's newest Server Timestamp, so every sync drains empty and the Gap never closes            | Stale cursor, bad lastOpen |
```
diegolmello: We undo this logic. Caring about this migration is over engineering. Users can just clear cache when the fix lands.

app/containers/MessageActions/index.tsx
```diff
@@ -249,7 +249,7 @@ const MessageActions = memo(
 
 						await db.write(async () => {
 							try {
-								await subRecord.update(sub => (sub.lastOpen = ts as Date)); // TODO: reevaluate IMessage
+								await subRecord.update(sub => (sub.ls = ts as Date)); // TODO: reevaluate IMessage
```
diegolmello:
```suggestion
								await subRecord.update(sub => (sub.ls = ts as Date));
```

app/lib/methods/subscriptions/room.ts
```diff
@@ -45,6 +44,7 @@ export default class RoomSubscription {
 	}
 
 	subscribe = async () => {
+		this.isAlive = true;
```
diegolmello: why?

app/lib/methods/loadMissedMessages.ts
```diff
@@ -1,12 +1,19 @@
...
+import { type RoomTypes, types } from './roomTypeToApiType';
 import sdk from '../services/sdk';
 import { store } from '../store/auxStore';
 import { getSubscriptionByRoomId } from '../database/services/Subscription';
+import { getMessageById } from '../database/services/Message';
+import { writeSyncWatermark } from './writeSyncWatermark';
 
 const count = 50;
 
+const isRoomType = (t: unknown): t is RoomTypes => typeof t === 'string' && t in types;
```
diegolmello: Add to roomTypeToApiType file

app/sagas/state.js
diegolmello: Grilled live — keeping it. Nothing else heals the rooms list after a silently dead socket (`stream-notify-user` has no replay, and `roomsRequest` previously rode only `LOGIN.SUCCESS` + pull-to-refresh). The REST delta is two usually-empty calls per foreground, same trade as the accepted open-room resync above.

OtavioStasiak:
I tested this PR following the ticket video instructions and using a newly created channel and I could not reproduce the issue!
Related to the #7298

**Passing test cases:**

* Foreground reconnect: Background the app for 10 minutes, then foreground → reconnects within 1s with no persistent "Waiting for network".
* Brief background (Control Center / Notification Center) → no "Connecting" flicker.
* Airplane mode then foreground → no busy-loop; the SDK's existing reopen schedule still applies.

**Issue still present:**

* Lock-screen call accept: With the device locked and the app suspended for ≥30s, accepting an incoming VoIP call from the iOS lock screen still results in the audio path not opening correctly, and the peer disconnects after ~10s with `timeout-remote-sdp`.

**Potential known issue:**

* Cold-start handoff: Kill the app → receive a call → accept it from the lock screen.

  * If microphone permission has **not** been granted yet, the call is accepted initially but closes shortly afterward.
  * After granting microphone permission and repeating the same flow, the call works as expected.

app/lib/methods/writeSyncWatermark.ts
diegolmello: Renamed to `updateLastOpen`; scrubbed the "Sync Watermark" glossary term and stale comments. 9ec8dcfcb

app/lib/methods/writeSyncWatermark.ts
diegolmello: Simplified the max-date loop to `Math.max(...timestamps)` while keeping the NaN guard. 9ec8dcfcb

app/sagas/state.js
diegolmello: Kept. Foreground REST resync is the only path that heals a silently dead socket when DDP lifecycle events never fire. Discussed live; no gating.

app/views/RoomView/index.tsx
diegolmello: Fixed. Init dispatch now uses `lastOpen` presence, not subscription-row presence. First loads go to `loadMessagesForRoom`; existing cursors go to `loadMissedMessages`. 403284818

CONTEXT.md
diegolmello: Renamed glossary head term to Last Open and updated aliases/line 261 note. 9ec8dcfcb

CONTEXT.md
diegolmello: Kept the self-heal. Added a 5-minute per-room cooldown on tombstone-triggered tail loads instead of dropping it; "clear cache" is a support trap. 59b0d8b34

app/containers/MessageActions/index.tsx
diegolmello: Already applied — `sub.ls = ts` is in the branch. No further change.

app/lib/methods/subscriptions/room.ts
diegolmello: Removed the dead `this.isAlive = true;` reset in `subscribe()`. 9ec8dcfcb

app/lib/methods/loadMissedMessages.ts
diegolmello: Moved `isRoomType` guard into `roomTypeToApiType.ts` and imported it. 9ec8dcfcb

app/views/RoomView/index.tsx
```diff
@@ -678,33 +692,53 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
...
-						this.setLastOpen(room.ls);
+				if (currentJoined && 'id' in currentRoom) {
+					if (currentRoom.alert || currentRoom.unread || currentRoom.userMentions) {
+						this.setLastSeen(currentRoom.ls);
 					} else {
-						this.setLastOpen(null);
+						this.setLastSeen(null);
 					}
-					readMessages(room.rid, newLastOpen, true).catch(e => console.log(e));
+					readMessages(currentRoom.rid, readReceiptTime).catch(e => console.log(e));
```
diegolmello: `readMessages` should use the sync cursor. Otherwise we're making the same mistake of relying on device clock.

app/sagas/state.js
```diff
@@ -26,7 +47,15 @@ const appHasComeBackToForeground = function* appHasComeBackToForeground() {
...
 		yield localAuthenticate(server);
-		checkAndReopen();
+		yield resyncSubscribedRoom();
+		// A silently dead socket dropped every `stream-notify-user` update; this delta fetch is
+		// the only thing that heals the rooms list. Throttle it so rapid foreground cycles don't
+		// hammer the endpoint while still healing after a genuine absence.
+		const now = Date.now();
+		if (now - lastRoomsRequestAt >= ROOMS_REQUEST_THROTTLE_MS) {
+			lastRoomsRequestAt = now;
+			yield put(roomsRequest());
```
diegolmello: The app is going to call this on loginSuccess, so we're going to call it twice. What if we wait for the login instead of force calling it here?

app/lib/methods/updateLastOpen.ts
```diff
@@ -3,13 +3,28 @@ import { getSubscriptionByRoomId } from '../database/services/Subscription';
...
+
+		const lastOpen = new Date(Math.max(...timestamps));
+
 		const subscription = await getSubscriptionByRoomId(rid);
 		if (!subscription) {
 			return;
 		}
+
+		// Deliberately not monotonic: server `_updatedAt` is authoritative, so an older value only
+		// heals a poisoned cursor. The worst outcome is re-fetching history that is already local.
```
diegolmello: remove

app/containers/MessageActions/index.tsx
diegolmello: Remove the comment

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -139,6 +144,12 @@ export async function loadMessagesForRoom(args: {
 			}
 			await updateMessages({ rid: args.rid, update: messages, loaderItem: args.loaderItem });
 		}
+
+		// Only the initial tail load reaches the newest history, so only it can state how far
+		// the room is synced. Older pages (`latest`) and gap fills (`loaderItem`) must not.
```
diegolmello: Remove. It's stating what's in the code.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -41,60 +50,84 @@ const getSyncMessagesFromCursor = async (
...
-	lastOpen,
 	updatedNext,
 	deletedNext
 }: {
 	rid: string;
-	lastOpen?: Date;
 	updatedNext?: number | null;
 	deletedNext?: number | null;
 }) {
+	const sub = await getSubscriptionByRoomId(roomId);
```
diegolmello, line 56: If `!sub`, we could early return.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -41,60 +50,84 @@ const getSyncMessagesFromCursor = async (
...
+
+	const roomType = sub?.t;
+	if (!isRoomType(roomType)) {
+		log(new Error(`loadMissedMessages: cannot resolve room type for ${roomId}`));
+		return;
+	}
+	// A tail load re-anchors the Last Open from a real server response; it never re-enters here.
+	await loadMessagesForRoom({ rid: roomId, t: roomType });
+}
+
```
diegolmello: This logic is wrong.
While `subscription.lastMessage` works properly, it doesn't mean all the messages until that one are fetched. We can't rely on `subscription.lastMessage` or max from messages, since we persist that lastMessage as a message.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -105,16 +138,29 @@ export async function loadMissedMessages(args: {
 			deleted,
 			deletedNext
 		}: { updated: ILastMessage[]; deleted: ILastMessage[]; updatedNext: number | null; deletedNext: number | null } = data;
+
+		// Snapshot before updateMessages: buildMessage mutates these rows and stamps a
+		// device-clock `_updatedAt` onto any row that lacks one.
+		const serverUpdatedAt = updated.map(message => ({ _updatedAt: message._updatedAt }));
```
diegolmello: Duplicated code. Should be a helper.

app/lib/methods/readMessages.ts
```diff
@@ -45,9 +45,6 @@ export async function readMessages(rid: string, ls: Date, updateLastOpen = false
 					s.userMentions = 0;
 					s.groupMentions = 0;
 					s.ls = ls;
```
diegolmello: What if we remove ls from here?
We're going to receive the new ls value from the server a few ms after `'subscriptions.read'` lands anyway.
We wouldn't have to worry about dates here.

app/views/RoomView/index.tsx
```diff
@@ -681,18 +681,19 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
...
 				});
 
 				// if room is joined
 				if (joined && 'id' in room) {
 					if (room.alert || room.unread || room.userMentions) {
-						this.setLastOpen(room.ls);
+						this.setLastSeen(room.ls);
 					} else {
-						this.setLastOpen(null);
+						this.setLastSeen(null);
```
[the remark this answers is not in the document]
diegolmello: Fixed in b7556ec2b — went further than a rename: `readMessages` no longer writes `ls` at all, and the `ls` parameter is gone. The server already stamps `ls` on `subscriptions.read` and the subscription stream delivers it (`subscriptions/rooms.ts` `Object.assign(s, tmp)`), so the optimistic device-clock write was redundant as well as skew-prone. All three call sites (RoomView, `subscriptions/room.ts`, encryption saga) were passing `new Date()`; none do now.

One correction to the finding: `ls` no longer feeds any sync state after this PR — the cursor is `lastOpen`, and `ls` had exactly one reader left (the unread separator). So the impact was cosmetic and self-correcting, not a sync-cursor regression.

app/views/RoomView/index.tsx
```diff
@@ -118,7 +118,7 @@ import { isInviteSubscription } from '../../lib/methods/isInviteSubscription';
 
 const EMPTY_HIDE_SYSTEM_MESSAGES: string[] = [];
 
-class RoomView extends Component<IRoomViewProps, IRoomViewState> {
+export class RoomView extends Component<IRoomViewProps, IRoomViewState> {
```
diegolmello, line 121: **Standards — test-only code in a non-test file.** This named export exists only so `index.test.tsx` can render the unconnected class; the real export is still the connected default. Prefer a test harness that renders the default (theme/redux providers) over widening production API for tests.

app/views/RoomView/index.tsx
```diff
@@ -678,21 +678,21 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
...
 				}
 			} else {
-				const newLastOpen = new Date();
 				await RoomServices.getMessages({
 					rid: room.rid,
-					t: room.t as RoomType,
-					...('lastOpen' in room && room.lastOpen ? { lastOpen: room.lastOpen } : {})
+					// A room with a sync cursor resumes from it; a room without one must
+					// pull recent history and establish the cursor from the response.
+					...('lastOpen' in room && room.lastOpen ? {} : { t: room.t as RoomType })
```
diegolmello: **Standards — control coupling.** The decision is now encoded in the *absence* of `t`: omitting it silently selects a different endpoint downstream. Simpler than the old union, but the intent lives in a hole rather than in a name. An explicit discriminator (or two named call sites) reads better and survives a future third mode.

app/views/RoomView/index.tsx
```diff
@@ -678,21 +678,21 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
...
 				if (joined && 'id' in room) {
 					if (room.alert || room.unread || room.userMentions) {
-						this.setLastOpen(room.ls);
+						this.setLastSeen(room.ls);
 					} else {
-						this.setLastOpen(null);
+						this.setLastSeen(null);
 					}
-					readMessages(room.rid, newLastOpen, true).catch(e => console.log(e));
+					readMessages(room.rid).catch(e => console.log(e));
```
diegolmello, line 693: **Spec — "Separator behaviour is unchanged" doesn't hold.** `setLastSeen(room.ls)` is unchanged, but `ls` is now *only* ever written by the server (stream) after `readMessages` stopped writing it. On an offline or slow-stream read the local row keeps a stale `ls` while `alert`/`unread`/`userMentions` are already cleared, so the separator can render above messages the user has read until the stream lands. Either this claim or the `ls` removal needs to change.

app/lib/methods/readMessages.ts
```diff
@@ -36,6 +36,9 @@ export async function readMessages(rid: string, ls: Date, updateLastOpen = false
 		// @ts-ignore
 		await sdk.post('subscriptions.read', { rid });
 
+		// `ls` is intentionally not written here: the server stamps it and the
+		// subscription stream delivers it, so the device clock never sets it.
+
```
diegolmello: **Spec — contradicts the PR description**, which says `readMessages` "no longer writes `lastOpen` at all (**it owns `ls`**)". It no longer writes `ls` either; the only remaining writers are `subscriptions/rooms.ts` (stream) and `MessageActions`. So the shipped behaviour is the opposite of that sentence — fix the sentence or restore the `ls` write.

**Standards — comment describes code that isn't here** (`no-noise.md`: comments are technical notes about code that exists; delete comments about absent code). The same invariant is copy-pasted into `loadNextMessages.ts` and `loadSurroundingMessages.ts`; CONTEXT.md's new Timestamp Trust Boundary section is the single place it belongs.

app/lib/methods/loadNextMessages.ts
```diff
@@ -36,6 +36,8 @@ export function loadNextMessages(args: ILoadNextMessages): Promise<void> {
 					};
 					messages.push(loadMoreItem);
 				}
+				// Jump paths never write the Last Open: a ts-ordered forward walk never sees the
+				// `_updatedAt` of pre-anchor messages, so any cursor derived from it would skip history.
```
diegolmello: **Standards — duplicated comment.** Byte-identical to the block in `loadSurroundingMessages.ts`, and a third variant sits in `readMessages.ts`. Three copies of one invariant drift independently; state it once in CONTEXT.md.

app/lib/methods/updateLastOpen.ts
```diff
@@ -3,13 +3,28 @@ import { getSubscriptionByRoomId } from '../database/services/Subscription';
 import log from './helpers/log';
 import { type TSubscriptionModel } from '../../definitions';
 
-export async function updateLastOpen(rid: string, lastOpen = new Date()): Promise<void> {
+/**
+ * Persists the Last Open as the newest server `_updatedAt` actually received, so the next
+ * `chat.syncMessages` resumes from a server-clock value the server can compare against.
+ * Must run on the RAW payload: normalizeMessage invents `_updatedAt` from the device clock.
+ */
+export async function updateLastOpen(rid: string, payload: { _updatedAt?: string | Date }[]): Promise<void> {
```
diegolmello: **Standards — Data Clumps / Primitive Obsession.** `{ _updatedAt?: string | Date }[]` is spelled inline here, in `loadMessagesForRoom.ts` and in `loadMissedMessages.ts`. CONTEXT.md now *names* this concept ("Server Timestamp") — give it a type and reuse it.

app/lib/methods/updateLastOpen.ts
```diff
@@ -3,13 +3,28 @@ import { getSubscriptionByRoomId } from '../database/services/Subscription';
...
 	try {
-		const db = database.active;
+		const timestamps = payload.map(m => new Date(m._updatedAt as string | Date).getTime()).filter(t => !Number.isNaN(t));
+		if (!timestamps.length) {
+			return;
+		}
+
+		const lastOpen = new Date(Math.max(...timestamps));
+
```
diegolmello: **Spec — empty payload never establishes a cursor.** A room whose initial tail load returns no messages returns early here and keeps `lastOpen` unset forever, so every subsequent open re-enters the full `loadMessagesForRoom` fallback and "resolves the cursor from the subscription row" never becomes true for it. Direction is safe, but worth either handling or stating as accepted.

app/lib/methods/loadMessagesForRoom.ts
```diff
@@ -73,6 +75,11 @@ async function load(args: {
 		}
 
 		const batch = data.messages as IMessage[];
+		if (batchesFetched === 1) {
+			// Snapshot before updateMessages: buildMessage mutates these rows and stamps a
+			// device-clock `_updatedAt` onto any row that lacks one.
+			firstBatch = batch.map(message => ({ _updatedAt: message._updatedAt }));
+		}
```
diegolmello: **Standards — Duplicated Code.** This block and its comment appear byte-identical in `loadMissedMessages.ts`. One `snapshotServerTimestamps(payload)` helper next to `updateLastOpen` removes both copies and the comment.

**Standards — Mysterious Name.** `firstBatch` holds timestamps, not messages.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -41,60 +46,53 @@ const getSyncMessagesFromCursor = async (
...
+	if (!cursor && !updatedNext && !deletedNext) {
+		const roomType = sub?.t;
+		if (!isRoomType(roomType)) {
+			log(new Error(`loadMissedMessages: cannot resolve room type for ${roomId}`));
+			return null;
+		}
+		await loadMessagesForRoom({ rid: roomId, t: roomType });
+		return null;
+	}
+
```
diegolmello, line 61: **Spec — the no-op the PR says it removed.** The description promises it "falls back to `loadMessagesForRoom` when there is no cursor instead of no-op'ing", but when the subscription row is missing or `t` is unrecognised this logs and `return null` — a silent no-op, on the exact path SUP-1078 describes (room first opened from a push notification). `getMessages.ts` has `t` in hand and deliberately drops it; pass it through so the fallback can't be starved.

**Standards — Divergent Change.** `load()` now resolves the cursor, heals a future-dated cursor, chooses a fallback loader *and* branches on server version. Extracting `resolveSyncCursor(roomId)` would leave one reason to change here.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -105,16 +103,26 @@ export async function loadMissedMessages(args: {
 			deleted,
 			deletedNext
 		}: { updated: ILastMessage[]; deleted: ILastMessage[]; updatedNext: number | null; deletedNext: number | null } = data;
+
+		// Snapshot before updateMessages: buildMessage mutates these rows and stamps a
+		// device-clock `_updatedAt` onto any row that lacks one.
+		const serverUpdatedAt = updated.map(message => ({ _updatedAt: message._updatedAt }));
+
```
diegolmello, line 96: **Standards — Duplicated Code** (see `loadMessagesForRoom.ts`): same snapshot shape, same comment. Also `serverUpdatedAt` is a list with a singular name.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -105,16 +103,26 @@ export async function loadMissedMessages(args: {
...
 				updatedNext,
 				deletedNext
 			});
 		}
+
+		// Only once the UPDATED cursor has drained. Advancing mid-pagination would skip the
+		// pages not yet fetched; `deleted` is never a source, its rows carry no new history.
+		if (!updatedNext) {
+			await updateLastOpen(args.rid, serverUpdatedAt);
+		}
```
diegolmello, line 115: **Spec — accepted cost worth stating.** This writes the *last* page's max `_updatedAt`, and `updateLastOpen` is deliberately non-monotonic, so if `chat.syncMessages` pages descend the final write lands below earlier pages and the cursor moves backwards. The PR's own table marks that direction "wasteful, safe", so not a defect — but it's an unstated consequence of the non-monotonic choice.

app/containers/MessageActions/index.tsx
diegolmello: **Spec — unspec'd change.** Correct under the `lastOpen`/`ls` split, but the mark-unread path isn't mentioned anywhere in the PR description. Worth a line, since it's now the only device-clock writer of `ls` left.

app/views/RoomView/services/getMessages.ts
```diff
@@ -2,23 +2,17 @@ import { loadMessagesForRoom } from '../../../lib/methods/loadMessagesForRoom';
...
-interface ILoadMissedMessagesParams extends IBaseParams {
-	lastOpen: Date;
-}
-
-const getMessages = (params: ILoadMissedMessagesParams | ILoadMessagesForRoomParams): Promise<void> => {
-	if ('lastOpen' in params) {
-		return loadMissedMessages(params);
+const getMessages = ({ rid, t }: IGetMessagesParams): Promise<void> => {
+	if (!t) {
+		return loadMissedMessages({ rid });
```
diegolmello, line 12: **Standards — the discarded `t`.** The old union at least made the two modes nameable. Now `t` is dropped on the sync path even though `loadMissedMessages` needs precisely that value to run its fallback (see the comment there) and re-reads it from the DB, where it may be absent. Threading `t` through removes both the re-read and its failure branch.

diegolmello (commented):
Two-axis review of this PR against its base (`purring-seahorse`): **Standards** (repo + project conventions, plus a Fowler smell baseline) and **Spec** (SUP-1078 + the PR description). Findings are inline. Worst on each axis: the test-only `export class RoomView`, and `loadMissedMessages` still silently no-op'ing when the subscription row / room type can't be resolved — which is the exact failure mode SUP-1078 reports.

One finding has no diff line to attach to: `app/views/RoomView/docs/ARCHITECTURE.md` keeps an Invariants section, and the new "init dispatches on cursor presence" predicate isn't recorded there even though `index.tsx` init behaviour changed.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -41,60 +46,53 @@ const getSyncMessagesFromCursor = async (
...
 }) {
+	const sub = await getSubscriptionByRoomId(roomId);
+	const persistedCursor = sub?.lastOpen;
+
+	// A cursor in the future was written from a skewed device clock; the server would report
+	// nothing newer than it, permanently hiding messages. Treat it as absent so it self-heals.
+	const cursor = persistedCursor && persistedCursor.getTime() > Date.now() ? undefined : persistedCursor;
+
+	// A room first opened from a push notification has no cursor; syncing from
+	// nothing would fetch nothing, so fall back to a full recent-history load.
```
diegolmello: > A room first opened from a push notification has no cursor

This is wrong. Some rooms will have.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -41,60 +46,53 @@ const getSyncMessagesFromCursor = async (
...
-	lastOpen?: Date;
 	updatedNext?: number | null;
 	deletedNext?: number | null;
 }) {
+	const sub = await getSubscriptionByRoomId(roomId);
+	const persistedCursor = sub?.lastOpen;
+
+	// A cursor in the future was written from a skewed device clock; the server would report
+	// nothing newer than it, permanently hiding messages. Treat it as absent so it self-heals.
+	const cursor = persistedCursor && persistedCursor.getTime() > Date.now() ? undefined : persistedCursor;
```
diegolmello: Remove this logic. We don't need to check if lastOpen is in a wrong state.

app/lib/methods/loadNextMessages.ts
diegolmello: Remove

app/lib/methods/loadSurroundingMessages.ts
```diff
@@ -50,6 +50,8 @@ export function loadSurroundingMessages({ messageId, rid }: { messageId: string;
 					}
 				}
 
+				// Jump paths never write the Last Open: a ts-ordered forward walk never sees the
+				// `_updatedAt` of pre-anchor messages, so any cursor derived from it would skip history.
```
diegolmello: Remove

app/lib/methods/readMessages.ts
diegolmello: Remove

app/lib/methods/updateLastOpen.ts
```diff
@@ -3,13 +3,28 @@ import { getSubscriptionByRoomId } from '../database/services/Subscription';
 import log from './helpers/log';
 import { type TSubscriptionModel } from '../../definitions';
 
-export async function updateLastOpen(rid: string, lastOpen = new Date()): Promise<void> {
+/**
+ * Persists the Last Open as the newest server `_updatedAt` actually received, so the next
+ * `chat.syncMessages` resumes from a server-clock value the server can compare against.
+ * Must run on the RAW payload: normalizeMessage invents `_updatedAt` from the device clock.
+ */
```
diegolmello: Remove

app/views/RoomView/services/getMessages.ts
```diff
@@ -2,23 +2,17 @@ import { loadMessagesForRoom } from '../../../lib/methods/loadMessagesForRoom';
 import { loadMissedMessages } from '../../../lib/methods/loadMissedMessages';
 import { type RoomTypes } from '../../../lib/methods/roomTypeToApiType';
 
-interface IBaseParams {
+interface IGetMessagesParams {
 	rid: string;
+	/** Omit for a subscribed room: it syncs from its own cursor instead of reloading history. */
```
diegolmello: Remove

app/views/RoomView/index.tsx
```diff
@@ -678,21 +678,21 @@ class RoomView extends Component<IRoomViewProps, IRoomViewState> {
 					this.consumeJumpParam(messageId);
 				}
 			} else {
-				const newLastOpen = new Date();
 				await RoomServices.getMessages({
 					rid: room.rid,
-					t: room.t as RoomType,
-					...('lastOpen' in room && room.lastOpen ? { lastOpen: room.lastOpen } : {})
+					// A room with a sync cursor resumes from it; a room without one must
+					// pull recent history and establish the cursor from the response.
```
diegolmello: Remove

app/lib/methods/loadMissedMessages.ts
```diff
@@ -41,60 +46,53 @@ const getSyncMessagesFromCursor = async (
...
+	// nothing would fetch nothing, so fall back to a full recent-history load.
+	if (!cursor && !updatedNext && !deletedNext) {
+		const roomType = sub?.t;
+		if (!isRoomType(roomType)) {
+			log(new Error(`loadMissedMessages: cannot resolve room type for ${roomId}`));
+			return null;
+		}
+		await loadMessagesForRoom({ rid: roomId, t: roomType });
+		return null;
+	}
```
diegolmello: Remove this new logic and keep getMessages as the gate to call loadMessagesForRoom or loadMissedMessages

app/lib/methods/loadMissedMessages.ts
diegolmello: Removed in 480e381be. Every cursor write is server-derived now, so a skewed device clock can no longer poison `lastOpen`. Accepted cost: a cursor already poisoned on a pre-upgrade install does not self-heal through this path — the server reports nothing newer than a future cursor, so the room stays stuck until real time passes the poisoned stamp.

app/lib/methods/loadMissedMessages.ts
diegolmello: Removed in 480e381be — `getMessages` is the sole gate. Parity check for the notification-tap case: the RoomView gate (`RoomView/index.tsx:685`) passes `t` whenever the cursor is absent, both for the reduced nav-params room (no `lastOpen` key) and for a full subscription with `lastOpen: null`, so a push-opened room lands in `loadMessagesForRoom` exactly as this fallback routed it — and `loadMessagesForRoom` establishes the cursor from the response. The only other `loadMissedMessages` caller is `room.ts` `handleConnection`, which can only run after init, i.e. after the cursor exists.

app/views/RoomView/services/getMessages.ts
diegolmello: The concrete half of this dissolved in `480e381be`: the fallback, the DB re-read of `sub.t`, and the `isRoomType` failure branch are gone — `loadMissedMessages` now reads only `sub.lastOpen`. Leaving the optional `t?` as is (no discriminated-union rename).

app/containers/MessageActions/index.tsx
diegolmello: Added a line to the PR body naming the mark-unread path as the only device-clock writer of `ls` left.

app/lib/methods/loadMessagesForRoom.ts
diegolmello: Done in `4c08b65b1` — both loaders now accumulate the max server `_updatedAt` across every batch/page and write the cursor once at the end. `loadMessagesForRoom` snapshots inside `fetchBatch` for all batches (fetch-order-proof); `loadMissedMessages` threads the accumulator through the pagination recursion and writes only on the final UPDATED page (a DELETED-only continuation can't write again; an abort mid-pagination writes nothing). Regression tests added in both loaders: a later batch/page with a newer `_updatedAt` wins.

app/lib/methods/loadMissedMessages.ts
diegolmello: The backwards walk is gone in `4c08b65b1` — the cursor now accumulates the max `_updatedAt` across all pages and is written once when the UPDATED cursor drains, so the last-page-wins cost this asked to document no longer exists. No PR-body line needed.

app/lib/methods/loadMissedMessages.ts
diegolmello: Done in `00d163ee6` — `load()` returns early when there is no subscription.

app/lib/methods/updateLastOpen.ts
diegolmello: Done in `18d3a8744` — the carrier is now `TServerTimestamps` in `updateLastOpen.ts`, reused at both call sites.

app/lib/methods/loadMessagesForRoom.ts
diegolmello: Done in `18d3a8744` — `snapshotServerTimestamps` lives beside `updateLastOpen` and both loaders use it; the duplicated block and its comment are gone. The accumulator is named `serverTimestamps` (it holds every batch since `4c08b65b1`, so `firstBatch` died with the first-batch rule).

app/lib/methods/loadMissedMessages.ts
diegolmello: Done in `18d3a8744` — same `snapshotServerTimestamps` extraction, and the list is now plural (`serverTimestamps`).

app/lib/methods/readMessages.ts
diegolmello: Removed in `9537a040b`.

app/lib/methods/loadNextMessages.ts
diegolmello: Removed in `9537a040b`. One flag for the record: this comment (and its twin in `loadSurroundingMessages`) stated why jump paths must never write the cursor — with both gone, nothing warns that adding an `updateLastOpen` call on a jump path silently skips history.

app/lib/methods/loadSurroundingMessages.ts
diegolmello: Removed in `9537a040b`.

app/views/RoomView/services/getMessages.ts
diegolmello: Removed in `9537a040b`.

app/views/RoomView/index.tsx
diegolmello: Removed in `9537a040b`.

app/lib/methods/loadMessagesForRoom.ts
diegolmello: Removed in `9537a040b`.

app/lib/methods/updateLastOpen.ts
diegolmello: Removed in `9537a040b`.

app/lib/methods/updateLastOpen.ts
diegolmello: Removed in `9537a040b`.

app/containers/MessageActions/index.tsx
diegolmello: Applied in `9537a040b` — same line, TODO trailer dropped.

app/lib/methods/loadMissedMessages.ts
diegolmello: That comment was deleted with the guards in `480e381be`.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -105,16 +129,29 @@ export async function loadMissedMessages(args: {
...
 		await updateMessages({ rid: args.rid, update: updated, remove: deleted });
 
 		if (deletedNext || updatedNext) {
 			loadMissedMessages({
 				rid: args.rid,
-				lastOpen: args.lastOpen,
 				updatedNext,
 				deletedNext
 			});
 		}
```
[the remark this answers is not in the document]
diegolmello: Done in `935a28641` — the continuation now carries a `.catch(log)`. Kept un-awaited on purpose: awaiting would serialise pagination behind the cursor write.

app/lib/methods/loadMissedMessages.ts
```diff
@@ -105,16 +129,29 @@ export async function loadMissedMessages(args: {
...
+
+		// Only once the UPDATED cursor has drained. Advancing mid-pagination would skip the
+		// pages not yet fetched; `deleted` is never a source, its rows carry no new history.
+		if (!updatedNext) {
+			await writeSyncWatermark(args.rid, serverUpdatedAt);
+			if (!updated.length) {
+				await healLyingCursor(args.rid);
+			}
+		}
```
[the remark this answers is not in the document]
diegolmello: `healLyingCursor` no longer exists on this branch, and since `4c08b65b1` a DELETED-only continuation page is explicitly guarded from re-entering `updateLastOpen`, so nothing re-runs per deleted page anymore.

app/lib/methods/updateLastOpen.ts
```diff
@@ -3,13 +3,28 @@ import { getSubscriptionByRoomId } from '../database/services/Subscription';
...
+export async function updateLastOpen(rid: string, payload: { _updatedAt?: string | Date }[]): Promise<void> {
 	try {
-		const db = database.active;
+		const timestamps = payload.map(m => new Date(m._updatedAt as string | Date).getTime()).filter(t => !Number.isNaN(t));
```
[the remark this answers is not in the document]
diegolmello: Done in `935a28641` — `null` rows are skipped before `Date` construction (the carrier type now admits `null`), and a null-only payload test asserts no write.

app/lib/methods/subscriptions/room.resumeSync.test.ts
```diff
@@ -0,0 +1,102 @@
...
+	_id: 'missed-1',
+	rid: RID,
+	msg: 'sent while the app was backgrounded',
+	ts: new Date(Date.UTC(2024, 0, 1, 12, 0, 0)).toISOString(),
+	u: { _id: 'user2', username: 'user2' }
+};
+
+const syncMessagesResponse = (updated: unknown[]) => ({
+	result: { updated, deleted: [], cursor: { next: null } }
+});
```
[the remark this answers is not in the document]
diegolmello: Done in `935a28641`.

app/lib/methods/subscriptions/room.resumeSync.test.ts
```diff
@@ -0,0 +1,102 @@
...
+		mockedUpdateMessages.mockResolvedValue(0);
+		mockedSdkGet.mockResolvedValue(syncMessagesResponse([missedMessage]) as never);
+	});
+
+	it('fetches and persists messages missed while backgrounded when the room has a sync cursor', async () => {
+		mockedGetSubscriptionByRoomId.mockResolvedValue({ lastOpen: new Date(Date.UTC(2024, 0, 1, 11, 0, 0)) } as never);
+
+		await new RoomSubscription(RID).handleConnection();
+
+		expect(mockedSdkGet).toHaveBeenCalledWith('chat.syncMessages', expect.objectContaining({ roomId: RID, type: 'UPDATED' }));
```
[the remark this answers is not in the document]
diegolmello: Done in `935a28641` — the cursor test now asserts `chat.syncMessages` receives the persisted `lastOpen` as `next`.

OtavioStasiak (approved):
LGTM

PR #7526, opened 2026-07-28
fix(voip): reconnect stale socket before native call accept

diegolmello:
## Proposed changes

With the device locked and the app suspended ≥30s, accepting an incoming VoIP call from the lock screen died at 10s with `timeout-remote-sdp`. iOS freezes the TCP socket silently; the SDK keeps reading `connected === true` through its 40s tolerance window, `send()` hangs forever awaiting `'open'`, and the only reconnection trigger (the foreground saga) never fires while the device stays locked. The native accept path had no connection gate at all, so everything needed for audio — inbound `remote-sdp`, outbound `local-sdp`/`local-state` — rode a dead socket while the REST-based native accept succeeded.

- **SDK patch**: `reopenNow()` forces a single shared reconnect — clears the pending reopen timer, tears down the orphan socket, rejects in-flight `send()` promises via a real `'disconnected'` emit (previously nothing ever emitted it, so sends hung forever), and preserves `Socket.subscriptions` so DDP login's `subscribeAll()` restores every stream. Forced reopens are serialized against concurrent `open()`. No forceReopen, no subscription wipe, no synthetic close event — `open()`'s synchronous `'connecting'` emit flips redux honestly.
- **`probe()`**: bounded raw-socket liveness check for the gray zone (20s < lastPing age < 40s), so healthy-but-quiet sockets don't get force-reconnected.
- **Accept gate** (`acceptNativeCallWithReadiness`): every native accept/replay path (warm accept, cold start, init replay, live-signal auto-answer) funnels through one gate — classify socket health, reopen if stale, wait for login readiness **and** media-signal subscription ack (server does no queueing; accepting before the ack loses post-accept signaling), then replay state signals and answer. 8s cap; on failure the native call is terminated and a hangup queued. Concurrent accepts for the same call abort the older gate without terminating the call.
- **Foreground saga**: reconnects immediately when foregrounding a stale socket (probe first in the gray zone) — app open after suspension no longer waits out the ping timeout, which also covers deep links and notification taps.

## Issue(s)

## How to test or reproduce

Automated: `TZ=UTC pnpm test` (2045 tests — new coverage in `ddpSocket.test.ts`, `acceptNativeCall.test.ts`, `state.test.ts`, `waitForLoginReady.socketHealth.test.ts`, `VoipCallLifecycle.integration.test.tsx`).

Manual: items 1–3 need physical devices (CallKit / ConnectionService lock-screen UI and WebRTC audio do not work on a simulator). Items 1 and 2 need a second account on a second device to place the call.

1. **Lock-screen accept after deep suspension** (iOS + Android) — log in, lock the device, wait at least 60s with the screen off. Place a VoIP call from the second account and accept it from the lock screen without unlocking. Expected: two-way audio within a few seconds, call survives past 10s, no `timeout-remote-sdp`, no silent call.
2. **App force-quit accept** (iOS + Android) — swipe the app away, then place a call from the second account and accept the native incoming UI. Expected: the app starts, logs in, and audio connects.
3. **Airplane mode on/off** (iOS + Android) — (a) during a call, turn airplane mode on for ~20s then off: the call ends on both sides, no orphan CallKit/native call entry remains, and the app returns to connected without a restart. (b) while idle, background the app, turn airplane mode on for ~60s then off, and foreground the app: it reconnects promptly and the room list loads.
4. **Background 1 minute on a stable connection** (iOS + Android, emulator is fine) — background the app for 60s without touching the network, then foreground it. Expected: no connection banner and no room-list refetch.

Before this PR, item 1 gave silence and then a drop at 10s with `timeout-remote-sdp`.

> Verified so far on physical hardware: item 2 (force-quit accept, one-minute call, no issues). The remaining items are still pending.

## Screenshots

N/A — connection-layer change, no UI.

## Types of changes

## Checklist

## Further comments

> **Merge order**: stacked on #7521 (base `purring-seahorse`). Merge only after #7521 lands, then retarget this PR to `develop`.

Deliberately out of scope — follow-ups to raise after this PR:

- **Mic-lock fix (media-signaling)**: `Session.requestInputTrackUpdate` drops track updates with no requeue; can independently produce the same 10s `timeout-remote-sdp` on a healthy socket. Needs a patch-package requeue + sequential `await processSignal`.
- **stateSignals re-poll** at +2s/+5s while `waiting-for-offer`, closing the sub-ack vs first-offer race on slow networks.
- **iOS cold-start accept** (app killed): `getInitialMediaCallEvents` skips `appInit`, so replay no-ops and no login ever runs.
- **App-wide ping-send deadline** so a hung ping can't kill the SDK's self-healing chain (rooms benefit too).
- **Server-side REST callee-hangup endpoint** — today a callee that fails to connect can only terminate locally; the caller rings until expiry.
- **Deep-link readiness gating** — deep-link/push paths still gate on server selection rather than live login state.

## Summary by CodeRabbit

* **New Features**
  * Added a readiness gate to native VoIP call acceptance (login, socket health, and media subscription readiness) before answering.
  * Improved foreground DDP recovery using socket staleness (reopen vs one-at-a-time probe), plus readiness-aware media subscription re-subscribe.
* **Bug Fixes**
  * Fixed VoIP permission change handling so the active call isn’t cleared during an active call.
  * Updated accepted VoIP-call handling to consistently use readiness-gated acceptance and improved error logging.
* **Tests**
  * Expanded integration and unit coverage for socket health/login readiness, DDP reconnect behavior, and VoIP acceptance flows.
* **Chores**
  * Added dev reconnect tracing markers and updated Jest transform/patch configuration.


app/lib/services/ddpSocket.test.ts
```diff
@@ -0,0 +1,328 @@
...
+		expect(socket.reopenPromise).toBeUndefined();
+
+		const secondPromise = socket.reopenNow();
+		expect(mockConnections).toHaveLength(2);
+
+		mockConnections[1].onopen();
+		await jest.runOnlyPendingTimersAsync();
+		await secondPromise;
+		jest.useRealTimers();
+	});
```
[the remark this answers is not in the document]
diegolmello: Fixed in `409c796df` — moved the restore into an `afterEach` on the `Socket.reopenNow` describe and dropped the trailing in-test `jest.useRealTimers()`.

app/lib/services/ddpSocket.test.ts
```diff
@@ -0,0 +1,328 @@
...
+			id: 'sub-mc',
+			name: 'stream-notify-user',
+			params: ['uid/media-calls'],
+			unsubscribe: jest.fn()
+		};
+
+		await jest.advanceTimersByTimeAsync(100);
+		await expect(promise).resolves.toBe(true);
+		expect(driver.subscribe).toHaveBeenCalledTimes(2);
+	});
```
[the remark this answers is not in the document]
diegolmello: Fixed in `f1e7fe13a` — added a case that inserts only `media-signal`, advances the poll, asserts the promise is still unresolved and that no resubscribe was issued, then inserts `media-calls` and asserts it resolves `true`. It passes because `attempt()` now requires every name (same commit).

app/lib/services/voip/acceptNativeCall.test.ts
```diff
@@ -0,0 +1,286 @@
...
+import sdk from '../sdk';
+
+const mockWaitForLoginReady = waitForLoginReady as jest.MockedFunction<typeof waitForLoginReady>;
+const mockGetState = useCallStore.getState as jest.Mock;
+const mockTerminateNativeCall = terminateNativeCall as jest.Mock;
+const mockDdp = () => sdk.current?.ddp as any;
+
+jest.mock('../waitForLoginReady', () => ({
+	waitForLoginReady: jest.fn()
+}));
```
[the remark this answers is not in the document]
diegolmello: Fixed in `409c796df` — removed the dead first registration, kept the `requireActual`-spreading one.

app/lib/services/voip/acceptNativeCall.ts
```diff
@@ -0,0 +1,129 @@
...
+			return;
+		}
+
+		const { call } = useCallStore.getState();
+		if (call?.callId !== callId) {
+			await mediaSession.answerCall(callId);
+		}
+	} finally {
+		cleanup();
+	}
```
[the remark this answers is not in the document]
diegolmello: Fixed in `4ba144544` — added `catch (error) { log(error); if (!controller.signal.aborted) handleFailure(callId, mediaSession); }` before the `finally`. The aborted guard is necessary: abort is signalled by `controller.signal.aborted` early returns, never by throwing, so terminating on an aborted gate would reintroduce the regression fixed in `83c9af6ca`.

patches/@rocket.chat+sdk+1.3.3-mobile.patch
```diff
@@ -45,7 +104,125 @@ index 19d31ae..9f752df 100644
...
++        if (timeout) clearTimeout(timeout as any)
++        delete this.reopenPromise
++        resolve()
++      }
++
++      this.once('open', cleanup)
++
++      this.createConnection().catch(() => {})
++
++      const timeout = setTimeout(() => cleanup(), 10000)
```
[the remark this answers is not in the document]
diegolmello: Declining — not reachable. `createConnection()` cannot emit `'open'` synchronously: its executor only constructs the WebSocket, tears down the old connection (nulling handlers before `close()`, so there is no synchronous `onclose` path either), assigns handlers, and emits `'connecting'`. `'open'` comes from `connection.onopen`, which the WS runtime always fires on a later tick, so the TDZ window can't be hit. `cleanup` also guards with `if (timeout)`. Agreed the ordering reads fragile, but there is no defect to fix here.

patches/@rocket.chat+sdk+1.3.3-mobile.patch
```diff
@@ -56,3 +233,78 @@ index 19d31ae..9f752df 100644
...
++    const findSubs = () => Object.keys(this.ddp.subscriptions || {})
++      .map(id => this.ddp.subscriptions[id])
++      .filter((sub: any) => (
++        sub &&
++        sub.name === topic &&
++        names.some(name => sub.params?.[0] === `${userId}/${name}`)
++      ))
++    const resubscribe = (subs: any[]) => Promise.all(
++      subs.map((sub: any) => this.subscribe(topic, sub.params[0], false, sub.id))
++    )
```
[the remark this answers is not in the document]
diegolmello: Fixed in `f1e7fe13a` — confirmed the id was dropped: only two args reached `Socket.subscribe(name, params, callback?, id?)`, so the server received the sub id as an extra stream arg *and* `Socket.subscribe` wrote a new `subscriptions[newId]` entry without removing the old one, growing duplicates on every readiness gate. The resubscribe now bypasses the driver wrapper and calls `this.ddp.subscribe(topic, sub.params, undefined, sub.id)` — the same 4-arg shape `subscribeAll` uses — reusing the stored `params` verbatim. Test expectations updated to assert against the real `ddp.subscribe` signature.

patches/@rocket.chat+sdk+1.3.3-mobile.patch
```diff
@@ -56,3 +233,78 @@ index 19d31ae..9f752df 100644
...
++        const subs = findSubs()
++        if (subs.length > 0) {
++          resubscribe(subs).then(finish)
++        }
++      }
++      const deadline = setTimeout(() => finish(false), timeoutMs)
++      const poll = setInterval(attempt, 100)
++      attempt()
++    })
++  }
```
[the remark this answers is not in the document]
diegolmello: Fixed in `f1e7fe13a` — `attempt()` now gates on `names.every(name => subs.some(sub => sub.params?.[0] === `${userId}/${name}`))`, and an `inFlight` flag set before `resubscribe` and cleared when it settles prevents the 100ms interval from issuing overlapping re-subscribes. The partial-state test case was added in `app/lib/services/ddpSocket.test.ts`.

patches/@rocket.chat+sdk+1.3.3-mobile.patch
diegolmello, line 142: **Bug (TDZ).** `cleanup()` closes over `timeout`, but `const timeout` is declared *after* `this.once('open', cleanup)` (line 138) and after `createConnection()`. A synchronous `'open'` emit runs `cleanup()` before the binding is initialized → `ReferenceError`. Move the `setTimeout` above the listener registration.

patches/@rocket.chat+sdk+1.3.3-mobile.patch
```diff
@@ -45,7 +104,125 @@ index 19d31ae..9f752df 100644
...
++        if (timeout) clearTimeout(timeout as any)
++      }
++
++      const onPong = () => {
++        if (this.lastPing <= lastPingAtStart) return
++        cleanup()
++        resolve(true)
++      }
++
++      this.once('pong', onPong)
```
diegolmello, line 174: **Bug.** `this.once('pong', onPong)` — but `onPong` early-returns when `lastPing` hasn't advanced. `once` has already consumed the listener at that point, so a first pong that doesn't advance `lastPing` leaves nothing listening and `probe()` can only ever time out to `false`. Use `on('pong', onPong)` plus an explicit `off` in the cleanup path.

app/lib/services/waitForLoginReady.ts
```diff
@@ -0,0 +1,71 @@
+import { store } from '../store/auxStore';
+
+export function classifySocketHealth(ddp: {
+	lastPing: number;
+	pingInterval?: number;
+	config?: { ping?: number };
+}): 'healthy' | 'probe' | 'reopen' {
+	const pingInterval = (ddp.pingInterval ?? ddp.config?.ping) || 10000;
+	const age = Date.now() - ddp.lastPing;
+	if (age > pingInterval * 2) {
```
diegolmello: **Spec mismatch.** PR body: *"bounded raw-socket liveness check for the gray zone (20s < lastPing age < 40s)"*. `pingInterval` resolves to `config.ping`, which defaults to `10000` (`sdk.ts` passes no `timeout`), so the real bands are 10–20s (probe) and >20s (reopen) — half the documented numbers. Same for the claimed "40s tolerance window" (`checkAndReopen` uses `ping * 2` = 20s). Either the body or the multipliers is wrong.

app/lib/services/waitForLoginReady.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Reads redux rather than `ddp.loggedIn`: `close` clears `meteor.connected`, while `ddp.loggedIn` survives it.
+// Neither survives a silent background death, so callers must bound their wait.
+export function isLoginReady(): boolean {
+	const state = store.getState();
+	return state.login.isAuthenticated && state.meteor.connected;
+}
+
+function onAbort(signal: AbortSignal | undefined, callback: () => void): void {
```
diegolmello: **Duplicated code.** This `onAbort` helper is byte-identical to `acceptNativeCall.ts:25-40`, comment included. `acceptNativeCall.ts` already imports from this module, so there's no cycle blocking extraction — export it here and drop the copy.

diegolmello (commented):
Two-axis review (standards + spec) of this branch vs `purring-seahorse`. Inline comments below.

- **Stale doc.** `app/lib/services/voip/docs/ARCHITECTURE.md:45` still says `tryAnswerIfNativeAcceptedNotification` triggers `answerCall(callId)`; the stream path now routes through the readiness gate.

Worst issues: two live bugs in the SDK patch (`reopenNow` TDZ, `probe` always-false), and the "8s cap" in the PR body is not actually bounded (~20s worst case, past the 10s `timeout-remote-sdp` window this PR exists to beat).

app/lib/services/connect.ts
```diff
@@ -463,6 +443,22 @@ function checkAndReopen() {
 	return sdk.current.checkAndReopen();
 }
 
+/** Classify DDP socket freshness from last pong. Returns 'fresh' when the SDK
+ *  lacks the new probe/reopen hooks so callers fall back to checkAndReopen. */
+export function getSocketStaleness(ddp: any): 'stale' | 'gray' | 'fresh' {
```
diegolmello: **`any`.** Repo style bans `any` unless strictly necessary; `VoipReadyDdp` in `acceptNativeCall.ts:14` already types this exact duck-shape. Reuse or narrow it.

Two more concerns on this function:

- **Middle Man / naming.** It only renames `classifySocketHealth`'s three values (`reopen→stale`, `probe→gray`, `healthy→fresh`). Two vocabularies for one concept, and `'gray'` is undefined at the call sites. Prefer one set of names.
- **Unknown reported as healthy.** When the SDK lacks the hooks it returns `'fresh'`. That's the safe fallback here, but `acceptNativeCall.ts:85-86` treats the *same* condition as a hard failure and terminates the call. Pick one behaviour.

app/lib/services/voip/acceptNativeCall.ts
```diff
@@ -0,0 +1,134 @@
...
+	const cleanup = () => {
+		if (activeGates.get(callId) === controller) {
+			activeGates.delete(callId);
+		}
+	};
+
+	try {
+		const ddp = sdk.current?.ddp as VoipReadyDdp | undefined;
+		if (!ddp || typeof ddp.reopenNow !== 'function' || typeof ddp.probe !== 'function' || ddp.lastPing == null) {
+			return handleFailure(callId, mediaSession);
```
diegolmello, line 67: **Inconsistent fallback.** An SDK without `reopenNow`/`probe`/`lastPing` goes to `handleFailure`, i.e. every native accept terminates the call. `getSocketStaleness` (`connect.ts:444-448`) treats the identical condition as `'fresh'` and falls back to `checkAndReopen`. A graceful fallback is the consistent choice — a stale-socket optimisation shouldn't be able to kill accepts on an unpatched SDK.

app/lib/services/voip/acceptNativeCall.ts
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		const action = classifySocketHealth(ddp);
+		if (action === 'reopen') {
+			await ddp.reopenNow();
+		} else if (action === 'probe') {
+			const alive = await ddp.probe(2000);
```
diegolmello: **Duplicated ladder / shotgun surgery.** This decide → `probe(2000)` → `reopenNow()` sequence, including the `2000` literal, is written again in `app/sagas/state.js:34-52`. Changing the policy now means editing a saga and a service in lockstep. Extract one helper.

app/lib/services/voip/acceptNativeCall.ts
```diff
@@ -0,0 +1,134 @@
...
+			waitForLoginReady(8000, controller.signal),
+			waitForMediaSignalSubs(ddp, 8000, controller.signal)
+		]);
+
+		if (controller.signal.aborted) {
+			return;
+		}
+
+		if (!loginReady || !mediaSubsReady || !mediaSession.isInitialized()) {
+			return handleFailure(callId, mediaSession);
```
diegolmello, line 92: **No real 8s cap.** PR body says the gate has an "8s cap". Worst case here is `reopenNow()` (own 10s internal timeout) + `probe(2000)` + `Promise.all([waitForLoginReady(8000), waitForMediaSignalSubs(8000)])` ≈ 20s — well past the 10s `timeout-remote-sdp` window this change exists to beat. Either bound the whole gate with one deadline or correct the description.

app/lib/services/voip/acceptNativeCall.ts
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		await mediaSession.applyRestStateSignals();
+
+		if (controller.signal.aborted) {
+			return;
+		}
+
+		const { call } = useCallStore.getState();
+		if (call?.callId !== callId) {
```
diegolmello, line 102: Guard reads inverted: this answers only when `call?.callId !== callId`, i.e. it *skips* `answerCall` when the store already holds this very call. `tryAnswerIfNativeAcceptedNotification` guards on `call == null` instead. Worth confirming intent and aligning the two.

app/lib/services/voip/acceptNativeCall.ts
```diff
@@ -0,0 +1,134 @@
...
+interface VoipReadyDdp {
+	reopenNow(): Promise<void>;
+	probe(timeoutMs: number): Promise<boolean>;
+	lastPing: number;
+	pingInterval?: number;
+	config?: { ping?: number };
+	waitForNotifyUserMediaSubs(timeoutMs: number): Promise<boolean>;
+}
+
+const activeGates = new Map<string, AbortController>();
```
diegolmello, line 21: Minor: a `Map` of gates keyed by `callId` plus a per-gate `AbortController`, when only one native call can be active at a time. Defensible for the concurrency race being fixed, but a single `{ callId, controller }` slot would express the invariant. Judgement call.

app/sagas/state.js
```diff
@@ -4,9 +4,12 @@ import log from '../lib/methods/helpers/log';
 import { localAuthenticate, saveLastLocalAuthenticationSession } from '../lib/methods/helpers/localAuthentication';
 import { APP_STATE } from '../actions/actionsTypes';
 import { RootEnum } from '../definitions';
-import { checkAndReopen } from '../lib/services/connect';
+import { checkAndReopen, getSocketStaleness } from '../lib/services/connect';
 import { setUserPresenceOnline, setUserPresenceAway } from '../lib/services/restApi';
 import { checkPendingNotification } from '../lib/notifications';
+import sdk from '../lib/services/sdk';
+
```
diegolmello: Module-level mutable state in a saga file. `isProbingSocket` survives logout and server switch, sits outside saga scope while the rest of the file is generator-driven, and can't be reset in tests without re-importing the module. Consider a saga-local guard or moving the re-entry guard into the probe helper itself.

app/sagas/state.js
```diff
@@ -26,7 +29,30 @@ const appHasComeBackToForeground = function* appHasComeBackToForeground() {
 	try {
 		const server = yield select(state => state.server.server);
 		yield localAuthenticate(server);
-		checkAndReopen();
+
+		const ddp = sdk.current?.ddp;
```
diegolmello: The saga reaches two levels into the SDK (`sdk.current?.ddp`, then `reopenNow`/`probe`) and orchestrates socket internals directly — with non-null safety resting on an implicit contract with `getSocketStaleness`. That orchestration belongs behind `connect.ts`; the saga should call one function.

app/sagas/login.js
```diff
@@ -260,7 +261,9 @@ const checkVoipPermission = async () => {
 		const canUseVoip = isVoipModuleAvailable() && (hasPermissions[0] || hasPermissions[1]);
 
 		if (!canUseVoip) {
-			mediaSessionInstance.reset();
+			if (!isInActiveVoipCall()) {
```
diegolmello, line 264: **Scope creep.** No bullet in the PR body covers VoIP-permission-change behaviour. It looks like a real fix (don't wipe an active call when permissions flip), but it's an independent change riding along — worth its own line in the description at minimum.

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -96,15 +111,24 @@ class MediaSessionInstance {
 				})
 		);
 		mediaSessionStore.setSendSignalFn((signal: ClientMediaSignal) => {
```
diegolmello, line 113: **Scope creep.** Swallowing `methodCall` rejections into `log` is an unrelated unhandled-rejection fix. Reasonable, but not covered by any bullet in the PR body.

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -96,15 +111,24 @@ class MediaSessionInstance {
...
 		if (!this.instance) {
 			throw new Error('Failed to create media session instance');
 		}
 
-		await this.applyRestStateSignals();
+		const { nativeAcceptedCallId } = useCallStore.getState();
+		if (nativeAcceptedCallId) {
+			this.acceptNativeCallWithReadiness(nativeAcceptedCallId).catch(error => {
+				log(error);
+			});
```
diegolmello, line 128: **Behaviour change beyond "init replay".** In the gated branch `init()` skips `applyRestStateSignals()` entirely, and the gate is fire-and-forget — so `init()` no longer awaits state-signal replay at all when `nativeAcceptedCallId` is set. Callers that relied on `init()` resolving with replay complete now race. Intentional?

app/lib/services/voip/MediaSessionInstance.ts
```diff
@@ -48,7 +49,7 @@ class MediaSessionInstance {
 	private storeTimeoutUnsubscribe: (() => void) | null = null;
 	private storeIceServersUnsubscribe: (() => void) | null = null;
 
-	private tryAnswerIfNativeAcceptedNotification(signal: ServerMediaSignal): void {
+	private tryAnswerIfNativeAcceptedNotification(signal: ServerMediaSignal, useGate = false): void {
```
diegolmello, line 52: Flag argument: `useGate = false` drives an if/else, and both call sites pass a literal (`false` at :84, `true` at :152). Either split into two methods or inject the accept function.

OtavioStasiak (approved):
LGTM

PR #7530, opened 2026-07-29
fix: video attachments not reproducing on iOS

OtavioStasiak:
## Proposed changes

Media attachments could keep pointing at the remote URL even after the file was already cached on disk, so opening them streamed from the server instead of playing the local file. On iOS that surfaced as a video thumbnail that spins
forever with no error.

  **Root cause.** `useFile` only merged local overrides for *non-persisted* messages, relying on `persistMessage` to write the downloaded uri back to the database for everything else. That write silently no-ops whenever no row matches the message id, which is the case for forwarded messages and for the Files/Mentions/Starred/Pinned lists —`MessagesView` builds attachments from the uploads REST payload, which has `_id` but no `id` and no message id at all, so `useMessageId()` is `undefined`. Downloading from the Files tab therefore left the room's message row holding the remote URL while the file sat fully downloaded on disk.

  **Why only iOS, and only some videos.** The broken state is identical on both platforms; the difference is what each player does when asked to *stream* the file. An MP4/MOV can't start until the player has the `moov` atom (sample
tables, byte offsets, durations). macOS screen recordings and `az_recorder` write `moov` at the *end* of the file — they can't know the sample table until recording stops — whereas the iPhone camera writes faststart (`moov` first).
Reaching a trailing `moov` over HTTP requires working Range requests. AVFoundation assumes Range works and simply waits: the item stays in `AVPlayerItemStatusUnknown`, so expo-av never fires `onLoad` (needs `ReadyToPlay`) or `onError` (needs `AVPlayerItemStatusFailed`), and `loading` stays `true`. ExoPlayer falls back to reading forward from byte 0 until it finds `moov` — wasteful, but it terminates, so Android just loaded slowly. The absence of an error alert was the tell: a 404/403 would have popped the view with `Error_play_video`.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1307

## How to test or reproduce

- Open the app;
- Send a screen Record from macOS;
- Try to reproduce it on RoomView;
 
## Screenshots
| Before | After |
| ----- | ----- |
| before_video_ios | after_video_ios |
## Types of changes

## Checklist

## Bug Fixes
- Fixed attachment URLs with spaces or special characters so images and videos load correctly.
- Prevented double-encoding of attachment links.
- Improved handling of protected and remote attachment URLs.
- Added safer behavior when media updates have nothing to save.

## Improvements
- Attachment overrides now merge consistently, even when source details change.
- Simplified media auto-download initialization.

## Tests
- Added coverage for URL encoding and attachment override behavior.


app/lib/methods/helpers/formatAttachmentUrl.ts
```diff
@@ -10,6 +10,15 @@ function setParamInUrl({ url, token, userId }: { url: string; token: string; use
 	return urlObj.toString();
 }
 
+// Idempotent encode: the server already sends encoded paths, and a plain encodeURI would turn `%20` into `%2520`.
+export const encodeAttachmentUrl = (url: string): string => {
```
Rohit3523, line 13: can we have a test for it, small exported function with three cases worth covering: plain URL with spaces (/Screen Recording.mov → /Screen%20Recording.mov), already-encoded URL (/Screen%20Recording.mov stays same — no double-encoding), and malformed escape (/%ZZ returns raw — no crash).

app/lib/methods/helpers/formatAttachmentUrl.ts
```diff
@@ -28,17 +37,17 @@ export const formatAttachmentUrl = (
 		}
 
 		if (attachmentUrl.includes('rc_token')) {
-			return encodeURI(attachmentUrl);
+			return encodeAttachmentUrl(attachmentUrl);
 		}
 
-		if (protectFiles) return setParamInUrl({ url: attachmentUrl, token, userId });
-		return attachmentUrl;
+		if (protectFiles) return encodeAttachmentUrl(setParamInUrl({ url: attachmentUrl, token, userId }));
```
Rohit3523, line 48: `setParamInUrl` already parses through WHATWG `URL` and re-serializes, which normalizes the path on its own — it percent-encodes what needs it and leaves existing escapes alone. So wrapping its output in `encodeAttachmentUrl` can only do one of two things: nothing, or damage. It can't improve the result.

The damaging case is a path that already contains an escaped reserved character. `decodeURI` deliberately does *not* decode the reserved set (`# & + , / : ; = ? @ $`), so the escape survives as literal text and `encodeURI` then escapes its `%`:

```
filename: a video #2.mov

server sends:           /file-upload/abc/a%20video%20%232.mov
setParamInUrl:          /file-upload/abc/a%20video%20%232.mov?rc_token=…     ✅
+ encodeAttachmentUrl:  /file-upload/abc/a%20video%20%25232.mov?rc_token=…   ❌
```

`%23` became `%2523`, so the server is asked for a file literally named `a video %232.mov` → 404. On `develop` this path returned `setParamInUrl(...)` untouched, so it'd be a new failure for filenames containing `#`, `&`, `+` or `,`. Line 51 is the same shape.

How often it actually bites depends on whether the server pre-encodes those characters in `title_link` — I haven't verified that, and if it sends them raw the wrapper is a harmless no-op. But since it has no upside at these two call sites either way, simplest to drop it:

```suggestion
		if (protectFiles) return setParamInUrl({ url: attachmentUrl, token, userId });
```

Optional follow-on: the `URL` round-trip is exactly what makes these two lines safe, so it may be worth using it for the helper itself —

```ts
export const encodeAttachmentUrl = (url: string): string => {
	try {
		return new URL(url).toString();
	} catch {
		return url;
	}
};
```

That still covers what this PR is fixing (`/Screen Recording.mov` → `/Screen%20Recording.mov`), still leaves `%20` alone on a second pass, and still returns `%ZZ.mov` unchanged — so all three tests you added keep passing, and the reserved-char case above stops being a hazard on the remaining call sites too.

Rohit3523:
### Code review

The `encodeAttachmentUrl` helper is the right primitive — `new URL(x).toString()` is idempotent where `encodeURI` isn't, so the reserved-character hazard from the earlier round is gone. Two issues with how it's wired up.

Found 2 issues:

**1. Inline message images still double-encode, and this PR breaks the one config that worked**

[`AttachmentView`](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/views/AttachmentView.tsx#L61-L64) drops its `encodeURI`, but the in-message image renderer still has one, fed directly from `formatAttachmentUrl`:

https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/containers/message/components/Attachments/Image/Image.tsx#L71-L73

The `uri` comes from `useMediaAutoDownload().url`, unmodified:

https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/containers/message/components/Attachments/Image/Container.tsx#L25-L27

For `Screen Recording.png` with `title_link = /file-upload/abc/Screen Recording.png`, which falls through to [line 51](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/lib/methods/helpers/formatAttachmentUrl.ts#L49-L51):

| `FileUpload_ProtectFiles` | | `formatAttachmentUrl` | after `encodeURI` at Image.tsx:72 |
|---|---|---|---|
| `false` | develop | `…/Screen Recording.png` (raw) | `Screen%20Recording.png` ✅ |
| `false` | **this PR** | `…/Screen%20Recording.png` | `Screen%2520Recording.png` ❌ |
| `true` | develop | `…/Screen%20Recording.png` | `Screen%2520Recording.png` ❌ |
| `true` | this PR | `…/Screen%20Recording.png` | `Screen%2520Recording.png` ❌ |

With `protectFiles` off, that row goes from working to 404. With it on it was already broken and stays broken — so as it stands, tapping through to the fullscreen viewer works but the inline thumbnail doesn't. Videos are unaffected; `Image.tsx:72` is the only remaining `encodeURI` in an attachment render path.

**2. `_originalUrl` bypasses the new encoding, so `encodeURI` can't simply be deleted**

https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/lib/methods/helpers/formatAttachmentUrl.ts#L33-L36

`useMediaAutoDownload` is the only caller that passes `_originalUrl`, so externally-hosted attachments return raw — and today `Image.tsx:72` is the only thing encoding them. Wrapping that return makes the "always encoded" contract asserted in [the new AttachmentView comment](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/views/AttachmentView.tsx#L61-L63) actually hold, and then #1 is a one-line deletion:

```ts
if (_originalUrl && !_originalUrl.startsWith(server)) {
    return encodeAttachmentUrl(_originalUrl);
}
```

The `file://` and base64 early returns can stay unencoded — `sanitizeFileName` rewrites `%20` to `_20`, so cached paths never contain spaces.

**Minor:** the malformed-escape test asserts correct behavior, but not for the stated reason — WHATWG `URL` doesn't throw on `%ZZ`, it passes it through, so the `catch` is never reached. The branch that does throw is a non-absolute URL (`new URL('/file-upload/1/x.png')`), reachable at [line 51](https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/c9abe98d1c3d1239ff9712d29095e247e044b175/app/lib/methods/helpers/formatAttachmentUrl.ts#L49-L51) when `baseUrl` is empty. Worth one more case so the fallback is actually covered.

app/containers/message/hooks/useFile.tsx
```diff
@@ -1,30 +1,16 @@
...
-				if (!message) {
-					setIsMessagePersisted(false);
-				}
-			}
-		};
-		checkMessage();
-	}, [messageId]);
+// Merges local overrides (the downloaded uri) over the `file` prop, unconditionally: persistMessage no-ops without a message row.
+export const useFile = (file: IAttachment) => {
+	'use memo';
```
Rohit3523: We can remove this 'use memo' as react compiler is memorising components without it now

Rohit3523 (approved):

app/containers/message/components/Attachments/Image/Image.tsx
```diff
@@ -69,7 +69,7 @@ export const MessageImage = ({ uri, status, encrypted = false, imagePreview, ima
 		<>
 			{showImage ? (
 				<View style={[containerStyle, borderStyle]}>
-					<Image autoplay={autoplayGifs} style={imageStyle} source={{ uri: encodeURI(uri) }} contentFit='cover' />
+					<Image autoplay={autoplayGifs} style={imageStyle} source={{ uri }} contentFit='cover' />
```
diegolmello, line 72: Isn't this a breaking change? It could impact other use cases. How does it handle already downloaded files?

app/containers/message/hooks/useFile.tsx
```diff
@@ -1,30 +1,14 @@
...
-				const message = await getMessageById(messageId);
-				if (!message) {
-					setIsMessagePersisted(false);
-				}
-			}
-		};
-		checkMessage();
-	}, [messageId]);
+// Merges local overrides (the downloaded uri) over the `file` prop, unconditionally: persistMessage no-ops without a message row.
+export const useFile = (file: IAttachment) => {
```
diegolmello, line 6: Looks like two different hooks, no?

app/containers/message/hooks/useMediaAutoDownload.tsx
```diff
@@ -80,7 +80,7 @@ export const useMediaAutoDownload = ({
 	const baseUrl = useBaseUrl();
 	const user = useMessageUser();
 	const [status, dispatchDownloadEvent] = useReducer(downloadStatusReducer, 'to-download');
-	const [currentFile, setCurrentFile] = useFile(file, id ?? '');
+	const [currentFile, setCurrentFile] = useFile(file);
```
diegolmello, line 83: Possibly a breaking change: if the same file name exists in two messages or two channels, it fails.

app/lib/methods/helpers/formatAttachmentUrl.ts
```diff
@@ -9,6 +9,14 @@ function setParamInUrl({ url, token, userId }: { url: string; token: string; use
 	urlObj.searchParams.set('rc_uid', userId);
 	return urlObj.toString();
 }
+// Idempotent encode: the server already sends encoded paths, and a plain encodeURI would turn `%20` into `%2520`.
```
diegolmello, line 12: What version of the server does it?
Remove this comment.

app/lib/methods/handleMediaDownload.ts
```diff
@@ -249,11 +261,14 @@ const persistMessage = async (messageId: string, uri: string, encryption: boolea
 			})
 		);
 	}
-	if (batch.length) {
-		await db.write(async () => {
-			await db.batch(batch);
-		});
+	if (!batch.length) {
+		// Expected when the attachment has no message row (forwarded, attachment lists); useFile keeps the local uri.
```
diegolmello, line 265: Remove

app/views/AttachmentView.tsx
```diff
@@ -59,9 +59,9 @@ const RenderContent = ({
 	}, [navigation]);
 
 	if (attachment.image_url) {
-		const url = formatAttachmentUrl(attachment.title_link || attachment.image_url, user.id, user.token, baseUrl);
-		const uri = encodeURI(url);
-		const isAnimated = attachment.image_type === 'image/gif' || /\.gif(\?|$)/i.test(url);
+		// formatAttachmentUrl already encodes the url; encoding again would turn `%20` into `%2520`.
```
diegolmello, line 62: This comment is repeated again

diegolmello (changes requested):

app/containers/message/hooks/useMediaAutoDownload.tsx
OtavioStasiak: `messageId` was never part of the file cache identity, so removing it doesn't introduce filename collisions. Files are cached under `${server}/${fileId}/${filename}`, meaning identical filenames from different uploads are stored separately. The `messageId` was only used to determine whether local overrides should be applied.

Where the real messageId argument goes, it's used only for DB persistence, never for paths:
  - app/containers/message/hooks/useMediaAutoDownload.tsx:140 — messageId: id ?? '' passed to downloadMediaFile
  - app/lib/methods/handleMediaDownload.ts — persistMessage, ~lines 245–271: getMessageById / getThreadMessageById to write the local uri onto the message row

app/containers/message/components/Attachments/Image/Image.tsx
OtavioStasiak: Not a breaking change. The encoding was moved upstream to formatAttachmentUrl, so MessageImage now always receives an already-normalized URL. Keeping encodeURI in the leaf was actually causing double-encoding (%20 → %2520) and 404s.

app/containers/message/hooks/useFile.tsx
OtavioStasiak: Yes, but we did not discarded the old logic... The old hook checked whether the message existed in the DB to decide whether to use the downloaded file’s local uri. That caused two issues: for attachments without a persisted DB row, the local uri could be discarded, and the async DB lookup could race with the download callback.

The new logic separates those responsibilities: persistMessage still updates the DB when a row exists, while useFile always applies the local uri. If persistence succeeds, the override is simply redundant; if it doesn’t, the override ensures the cached file is still used. This also removes an unnecessary DB read per attachment.

diegolmello (commented):
Two blockers, both from the same root cause: the URL is encoded too early.

`formatAttachmentUrl`'s output is used for four different jobs, and only one of them wants an encoded URL:

1. the download address — `handleMediaDownload.ts:296` — **needs encoding**
2. the file name on disk — `:291` — needs to stay stable
3. the key that finds the Attachment row — `:309` — needs to stay stable
4. the emitter channel name — `:311` — needs to stay stable

Encoding inside `formatAttachmentUrl` sits upstream of all four, so it moves all four. Job 3 broke and this PR patches it inside `matchDownloadUrl`. Job 2 broke the same way and is still unpatched — that's blocker 2.

Details inline. There are also three smaller things worth doing:

- Delete the `console.log` rather than switching it to `log()` (inline).
- The comment at `useFile.tsx:5` will rot (inline). The ones at `formatAttachmentUrl.ts:34-35` and `handleMediaDownload.ts:212-213` are load-bearing — keep those.
- The "which server version does this?" thread was closed by deleting the comment, not by answering it. Still open as far as I can tell.

One note on `useMediaAutoDownload.tsx:86`: it reads `file.title_link` before the merged override, so the prop wins and the override only reaches the player via `showAttachment(currentFile)`. That still matches the reported bug, so it isn't a blocker — but the test comment saying overrides always win isn't true for `url`.

The `useFile` change itself is good. Keep it as is.

app/lib/methods/helpers/formatAttachmentUrl.ts
```diff
@@ -19,26 +27,32 @@ export const formatAttachmentUrl = (
...
 
 		if (attachmentUrl.includes('rc_token')) {
-			return encodeURI(attachmentUrl);
+			return encodeAttachmentUrl(attachmentUrl);
 		}
 
-		if (protectFiles) return setParamInUrl({ url: attachmentUrl, token, userId });
-		return attachmentUrl;
+		if (protectFiles) return encodeAttachmentUrl(setParamInUrl({ url: attachmentUrl, token, userId }));
+		return encodeAttachmentUrl(attachmentUrl);
```
diegolmello: **Blocker 1 — the encoding is in the wrong place.**

In short: one string does four jobs, three of them need it to stay unchanged, and this encodes it for all four.

The tell is `matchDownloadUrl` below. It broke because of this change, and the repair decodes both sides before comparing. When you have to decode at the consumer, the encoding happened too early. `getFilePath` broke in exactly the same way and nobody caught it (see the comment there).

I think the real fix in this PR is **`new URL` instead of `encodeURI`**, not **moving the encode to the source**. The old leaf `encodeURI` turned `setParamInUrl`'s `%20` into `%2520` — that's a genuine bug and worth fixing. But `new URL().toString()` is idempotent, so encoding at the leaf is safe now. Moving it upstream was never needed, and it's what produced all the fallout.

**To see the coupling:**

1. Restore `matchDownloadUrl` (`handleMediaDownload.ts:214-218`) from `develop`.
2. Run the app. Send an Attachment with a space in the name.
3. Tap it, let it download.
4. Read the Message row — `title_link` still holds the remote URL.

The same test passes on `develop`. The early encode is the only difference.

**Suggested fix — encode at the sink:**

1. Revert this file to `develop`, keeping the new `encodeAttachmentUrl` export.
2. Call `encodeAttachmentUrl` at the three places that hand a URI to a native player: `Image/Image.tsx:72`, `AttachmentView.tsx:62`, `AttachmentView.tsx:76`.
3. Revert `matchDownloadUrl` and delete `decodeUrl`.

That makes blocker 2 disappear and removes six of the seven `encodeAttachmentUrl` calls.

**If you'd rather keep it central**, then all three of these:

- **One exit point, not seven wrapped returns.** Seven is how blocker 2 stayed hidden:
  ```ts
  if (attachmentUrl && isImageBase64(attachmentUrl)) return attachmentUrl;
  return encodeAttachmentUrl(resolveAttachmentUrl(attachmentUrl, userId, token, server, _originalUrl));
  ```
  Move the current body into `resolveAttachmentUrl` unchanged. Same behaviour, and the rule is readable in one line instead of provable by audit.
- Fix `getFilePath` (see the other comment).
- Add a test for `matchDownloadUrl` — the rewrite loosens the match and nothing covers it.

app/lib/methods/handleMediaDownload.ts
```diff
@@ -200,10 +200,22 @@ export async function cancelDownload(messageUrl: string): Promise<void> {
...
+	}
+};
+
+// `downloadUrl` has been through formatAttachmentUrl (so spaces are `%20`) while the attachment rows carry the
+// server value verbatim. Compare decoded so both forms line up regardless of which one the server sent.
+export const matchDownloadUrl = (att: IAttachment, downloadUrl: string) => {
+	const target = decodeUrl(downloadUrl);
+	const matches = (attachmentUrl?: string) => !!attachmentUrl && target.includes(decodeUrl(attachmentUrl));
+	return matches(att.image_url) || matches(att.audio_url) || matches(att.video_url);
+};
```
diegolmello: **Blocker 2 — every cached Attachment with a space in its name gets orphaned.**

(Commenting here because `getFilePath:126` isn't in the diff — that's the actual site.)

In short: the file on disk is named after the URL. This PR changes the URL. So the app looks under a new name, doesn't find it, and downloads again. The old file stays on disk forever.

`getFilePath:126` derives the name from the URL, and `sanitizeLikeString` (`app/lib/database/utils.ts:7`) replaces every non-alphanumeric character with `_` — including `%`.

For an Attachment named `Screen Recording.mov`:

| | URL segment | On disk |
|---|---|---|
| `develop` | `Screen Recording.mov` | `Screen_Recording.mov` |
| this branch | `Screen%20Recording.mov` | `Screen_20Recording.mov` |

The folder is fine — `getFolderPath:136` uses the file id segment, which has nothing to encode. Only the name drifts.

So every existing user re-downloads those Attachments once, and the old files leak. It's self-consistent from then on, so not a permanent defect — but it ships to everyone and the PR body doesn't mention it.

**Steps to reproduce:**

1. On `develop`, run the app. Send an Attachment named `Screen Recording.mov`.
2. Tap it, let it download.
3. In the app document directory, confirm `Screen_Recording.mov`.
4. Check out this branch and run again — **do not clear app data**.
5. Tap the same Attachment. It downloads a second time.
6. Both `Screen_Recording.mov` and `Screen_20Recording.mov` are now on disk. The first is never read again.

**Fix.** Blocker 1's fix removes this entirely. If the encode stays central, then both of:

1. Decode in `getFilePath`:
   ```ts
   const urlWithoutQueryString = decodeUrl(urlToCache.split('?')[0]);
   ```
   That prevents the drift. It won't recover files already orphaned by a build with this change.
2. Say in the PR body that these Attachments will download once more.

Separately on `matchDownloadUrl` itself: the rewrite loosens matching (`decodeUrl` turns `%2F` into `/` on both sides before a substring test) and there's no test for it. Low risk since server upload paths won't collide, but `decodeUrl` also swallows exceptions, which the name doesn't suggest.

app/lib/methods/handleMediaDownload.ts
```diff
@@ -251,9 +263,11 @@ export const persistMessage = async (messageId: string, uri: string, encryption:
 				})
 			);
 		}
-		if (batch.length) {
-			await db.batch(batch);
+		if (!batch.length) {
+			console.log('[handleMediaDownload] no message row for attachment, cached uri not persisted');
```
diegolmello: Delete this rather than keeping it — nothing reads it, and it fires on a path the PR itself describes as expected (forwarded Messages, and the Files/Mentions/Starred/Pinned lists).

Worth saying explicitly: don't swap it for `log()`. That goes to Bugsnag, which would turn an expected path into reporting noise.

app/containers/message/hooks/useFile.tsx
```diff
@@ -1,30 +1,14 @@
...
-			if (!threadMessage) {
-				const message = await getMessageById(messageId);
-				if (!message) {
-					setIsMessagePersisted(false);
-				}
-			}
-		};
-		checkMessage();
-	}, [messageId]);
+// Merges local overrides (the downloaded uri) over the `file` prop, unconditionally: persistMessage no-ops without a message row.
```
diegolmello: This one will rot — it asserts how `persistMessage` behaves in a different module, so it goes stale the moment that changes (and it just did change, in the develop merge).

The other comments in this PR are fine by me: `formatAttachmentUrl.ts:34-35` and `handleMediaDownload.ts:212-213` both explain non-obvious encoding rules that the code can't say on its own. Keep those.

diegolmello (approved):

PR #7545, opened 2026-08-04
chore: switch React Compiler to infer mode

diegolmello:
## Proposed changes

React Compiler ran in `annotation` mode: only the 172 files hand-marked with `'use memo'` were compiled — 201 functions of the ~1,300 the compiler can cover. Every new component was uncompiled unless the author remembered the directive, so most of the app missed the compiler's memoization and the directive set drifted out of date.

This PR switches `babel-plugin-react-compiler` to `compilationMode: 'infer'` (the compiler's own default), so every component and hook is compiled automatically, and deletes the now-redundant directives. The compiler skips what it cannot prove safe, so unmodified code keeps working uncompiled.

Measured evidence:

- **Compile inventory:** 201 → 1,097 compiled functions (+896), zero transform errors. The 126 bail-outs are safe by design (top reasons: "This value cannot be modified", "Cannot access refs during render") — those components render uncompiled.
- **Version:** plugin bumped to 1.0.0 stable in the same PR. Deltas vs the previous rc.3 measured as trivial (−3 compiled functions, −2 bails, identical gates).
- **Jest:** the plugin moves into a babel `overrides` entry with `exclude` matching `jest.setup.js` and test files, so `jest.mock()` factories keep working while app source stays compiled under jest. (The `exclude` form is required: babel merges an override's `plugins` with the root list instead of replacing it.)
- **Hook tests:** the three suites that fail only under infer (`useShortnameToUnicode`, `usePreviewFormatText`, `useNewMediaCall`) now call hooks through `renderHook`, the pattern five other hook suites already use.
- **Directives:** `'use memo'` stripped from all 172 files, deletions-only. The directives are inert under infer; keeping them implies annotation mode still matters.

Escape hatch: any component that misbehaves at runtime can opt out with `'use no memo'`, recorded here. **None were needed.**

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1481

## How to test or reproduce

Gates: `pnpm format-lint`, `tsc`, and the full jest suite against compiled app code — 227/227 suites (2,066 tests, 411 snapshots).

Device smoke checks (debug build of this branch, Android API 34 emulator, `open.rocket.chat`):

- RoomView — open a room, scroll history (older pages load), send a message (receipt flips to read), receive a REST-posted message live via stream (~2s), add a 👍 reaction from the long-press actions sheet, open a thread and send a thread reply
- RoomsListView — realtime reorder (the room jumps to top on stream receive), unread badges, search filtering

All flows pass. One anomaly observed, not attributable to this PR: the `rocketchat://auth` deeplink stalled before dispatching login (the deepLinking saga is untouched here and is not a compiled component); flagged for a separate develop-side investigation.

## Types of changes

## Checklist

## Further comments

No re-render A/B was run: the risk bar for this change is behavior parity, not perf measurement. If a NO-GO-grade surprise appears later, the revert is one config value; the directive strip is cosmetic and can stay.

## Summary by CodeRabbit

* **Refactor**
  * Updated rendering optimization across navigation, messaging, rooms, calls, and settings without changing visible functionality or interactions.
  * Removed obsolete optimization directives throughout the app.

* **Bug Fixes**
  * Improved rendering-tool configuration compatibility by applying it selectively outside test files.

* **Tests**
  * Modernized hook tests while preserving existing coverage and assertions.

* **Chores**
  * Updated rendering optimization tooling to its stable release.


OtavioStasiak (approved):
LGTM

PR #7546, opened 2026-08-04
fix(db): move sendMessage reads and prepares inside the writer lock

OtavioStasiak:
## Proposed changes

`sendMessage` and `changeMessageStatus` read records and called `prepareUpdate`/`prepareCreate` outside
`db.write`, committing the batch in a separate write later. A concurrent writer touching the same cached record during
that window left the prepared records stale, so the commit threw `Cannot update a record with pending changes` (reaching Bugsnag) and the pending change was lost — a  sent message stuck in TEMP, a decrypted message still encrypted.

  Both functions now do their reads, prepares and batch inside a single `db.write` callback. Encryption and the network call stay outside the lock, so nothing is held longer than needed. Same approach already used by `RoomSubscription.updateMessage`.

  Adds one regression test per function: a concurrent writer races the same record  and the batch must commit without a "pending changes" throw. Both fail on the current code and pass with the fix. Signatures and callers unchanged.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1463

## How to test or reproduce

  - `TZ=UTC pnpm test app/lib/methods/sendMessage.test.ts` — both tests pass; revert
  `sendMessage.ts` and they fail with `Cannot update a record with pending changes`
  - Send a message in a busy room (lots of incoming stream activity) — no message
  stuck in the temp/sending state
  - Send a thread reply, both in a new thread and an existing one
  - Send in an E2EE room — message decrypts instead of staying encrypted
  - Resend a failed message from the message actions
  - Send from the share extension (text-only share into a room)
  
## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message-sending reliability when multiple updates occur simultaneously.
  * Prevented pending-change errors during concurrent message or subscription updates.
  * Ensured message, thread, status, and draft changes are committed together consistently.
  * Preserved expected message content and delivery statuses during concurrent activity.

* **Tests**
  * Added comprehensive coverage for concurrent sending, status updates, encryption handling, and database commits.


app/lib/methods/sendMessage.ts
```diff
@@ -110,120 +112,124 @@ export async function sendMessage(
...
-			try {
-				// Find thread message header in Messages collection
-				tMessageRecord = await msgCollection.find(tmid);
-				batch.push(
-					tMessageRecord.prepareUpdate(m => {
-						m.tlm = messageDate;
-						if (m.tcount) {
-							m.tcount += 1;
+		// Serialize reads, prepares and the batch under the writer lock so a concurrent write on the
+		// same records can't leave them prepared-but-stale by the time we commit.
```
diegolmello: I think this comment is not worth it

app/lib/methods/sendMessage.ts
```diff
@@ -14,33 +14,36 @@ const changeMessageStatus = async (id: string, status: number, tmid?: string, me
...
-				if (message) {
-					tm.mentions = message.mentions;
-					tm.channels = message.channels;
-				}
-			})
-		);
-	}
 
+	// Serialize reads, prepares and the batch under the writer lock so a concurrent write on the
+	// same records can't leave them prepared-but-stale by the time we commit.
```
diegolmello: I think this comment is not worth it

diegolmello (approved):

PR #7548, opened 2026-08-04
fix(db): move decryptPendingMessages prepares inside the writer lock

OtavioStasiak:
## Proposed changes

decryptPendingMessages` decrypted each pending e2e message and called `prepareUpdate` outside `db.write`, committing the batch in a separate write later. A concurrent writer touching the same cached record during that window — a new message arriving in the room, for instance — left the prepared records stale, so the commit threw `Cannot update a record with pending changes` (reaching Bugsnag) and the message stayed encrypted.

  Decryption now happens first, outside the lock, and the `prepareUpdate` calls plus the `db.batch` run inside a single `db.write` callback. Records whose `prepareUpdate` throws are now filtered out instead of being batched as `null`.

  Adds one regression test: a concurrent writer races the same record and the batch must commit without a "pending changes" throw. It fails on the current code and passes with the fix. Signature and both callers (`Encryption.initialize`, `createOrUpdateSubscription`) unchanged.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1464

## How to test or reproduce

- `TZ=UTC pnpm test app/lib/encryption/encryption.test.ts` — all pass; revert `encryption.ts` and the new test fails with
  `Cannot update a record with pending changes`
  - Open an E2EE room with pending encrypted messages (e.g. kill the app before decryption finishes, or enter the E2EE
  password after opening the room) — they all decrypt, none stay encrypted
  - Do the above while new messages arrive in that room, so decryption overlaps with incoming writes
  - Enter an E2EE room from the rooms list — `createOrUpdateSubscription` triggers a per-room decrypt
  - Cold start with several E2EE rooms having pending messages — `Encryption.initialize` decrypts them all
  
## Types of changes

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when decrypting pending messages, including situations where multiple updates affect the same message.
  - Prevented failed or incomplete decryption updates from being committed, reducing database errors and ensuring only valid changes are saved.
  - Ensured successful message updates continue to be processed even when another update fails.

- **Tests**
  - Added coverage for concurrent message updates, partial failures, and successful batch processing of decrypted messages.


app/lib/encryption/encryption.ts
```diff
@@ -335,12 +335,13 @@ class Encryption {
...
 				...messagesToDecrypt,
 				...threadsToDecrypt,
 				...threadMessagesToDecrypt
 			];
-			toDecrypt = (await Promise.all(
+
+			const decrypted = await Promise.all(
 				toDecrypt.map(async message => {
 					const { t, msg, tmsg, attachments, content } = message;
 					let newMessage: Partial<TMessageModel> = {};
```
diegolmello, line 347: Minor, while you're in here: the `if (message.subscription)` guard on the next line is always true. `Message.js:23` is `@relation('subscriptions', 'rid')`, and the getter returns a Relation object no matter what `rid` is — it never consults it. So `newMessage` is never left as `{}` by that branch.

Either delete the guard, or make it a real `rid` check. As written, a row with a null `rid` reaches `decryptMessage` as `rid: undefined` anyway.

app/lib/encryption/encryption.ts
```diff
@@ -357,20 +358,25 @@ class Encryption {
...
-						);
-					} catch {
-						return null;
-					}
+					return { message, newMessage };
 				})
-			)) as (TThreadModel | TThreadMessageModel)[];
+			);
 
 			await db.write(async () => {
```
diegolmello, line 365: `rooms.ts:224-225` calls this on every subscription stream event. When there's nothing pending, each event still grabs the writer lock and calls `db.batch([])` for nothing.

One line, before the `db.write`:

```ts
if (!decrypted.length) return;
```

app/lib/encryption/encryption.ts
```diff
@@ -357,20 +358,25 @@ class Encryption {
...
+					.map(({ message, newMessage }) => {
+						try {
+							return message.prepareUpdate(
+								protectedFunction((m: TMessageModel) => {
+									Object.assign(m, newMessage);
+								})
+							);
+						} catch {
+							return null;
+						}
```
diegolmello, line 376: Worth a `log(e)` here now.

Before this PR this catch was mostly eating "pending changes" — expected noise, fine to swallow. After the fix it can't be: `Model/index.js:104` is the only pending-state throw, `Database/index.js:125` clears the state on commit, and the writer lock means nothing interleaves between the synchronous prepares and the batch. That throw is now structurally impossible — which is the whole point of the fix.

So `log(e)` will be silent in practice, and that's exactly why it's worth having: a future hit means something genuinely new, and right now it'd be invisible while the message stays encrypted forever.

(`protectedFunction` already logs anything thrown inside the updater, so this catch only hides the invariant.)

app/lib/encryption/encryption.ts
```diff
@@ -357,20 +358,25 @@ class Encryption {
...
+					})
+					.filter((record): record is TThreadModel | TThreadMessageModel => record !== null);
```
diegolmello, line 378: Two problems, one fix — delete the filter entirely.

The predicate is wrong: `TMessageModel` records are in this array too (they come from the `messages` collection), so the narrowing is a hand-asserted lie. It's inherited from the `as (TThreadModel | TThreadMessageModel)[]` cast you removed — the cast is gone but the wrong type survived it.

And the filter isn't needed at all. `db.batch` ignores falsy entries by design — `Database/index.js:91-93`, documented at :72-73 — and `updateMessages.ts:217` already relies on that. So `await db.batch(prepared)` with nulls in it is legal, and the bad type goes away with the filter.

app/lib/encryption/encryption.test.ts
```diff
@@ -141,3 +166,83 @@ describe('Encryption.encryptMessage', () => {
...
+		concurrentGate.resolve();
+
+		await expect(Promise.all([concurrentWrite, decrypting])).resolves.toBeDefined();
+
+		// The decrypted update reached db.batch and nothing was left prepared-but-uncommitted.
+		const committed = mockDbBatch.mock.calls.map(call => call.flat()).some(items => items.includes(record));
+		expect(committed).toBe(true);
+		expect(record.msg).toBe('plain');
+		expect(record.e2e).toBe('done');
+		expect(record._preparedState).toBeNull();
```
diegolmello, line 246: This test proves "doesn't throw", which is the important half. The other half of the contract is uncovered: a record whose `prepareUpdate` throws inside the lock gets skipped and the batch still commits with the rest.

That branch *is* the fix, and your `makeMessageRecord` fake already throws on a second prepare, so it's a cheap deterministic test — pre-set `_preparedState` on one of two records and assert only the other reaches `db.batch`.

diegolmello (changes requested):
The fix is right. `prepareUpdate` + `db.batch` inside one `db.write`, decryption outside it — that's the correct shape, and the regression test genuinely fails on the old code.

Requesting changes on one thing: the same bug is still in this file, 20 lines below the fix.

## Blocker: the twin is still broken

`decryptPendingSubscriptions` (`app/lib/encryption/encryption.ts:409-421`) does exactly what you just fixed — `sub.prepareUpdate` inside `Promise.all`, outside the lock, with `db.write` wrapping only the `db.batch` at :423. Same crash, same Bugsnag report, different stack. Fixing one and shipping the other means a second PR for the same ticket.

There's a second reason beyond the concurrent-writer race: `Model/index.js:122-131` asserts in dev that a prepared record reaches `batch()` **synchronously**. The subs path awaits `decryptSubscription` and then `db.write` after preparing, so it trips that invariant on its own, with no other writer involved.

Mirror the shape you just landed here. I wouldn't extract a shared helper — two call sites with different updaters, the abstraction costs more than it saves.

## Follow-up ticket, not this PR

`app/lib/methods/subscriptions/rooms.ts:158` has the same unlocked-prepare bug in a hotter path: prepares a subscription update, awaits `getMessageById` at :195, prepares a **message** update at :200, then finally opens `db.write` at :218 — and calls `decryptPendingMessages(tmp.rid)` at :225. That's the thing actually racing this decrypt path.

## Two things I checked so nobody else flags them

The `setImmediate` ordering in the test is fine. Everything between entry and `db.write` is microtasks — resolved-promise `fetch()` calls and the mocked `decryptMessage` — and `setImmediate` is a macrotask, so it drains all of them. An extra `await` of a resolved promise can't turn it green; it'd take a real timer or IO in that path.

The stateful `database` mock backing the five older `encryptMessage` tests is fine too. It's the only honest way to reproduce the writer lock without native WatermelonDB.

diegolmello (approved):

PR #7549, opened 2026-08-04
fix: Reopen a frozen DDP socket by correlating the health round trip

Rohit3523:
## Proposed changes

After the phone sleeps long enough for the DDP ping to go stale, an open room stops showing new messages for the rest of the session. Leaving and re-entering makes them appear — which looks like a rendering bug but is not: that is `RoomView` refetching over REST while the server-side stream stays dead.

Two pieces of evidence used to answer "is this socket still usable?" were wrong. Either one alone can leave a dead session in place.

**1. `probe()` had no request/response correlation.** It sent `{msg:'ping'}` with no id and resolved on the first `pong` event. When the OS unfreezes a backgrounded app it flushes the frames buffered during the freeze, so a pong answering a *pre-freeze* ping — or one the client provoked by auto-replying to a queued server ping — vouched for a session the server had already dropped. The `lastPing <= lastPingAtStart` check could not close this, because `onMessage` refreshes `lastPing` before emitting: it only proved *something* arrived, not that the server answered *us*.

**2. `classifySocketHealth` aged the socket against `lastPing`,** which `onMessage` refreshes for **every** inbound frame — so it measures traffic, not heartbeat health. One flushed message made a socket frozen for minutes look seconds old, sending it to a round trip instead of straight to a reopen.

When both misfire nothing is reopened: no `close`, no `connecting`, redux stays `connected`, the resume login never runs, `subscribeAll()` never re-sends the streams, and the room is deaf until something else triggers a sync.

This PR makes the ping carry an id and accept only the pong echoing it, and adds a `lastPongAt` clock that only a real server answer advances.

| File | Change |
|---|---|
| `patches/@rocket.chat+sdk+1.3.3-mobile.patch` | `probe()` puts an id on its ping and accepts only the matching pong, registered with `on` rather than `once` so an uncorrelated pong cannot consume the listener. Adds `lastPongAt` — advanced only by a pong and by the handshake reply, zeroed on `reopenNow` — exposed via a `DDPDriver` getter. |
| `app/lib/services/socketHealth.ts` | Ages against `lastPongAt`, falling back to `lastPing` on a driver without the patch. |
| 4 test files | Deterministic repros for both defects, plus the new correlation contract. |

The patch was regenerated from a pristine package extract and verified to reproduce `node_modules` byte-for-byte.

### Correction to the root cause on the ticket

The ticket attributes this to the reopen emitting no `close`, leaving redux connected and skipping the login. **That does not hold.** `createConnection()` emits **`connecting`**, which `connect.ts` maps to `connectRequest()` and clears `meteor.connected` — so a real reopen *does* run the resume login. Verified two ways: a test in `connect.test.ts` driving the real `connect` reducer, and a device wire trace showing `method: login {resume}` followed by 25 `sub` frames.

The observations on the ticket are right; the attribution is one layer too low. The truth is simpler — **no reopen happened at all**, which by itself explains both the missing `close` and the missing login request. Flagging it because a fix aimed at the stated cause would have changed nothing.

## Issue(s)

[NATIVE-1471](https://rocketchat.atlassian.net/browse/NATIVE-1471)

## How to test or reproduce

Affected screen: **RoomView** (any open room). Android emulator:

1. Log user A into the app, open a room shared with user B.
2. Freeze the device: screen off, then `adb shell dumpsys battery unplug`, then `adb shell dumpsys deviceidle force-idle`. Plain screen-off does **not** starve the socket.
3. Hold the freeze in the **90–170s** window. Under 90s the ping never goes stale; past ~200s the session ends on the login screen and heals correctly, which is a different outcome — "lock it longer to be safe" is wrong advice here.
4. While frozen, send a message from user B (REST or a second client).
5. Unlock with `adb shell wm dismiss-keyguard`. (`input keyevent 82` opens the RN dev menu on debug builds.)

**Before:** the room does not show B's message, and messages sent live after unlock are also missing. Leaving and re-entering the room makes them appear.

**After:** B's message appears in the open room without leaving it, and messages sent live after unlock arrive too.

Signals beyond the UI — after unlock the wire shows, in order:

```
connect  →  method: login {resume}  →  sub stream-room-messages <rid>
```

Tip for repeated cycles: use `adb shell input keyevent 223` / `224` (SLEEP / WAKEUP) rather than `26` (POWER). 26 is a toggle and drifts out of phase, which silently inverts the freeze phases.

## Screenshots

Not applicable — no visual change. The observable difference is message delivery; evidence is the wire trace above and the soak results below.

## Types of changes

## Checklist

- [ ] I have added necessary documentation (if applicable) — n/a
- [x] Any dependent changes have been merged and published in downstream modules — n/a, the SDK change ships in this repo's `patches/`

### Why this approach

A liveness check that cannot recognise its own answer is not a liveness check. DDP ping/pong carries an optional id and the server echoes it, so correlating is the smallest change that makes the round trip mean what its name says. It fails safe: an unanswered probe reopens rather than trusting a dead socket.

`lastPongAt` is defence in depth — with correlation in place, either fix alone resolves the reported symptom. It also stops a misleadingly-named field from being read as "last heartbeat" when it means "last frame".

Alternative considered and rejected: always reopen after any background gap. That discards the round-trip check and pays a full resume login, `subscribeAll()` and rooms refetch on every app switch.

### Verification

- **Unit/integration:** full suite green (2108 passing). The new repros fail without the fix and pass with it, driving the real patched driver over a mocked WebSocket. One red storyshot (`HeadersWithFormatting`) also fails on clean `develop`.
- **Live server:** confirmed Rocket.Chat echoes the ping id — `{msg:'ping',id:'probe-0'}` → `{msg:'pong',id:'probe-0'}`. This is a prerequisite: without the echo, `probe()` would always time out and every gray-zone foreground would pay a needless reconnect.
- **Device soak:** 30 consecutive freeze cycles on an Android emulator, durations rotating through 95/110/125/140/155/168s to sample the whole window. All 30 produced the full recovery ladder, and all 30 delivered a message posted *after* unlock over the live stream.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved detection of frozen or unresponsive connections, including cases where stale network data could appear healthy.
  - Prevented outdated connection responses from interfering with newer connections.
  - Improved reconnection reliability and ensured interrupted sessions can resume authentication smoothly.

- **Improvements**
  - Added more reliable connection health checks and coordinated reconnection handling.
  - Improved restoration of subscriptions, including user media updates, after reconnecting.

[NATIVE-1471]: https://rocketchat.atlassian.net/browse/NATIVE-1471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Rohit3523:
@coderabbitai review

Rohit3523:
@coderabbitai review

Rohit3523:
## Baseline results on unfixed `develop`

Ran the ticket's reproduction 30 times on each side, same harness, same emulator, freeze durations rotating through 95/110/125/140/155/168s.

| | blind cycles |
|---|---|
| unfixed `develop` (clean install from lockfile) | **1 / 30** (cycle 3) |
| this branch | **0 / 30** |

"Blind" = a message posted *after* unlock never arrived over `stream-room-messages`, i.e. the ticket's worst-case symptom.

### This does not statistically demonstrate the fix

0/30 vs 1/30 is a single event of difference; Fisher's exact gives p ≈ 1.0. At the ticket's own rate (3/28 ≈ 11%) you would need far more cycles per arm to separate the two. So the soak shows **no regression** and confirms the defect is still reachable on `develop`, but the evidence that this change fixes it rests on the deterministic unit/integration repros, which fail without it and pass with it.

### And the one baseline failure is probably a different defect

Cycle 3's wire trace:

```
517s OUT connect      <- unfroze exactly on schedule (125s freeze, valid window)
519s OUT login
519s OUT sub-room     <- stream re-subscribed
 ... BLIVE-3 posted ~110s later, never received ...
827s OUT connect      <- cycle 4
```

It reopened, re-authenticated and re-subscribed, and still lost the message. This PR makes a frozen socket reopen; it cannot help a case where the reopen already happened. Candidate causes: the `sub` frame written but never registered server-side, a re-subscribe rejected as a duplicate id, or the sync/resubscribe ordering handled on `room-resync-after-resubscribe`. Not diagnosed — I did not record `ready`/`nosub` frames.

### Harness caveat

11 of 30 baseline cycles overran their script because `adb`/`curl` intermittently stall on a dozing device. Cycle 3 was one of them, but its stall landed *after* unfreeze (proven by the on-time `connect` at 517s), so its freeze was a valid 125s and the result stands. For other stalled cycles the freeze itself overran, putting them outside the 90–170s window. Any future run should wrap every `adb`/`curl` in `timeout` so a hang cannot reshape a cycle.

PR #7552, opened 2026-08-05
fix: run handleDelete finds and prepares inside the writer lock

OtavioStasiak:
## Proposed changes

handleDelete in MessageErrorActions prepared its WatermelonDB changes before acquiring the writer lock. On the thread branch (tmid set) it called message.prepareDestroyPermanently() and then awaited three find calls  messages.find(message.id), messages.find(tmid), threads.find(tmid) — all outside db.write, only opening the write at the very end to run the batch.

  That leaves a window where the record carries a pending prepared change but nothing holds the lock. If a saga writes the same record during it, one of two things happens:

  - the concurrent writer throws Cannot update a record with pending changes (...), or
  - it commits first, which resets our record's _preparedState to null, so our db.batch then throws Cannot batch a record that doesn't have a prepared 
  create/update/delete.

  Either way, deleting a failed thread message fails and the thread count / thread record are left inconsistent. The plain (non-thread) branch had noawait between prepare and batch, so it was never exposed.

## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1467

## How to test or reproduce

- Run TZ=UTC npx jest app/containers/MessageErrorActions.test.tsx — 3 tests, all pass on this branch.
  - To see the bug the tests catch: git stash push app/containers/MessageErrorActions.tsx, re-run the command (race test fails with Cannot update a 
  record with pending changes (thread_messages#msg-1)), then git stash pop.
  - Affected screen: RoomView's failed-message action sheet (app/views/RoomView/index.tsx:1572) — specifically when the view is opened as a thread
  (tmid set).
  - Manual repro (timing-dependent, window is only as long as the three find calls): open a thread → go offline → send a message so it enters the
  failed state → have a second user post to the same thread (or reconnect so the sync saga writes it) → tap the failed message and choose Delete while
  that write lands.
  - Before the fix: the delete silently fails (the error is swallowed by log), the failed message stays in the list, and the thread header keeps a
  stale tcount.
  - Unchanged-path check: tap a failed message in the main channel view (not a thread) and delete it — it disappears as before.
  - Not verified by running the app — the manual steps are reasoning about the window, only the unit-level behavior was observed.
  
## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deletion of failed messages by completing related message and thread updates safely within a single database transaction.
  * Preserved thread counts and cleanup behavior when deleting messages, including messages without threads.
  * Prevented conflicting database writes during deletion.
  * Improved reliability when multiple message actions occur at the same time, helping ensure changes are applied consistently without incomplete cleanup.


app/containers/MessageErrorActions.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+const tick = () => Promise.resolve();
+
+/**
+ * Minimal WatermelonDB stand-in that reproduces the invariants we care about:
+ * - a record can only hold one prepared change at a time ("pending changes")
+ * - batch() only accepts records whose prepared change is still pending
+ * - batch() must run inside a writer
+ * - write() serializes writers (the writer lock)
+ */
+class FakeModel {
```
diegolmello, line 37: These mocks could be reused

diegolmello (approved):

PR #7553, opened 2026-08-05
fix: re-arm reconnect ladder when a forced socket reopen fails

Rohit3523:
## Proposed changes

When the app was brought back to the foreground while still offline (e.g. airplane mode), `reopenNow()` performs the immediate socket attempt but cancels the retry ladder timer first. If that attempt failed, nothing was left scheduled — and a stale timer id disabled the ladder for the rest of the session — so the session stayed disconnected until the *next* foreground event, even after connectivity returned.

Now a failed forced reopen re-arms the retry ladder and drops the stale id, so the app reconnects on its own once the network is back.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1461

## How to test or reproduce

1. Open the app, move to a room, enable airplane mode.
2. Move the app to the background, and from another account send a few messages to that room.
3. Move the app back to the foreground, with airplane mode still on.
4. Disable airplane mode.

Expected: within one retry interval the room header stops showing "Waiting for network…" and the messages sent in step 2 load.

Without this fix the header stays on "Waiting for network…" indefinitely — `meteor.connected` never returns to `true` — and the messages never arrive, until another background → foreground cycle forces a reconnect. Pull-to-refresh in RoomsListView loads the messages over REST but leaves the socket dead, which is why the room can look partly recovered while the banner persists.

Note on timing: the retry interval is 5s on a release build and 20s in debug (`reopen: __DEV__ ? 20000 : 5000` in `app/lib/services/sdk.ts`), so on a debug build allow ~20s after disabling airplane mode before judging the result.

## Types of changes

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

## Further comments

**Root cause.** `reopenNow()` in the patched DDP driver gets its "immediate" attempt by cancelling the pending ladder timer — but it only called `clearTimeout()`, never deleting the `openTimeout` id. `reopen()` guards on `if (this.openTimeout) return`, so `openTimeout` is not merely a timer handle: it is the flag for "a tick is already scheduled". A cleared-but-still-set id therefore silenced the entire ladder for the rest of the session, including the re-arm `onClose` attempts on every later close. So a single foreground while offline did not just cancel one tick — it permanently removed the app's ability to schedule another.

On top of that, when the forced attempt failed (device offline / server unreachable), the old cleanup path resolved the shared promise and left nothing scheduled, so that one failed reconnect meant the session sat disconnected with no retry path until the next explicit trigger.

**Why this fix.** The ladder was already the correct place to own retry scheduling; forcing a reopen just needed to not sabotage it. So instead of adding a parallel retry timer in the caller, `reopenNow()` now (a) deletes the cancelled id so `reopen()` can schedule again, and (b) re-arms the ladder — `this.reopen()` — whenever the attempt did not produce a live socket (on create error or the 10s deadline). A successful `open` deliberately skips the re-arm so no redundant tick is left behind. `reopenPromise` is deleted before the ladder is re-armed so a ladder tick calling `open()` cannot short-circuit onto the already-settled reopen promise.

The two parts are complementary. Deleting the id alone would let `onClose` recover the ladder, but a socket that never opens may never emit `close`, so the re-arm covers that case. The re-arm alone does nothing at all while the stale id is still present — an earlier revision of this fix had only the re-arm, passed the whole unit suite, and was a complete no-op on device.

**Verification.** Beyond the unit tests, verified on an Android emulator against a live server: three airplane-mode cycles following the steps above all recovered — banner cleared, `meteor.connected` back to `true`, and every message sent during the offline window arrived. A control cycle with no background/foreground step confirms the path that already worked is not regressed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved connection recovery after interruptions, timeouts, and connection errors.
  - Prevented outdated connections from interfering with active sessions.
  - Added connection health checks and reduced connection timeout delays.
  - Improved automatic restoration of user media signals and media call subscriptions after reconnecting.
  - Reduced redundant reconnect attempts and improved handling of simultaneous connection requests.


Rohit3523:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > patches/@rocket.chat+sdk+1.3.3-mobile.patch (1)> `75-76`: _🎯 Functional Correctness_ | _🟠 Major_ | _⚡ Quick win_
> > > **Do not resolve `open()` after a failed forced reconnect.**
> > > If `reopenNow()` fails, it resolves `reopenPromise` after it re-arms the reconnect ladder. This branch then resolves `open()` with `this.connection`, even when that socket is not open. A caller can treat `open()` as successful and send on a failed or connecting socket.
> > > Reject when `this.connected` is false, or wait for a subsequent `open` event before resolving.
> > > 
> > > Proposed fix
> > > ```diff
> > >  if (this.reopenPromise) {
> > > -  return this.reopenPromise.then(() => resolve(this.connection)).catch(reject)
> > > +  return this.reopenPromise
> > > +    .then(() => {
> > > +      if (!this.connected || !this.connection) {
> > > +        throw new Error('Forced reconnect did not establish a socket')
> > > +      }
> > > +      resolve(this.connection)
> > > +    })
> > > +    .catch(reject)
> > >  }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > 
> > > In `@patches/`@rocket.chat+sdk+1.3.3-mobile.patch around lines 75 - 76, Update the
> > > reopenPromise branch in open() so it does not resolve after a failed forced
> > > reconnect: after reopenPromise settles, resolve only when this.connected is
> > > true, otherwise reject (or wait for a subsequent open event). Ensure open()
> > > never reports success with a failed or still-connecting this.connection.
> > > ```
> 
> 🧹 Nitpick comments (2)
> > patches/@rocket.chat+sdk+1.3.3-mobile.patch (2)> `334-334`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Replace `any` with `DDPDriver`.**
> > > `ddp` exposes the new DDP-specific API. `any` disables type checking at this public boundary. Import `DDPDriver` as a type and declare `ddp?: DDPDriver`.
> > > As per coding guidelines, TypeScript code must use type safety for object shapes.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > 
> > > In `@patches/`@rocket.chat+sdk+1.3.3-mobile.patch at line 334, Replace the any
> > > type on the public ddp property with DDPDriver, importing DDPDriver as a type
> > > from the appropriate module. Preserve the property’s optional nature while
> > > restoring type checking for the DDP-specific API.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Coding guidelines_
> > > 
> > > `290-294`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Use `async` and `try`/`catch` for subscription acknowledgements.**
> > > `resubscribe` handles an asynchronous operation with a `.then()` chain. Make it an `async` function and use `try`/`catch`.
> > > As per coding guidelines, prefer `async`/`await` and use explicit `try`/`catch` handling for async operations.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > 
> > > In `@patches/`@rocket.chat+sdk+1.3.3-mobile.patch around lines 290 - 294, Update
> > > the resubscribe function to be async and await the Promise.all subscription
> > > acknowledgements inside a try/catch block. Return true after all subscriptions
> > > succeed and false from the catch path when any subscription fails, preserving
> > > the current behavior without the .then()/.catch() chain.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > _Source: Coding guidelines_
> 
> ```
> 
> Outside diff comments:
> In `@patches/`@rocket.chat+sdk+1.3.3-mobile.patch:
> - Around line 75-76: Update the reopenPromise branch in open() so it does not
> resolve after a failed forced reconnect: after reopenPromise settles, resolve
> only when this.connected is true, otherwise reject (or wait for a subsequent
> open event). Ensure open() never reports success with a failed or
> still-connecting this.connection.
> 
> ---
> 
> Nitpick comments:
> In `@patches/`@rocket.chat+sdk+1.3.3-mobile.patch:
> - Line 334: Replace the any type on the public ddp property with DDPDriver,
> importing DDPDriver as a type from the appropriate module. Preserve the
> property’s optional nature while restoring type checking for the DDP-specific
> API.
> - Around line 290-294: Update the resubscribe function to be async and await the
> Promise.all subscription acknowledgements inside a try/catch block. Return true
> after all subscriptions succeed and false from the catch path when any
> subscription fails, preserving the current behavior without the .then()/.catch()
> chain.
> ```
> 
> ℹ️ Review info
> ⚙️ Run configuration
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro Plus
> 
> **Run ID**: `e1b538ac-44e1-47cd-b356-0ac1bbfe8958`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [140a44e](https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/140a44e9dc54b4be186af3c1e1dabe525417a752) and [19526fd](https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/19526fdd4eab7c9cdcae595c0ab686c0b9e9125c).
> 
> 📒 Files selected for processing (2)
> * `app/lib/services/ddpSocket.test.ts`
> * `patches/@rocket.chat+sdk+1.3.3-mobile.patch`
> 
> 📜 Review details

none are being changed in this PR, with reasoning below.

1. open() resolving after a failed forced reopen

Correct about the contract, but out of scope here and not currently reachable. This PR only changes reopenNow; open() is pre-existing patch content from the earlier socket-health work.

open() has three callers and only reopen() consumes its resolution — checkAndReopen ignores it, DDPDriver.connect only uses .catch. So a false resolve costs exactly one thing: reopen()'s catch doesn't re-arm the ladder. This PR makes reopenNow's cleanup re-arm the ladder unconditionally whenever the attempt didn't produce a live socket, which covers that window, and once cleanup has run reopenPromise is deleted so open() no longer short-circuits. I couldn't construct a case where the false resolve strands a session after this change.

The suggested fix also needs a companion change: checkAndReopen calls this.open() with no .catch, so making open() reject would trade this wart for an unhandled rejection. I'll fold both into the next change that touches this area.

2. ddp?: any → DDPDriver

This wouldn't restore any type checking, so I'd rather not add an annotation that implies it did. The package ships index.d.ts containing only declare module '@rocketchat/sdk'; — not even the correct package name, the dot is missing — and package.json#types points at it, so the whole module resolves to any in app code. tsc --noEmit accepts all of the following without error:


sdk.current.thisPropertyDoesNotExist
sdk.current.ddp.alsoNotReal.deeplyFake()
const c: number = sdk.current.ddp
Separately, app/lib/services/sdk.ts declares the client as typeof Rocketchat — the constructor type, not the instance — so .ddp wouldn't resolve even with real declarations. Making this meaningful requires proper declarations plus fixing the instance typing, which will surface a lot of previously unchecked call sites. That deserves its own PR rather than a drive-by in a reconnect bugfix. (It would also need import type { DDPDriver }, since ddp is assigned from a dynamic import().)

3. resubscribe async/await

Pre-existing code inside waitForNotifyUserMediaSubs, untouched by this PR. Promise.all(...).then(() => true).catch(() => false) is behaviourally identical to the async/await form, so this is purely stylistic — better left to whoever next edits that function.

PR #7554, opened 2026-08-05
fix(e2ee): re-fetch subscription inside the write in toggleRoomE2EE

OtavioStasiak:
## Proposed changes

toggleRoomE2EE fetched the subscription record once, before showing the confirm Alert, and then reused that same instance for all three of its writes: the optimistic update, the Cancel revert, and the revert after a failed saveRoomSettings.

  The last two happen after the record has crossed user input (the Alert can stay open indefinitely) and a network round trip. If a subscription stream event updates the row in that window, the captured instance is stale, WatermelonDB rejects record.update(), and the revert is silently swallowed by the surrounding catch — leaving the switch showing an encrypted state the server never agreed to.

  Each update path now re-fetches the subscription with getSubscriptionByRoomId(rid) inside its own db.write callback and updates the fresh record, bailing out if the row no longer exists. optimisticUpdate takes a rid instead of a record; toggleRoomE2EE(rid) and all of its callers are unchanged.

  Added toggleRoomE2EE.test.ts, which mocks database.active with a store that reproduces WatermelonDB's staleness check (a handle remembers the version it was fetched at and throws record has pending changes if another writer bumped it). It covers a concurrent write landing while the Alert is open, one landing during the failing save request, and the happy path. Both concurrency tests fail against the previous code.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1468

## How to test or reproduce

 - Open a room where you can toggle E2E encryption (Room Actions → Encrypted, or the E2EE Toggle Room screen) and tap the switch — the switch flips
  immediately and the confirm Alert appears.
  - With the Alert still open, have the room's subscription updated from elsewhere (change a room setting from another client/admin, or receive a
  message in the room) so a stream event writes to that subscription row.
  - Tap Cancel. The switch must go back to its original value. Before this fix it stayed on the optimistic value, because the revert wrote to a record
  fetched before the Alert opened.
  - Repeat, but tap the confirm button (Enable/Disable) while offline or with the server unreachable so saveRoomSettings fails, again with a stream
  update landing during the request. The switch must revert.
  - Confirm the happy path is untouched: toggle, confirm, and the room stays encrypted/decrypted as chosen after a reload.
  - Unit tests: TZ=UTC pnpm test app/lib/encryption/helpers/toggleRoomE2EE.test.ts
  
## Types of changes

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved room encryption toggling to keep the displayed setting accurate when changes are cancelled, dismissed, or fail to save.
  - Added safeguards for concurrent updates and rooms that are no longer available.
  - Prevented encryption settings from becoming inconsistent after interrupted or unsuccessful changes.
  - Ensured successful encryption changes are saved and reflected correctly.


app/lib/encryption/helpers/toggleRoomE2EE.ts
```diff
@@ -7,12 +7,17 @@ import log from '../../methods/helpers/log';
...
 import { type TSubscriptionModel } from '../../../definitions';
 
-const optimisticUpdate = async (room: TSubscriptionModel, value: TSubscriptionModel['encrypted']) => {
+const optimisticUpdate = async (rid: string, value: TSubscriptionModel['encrypted']): Promise<void> => {
 	try {
 		const db = database.active;
 
 		// Instantly feedback to the user
 		await db.write(async () => {
+			// Fetch the room again: a stream event may have updated the record while the alert was open or while the request was in flight
```
diegolmello, line 16: Remove

diegolmello (approved):

PR #7476, opened 2026-07-09
ci: route Maestro e2e selection through sniffler impact analysis

diegolmello:
## Proposed changes

E2E test selection now runs through [sniffler](https://github.com/callstackincubator/sniffler) impact analysis, so a regular dev PR only executes the Maestro flows its diff actually touches instead of the full 14-shard suite every time. Release-cut PRs (base `master` or a `release` label) still run all 14. The design's hard rule is that **under-selection is impossible** — every uncertainty (no base ref, shallow clone, sniffler error, unparseable output, a `runAllWhenChanged` hit, a release lane) falls back to the full suite; the optimization only ever narrows, never risks missing coverage.

**CI (`build-pr.yml`)**
- `e2e-shards` now runs first and narrows the shard matrix via `select-impacted-shards.sh`. All six fail-safe guards fall back to the full 14.
- A confident zero (no impacted flow — e.g. a docs-only PR) sets `should_run=false` and skips the whole e2e stage: no manual approval, no native build, no flows.
- A new `always()` `e2e-result` job is the single required e2e check: green on a confident-zero skip, green when both run jobs pass, red otherwise. It never hangs "Waiting for status".

**Local**
- `pnpm e2e:changed <android|ios>` runs the impacted flows against a booted device with the app installed. It is working-tree-aware (committed + uncommitted + untracked) and mirrors CI's `util` / wrong-platform tag excludes.

**Coverage + freshness**
- `.sniffler/config.json` + `test-map.json` map the runtime-decoupled surface the import graph can't trace: domain sagas anchored to their flows; global sagas, notification/stream routers, navigation stacks, the root container, the store, and the root reducer in `runAllWhenChanged`.
- `validate-test-map.js` (wired into `eslint.yml`) fails a PR on orphan flows, dangling globs, stale globals, or any uncovered saga/stack, and warns on uncovered views — so the map can't silently rot.

**Tests**
- Jest unit tests for the selection tooling — `select-impacted-shards.sh`, `e2e-changed.sh`, and `validate-test-map.js` — so the logic that gates e2e is exercised on every `ESLint and Test` run, not just verified by hand. A shell harness stubs the external deps (`sniffler` / `git` / `maestro`) on `$PATH`; per-branch fixture maps drive each validator check.
- Each validator failure mode (orphan flow, dangling glob, uncovered view, stale global, decoupled gap) has a dedicated fixture, plus a clean-pass fixture.

**Other**
[...]

PR #7555, opened 2026-08-05
fix(ci): select shards for changes outside app and honor the release label

diegolmello:
## Proposed changes

The sniffler shard selection from #7476 has two separate faults. I found both during work on NATIVE-1461. The pull request for that task selected no shards, and CI skipped all Maestro flows.

**A change outside `app/` selected no shard.** The `source.roots` value is `["app"]`. Therefore no file outside `app/` enters the import graph. The `runAllWhenChanged` list held only paths inside `app/`, plus `package.json` and `pnpm-lock.yaml`. A change to a runtime file outside `app/` matched no flow. The `select-impacted-shards.sh` script then reported a confident-zero diff.

The diff for NATIVE-1461 changes `patches/@rocket.chat+sdk+1.3.3-mobile.patch`. That patch changes the DDP socket, and every flow depends on it. Sniffler returned `recommendedTests: []`. The same fault applied to `android/`, `ios/`, `metro.config.js`, `babel.config.js`, `app.json`, `index.js`, `react-native.config.js` and `.maestro/`.

This change also makes `validate-test-map.js` accept a glob. That part is necessary, not optional. Sniffler reads each `runAllWhenChanged` item as a pattern. The earlier `existsSync` check refuses every glob as stale and fails CI.

**The release label did not force the full set of shards.** The gate used `contains(github.event.pull_request.labels.*.name, 'release')`. The `contains()` function does an exact match on each item of an array. Therefore it needs a label with the exact name `release`. Our label has the name `💎 release`.

The same line has a second fault. The `build-pr.yml` file declares no `types:` key. Therefore the workflow does not start when a user adds a label, and a new run reads the old event data. Even the exact name works only if the label was on the pull request before the last push. The job now reads the labels with `gh pr view` during the run. This corrects both faults. It also avoids a `labeled` trigger, which starts a full Build PR each time a user changes a label.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1484

## How to test or reproduce

Run sniffler against the two commits of the NATIVE-1461 pull request:

```
pnpm exec sniffler impact --base 140a44e9dc --head 88700b52b8 --format json
```

- Before the change: `recommendedTests: []`
- After the change: 55 flows, with `reason: run-all` from `patches/**`

Then run the CI script:

```
IS_RELEASE_LANE=false BASE_REF=develop HEAD_SHA=88700b52b8 \
  FULL_SHARDS='[1,2,3,4,5,6,7,8,9,10,11,12,13,14]' \
  bash .github/scripts/select-impacted-shards.sh
```

- Before the change: `shards=[] should_run=false`
- After the change: `shards=[1,2,3,4,5,6,7,8,9,10,11,12,13,14] should_run=true`

To test the label, add `💎 release` to a pull request. The `Detect release label` step must print `release=true`. The preflight must then select all 14 shards.

## Screenshots

Not applicable. This change affects only CI.

## Types of changes

## Checklist

## Further comments

The test count in `.github/scripts/__tests__` goes from 38 to 40. Both new tests can fail. I ran them against the earlier validator. The glob test failed, and the test that uses the real repository also failed, because the earlier `existsSync` check refuses the new globs.

One item is suitable for a separate task. The header comment of `select-impacted-shards.sh` says that under-selection is not possible. That statement is true only if the test-map covers each runtime path. This fault shows that it did not. Either correct the comment, or let the validator check each top-level runtime path.

## Summary by CodeRabbit

- **CI Improvements**
  - Pull request builds now identify release labels more reliably, including labels with decorative prefixes, to select the appropriate test shards.
  - Test suites automatically rerun when relevant configuration, patch, mobile platform, automation, or workflow files change.
  - Test-map validation now supports glob patterns, including matches for hidden files, while detecting patterns that match nothing.

- **Tests**
  - Added coverage to verify valid glob handling and stale-pattern detection.


app/views/DirectoryView/index.tsx
```diff
@@ -214,3 +214,5 @@ const DirectoryView = ({ navigation }: IDirectoryViewProps): ReactElement => {
 };
 
 export default DirectoryView;
+
+// smoke test: narrow app change
```
Rohit3523: We can revert this, as it was added for testing

app/views/ForgotPasswordView.tsx
```diff
@@ -99,3 +99,5 @@ const ForgotPasswordView = ({ route }: ForgotPasswordViewProps): ReactElement =>
 };
 
 export default ForgotPasswordView;
+
+// smoke test: narrow app change
```
Rohit3523: We can revert this, as it was added for testing

.github/workflows/build-pr.yml
```diff
@@ -94,10 +95,26 @@ jobs:
...
+      # at push time (a label added afterwards, or a re-run, would never see it),
+      # and `contains()` over an array is an exact element match — it would miss
+      # the decorated label name ("💎 release").
+      - name: Detect release label
+        id: label
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+        run: |
+          if gh pr view "$PR_NUMBER" --json labels --jq '.labels[].name' | grep -qi 'release'; then
```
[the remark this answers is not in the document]
Rohit3523: This change suggested by coderabbit is good, instead of matching with all 'release', we can strictly match with '💎 release'

.github/scripts/testlib/fixtures/maps/stale-global/patches/some.patch
Rohit3523, line 1: Is this file for testing the CI? Can we remove it?

.github/workflows/build-pr.yml
```diff
@@ -94,10 +95,26 @@ jobs:
...
+            echo "release=true" >>"$GITHUB_OUTPUT"
+          else
+            echo "release=false" >>"$GITHUB_OUTPUT"
+          fi
```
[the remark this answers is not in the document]
Rohit3523: If the lookup fails, it's a good idea to set release false

.github/scripts/testlib/fixtures/maps/stale-global/patches/some.patch
diegolmello: no, it's needed

Rohit3523 (approved):

PR #7557, opened 2026-08-06
fix: re-fetch message inside the write in getThreadName

OtavioStasiak:
## Proposed changes

In `getThreadName`, the "no thread yet" branch prepared its update on a message record fetched *before* the network call and decryption. A sync write can update or delete that row during the gap, making `prepareUpdate` throw on a record with pending changes.

  The message is now re-fetched inside the `db.write` callback, and the update is prepared on that fresh record. `getSingleMessage` and `Encryption.decryptMessage` stay outside the lock, and the function signature and callers are unchanged.

  Added a regression test that mocks `database.active` and simulates a concurrent writer during the gap, the function still commits without throwing.
  
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1469

## How to test or reproduce

 - Run `TZ=UTC npx jest app/lib/methods/getThreadName.test.ts` — passes on this branch.
  - Stash `app/lib/methods/getThreadName.ts` and re-run — the test fails, showing the update was prepared on the stale record.
  - Manually: open a thread message from a room where the parent thread isn't cached yet (jump to message from Search/Messages view, or tap a thread
  reply after a fresh cache clear) while messages are syncing. The thread name should resolve and no "Cannot update a record with pending changes"
  error should be logged.
  - No UI changes — affects RoomView, SearchMessagesView and MessagesView only through the thread name they display.

## Types of changes

## Checklist

## Bug Fixes
- Improved thread naming to use the latest available message information when updates occur concurrently.
- Prevented stale thread names from being saved during rapid message changes.
- Improved fallback behavior when attachment titles are used or remote name retrieval fails.

## Tests
- Added coverage for concurrent updates, cached-name handling, attachment-title fallbacks, thread creation, and remote-fetch failures.


app/lib/methods/getThreadName.ts
diegolmello, line 19: You're introducing a different coding style. What about changing messageRecord to let instead of creating freshMessageRecord?

app/lib/methods/getThreadName.test.ts
```diff
@@ -0,0 +1,103 @@
...
+		jest.clearAllMocks();
+		batch.mockResolvedValue(undefined);
+		(database as any).active = {
+			get: jest.fn(() => threadCollection),
+			write: jest.fn((fn: () => Promise<void>) => fn()),
+			batch
+		};
+	});
+
+	it('re-fetches the message inside the write so a concurrent writer during the network gap does not break the update', async () => {
```
diegolmello, line 77: Write tests for the other use cases

diegolmello (changes requested):

diegolmello (approved):

PR #7563, opened 2026-08-10
fix: attachments with # in the filename fail to download

OtavioStasiak:
## Proposed changes

An attachment whose filename contains `#` couldn't be downloaded or played — the request reached the server with a truncated path and returned the wrong resource or a 404.

  The server builds the attachment path with `encodeURI(file.name)` ([sendFileMessage.ts#L57](https://github.com/RocketChat/Rocket.Chat/blob/develop/apps/meteor/server/meteor-methods/messages/sendFileMessage.ts#L57)),
  and `encodeURI` doesn't escape `#`. So a file named `a video #2.mov` arrives as:

      /file-upload/<id>/a%20video%20#2.mov

  Every consumer downstream reads that `#` as the fragment delimiter: `setParamInUrl`'s `new URL()` splits it into `path = /file-upload/<id>/a%20video%20` + `hash = #2.mov` and re-emits the tail after the auth params, and HTTP drops the fragment before connecting. The server only ever sees `a video `.

  `formatAttachmentUrl` now escapes `#` → `%23` on server-origin attachment urls, before anything parses them. Attachment urls never carry a meaningful fragment, so this is unambiguous.

  Notes on placement:
  - On the `rc_token` branch the escape runs *after* `encodeURI`, not before — `encodeURI` leaves `#` raw but does escape `%`, so pre-escaping would  have produced `%2523`.
  - The external `_originalUrl` early return is untouched: that's not a file-upload path, so a `#` there can be a genuine fragment.
  - `?` in a filename breaks the same way, but is deliberately left alone — the server does emit real query strings (signed-url expiry), so there's no safe string-level rule to tell the two apart.
  Covered by a new `formatAttachmentUrl.test.ts` (relative, absolute, `FileUpload_ProtectFiles`, pre-authed and CDN-prefixed urls, plus already-escaped / external / base64 / `file://` passthroughs).
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1486

## How to test or reproduce

  1. Upload a file named `a video #2.mov` to a channel.
  2. Tap it — before this change the download fails or resolves to the wrong file; after, it opens normally.
  3. Repeat with **File Upload → Protect Uploaded Files** enabled, which exercises the `setParamInUrl` path.
  
## Types of changes

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed attachment URLs containing `#` characters so images and videos load correctly.
  * Improved attachment URL handling across protected, authenticated, CDN-hosted, external, data, and local file sources.
  * Corrected GIF detection for formatted URLs, including URLs with query strings, fragments, and escaped fragment characters.
  * Improved reliability when displaying image and video attachments with complex or encoded URLs.

### DIFF
--- a/app/lib/methods/helpers/formatAttachmentUrl.ts
+++ b/app/lib/methods/helpers/formatAttachmentUrl.ts
@@ -10,6 +10,9 @@ function setParamInUrl({ url, token, userId }: { url: string; token: string; use
 	return urlObj.toString();
 }
 
+// The server encodes the path with `encodeURI(file.name)`, which leaves `#` raw and truncates the url at the fragment.
+const escapeFragmentDelimiter = (url: string) => url.replace(/#/g, '%23');
+
 export const formatAttachmentUrl = (
 	attachmentUrl: string | undefined,
 	userId: string,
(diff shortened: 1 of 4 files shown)


app/lib/methods/helpers/formatAttachmentUrl.ts
```diff
@@ -10,6 +10,9 @@ function setParamInUrl({ url, token, userId }: { url: string; token: string; use
 	return urlObj.toString();
 }
 
+// The server encodes the path with `encodeURI(file.name)`, which leaves `#` raw and truncates the url at the fragment.
```
diegolmello, line 13: The server does that? Are you sure? What version started doing that?

app/views/AttachmentView.tsx
```diff
@@ -59,9 +59,9 @@ const RenderContent = ({
 	}, [navigation]);
 
 	if (attachment.image_url) {
-		const url = formatAttachmentUrl(attachment.title_link || attachment.image_url, user.id, user.token, baseUrl);
-		const uri = encodeURI(url);
-		const isAnimated = attachment.image_type === 'image/gif' || /\.gif(\?|$)/i.test(url);
+		const uri = formatAttachmentUrl(attachment.title_link || attachment.image_url, user.id, user.token, baseUrl);
+		// `#` is escaped to `%23` before it reaches here, so both spellings can follow the extension.
```
diegolmello, line 63: Remove

diegolmello (changes requested):

app/lib/methods/helpers/formatAttachmentUrl.ts
```diff
@@ -10,6 +10,9 @@ function setParamInUrl({ url, token, userId }: { url: string; token: string; use
...
+const escapeFragmentDelimiter = (url: string) => url.replace(/#/g, '%23');
```
diegolmello, line 14: What other symbols should be escaped as well?

diegolmello (changes requested):

PR #7579, opened 2026-08-17
fix(voip): unblock incoming calls on Android below API 33

OtavioStasiak:
## Proposed changes

The bug: on Android phones older than API 33, the app never called RNCallKeep.setup. Why? Because index.js first checked "does this device have android.software.telecom?" — and that feature flag literally doesn't exist before API 33. So the check always said no, setup never ran, and after the first call there was a leftover Telecom connection sitting there jamming the door. Every call after that got blocked.

  What the PR does, file by file:

  index.js — throw away the FEATURE_TELECOM check and just always run RNCallKeep.setup on Android. Devices
  that genuinely can't do Telecom already give up on the native side in registerPhoneAccount, so that JS
  check was doing nothing except breaking old phones.

  VoipModule.kt + NativeVoip.ts + VoipModule.mm — new method disconnectNativeCall(callId). It grabs the
  connection straight out of VoiceConnectionService and calls onDisconnect() on it, going around
  RNCallKeep's JS bookkeeping entirely. On iOS it's an empty stub, since CallKit gets cleaned up through
  RNCallKeep there anyway.

  terminateNativeCall.ts — two things. First, order matters: the native disconnect has to happen before
  RNCallKeep.endCall, because endCall yanks the connection out of VoiceConnectionService's map and then
  there's nothing left to disconnect. Second, calling it twice is now safe — it keeps a set of the last 32
  call ids it already killed. Needed because three different places all try to tear the call down:
  useCallStore.endCall, acceptNativeCall, and the MediaSessionInstance handlers.

  useCallStore.ts — call.hangup() is now in a try/catch that logs. Before, if hangup blew up it took the
  rest of the cleanup down with it.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1460

## How to test or reproduce

- open the app in a device android API < 33;
- receive a call;
- accept the call;
- press end button;
- receive another call;
- it must work;

## Types of changes

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Android call termination by reliably disconnecting native call connections before cleanup.
  - Prevented duplicate termination attempts and continued cleanup when hangup or native disconnection errors occur.
  - Preserved expected iOS call teardown behavior through the existing call management flow.

- **Compatibility**
  - Android call handling now works on devices without Telecom support.

- **Tests**
  - Added coverage for teardown ordering, duplicate calls, error handling, fallback behavior, and iOS platform differences.


android/app/src/main/java/chat/rocket/reactnative/voip/VoipModule.kt
```diff
@@ -199,6 +200,20 @@ class VoipModule(reactContext: ReactApplicationContext) : NativeVoipSpec(reactCo
         }
     }
 
+    override fun disconnectNativeCall(callId: String) {
+        try {
+            when (val connection = VoiceConnectionService.getConnection(callId)) {
+                null -> Log.d(TAG, "disconnectNativeCall: no connection for $callId")
+                else -> {
+                    connection.onDisconnect()
```
diegolmello: **Incomplete teardown: terminating a ringing call leaves the full-screen UI up and ringing**

- **Severity:** high — user-visible stuck ringing UI.
- **Likelihood:** high — any terminate that lands while the call is still ringing.
- **ELI5:** this hangs up the call in the phone system but forgets to tell the ringing screen, the notification, and the background service to go away. So the phone keeps ringing at a call that no longer exists.

Every sibling of this helper is paired at its call site with the full dismissal set. Compare `VoipNotification.kt:570-576`:

```kotlin
disconnectIncomingCall(callId, false)
VoipCallService.stopService(appContext)
cancelById(appContext, payload.notificationId)
LocalBroadcastManager.getInstance(appContext).sendBroadcast(Intent(ACTION_DISMISS) ...)
```

This helper does `connection.onDisconnect()` and nothing else. `IncomingCallActivity.kt:108-109` registers only `ACTION_TIMEOUT` and `ACTION_DISMISS`, so with no broadcast the activity never finishes.

Also note the `else ->` branch: siblings special-case `is VoiceConnection` to pick the right disconnect (`onReject()` for a declined call, `reportDisconnect(DISCONNECT_REASON_MISSED)` for a timeout). This one always uses `onDisconnect()`, so a ringing-call terminate reports a wrong `DisconnectCause`. Milder than the missing broadcast, but it is the reason the `// TODO: unify these three functions` at `VoipNotification.kt:397` is now four.

**Repro:**
1. Android device, inbound VoIP push → full-screen incoming UI is ringing.
2. Trigger termination through the JS path (`terminateNativeCall`) rather than a notification action — e.g. hang up from the other device, or end from the app.
3. `disconnectNativeCall` runs; no `ACTION_DISMISS` is broadcast.
4. The full-screen `IncomingCallActivity` stays on screen and keeps ringing.
5. `adb shell dumpsys activity activities | grep IncomingCallActivity` confirms it is still resumed.

**Fix:** either broadcast the dismissal set here, or reuse the existing `disconnectIncomingCall` + stopService + cancelById + `ACTION_DISMISS` sequence instead of adding a fourth variant.

android/app/src/main/java/chat/rocket/reactnative/voip/VoipModule.kt
```diff
@@ -199,6 +200,20 @@ class VoipModule(reactContext: ReactApplicationContext) : NativeVoipSpec(reactCo
         }
     }
 
+    override fun disconnectNativeCall(callId: String) {
```
diegolmello: **Telecom `Connection` mutated off the main looper**

- **Severity:** medium — nondeterministic teardown failure, the worst kind to debug here.
- **Likelihood:** low-to-medium — depends on which JS thread the bridge call lands on; a TurboModule sync method runs on the caller's thread, not the UI thread.
- **ELI5:** the phone system expects to be poked from one specific thread. This pokes it from whichever thread happens to be free, which usually works and occasionally does not.

Every other disconnect in this codebase is posted to the main looper — `VoipNotification.kt:564` wraps `disconnectIncomingCall` in `Handler(Looper.getMainLooper()).post { ... }`, and this class already holds `fgsHandler = Handler(Looper.getMainLooper())` at `VoipModule.kt:117`. The new helper posts nothing.

The `catch (e: Exception)` makes this worse rather than better: a threading failure gets logged and swallowed, so the connection stays ACTIVE and the caller believes teardown succeeded.

**Repro:** hard to trigger on demand — that is the point. Under load (or with the JS call arriving on a non-UI thread) the `onDisconnect()` is dropped, `disconnectNativeCall: failed to disconnect` appears in logcat, and `dumpsys telecom` still shows the connection.

**Fix:** `fgsHandler.post { ... }` around the body, matching `VoipNotification.kt:564`.

app/lib/services/voip/terminateNativeCall.ts
```diff
@@ -17,3 +51,11 @@ export function terminateNativeCall(callId: string): void {
 		}
 	}
 }
+
+/**
+ * Resets module-scoped state for testing purposes.
+ * NOT intended for production use.
+ */
+export function resetTerminateNativeCallForTesting(): void {
```
diegolmello: **Test-only export in a production module**

- **Severity:** low — no runtime impact.
- **Likelihood:** n/a — it is in the shipped bundle unconditionally.
- **ELI5:** a door labelled "staff only, do not use" built into the customer entrance. Nothing stops someone walking through it.

This is against the standing coding-style rule (no test-only code in non-test files), and the doc comment saying "NOT intended for production use" is the tell — a comment is being used where structure should be.

**Fix:** move the dedupe set to the module that already owns VoIP sentinels, alongside `clearVoipAcceptDedupeSentinels` in `MediaCallEvents.ts`, and let tests reset it via the same `resetVoipState()` production entry point the rest of the teardown uses. That deletes this export and resolves the sibling finding on line 10 at the same time.

app/lib/services/voip/terminateNativeCall.ts
```diff
@@ -3,12 +3,46 @@ import RNCallKeep from 'react-native-callkeep';
 
 import NativeVoipModule from '../../native/NativeVoip';
 
+// Termination is triggered from several independent paths (useCallStore.endCall,
+// acceptNativeCall, and the MediaSessionInstance event handlers), so the same
+// callId arrives more than once. Bounded because the app is long-lived.
+const MAX_TRACKED_CALL_IDS = 32;
+const terminatedCallIds = new Set<string>();
```
diegolmello: **No production path clears `terminatedCallIds`**

- **Severity:** low — bounded by the 32-entry LRU below, so not a leak.
- **Likelihood:** low — needs a callId to recur after 32 intervening terminations, or a logout/account switch that should have reset it.
- **ELI5:** we keep a list of calls we already hung up so we do not hang up twice. Nothing ever wipes that list on logout — only the tests can wipe it.

The asymmetry is the real problem: `resetVoipState.ts:6` clears the analogous accept-dedupe sentinel precisely so that "Android cold-start and re-delivery paths are not poisoned by a prior call." This sentinel has the same failure mode and no equivalent clear, and the only reset that exists is the test-only export below.

**Repro (constructed):**
1. Terminate a call with id `X`.
2. Log out and log back in (or switch server) — `resetVoipState()` runs, clearing the accept sentinels but not this Set.
3. Receive a call that reuses id `X`.
4. `terminateNativeCall('X')` returns at the guard; no native disconnect, no `endCall`, no `stopVoipCallService`.
5. Connection stranded ACTIVE.

**Fix:** clear it from `resetVoipState()` together with the accept sentinels — see the note on line 59; one move covers both.

app/lib/services/voip/MediaSessionInstance.test.ts
```diff
@@ -137,7 +137,8 @@ jest.mock('../../native/NativeVoip', () => ({
 	default: {
 		stopNativeDDPClient: jest.fn(),
 		startVoipCallService: (callId: string) => mockStartVoipCallService(callId),
-		stopVoipCallService: () => mockStopVoipCallService()
+		stopVoipCallService: () => mockStopVoipCallService(),
+		disconnectNativeCall: jest.fn()
```
diegolmello: **This mock entry is dead, and the wider test isolation for the new dedupe Set is missing**

- **Severity:** medium — the new module-level Set is order-dependent state shared across test files, which is how green suites start hiding real regressions.
- **Likelihood:** high — already true on this branch.
- **ELI5:** we added a note-to-self that survives between tests, then only told one test file how to erase it. The other files reuse the same call names, so whichever runs second may silently do nothing.

Two separate problems:

1. **This line is dead.** This file mocks `./terminateNativeCall` wholesale at `:23-25`, so `NativeVoipModule.disconnectNativeCall` is never reached from here. Harmless but misleading — it reads like coverage that does not exist.
2. **Isolation is incomplete.** `resetTerminateNativeCallForTesting()` is called in exactly one file, `VoipCallLifecycle.integration.test.tsx:400` — and that file's own `NativeVoip` mock at `:139-142` is `{ stopNativeDDPClient: jest.fn() }`, missing both `disconnectNativeCall` and `stopVoipCallService`. Meanwhile `useCallStore.test.ts`, `useCallStore.ios.test.ts` and `CallView/index.test.tsx` all gained the `disconnectNativeCall` mock but never reset the Set, and they reuse callIds.

**Repro:**
1. Run the VoIP suites in one Jest worker: `TZ=UTC pnpm test app/lib/services/voip app/views/CallView app/containers/NewMediaCall --runInBand`.
2. Add a second assertion in `useCallStore.test.ts` that terminates a callId an earlier test in the same file already terminated.
3. It fails — the guard short-circuits — even though the production behavior under test is fine.
4. Reorder the files and different assertions fail. Order-dependence confirmed.

**Fix:** once the Set moves behind `resetVoipState()` (see the comments on `terminateNativeCall.ts`), reset it in each suite's `beforeEach` through that production API, and drop this dead line.

ios/Libraries/VoipModule.mm
```diff
@@ -124,6 +124,10 @@ - (void)startVoipCallService:(NSString *)callId
     resolve(nil);
 }
 
+// Android-only Telecom teardown fallback. iOS tears CallKit down through RNCallKeep.
+- (void)disconnectNativeCall:(NSString *)callId {
```
diegolmello: **Pre-existing, not introduced here — but this stub highlights it: `stopVoipCallService` has no ObjC implementation**

- **Severity:** medium if ever reached — unrecognized-selector crash, not a soft failure.
- **Likelihood:** very low today — unreachable behind the `Platform.OS === 'android'` guard at `terminateNativeCall.ts:44`.
- **ELI5:** the shared contract promises iOS can do two things. This PR correctly adds a do-nothing stub for the new one, but the older promise still has no stub at all — call it on iOS and the app crashes rather than doing nothing.

Flagging for the record, not as PR feedback: `app/lib/native/NativeVoip.ts:57` declares `stopVoipCallService()` in the Spec, and this file implements `startVoipCallService` but not `stopVoipCallService`. Adding the `disconnectNativeCall` stub here is exactly right; the gap next to it is the one worth noticing, because the only thing preventing a crash is a platform check in JS. Anyone who later collapses that guard — a reasonable-looking cleanup, since two of the three natives are now no-ops on iOS — gets a crash.

**Repro:**
1. Remove the `Platform.OS === 'android'` guard around `NativeVoipModule.stopVoipCallService()` in `terminateNativeCall.ts:44`.
2. Run on an iOS device/simulator and end a call.
3. `-[VoipModule stopVoipCallService]: unrecognized selector sent to instance`.

**Fix:** add the empty stub alongside this one, with the same one-line comment explaining why it is empty.

diegolmello (commented):
Review of the native-teardown change. Each finding below carries severity, likelihood, an ELI5, and repro steps. Findings on files this PR does not touch could not be attached inline, so they are here.

The inline comments cover: the incomplete Telecom teardown helper, its threading, the test-only production export, the unclearable dedupe set, the dead test mock, and the iOS stub.

---

### A. `MediaSessionInstance.endCall` can skip the whole teardown it just made more important

`app/lib/services/voip/MediaSessionInstance.ts:281,283`

- **Severity:** high — strands exactly the ACTIVE Telecom connection this PR exists to clear.
- **Likelihood:** medium — needs `reject()`/`hangup()` to throw (dead socket, disposed session), which is the same condition `pendingHangups` was added to survive.
- **ELI5:** we hang up the call, then clean up the phone system. If hanging up trips, we never get to the cleanup — and the phone system stays convinced you are still on a call, so no one can ring you again.

The PR added `try/catch` around `call.hangup()` in `useCallStore.ts:313-317` but not here. A throw at `:281`/`:283` skips `terminateNativeCall(callId)` at `:286` and the four cleanup lines after it. This path has no other teardown, so nothing else clears the connection.

**Repro:**
1. Answer an inbound call on Android.
2. Kill connectivity so the signalling session is dead but `getCallData(callId)` still returns the call.
3. End the call from the UI (`MediaSessionInstance.endCall`).
4. `mainCall.hangup()` throws; `terminateNativeCall` never runs.
5. `adb shell dumpsys telecom | grep -A5 SelfManaged` still shows the connection ACTIVE.
6. Send a new inbound call — rejected as busy. Only `am force-stop` clears it.

**Fix:** wrap it the way `useCallStore` now does, or better — the three teardown sequences (`MediaSessionInstance.endCall`, `useCallStore.endCall`, `handleVoipAcceptFailed`) are now three hand-rolled subsets of the same five steps. One shared teardown function removes this whole class of bug.

---

### B. Accept-failure path bypasses the native disconnect entirely

`app/sagas/deepLinking.js:89-92`

- **Severity:** high — same stranded-connection outcome as A.
- **Likelihood:** medium — every failed native accept.
- **ELI5:** when answering a call fails, we tell the old, unreliable helper to clean up — the exact helper this PR concluded cannot be trusted.

`handleVoipAcceptFailed` calls raw `RNCallKeep.endCall(callId)`. No `disconnectNativeCall`, no `stopVoipCallService`. The PR's own premise is that CallKeep's JS module state may not be initialized on this path, so this is the one place most likely to need the native fallback.

**Repro:**
1. Cold-start via a VoIP push on Android and accept from the full-screen incoming UI.
2. Force the accept to fail (server rejects, or no network).
3. `handleVoipAcceptFailed` runs → `RNCallKeep.endCall` only.
4. `dumpsys telecom` still lists the connection.
5. Next inbound push is rejected as busy.

**Fix:** call `terminateNativeCall(callId)` instead of `RNCallKeep.endCall`.

---

### C. Unverified: `setSettings` now persists `foregroundService` on API < 33

`index.js:26` (interacts with vendored `RNCallKeepModule.java:353`, `VoiceConnectionService.java:362,368`)

- **Severity:** medium if confirmed — a call notification that outlives the call.
- **Likelihood:** unknown — **not yet reproduced on a device.**
- **ELI5:** now that setup runs on older Androids, CallKeep may start showing its own call notification there for the first time, and the code meant to take it down looks like it passes the wrong kind of number.

Removing the JS gate means `setSettings(options)` runs on API < 33 for the first time, persisting `foregroundService` and making `isForegroundServiceConfigured()` true. `stopForeground(128)` appears to pass the service *type* where the flags argument belongs, and `128 & STOP_FOREGROUND_REMOVE(1) == 0`, so that call would not remove the notification. Not permanent — it still dies with the ConnectionService.

**Repro (to run before writing any fix):**
1. Boot an API 30 (or any <33) emulator.
2. Install this branch's debug build.
3. Receive and answer a VoIP call, then end it.
4. Check whether a CallKeep foreground notification remains in the shade after the call ends.
5. `adb shell dumpsys activity services | grep -i callkeep` to see whether the service is still up.

If the notification does not linger, close this out — the reasoning is static-read only.

---

### Two structural notes

1. `terminateNativeCall` now owns a dedupe Set, an LRU, a platform branch, three `try/catch` blocks and a test-only export. The dedupe belongs with the other VoIP sentinels, next to `clearVoipAcceptDedupeSentinels` and cleared by `resetVoipState` — that single move resolves both inline findings on this file.
2. Findings A and B are both symptoms of the missing single teardown path described in A.

index.js
```diff
@@ -23,7 +22,8 @@ if (process.env.USE_STORYBOOK) {
 
 	LogBox.ignoreAllLogs();
 
-	if (Platform.OS === 'android' && DeviceInfo.hasSystemFeatureSync('android.software.telecom')) {
+	// Never gate on FEATURE_TELECOM (API 33+): non-Telecom devices bail natively in registerPhoneAccount
```
diegolmello: remove

index.js
```diff
@@ -23,7 +22,8 @@ if (process.env.USE_STORYBOOK) {
 
 	LogBox.ignoreAllLogs();
 
-	if (Platform.OS === 'android' && DeviceInfo.hasSystemFeatureSync('android.software.telecom')) {
```
diegolmello, line 26: You should be aware of this commit and prevent a regression https://github.com/RocketChat/Rocket.Chat.ReactNative/commit/540d5e5aaa898ec98c2bfaff87a101a7b39f1538

index.js
```diff
@@ -23,7 +22,8 @@ if (process.env.USE_STORYBOOK) {
...
+	// Never gate on FEATURE_TELECOM (API 33+): non-Telecom devices bail natively in registerPhoneAccount
+	if (Platform.OS === 'android') {
```
diegolmello: Ran a few code-reviews with agents and they say this is the only code changes needed to fix the original bug and everything else is out of scope.

diegolmello (changes requested):

index.js
```diff
@@ -23,7 +23,11 @@ if (process.env.USE_STORYBOOK) {
 
 	LogBox.ignoreAllLogs();
 
-	if (Platform.OS === 'android' && DeviceInfo.hasSystemFeatureSync('android.software.telecom')) {
+	// android.software.telecom is only declared from API 33, so querying it skipped setup() on older devices (#7334)
+	const supportsTelecom =
+		Platform.OS === 'android' && (Number(Platform.Version) < 33 || DeviceInfo.hasSystemFeatureSync('android.software.telecom'));
```
diegolmello: This is now treating every pre-33 device as Telecom-compatible.

index.js
```diff
@@ -23,7 +23,11 @@ if (process.env.USE_STORYBOOK) {
 
 	LogBox.ignoreAllLogs();
 
-	if (Platform.OS === 'android' && DeviceInfo.hasSystemFeatureSync('android.software.telecom')) {
+	// android.software.telecom is only declared from API 33, so querying it skipped setup() on older devices (#7334)
```
diegolmello: This comment gets outdated as soon as this PR merges. Remove.

diegolmello (changes requested):

diegolmello (approved):

PR #7581, opened 2026-08-19
fix: crash in ShareView when a shared file no longer exists

Rohit3523:
## Proposed changes

This PR fixes two issues:

1. Sharing files such as audio, HTML, and ZIP files causes the app to crash instantly.
2. In rare cases, when a file has an invalid URI, Expo fails to locate the file, which triggers the `Cannot set property 'canUpload' of null` error.

I have added an E2E test for file sharing that shares different types of files to ensure that our app can successfully open the Share View.

Android CI Run: https://github.com/RocketChat/Rocket.Chat.ReactNative/actions/runs/35013837739/job/104560624890?pr=7581

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1193

### Steps to test fix 1

1. Share an audio file from the Files app to Rocket.Chat.
2. The app should open and display the Share View, allowing you to select a channel without any errors or crashes.

### Steps to test fix 2

For this case, we need to use an emulator to send an invalid file URI.

1. Based on the OS, run the following command:
- Android - `adb shell am start -a android.intent.action.VIEW -d 'rocketchat://shareextension?mediaUris=file:///nonexistent.jpg' chat.rocket.android`
- iOS - `xcrun simctl openurl booted 'rocketchat://shareextension?mediaUris=file:///nonexistent.jpg'`

2. The app should display **"The shared file could not be found"** and navigate back to the Rooms List without crashing.

## Screenshots
| OS | Before | After |
| --- | --- | --- |
| Android | <video src="https://github.com/user-attachments/assets/c104d4a4-75e4-4d6a-b6ff-8f6c776b60fa" width="100%" controls></video> | <video src="https://github.com/user-attachments/assets/938150af-91c5-4f6f-bb26-7bb8328650ec" width="100%" controls></video> |
| iOS | <video src="https://github.com/user-attachments/assets/7cc937b7-0121-4922-b56e-f3c61bd58ab0" width="100%" controls></video> | <video src="https://github.com/user-attachments/assets/c9d82d9b-c8c3-4dd6-b399-f6a4fff2e4a3" width="100%" controls></video> |
| Android (Sharing audio file) | <video src="https://github.com/user-attachments/assets/560c44bc-0a95-4b87-8ef7-38b5513ee8d1" width="100%" controls></video> | <video src="https://github.com/user-attachments/assets/2c782052-ed70-4974-98d8-df63b3e3fe0b" width="100%" controls> |

## Types of changes

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Sharing now prevents navigation when media files are unavailable or invalid and displays a localized message.
  - Invalid attachment entries are safely ignored.
  - File previews handle unexpected file-type information without errors.
  - Long shared text is handled as a file when needed.
  - Shared contact files and audio attachments are handled correctly.

- **Localization**
  - Added the missing “shared file could not be found” message across supported languages.

- **Tests**
  - Expanded coverage for attachment validation, sharing behavior, navigation, and file previews.


Rohit3523:
@coderabbitai review

diegolmello (commented):
How can a user reproduce the share of a file that doesn't exist?
On the PR body, you mention scripts, but that's an edge case, right?

Rohit3523:
> How can a user reproduce the share of a file that doesn't exist? On the PR body, you mention scripts, but that's an edge case, right?

The file may still exist for the user, but our app might not be able to access it because it’s in a path we don’t have access to, was removed by the source app or the source app provided an invalid URI.

The deeplink script is just a way to reproduce this condition consistently by intentionally passing an invalid file URI.

diegolmello (changes requested):
I'm not convinced this is a valid fix. 
We found this on bugsnag, so it's something that actually happened to the users. 
There's a higher chance of one app like WhatsApp not sharing the images the way we are expecting on our share extension and causing the bug.

If we just merge the way it is, we are going to be hiding the error from bugsnag, but the users will still see it happening.

Rohit3523:
@coderabbitai review

app/views/ShareListView/ShareListView.test.tsx
```diff
@@ -0,0 +1,120 @@
+jest.mock('../../lib/database', () => ({
```
OtavioStasiak: app/views/ShareListView/ShareListView.test.tsx:1,12,22 and app/views/ShareView/Preview.test.tsx:13 — module specifiers
     '../../lib/database', '../../lib/methods/helpers/showToast', '../../containers/ImageViewer'. .oxlintrc.json's
     no-restricted-imports bans the ../../** group ("Use a ~/ alias for imports two or more directories up"); the rule doesn't
     fire only because the strings sit inside jest.mock/require. The neighbouring app/views/ShareView/ShareView.test.tsx:17-44
     mocks the same modules as ~/lib/database, ~/lib/navigation/appNavigation, etc.

ios/ShareRocketChatRN/ShareRocketChatRN.swift
```diff
@@ -12,6 +12,45 @@ import MobileCoreServices
...
+
+    private func uniqueFileURL(in groupURL: URL, filename: String) -> URL {
+        var dest = groupURL.appendingPathComponent(sanitizeFileName(filename))
+        while FileManager.default.fileExists(atPath: dest.path) {
+            dest = groupURL.appendingPathComponent("\(UUID().uuidString)-\(sanitizeFileName(filename))")
+        }
+        return dest
+    }
+
+    // ponytail: copy, don't Data(contentsOf:) — extension limit ~120MB is RAM, not file size
```
OtavioStasiak: ios/ShareRocketChatRN/ShareRocketChatRN.swift:37 and android/.../share/ShareActivity.kt:58 — // ponytail: … comments. These
     two are the only occurrences of "ponytail" anywhere in the repo; it's an agent-tooling marker, not a code comment. Worse,
     the Kotlin one is factually wrong: it claims "Bundle.get has no deprecated overloads", but BaseBundle.get(String) carries
     @Deprecated since API 33 (confirmed in the android-35 platform sources), and the code calls exactly that
     (intent.extras?.get(...)) at compileSdk 36. Per your standing rule against speculative explanatory comments, both should go
     — or the Kotlin one be replaced with the real reason for avoiding the typed getters.

OtavioStasiak (changes requested):
.sniffler/test-map.json:67-78 — the new entry's dependsOn lists only JS views, yet the branch's substance is
     ShareActivity.kt and ShareRocketChatRN.swift. Precedent for native globs exists in that file (line 135, the biometric
     entry), so this E2E won't be selected when the native code it actually exercises changes.

.maestro/tests/share-extension/utils/share-file.yaml
```diff
@@ -0,0 +1,104 @@
+appId: ${APP_ID}
+name: Shre Extension
```
OtavioStasiak: .maestro/tests/share-extension/utils/share-file.yaml:2 — name: Shre Extension typo. Also duplicated flow names:
     download-file.yaml:2 and download-files.yaml:2 are both Download File; verify-download.yaml:2 and download-and-verify.yaml:2
     are both Verify Download. Shard logs identify flows by name.

.maestro/tests/share-extension/utils/share-file.yaml
```diff
@@ -0,0 +1,104 @@
...
+tags:
+  - 'util'
+
+---
```
OtavioStasiak, line 6: .maestro/tests/share-extension/share-extension-deeplink.yaml:12 — createRandomRoom(user, pass, output.user.username) passes
     a username into the type param (data-setup.js:113). Every other call site passes nothing, 'p', or 'c'. It only works because
     anything ≠ 'c' falls through to groups.create — and the new roomMessageCount (data-setup.js:223) hardcodes groups.messages,
     so the assertion silently depends on that accident.

.maestro/tests/share-extension/utils/download-file.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - runFlow:
+          when:
+            visible:
+              text: '.*Type English and Hindi'
+          commands:
+            - tapOn:
+                text: 'Continue'
+                index: 1
+
+- inputText: https://mobile.qa.rocket.chat/file-upload/${file}?rc_uid=${output.login.userId}&rc_token=${output.login.authToken}&download=1
```
OtavioStasiak: .maestro/tests/share-extension/utils/download-file.yaml:42 — hardcodes https://mobile.qa.rocket.chat; .maestro/README.md
     makes data.server the single source and data.js:2 already holds it. Relatedly download-files.yaml:36-88 bakes in 12 fixed
     upload ids against "use random values instead".

.maestro/tests/share-extension/utils/download-and-verify.yaml
```diff
@@ -0,0 +1,20 @@
...
+      file: ${file}
+
+- runFlow:
+    when:
+      platform: iOS
+    commands:
+      - runFlow:
+          file: 'verify-download.yaml'
+          env:
+            file: ${file}
```
OtavioStasiak: .maestro/tests/share-extension/utils/download-and-verify.yaml:20 — no trailing newline; all five siblings have one.

.sniffler/test-map.json
```diff
@@ -64,6 +64,19 @@
 			"app/sagas/inviteLinks.js"
 		]
 	},
+	{
```
OtavioStasiak: .sniffler/test-map.json:68 — entry inserted between assorted/deeplink.yaml and assorted/delete-server.yaml, breaking the
     by-directory grouping every other entry follows.

OtavioStasiak (changes requested):

.maestro/scripts/data-setup.js
```diff
@@ -220,6 +220,20 @@ const deleteCreatedUsers = () => {
     }
 };
 
+const roomMessageCount = (username, password, roomId) => {
```
OtavioStasiak: .maestro/scripts/data-setup.js:24 roomMessageCount — it calls groups.messages, which only serves a Group (private Room,
     type 'p'). CONTEXT.md draws a hard Room/Channel/Group distinction; a Room-named helper that silently returns 0 for a
     Channel or DM is exactly the conflation the glossary forbids.

.github/scripts/run-maestro.sh
```diff
@@ -112,6 +112,10 @@ done < "$MAPFILE"
 echo "Main run will execute:"
 printf '  %s\n' "${FLOW_FILES[@]}"
 
+if [ "$SHARD" = "9" ] && [ "$PLATFORM" = "android" ]; then
```
OtavioStasiak: .github/scripts/run-maestro.sh:115 — hardcoded SHARD = 9 && PLATFORM = android with no back-reference from the flow, and a
     bare node scripts/push-downloads-to-sim.js under set -euo pipefail. Every other prerequisite in this file emits a
     classified ::error title=…::… environment failure, not an app or test regression with a distinct exit code; a
     fixture-download failure will instead kill the shard with a raw stack trace.

app/views/ShareListView/index.tsx
```diff
@@ -216,7 +237,7 @@ class ShareListView extends Component<IShareListViewProps, IState> {
 			headerRight: () =>
 				this.airGappedReadOnly ? null : (
```
OtavioStasiak, line 238: Dead test hooks — testID='share-list-search' / 'share-list-search-button' (ShareListView/index.tsx:224,240) are referenced
     by no flow or unit test.

OtavioStasiak (changes requested):

OtavioStasiak (approved):
LGTM

PR #7588, opened 2026-08-20
feat: floating date separator

OtavioStasiak:
## Proposed changes

 **1. Date separator design.** The date is now rendered as a pill (`buttonBackgroundSecondaryDefault` background,
  `buttonFontSecondary` text, `borderRadius: 16`) between the two divider lines, matching web's `<Bubble small
  secondary>` and the design applied in #7567. Only the styling is ported — dates keep the existing `LL` format, with
  no ordinal suffix, so `app/lib/dayjs` is untouched. The unread separator states are unchanged.

  **2. Floating date separator.** Like web, a date bubble now floats at the top of the message list: it fades in while
  you scroll, tracks the topmost visible message so it updates as you cross day boundaries, and fades out ~1s after
  the list settles. It reuses the exact same pill component as the inline separator, so the two are visually
  identical.

 ## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1491

## How to test or reproduce

- Open the app;
- Go to room;
- Scroll up to older messages;
- Open thread with many messages;
- Scroll up;

### Separator

| Android | iOS |
| ----- | ----- |
| Screenshot_1787588691 | Simulator Screenshot - iPhone 16 - 2026-08-24 at 13 24 57 |

### RoomView

| Android | iOS |
| ----- | ----- |
| android1 | ios2 |

### Threads

| Android | iOS |
| ----- | ----- |
| threadandroid | threadios |

## Types of changes

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added themed date separators with localized date labels and divider styling in message lists.
  - Added a floating date indicator that appears while scrolling through room messages and fades smoothly.
  - Improved date separator consistency across message views and loading states.

- **Bug Fixes**
  - Refined floating date visibility behavior to prevent overlapping animations and ensure delayed hiding after scrolling.


app/containers/DateSeparator.tsx
```diff
@@ -0,0 +1,57 @@
...
+		paddingHorizontal: 8,
+		paddingVertical: 2,
+		borderRadius: 16
+	},
+	marginHorizontal: {
+		marginHorizontal: 14
+	}
+});
+
+// The date pill itself, shared with the floating date bubble so both look exactly the same.
```
diegolmello: Remove

app/containers/DateSeparator.tsx
```diff
@@ -0,0 +1,57 @@
...
+		fontSize: 14,
+		...sharedStyles.textMedium
+	},
+	label: {
+		paddingHorizontal: 8,
+		paddingVertical: 2,
+		borderRadius: 16
+	},
+	marginHorizontal: {
+		marginHorizontal: 14
```
diegolmello: multiple of 4

app/views/RoomView/List/components/FloatingDateSeparator.tsx
```diff
@@ -0,0 +1,33 @@
...
+	container: {
+		position: 'absolute',
+		top: BUBBLE_TOP_MARGIN,
+		left: 0,
+		right: 0,
+		alignItems: 'center'
+	}
+});
+
+// Floating date bubble, like web's: fades in while the list scrolls and out once it settles.
```
diegolmello: Remove

app/views/RoomView/List/components/List.tsx
```diff
@@ -26,13 +28,43 @@ const styles = StyleSheet.create({
...
+			return;
+		}
+		isBubbleFadingIn.set(true);
+		bubbleOpacity.set(
+			withTiming(1, { duration: 150 }, (): void => {
+				isBubbleFadingIn.set(false);
+				hideBubble();
+			})
+		);
+	};
```
diegolmello: I feel like this logic shouldn't be inside RoomView. They should go inside the floating separator.

app/containers/Separator/FloatingDateSeparator.tsx
diegolmello, line 1: This could go to a `containers/Separator` folder with DateSeparator, MessageSeparator and FloatingDateSeparator.

diegolmello (changes requested):
Missing unread separator UI on pr body

app/containers/Separator/useFloatingDateOpacity.ts
```diff
@@ -0,0 +1,42 @@
...
+			return;
+		}
+		isFadingIn.set(true);
+		opacity.set(
+			withTiming(1, { duration: FADE_IN_DURATION }, (): void => {
+				isFadingIn.set(false);
+				hide();
+			})
+		);
+	};
```
diegolmello: **The fade-out is anchored to fade-in completion, not to the last scroll.**

`hide()` is scheduled from the fade-in completion callback, and `show()` only re-arms it when `opacity.get() === 1`. During the 150ms fade-in the `isFadingIn` guard drops every scroll event. So if a scroll continues past that window, the 1s delay stays anchored to when the fade-in finished — the bubble can start fading out while the user is still scrolling.

The intended behaviour is a debounce off the last scroll event. Worth confirming on a device rather than by reading the code.

Two things that are *not* problems, for the record: the `opacity.get() === 1` exact float compare is safe (reanimated snaps `withTiming` to `toValue` on finish, and opacity sits at exactly 1 through the `withDelay`), and re-entering mid-fade-out is a 150ms direction reversal rather than a restart from zero — the only artifact there is inconsistent apparent velocity.

app/containers/Separator/useFloatingDateOpacity.ts
```diff
@@ -0,0 +1,42 @@
...
+const HIDE_DELAY = 1000;
+const FADE_IN_DURATION = 150;
+const FADE_OUT_DURATION = 300;
+
+interface IUseFloatingDateOpacity {
+	opacity: SharedValue<number>;
+	show: () => void;
+}
+
+export const useFloatingDateOpacity = (): IUseFloatingDateOpacity => {
```
diegolmello: This hook has exactly one consumer (`List.tsx`) and is useless without `useFloatingDate`, which lives in `RoomView/List/hooks/`. One behaviour split across two trees means any change to the fade edits both.

Consider merging into `useFloatingDate`, returning `{ ts, opacity, show, viewabilityConfigCallbackPairs }`, and deleting this file. That also collapses the 7 wiring points currently in `List.tsx` (lines 13, 17, 31, 32, 41, 77, 79).

app/containers/Separator/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as DateSeparator, DateSeparatorLabel } from './DateSeparator';
+export { default as FloatingDateSeparator } from './FloatingDateSeparator';
+export { default as MessageSeparator } from './MessageSeparator';
+export { useFloatingDateOpacity } from './useFloatingDateOpacity';
```
diegolmello: The barrel publishes more than anything consumes: `useFloatingDateOpacity` has one consumer, and `DateSeparatorLabel`'s only consumers are in-folder (`DateSeparator.tsx`, `FloatingDateSeparator.tsx`). Worth trimming both.

Keeping the `SharedValue<number>` prop on `FloatingDateSeparator` is the right call though — the caller is a worklet inside `useAnimatedScrollHandler`, so a ref-based `show()` would need `scheduleOnRN` and push the animation back to the JS thread.

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,43 @@
...
+import { type TAnyMessageModel } from '../../../../definitions';
+
+type TViewabilityConfigCallbackPairs = NonNullable<FlatListProps<TAnyMessageModel>['viewabilityConfigCallbackPairs']>;
+
+interface IUseFloatingDate {
+	ts: Date | string | null;
+	viewabilityConfigCallbackPairs: TViewabilityConfigCallbackPairs;
+}
+
+// The list is inverted, so the topmost visible row is the one with the highest index.
```
diegolmello: Per the repo comment convention, a comment that explains what the code does is a signal the name is hiding the logic — this function's whole job is *maximize index*. Renaming to something like `getHighestIndexViewableTs` lets the comment go.

(Same comment is duplicated in the test file.)

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,43 @@
...
+export const getTopVisibleTs = (viewableItems: ViewToken<TAnyMessageModel>[]): Date | string | null => {
+	let top: ViewToken<TAnyMessageModel> | undefined;
+	viewableItems.forEach(token => {
+		if (!token.isViewable || !token.item?.ts || token.index == null) {
+			return;
+		}
+		if (!top || token.index > top.index!) {
+			top = token;
+		}
+	});
```
diegolmello: Hand-rolled max-by forces the non-null assertion on `top.index!` at line 20, even though `token.index == null` is already guarded on line 17. Accumulating `{ index, ts }` instead of the token — or a `reduce` / filter+max — drops the assertion.

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,43 @@
...
+	return top?.item.ts || null;
+};
+
+export const useFloatingDate = (): IUseFloatingDate => {
+	const [ts, setTs] = useState<Date | string | null>(null);
+
+	const [viewabilityConfigCallbackPairs] = useState<TViewabilityConfigCallbackPairs>(() => [
```
diegolmello, line 44: `useState` with no setter — `useRef(...).current` says the intent (this never changes) more directly.

Worth keeping it non-reactive deliberately: RN throws `Changing viewabilityConfigCallbackPairs on the fly is not supported` (`invariant` in `FlatList.js`), so this identity must stay stable. That invariant shouldn't rest on the React Compiler, which could bail later.

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,43 @@
...
+		{
+			viewabilityConfig: { itemVisiblePercentThreshold: 0 },
+			onViewableItemsChanged: ({ viewableItems }) => {
+				const next = getTopVisibleTs(viewableItems);
+				if (next) {
+					setTs(next);
+				}
```
diegolmello, line 57: `setTs` is only called when `next` is truthy, so `ts` is sticky for the list's lifetime. On a room switch, or a viewability report with no dated rows, the next fade-in shows the stale date before the new one lands. Tracking the topmost message should include tracking to `null`.

Separately: a day-key bail here (`dayjs(next).format('L')` matching the previous key) would cut renders from per-topmost-message to per-day-boundary. Note this is *not* about `Date` identity — WatermelonDB's `@date` decorator memoizes `Date` objects by raw ms, so `item.ts` is already referentially stable and `useState` bails on repeats. The win is purely fewer distinct writes. `minimumViewTime` alongside `itemVisiblePercentThreshold: 0` may help too.

app/views/RoomView/List/components/List.tsx
```diff
@@ -26,13 +28,17 @@ const styles = StyleSheet.create({
 const List = ({ listRef, jumpToBottom, isAnchored, ...props }: IListProps) => {
 	const [scrolledPastLimit, setScrolledPastLimit] = useState(false);
 	const { isAutocompleteVisible } = useRoomContext();
+	const { ts, viewabilityConfigCallbackPairs } = useFloatingDate();
+	const { opacity: bubbleOpacity, show: showBubble } = useFloatingDateOpacity();
```
diegolmello: "bubble" is invented here and appears nowhere else in the codebase — CONTEXT.md names the concept *Date Separator*. Suggest `floatingDateOpacity` / `showFloatingDate` (and `floatingDateOpacity` at line 79).

app/views/RoomView/List/components/List.tsx
```diff
@@ -68,7 +74,9 @@ const List = ({ listRef, jumpToBottom, isAnchored, ...props }: IListProps) => {
 				onScroll={scrollHandler}
 				{...props}
 				{...scrollPersistTaps}
+				viewabilityConfigCallbackPairs={viewabilityConfigCallbackPairs}
```
diegolmello, line 85: `viewabilityConfigCallbackPairs` sits after `{...props}`, and `IListProps extends FlatListProps`, so a caller passing its own pairs would have them silently dropped with no type error. Inert today — no caller passes one — but cheap to make safe:

```tsx
viewabilityConfigCallbackPairs={[...(props.viewabilityConfigCallbackPairs ?? []), ...viewabilityConfigCallbackPairs]}
```

app/containers/Separator/MessageSeparator.tsx
```diff
@@ -1,11 +1,12 @@
...
-import dayjs from '../lib/dayjs';
-import I18n from '../i18n';
-import sharedStyles from '../views/Styles';
-import { themes } from '../lib/constants/colors';
-import { useTheme } from '../theme';
+import dayjs from '../../lib/dayjs';
+import I18n from '../../i18n';
+import sharedStyles from '../../views/Styles';
+import { themes } from '../../lib/constants/colors';
+import { useTheme } from '../../theme';
```
diegolmello, line 7: This file uses `const { theme } = useTheme()` + `themes[theme]` (lines 34, 41-42), while the new sibling `DateSeparator.tsx` uses `const { colors } = useTheme()` — the repo convention. Pre-existing behaviour, but the file moved into this new folder in this PR, so it's now two conventions side by side in one directory.

app/containers/Separator/MessageSeparator.tsx
```diff
@@ -48,14 +49,8 @@ const MessageSeparator = ({ ts, unread }: { ts?: Date | string | null; unread?:
...
 	}
-	if (ts) {
-		return (
-			<View style={styles.container}>
-				<View style={[styles.line, { backgroundColor: themes[theme].strokeLight }]} />
-				<Text style={[styles.text, { color: themes[theme].fontSecondaryInfo }, styles.marginHorizontal]}>{date}</Text>
-				<View style={[styles.line, { backgroundColor: themes[theme].strokeLight }]} />
-			</View>
-		);
+	if (ts && date) {
```
diegolmello: `date` is derived from `ts` on line 40, so `if (ts && date)` is just `if (ts)` — the `&& date` implies an independent failure mode that doesn't exist.

app/containers/Separator/DateSeparator.tsx
```diff
@@ -0,0 +1,56 @@
...
+const styles = StyleSheet.create({
+	container: {
+		flexDirection: 'row',
+		alignItems: 'center',
+		margin: 12
+	},
+	line: {
+		height: 1,
+		flex: 1
+	},
```
diegolmello, line 12: `container` and `line` here are byte-identical to `MessageSeparator.tsx:12-24` in the same new folder; `text` differs only in `textBold` vs `textMedium`. Worth extracting to a shared `app/containers/Separator/styles.ts`.

If that extraction uses `gap: 12`, it also deletes both `marginHorizontal` wrappers and `marginRight` — at the cost of one visual delta, the unread branch's spacing going 14 - 12.

app/containers/Separator/FloatingDateSeparator.tsx
```diff
@@ -0,0 +1,33 @@
...
+const FloatingDateSeparator = ({ ts, opacity }: { ts?: Date | string | null; opacity: SharedValue<number> }) => {
+	const style = useAnimatedStyle(() => ({ opacity: opacity.get() }));
+
+	if (!ts) {
+		return null;
+	}
+
+	return (
+		<Animated.View pointerEvents='none' style={[styles.container, style]}>
+			<DateSeparatorLabel date={dayjs(ts).format('LL')} />
```
diegolmello: `dayjs(ts).format('LL')` is duplicated here and at `MessageSeparator.tsx:40`. Exporting a `formatSeparatorDate` helper from `DateSeparator.tsx` and having `DateSeparatorLabel`/`DateSeparator` take `ts` instead of a preformatted string would give both callers one source of truth.

app/containers/Separator/DateSeparator.tsx
```diff
@@ -0,0 +1,56 @@
...
+	text: {
+		fontSize: 14,
+		...sharedStyles.textBold
```
diegolmello, line 15: This switches every date separator from `textMedium` (current `develop`) to `textBold`. Defensible under matching web's `<Bubble small secondary>`, but the PR description says "Only the styling is ported" without calling out the weight change, and the design precedent in #7567 didn't include it. Worth a line in the description either way.

diegolmello (commented):
Two-axis review against `develop` — **Standards** (repo conventions + code smells) and **Spec** (PR description and NATIVE-1491). Comments inline.

The one worth looking at first is the fade-out timing in `useFloatingDateOpacity.ts` — it's the only finding that is user-visible, and it wants a device check rather than a code read. Everything else is cleanup or naming.

For the record, memoization suggestions were checked against the React Compiler and dropped: `compilationMode: 'infer'` is on, and all four touched files compile without bailing (worklets survive intact). The one carve-out is `viewabilityConfigCallbackPairs`, which must stay stable because of an RN invariant, not for performance.

app/containers/Separator/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as FloatingDateSeparator } from './FloatingDateSeparator';
+export { default as MessageSeparator } from './MessageSeparator';
```
diegolmello: No barrel files

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,76 @@
...
+		}
+	]);
+
+	const hide = (): void => {
+		'worklet';
+
+		opacity.set(withDelay(HIDE_DELAY, withTiming(0, { duration: FADE_OUT_DURATION })));
+	};
+
+	const show = (): void => {
```
diegolmello: **Severity: High** - the native scroll handler is re-registered on every render.

`show` and `hide` are re-created on every render, so the worklet passed to `useAnimatedScrollHandler` has a fresh identity each time. With no dependency array, reanimated 4.1.3 derives dependencies from the worklet's `__closure` and compares them by identity (`areWorkletsEqual` in `src/hook/utils.ts`), so the native scroll handler is unregistered and re-registered on every render of `List`.

Steps to reproduce:

1. Open a room and start scrolling.
2. Have another user send messages to that room while you scroll, so `List` re-renders during the gesture.
3. Each re-render tears down and re-attaches the native handler. The visible symptom is intermittent: the pill occasionally fails to fade during a scroll, and scroll frames can be dropped.

Fix: wrap `show` and `hide` in `useCallback` keyed on the shared values, so the worklet identity is stable.

app/views/RoomView/List/components/List.tsx
```diff
@@ -26,13 +28,16 @@ const styles = StyleSheet.create({
...
+	const { ts, opacity: floatingDateOpacity, show: showFloatingDate, viewabilityConfigCallbackPairs } = useFloatingDate();
+
 	const scrollHandler = useAnimatedScrollHandler({
 		onScroll: event => {
 			if (event.contentOffset.y > SCROLL_LIMIT) {
 				scheduleOnRN(setScrolledPastLimit, true);
 			} else {
 				scheduleOnRN(setScrolledPastLimit, false);
 			}
+			showFloatingDate();
```
diegolmello: **Severity: High** - the pill fades in on scrolls the user did not perform.

`showFloatingDate()` is called unconditionally from `onScroll`, which also fires for `scrollToOffset` / `scrollToIndex` (jump to bottom, jump to message) and for `maintainVisibleContentPosition` autoscroll at the live tail. So a date pill fades in over the newest messages when the user is not scrolling at all.

Steps to reproduce:

1. Open a room containing messages from more than one day.
2. Stay parked at the bottom, at the live tail. Do not touch the screen.
3. Have another user send a message.
4. A date pill fades in over the newest messages.
5. Separately, tap the jump-to-bottom FAB from further up the history. Same flash.

Fix: gate the call on a "user is dragging" flag set from `onScrollBeginDrag` and `onMomentumScrollBegin`. Clear it on `onScrollEndDrag` as well as `onMomentumScrollEnd`: a drag released at near-zero velocity emits no momentum event on either platform, so clearing only on momentum leaves the flag stuck true and the flash comes back intermittently. All four handler keys must be present on the same `useAnimatedScrollHandler`, since it subscribes native events by key presence, and none may be conditional.

Note this is a deliberate improvement over web rather than parity with it. Web calls its equivalent from a plain `onScroll` with no drag gate (`apps/meteor/client/views/room/MessageList/MessageList.tsx`), but web's list has no `maintainVisibleContentPosition` autoscroll hammering the handler at the tail.

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,76 @@
...
+	const opacity = useSharedValue(0);
+	const isFadingIn = useSharedValue(false);
+
+	const { current: viewabilityConfigCallbackPairs } = useRef<TViewabilityConfigCallbackPairs>([
+		{
+			viewabilityConfig: { itemVisiblePercentThreshold: 0 },
+			onViewableItemsChanged: ({ viewableItems }) => {
+				const next = getHighestIndexViewableTs(viewableItems);
+				const nextDayKey = next ? dayjs(next).format('L') : null;
+				if (nextDayKey === dayKey.current) {
```
diegolmello, line 55: **Severity: Medium** - the pill blinks off and returns at full opacity with no fade.

When `viewableItems` arrives empty the hook sets `ts` to null, which unmounts the overlay while `opacity` is still near 1. `opacity` lives in the hook, so it survives the unmount and the next date remounts at full opacity with no fade-in.

`ViewabilityHelper` can emit an empty `viewableItems` with a non-empty list when the render window does not overlap the viewport, which happens on a fast fling. Note every row carries a `ts` (loader rows are stamped with one in `loadMessagesForRoom` and `loadSurroundingMessages`), so an empty batch is the only way this triggers.

Steps to reproduce:

1. Open a room with several hundred messages spanning multiple days.
2. Fling the list upward hard, fast enough to outrun cell rendering and show blank space.
3. The pill vanishes mid-fling and snaps back at full opacity instead of fading.

Fix: hold the last known date instead of clearing it, and let `opacity` alone decide visibility. Web does exactly this and says why: "We always keep the previous date if we don't have a new one, so when the bubble disappears it doesn't flicker" (`useDateScroll.ts`).

app/containers/Separator/FloatingDateSeparator.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+const FloatingDateSeparator = ({ ts, opacity }: { ts?: Date | string | null; opacity: SharedValue<number> }) => {
+	const style = useAnimatedStyle(() => ({ opacity: opacity.get() }));
+
+	if (!ts) {
+		return null;
+	}
+
+	return (
+		<Animated.View pointerEvents='none' style={[styles.container, style]}>
```
diegolmello: **Severity: Medium** - the floating pill is not hidden from screen readers.

`pointerEvents='none'` governs touch hit-testing only. The view stays in the accessibility tree, and the `Text` inside `DateSeparatorLabel` is accessible by default, so the date becomes a swipe stop floating over the message list.

This also defeats an existing accommodation: `List` deliberately hides the FlatList from screen readers while the composer autocomplete is open, and this overlay is a sibling of that FlatList, so it stays exposed and floats above the autocomplete.

Steps to reproduce:

1. Enable VoiceOver on iOS or TalkBack on Android.
2. Open a room with messages from more than one day and scroll until the pill is visible.
3. Swipe through elements. The floating date is announced as its own stop, on top of the message list.
4. Open the composer autocomplete. The message list is correctly hidden from the screen reader, but the floating date is still reachable above it.

Fix: add `accessibilityElementsHidden` (iOS) and `importantForAccessibility='no-hide-descendants'` (Android) to this `Animated.View`. Put them here and not on `DateSeparatorLabel`: the inline `DateSeparator` shares that label, and there the date should stay announced, since it is the only marker of a day change in the reading order.

app/containers/Separator/DateSeparator.tsx
```diff
@@ -0,0 +1,46 @@
...
+	},
+	label: {
+		paddingHorizontal: 8,
+		paddingVertical: 2,
+		borderRadius: 16
+	}
+});
+
+export const formatSeparatorDate = (ts: Date | string): string => dayjs(ts).format('LL');
```
diegolmello: **Severity: Medium** - English dates are missing the ordinal day, and this conflicts with #7567.

`dayjs(ts).format('LL')` renders "November 9, 2017". Web renders "November 9th, 2017": its date divider formats through date-fns `PPP`, and date-fns 4.1.0 defines the en-US long format as `"MMMM do, y"`, where `do` is the ordinal day. dayjs `LL` never produces an ordinal.

#7567 already fixes this by adding `formatLongDate` to `app/lib/dayjs`, which applies `MMMM Do, YYYY` for ordinal-day locales and `LL` elsewhere. That PR modifies `app/containers/MessageSeparator.tsx`, which this PR deletes, so the two collide, and if this one merges second the ordinal fix is silently reverted on the separator it was written for.

Steps to reproduce:

1. Set the app language to English.
2. Open a room with messages from more than one day.
3. The separator reads "November 9, 2017". Compare against the same room on web, which reads "November 9th, 2017".

Fix: use a single shared formatter in `app/lib/dayjs` rather than a local one here, and settle the merge order with #7567 so one of the two supersedes the other explicitly.

app/containers/Separator/DateSeparator.tsx
```diff
@@ -0,0 +1,46 @@
...
+	return (
+		<View style={[styles.label, { backgroundColor: colors.buttonBackgroundSecondaryDefault }]}>
+			<Text style={[styles.text, { color: colors.buttonFontSecondary }]}>{formatSeparatorDate(ts)}</Text>
+		</View>
+	);
+};
+
+const DateSeparator = ({ ts }: { ts: Date | string }): ReactElement => {
+	const { colors } = useTheme();
+	const lineStyle = { backgroundColor: colors.strokeLight };
```
diegolmello: **Severity: Low** - divider line colour does not match web.

`strokeLight` is `#CBCED1`. Web's divider rule is `stroke-extra-light` `#ebecef`, and the palette already has `strokeExtraLight: '#EBECEF'`, an exact match.

Steps to reproduce:

1. Open a room with messages from more than one day, in light theme.
2. Compare the horizontal rule either side of the date against web. The rule here is visibly darker.

Fix: `colors.strokeExtraLight`.

app/views/RoomView/List/definitions.ts
```diff
@@ -8,7 +8,7 @@ export type TListRef = RefObject<FlatList<TAnyMessageModel> | null>;
 
 export type TMessagesIdsRef = RefObject<string[]>;
 
-export interface IListProps extends FlatListProps<TAnyMessageModel> {
+export interface IListProps extends Omit<FlatListProps<TAnyMessageModel>, 'onScroll' | 'viewabilityConfigCallbackPairs'> {
```
diegolmello: **Severity: Low** - only two of the four viewability props are omitted.

`FlatList` asserts `!(onViewableItemsChanged && viewabilityConfigCallbackPairs)`, and it also asserts that `viewabilityConfigCallbackPairs` never changes identity. Since `List` now owns `viewabilityConfigCallbackPairs`, a caller passing `onViewableItemsChanged` or `viewabilityConfig` would type-check and then throw at runtime.

No caller does this today, so this is prevention rather than a live bug.

Fix: omit the whole family: `'onScroll' | 'viewabilityConfigCallbackPairs' | 'onViewableItemsChanged' | 'viewabilityConfig'`.

diegolmello (changes requested):
Reviewed against web `develop` (`e3b6f3db5c`) and Fuselage 0.87.0, which is what `apps/meteor/package.json` pins.

The feature works and the structure is good. Seven things below, two of which I consider blocking: the scroll handler is torn down and re-attached on every render, and the pill fades in on scrolls the user did not perform.

Three corrections to the PR description:

1. "The unread separator states are unchanged" is not accurate. The shared container replaced `marginRight: 14` / `marginHorizontal: 14` with `gap: 12`, so unread separator spacing changed too. Visible in the `Message.test.tsx.snap` diff.
2. The description does not mention that the inline Date Separator is restyled app-wide: font weight 500 to 700, text colour `fontSecondaryInfo` to `buttonFontSecondary`, and a new `buttonBackgroundSecondaryDefault` pill background. That is a bigger visual change than the new overlay and reviewers should be told.
3. "It fades in while you scroll" is true of any scroll, including programmatic ones. See the comment on `List.tsx`.

Related: #7567 is open and touches the same component and the same date format. Whichever of the two merges second wins. Worth settling the order before either lands.

app/containers/Separator/styles.ts
```diff
@@ -0,0 +1,14 @@
+import { StyleSheet } from 'react-native';
+
+export const separatorStyles = StyleSheet.create({
+	container: {
+		flexDirection: 'row',
+		alignItems: 'center',
+		margin: 12,
+		gap: 12
```
diegolmello: **Blocking: shared styles unintentionally change Unread Separator spacing.**

`gap: 12` applies to every separator using this container. Both Unread Separator variants previously used 14-point spacing, through `marginHorizontal: 14` for the combined variant and `marginRight: 14` for the unread-only variant. The snapshot diff confirms both values were replaced.

Keep only layout common to every separator in `separatorStyles`. Give Date Separator its new 12-point spacing while preserving the existing Unread Separator spacing. The shared abstraction should not couple unrelated visual policies.

app/views/RoomView/List/hooks/useFloatingDate.ts
```diff
@@ -0,0 +1,81 @@
...
+			return top;
+		}
+		return !top || index > top.index ? { index, ts: item.ts } : top;
+	}, null)?.ts ?? null;
+
+export const useFloatingDate = (): IUseFloatingDate => {
+	const [ts, setTs] = useState<Date | string | null>(null);
+	const dayKey = useRef<string | null>(null);
+	const opacity = useSharedValue(0);
+	const isFadingIn = useSharedValue(false);
```
diegolmello: **High: the Floating Date Separator lifecycle is split between the list and this hook.**

`List.tsx` owns user-scroll state and gesture transitions, while this hook owns fade state, delayed hiding, and an imperative `show` API. That makes the list understand feature-specific animation rules, and the tests only exercise date selection rather than the stateful behavior.

Move the complete lifecycle into `useFloatingDate`. A restartable sequence such as `withSequence(withTiming(...), withDelay(...))` can replace `isFadingIn`, `hide`, and the opacity branches. Then test the animation restart and gesture transitions through the hook's public contract.

app/lib/dayjs/index.ts
```diff
@@ -36,6 +37,15 @@ dayjs.extend(timezone);
 dayjs.extend(calendar);
 dayjs.extend(relativeTime);
 dayjs.extend(localizedFormat);
+dayjs.extend(advancedFormat);
```
diegolmello: **High: this conflicts with the PR contract.**

The PR description says separators keep `LL` without ordinal suffixes and that `app/lib/dayjs` remains untouched. This installs `advancedFormat`, and `formatLongDate` changes English Date Separators and combined Unread Separators from `November 10, 2017` to `November 10th, 2017`. The snapshot diff confirms the behavior change.

Either restore `LL` and remove this global library change, or update the PR description and confirm that the broader formatting policy is intentional.

diegolmello (changes requested):
Thermo-nuclear code-quality review against `develop`. Three structural issues remain blocking.

diegolmello (approved):

PR #7601, opened 2026-08-25
fix: jump to message fails on Files (messagesView)

OtavioStasiak:
## Proposed changes

depends on https://github.com/RocketChat/Rocket.Chat/pull/41939
Tapping a file in **Room → Files** navigated to the room but never scrolled to the message containing that file.
  
  The Files list is built from `rooms.files`, which returns **upload records** — each row's `_id` is the *file* id, not the id of the message that carries it. `jumpToMessage` was passing that file id straight through as `jumpToMessageId`, so `RoomView` looked for a message that doesn't exist and the jump silently no-op'd.

  Now, when the list is the Files list, we resolve the real message id first via the new `chat.getMessageByFileId` endpoint, and only then navigate. If it can't be resolved we surface an error instead of navigating to nowhere.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/SUP-1106

## How to test or reproduce

- Open the app;
- Go to Room;
- Go to Files;
- Tap on File message;

## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Files in the Messages view now open the associated message when supported by the server version.
  * Added support for retrieving messages by file ID.

* **Bug Fixes**
  * Failed file-to-message lookups now display an error and prevent incorrect navigation.
  * Thread naming now uses the resolved message.


app/lib/services/restApi.ts
```diff
@@ -786,6 +786,10 @@ export const getFiles = (roomId: string, type: SubscriptionType, offset: number)
 	});
 };
 
+export const getMessageByFileId = (fileId: string) =>
+	// RC 8.10.0
```
diegolmello, line 790: We need to wait the backend PR to merge so we're sure about the version

app/views/MessagesView/index.tsx
```diff
@@ -136,11 +145,31 @@ class MessagesView extends Component<IMessagesViewProps, IMessagesViewState> {
 		navigation.navigate('RoomInfoView', navParam);
 	};
 
+	resolveMessageId = async (item: IMessage): Promise<string | null> => {
+		const { route, serverVersion } = this.props;
+		if (route.params?.name !== 'Files' || compareServerVersion(serverVersion, 'lowerThan', '8.9.0')) {
```
diegolmello, line 150: There's a gap that makes 8.9.* call the new logic, but your comment on the other file says 8.10.0.

diegolmello (changes requested):
Also missing an e2e test

PR #7609, opened 2026-08-26
refactor: import the 2FA cancellation guard from its leaf module

diegolmello:
## Proposed changes

`TwoFactorCancelledError` and `isTwoFactorCancelled` live in the leaf module `app/lib/services/twoFactorCancelled.ts`, which exists to break the `log → twoFactor → containers/TwoFactor → log` cycle. Until now `app/lib/services/twoFactor.ts` also re-exported both symbols, so half the consumers reached the guard through the barrel instead.

That second path was not just untidy — it was a latent crash. Five suites partially mock `services/twoFactor` with `{ twoFactor: jest.fn() }` and no `requireActual`:

- `app/sagas/__tests__/selectServer.sdkHost.test.ts`
- `app/sagas/__tests__/foregroundResume.integration.test.ts`
- `app/lib/methods/subscriptions/__tests__/roomSubscription.integration.test.ts`
- `app/lib/services/__tests__/connect.integration.test.ts`
- `app/views/ProfileView/index.test.tsx`

Under those mocks, any module reached transitively that imported the guard from the barrel got `undefined` and threw on call — including the backstop guards in `app/lib/methods/helpers/log/index.ts` and `helpers/info.ts`, which are precisely what keeps a deliberate 2FA cancellation from filing a Bugsnag report.

This PR points every consumer at the leaf and deletes the re-export, leaving one import path that no suite mocks.

## Issue(s)

None — internal refactor.

## How to test or reproduce

No behaviour change; the 15 guard sites are byte-identical apart from their import line. Covered by the existing suites:

```
TZ=UTC pnpm test
```

## Screenshots

n/a — no UI change.

## Types of changes

## Checklist

## Further comments

Targets `new-sdk` (#7574), which this branch is stacked on.

#7607 changes `ProfileView/index.tsx` and `ChangePasswordView/index.tsx` for the 2FA retry protocol and touches the same import blocks, so expect a small conflict on whichever lands second.

The third commit drops a `...jest.requireActual('./twoFactor')` spread in `app/lib/services/sdk.test.ts` that only existed to preserve the re-exported guard; `twoFactor.ts` now exports a single symbol, which the following line already overrides.

## Summary by CodeRabbit

* **Refactor**
  * Centralized two-factor cancellation handling through a dedicated service.
  * Updated authentication, account, profile, security, and room-management flows to use the shared cancellation handling.
  * Preserved existing behavior for cancelled two-factor authentication requests.

* **Tests**
  * Updated related tests and mocks to reflect the reorganized two-factor services.


app/containers/TwoFactor/index.test.tsx
```diff
@@ -1,7 +1,8 @@
 import { act, fireEvent, render, waitFor } from '@testing-library/react-native';
 
 import TwoFactor from '.';
-import { isTwoFactorCancelled, twoFactor } from '../../lib/services/twoFactor';
+import { twoFactor } from '../../lib/services/twoFactor';
+import { isTwoFactorCancelled } from '../../lib/services/twoFactorCancelled';
```
diegolmello: Create `lib/services/twoFactor` folder and move twoFactor and isTwoFactorCancelled to it. No barrel.

diegolmello (commented):
app/lib/methods/helpers/twoFactorCancellation.test.ts is colocated incorrectly.

`import { TWO_FACTOR } from '../../containers/TwoFactor';` Move it to constants folder

app/containers/TwoFactor/index.test.tsx
diegolmello: Done in 6042c2f437 — `app/lib/services/twoFactor/` now holds `twoFactor.ts` and `twoFactorCancelled.ts`, no index barrel. Every consumer imports the file it needs directly.

diegolmello:
Addressed in 6042c2f437:

- **`lib/services/twoFactor/` folder, no barrel** — `twoFactor.ts` and `twoFactorCancelled.ts` moved in; consumers import the specific file.
- **`TWO_FACTOR` moved to constants** — now `app/lib/constants/twoFactor.ts`, matching the flat no-barrel layout of that folder. `containers/TwoFactor/index.tsx` imports it instead of defining it, so the service no longer reaches into a container for a string constant.
- **Test colocated with what it tests** — `twoFactorCancellation.test.ts` moved from `lib/methods/helpers/` to `lib/services/twoFactor/`.

Full suite green: 254 suites / 2294 tests. `tsc` and `oxfmt` clean, and `no-cycle` reports nothing on these paths.

PR #7619, opened 2026-08-31
chore: remove all import/no-cycle warnings

diegolmello:
## Proposed changes

Oxlint reported 101 `import/no-cycle` warnings. Circular imports make module load order load-bearing: whichever module happens to be imported first wins, which is how an `Avatar` header ended up depending on a barrel re-export reaching `helpers/fetch`. This breaks every cycle, taking the count to 0 with no behavior change.

- **Barrels and constants.** `lib/methods/helpers` no longer re-exports `fileDownload`, which pulled `containers/Toast` and `lib/encryption` into every consumer of the barrel; the 5 call sites import the module directly. List, MessageComposer/Toolbar and message/Attachments import siblings directly instead of routing back through their own barrel, so `Attachments/components.ts` is removed. `INAPP_NOTIFICATION_EMITTER` and `TWO_FACTOR` move to `lib/constants`, so `lib/` no longer imports `containers/` for an event name.
- **Markdown recursion.** `Inline` owns the grammar switch and the recursion; `Bold`, `Italic`, `Strike` and `Link` take children and provide their emphasis through `MarkdownContext`. `Heading` delegates to `Inline`. `useMarkdownContext` accepts a text style and memoizes the merged value, and `Markdown` memoizes its context value so the existing memo on `AtMention` and `Hashtag` pays off.
- **Nested quotes.** `Attachments` owns the nested-quote recursion and picks a variant; `Reply` takes the nested lists as children and derives its message from the attachment. `Quote` is gone, callers pass the variant directly.
- **`services/restApi` <-> `lib/encryption`.** The push device token moves to a `notifications/deviceToken` leaf that both `notifications/push` and restApi read. `e2eResetOwnKey` no longer stops the rooms stream; `E2EEncryptionSecurityView` calls `unsubscribeRooms` before resetting. `editMessage` moves to `lib/methods/editMessage`, where encrypting before the post belongs. `EncryptionRoom` takes its Encryption client through the constructor as an `IEncryption`.
- **`Avatar`** headers no longer depend on module load order.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1522

## How to test or reproduce

`pnpm lint` reports 0 `import/no-cycle` warnings. Smoke the touched surfaces: markdown rendering (bold/italic/strike/links, mentions, headings), quoted and nested-quoted messages, attachments, message editing (including in an E2EE room), push registration, and E2E key reset from Security and Privacy.

## Types of changes

## Checklist

## Further comments

The 21 updated Avatar snapshots are all the same field: `Avatar` reads `RocketChatSettings.customHeaders`, populated by a module-level side effect in `helpers/fetch.ts`. Those suites were loading that module only because the helpers barrel happened to reach it, so the field is now `undefined` in every Avatar-rendering suite instead of some. Production is unaffected: `restApi`, `connect`, `selectServer` and `getSettings` all import `helpers/fetch` during server select and login.

Local: `pnpm lint` clean of cycles; `TZ=UTC pnpm test` 239 suites / 2200 tests / 416 snapshots passing.

## Summary by CodeRabbit

* **New Features**
  * Added support for editing messages with encrypted content when encryption is available.
  * Improved rendering of nested formatting in Markdown, including links, emphasis, mentions, hashtags, and headings.
  * Added nested quote attachment rendering for replies.
  * Added centralized device-token handling for push notifications.

* **Bug Fixes**
  * Improved avatar image loading when custom request headers are unavailable.
  * Preserved attachment descriptions when displaying quoted replies.

* **Refactor**
  * Consolidated notification and security event configuration for more consistent behavior.


OtavioStasiak (approved):
LGTM

PR #7620, opened 2026-08-31
fix: reaction list cuts off last reactors behind the nav bar

Rohit3523:
## Proposed changes
The "who reacted" bottom sheet (`ReactionsList`) opens with a fixed `snaps: ['50%']`. TrueSheet always anchors a bottom sheet's bottom edge to the raw screen edge regardless of snap height, so on Android the last row of a long reactor list rendered partly behind the nav bar — invisible and untappable. Reported: with 30 users reacting to the same emoji, only ~27 were reachable.

`UsersList` and `AllTab`'s `FlatList`s now pad their content by `insets.bottom` plus a fixed breathing-room buffer (`BOTTOM_SHEET_BREATHING_ROOM`, empirically verified on-device) instead of a flat `paddingBottom: 30`, so the last row always scrolls clear of the nav bar / home indicator.

## Issue(s)
https://rocketchat.atlassian.net/browse/SUP-1114

## How to test or reproduce
1. React to a message with the same emoji from 25-30+ different users.
2. Long-press the reaction pill to open the "who reacted" sheet.
3. Switch to the per-emoji tab and scroll to the last reactor.

Before the fix: last row(s) are cut off/hidden behind the nav bar and not tappable.
After: last row is fully visible and reachable with margin.

## Screenshots
| OS | Before | After |
| --- | --- | --- |
| Android | <video src="https://github.com/user-attachments/assets/d3fbabec-aa01-4d55-8b5a-1d9a2f2d60ee" controls width="100%" muted></video> | <video src="https://github.com/user-attachments/assets/4f0f3267-68b0-4cc1-9492-dbce7dd65ff3" controls width="100%" muted></video> |
| iOS | <video src="https://github.com/user-attachments/assets/b4715d8e-d76a-47b2-b1d7-5a55b3933327" controls width="100%" muted></video> | <video src="https://github.com/user-attachments/assets/eef434e1-250d-4c0b-b4c3-e3bed9cb8107" controls width="100%" muted></video> | 

| Screen | Android | iPhone | iPhone SE | Tablet |
| --- | --- | --- | --- | --- |
| Server list |  |  |  |  |
| Server history |  |  |  |  |
| Emoji picker |  |  |  |  |
| Add reaction |  |  |  |  |
| Message delete |  |  |  |  |
| Reaction list |  |  |  |  |
| Directory filter |  |  |  |  |

## Types of changes

## Checklist

## Bug Fixes
- Improved reaction list scrolling and bottom spacing on devices with safe-area insets.
- Added extra breathing room so the final reactors are easier to view.
- Improved reaction tab rendering for more reliable access to reaction lists.

## Tests
- Expanded coverage to verify scrolling until the last reactor is visible.
- Added support for testing reaction lists with multiple reactors and emojis.


.maestro/tests/room/reaction-list-scroll-to-last-user.yaml
```diff
@@ -0,0 +1,65 @@
+appId: ${APP_ID}
+name: Reaction List Scroll To Last User
+jsEngine: graaljs
+onFlowStart:
+  - runFlow: '../../helpers/setup.yaml'
+onFlowComplete:
+  - evalScript: ${output.utils.deleteCreatedUsers()}
+tags:
+  - test-12
```
diegolmello: Change this to a tag with less tests and runs faster than 12.

app/containers/ReactionsList/UsersList.tsx
```diff
@@ -19,7 +22,7 @@ const UsersList = ({ emoji, usernames, names }: { emoji: string; usernames: stri
 	return (
 		<FlatList
 			data={users}
-			contentContainerStyle={styles.listContainer}
+			contentContainerStyle={[styles.listContainer, { paddingBottom }]}
```
diegolmello: Thanks for digging into this. The bug is real, but I think the fix belongs in the action sheet itself, not in the reactions list.

**Why the rows get hidden**

On Android, TrueSheet makes the sheet a bit taller than what you see: it adds the height of the navigation bar to the bottom, and expects the app to pad that space back. Our sheet also renders the drag handle inside that same area without accounting for it. So any content we place in the sheet ends up 28dp (handle) + nav bar height too tall, and the last rows fall behind the nav bar. On a device with a 48dp nav bar that's 76dp, which is about what the before video shows.

`BottomSheetContent` already pads by the nav bar height when the sheet shows a list of options, but it does nothing when we pass our own children. That's why several other sheets in the app each add their own bottom padding, all with different numbers.

**What I'd like instead**

Fix it once in `BottomSheetContent`, in the branch that renders children:

```diff
+import { HANDLE_HEIGHT } from './useActionSheetDetents';
 ...
+		const contentPaddingBottom = bottom + (isAndroid ? HANDLE_HEIGHT : 0);
 ...
-				style={fullContainer && !(hugContent && isAndroid) ? [styles.fullContainer, minHeightStyle] : undefined}
+				style={[
+					fullContainer && !(hugContent && isAndroid) ? [styles.fullContainer, minHeightStyle] : undefined,
+					{ paddingBottom: contentPaddingBottom }
+				]}
```

Padding works better than shrinking the height here, because the lists inside use `flex: 1` and will just fill the smaller area.

Then:
- Revert `UsersList` and `AllTab` to the plain `styles.listContainer` and remove `BOTTOM_SHEET_BREATHING_ROOM`. If you want some space under the last row, a small fixed padding in the style is fine, but it shouldn't depend on insets anymore.
- Remove the padding workarounds in `AttachmentActionSheet`, `NewMediaCall` and `EmojiPicker` (when shown in a sheet), otherwise they'll get padded twice. A separate PR is fine if you prefer, as long as it lands together with this one.
- `ReactionsList/styles.ts` lost its final newline, lint will complain.
- Update the snapshots.

**Please check on a device**

- Android with a 3-button nav bar: last reactor visible and tappable, no big empty space under it.
- Sheets that size themselves to their content (attachments, servers history): they'll grow by the padding, just make sure they still look right.
- iOS: it should only need the safe area padding. Confirm the home indicator doesn't cover the last row.

The testID additions and the Maestro test are good, keep them.

diegolmello (dismissed):

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -1,11 +1,25 @@
...
+	bottom,
+	fullContainer,
+	hugContent,
+	scrollEnabled
+}: {
+	bottom: number;
+	fullContainer?: boolean;
+	hugContent?: boolean;
+	scrollEnabled?: boolean;
+}): number => Math.max(32, bottom) + (isAndroid && fullContainer && !hugContent && !scrollEnabled ? HANDLE_HEIGHT : 0);
```
OtavioStasiak: unnamed magic number

app/containers/ActionSheet/useActionSheetDetents.test.tsx
```diff
@@ -86,3 +94,43 @@ describe('useActionSheetDetents', () => {
 		expect(result.current.detents).toEqual([0.15]);
 	});
 });
+
+describe('getSheetContentPaddingBottom', () => {
+	const bottom = 32;
```
OtavioStasiak: the fixture collides with the floor, so 5 of 6 new cases assert nothing. const bottom = 32; is exactly the Math.max(32, …)
  floor, so every toBe(bottom) assertion (105, 109, 121, 127, 133) passes identically against a plain bottom implementation. The test names ("returns the safe-area
  bottom…") are unverified, and the floor itself has no test. Use bottom: 48 and add a bottom: 0 → 32 case.

app/containers/ActionSheet/ActionSheet.tsx
```diff
@@ -133,6 +142,7 @@ const ActionSheet = memo(
 					ref={sheetRef}
 					detents={detents}
 					maxHeight={maxHeight}
+					insetAdjustment='automatic'
```
OtavioStasiak: insetAdjustment='automatic' (ActionSheet.tsx:145) is a pure no-op — already the default in JS, the Android ViewManager and the controller.

app/containers/ActionSheet/ActionSheet.tsx
```diff
@@ -116,6 +114,17 @@ const ActionSheet = memo(
 		const hasSnaps = !!effectiveSnaps?.length;
 		const disableContentPanning = data?.enableContentPanningGesture === false;
 		const isScrollable = hasOptions || (hasSnaps && !disableContentPanning);
+		const contentScrollEnabled = hasOptions ? scrollEnabled : isScrollable;
+
+		const handleContentLayout = ({ nativeEvent: { layout } }: LayoutChangeEvent) => {
```
OtavioStasiak, line 119: double check:
 1. AttachmentActionSheet regression — ComposerAttachments.tsx:33-34 enables panning, so contentScrollEnabled=true → padding drops to max(32, bottom), replacing the
     old bottom + top + 48. (See correction below.)
  2. handleContentLayout cancels the fix for dynamic sheets. ActionSheet.tsx:126 subtracts the padding back out, so for no-snaps/no-options sheets the detent is
     unchanged and the padding just overflows into overflow:hidden — zero clearance gained.

OtavioStasiak (changes requested):

app/containers/ActionSheet/ActionSheet.tsx
Rohit3523: Verified both, and both are fine: the attachment sheet resolves to max(32, bottom) instead of the old bottom + top + 48, but Save is still reachable on Android and unchanged (flush) on iPhone SE, which I confirmed on-device. And the subtracted padding on dynamic sheets just restores develop rendering — that's what kills the Workspaces gap — while the snapped Reactions sheets never use that measurement, so the original fix is untouched.

app/containers/ActionSheet/useActionSheetDetents.ts
```diff
@@ -1,10 +1,28 @@
 import type { SheetDetent } from '@lodev09/react-native-true-sheet';
 import { useMemo } from 'react';
 
+import { isAndroid } from '../../lib/methods/helpers';
+
 const ACTION_SHEET_MIN_HEIGHT_FRACTION = 0.15;
 const ACTION_SHEET_MAX_HEIGHT_FRACTION = 0.75;
 const SCROLL_ENABLED_THRESHOLD = 0.6;
 export const HANDLE_HEIGHT = 28;
+// home-button devices returns a 0 bottom inset, hiding content off-screen without this floor.
```
OtavioStasiak: remove, this comment, run e2e tests and you can merge!

OtavioStasiak (approved):
Approved.

PR #7625, opened 2026-09-01
fix: UNIQUE constraint failed on threads.id

diegolmello:
## Proposed changes

`getThreadName` re-read the local thread with `getThreadById` *outside* `db.write` (the comment there said "check it again to avoid race condition"). Two callers resolving the same `tmid` — several messages in a room pointing at the same thread, all rendering at once — could both pass that check, both enter the writer, and both `prepareCreate` a thread with the same id. The second batch then hit the sqlite unique index and threw:

```
Failed to execute db update - sqlite error 1555 (UNIQUE constraint failed: threads.id)
  app/lib/methods/getThreadName.ts:38:19
```

The re-check now runs inside the writer, where WatermelonDB serializes writes, so it is authoritative. When the thread is already there, the caller just updates `tmsg` instead of creating a duplicate.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1585

## How to test or reproduce

- Open a room with several messages replying to the same thread, where that thread's parent message is not yet in the local database
- Before: the second concurrent `getThreadName` logs `sqlite error 1555 (UNIQUE constraint failed: threads.id)` and the thread name does not render on that message
- After: the thread is created once, both messages get their `tmsg`

Covered by a new unit test that races two `getThreadName` calls for the same `tmid` against a serialized writer and a batch that enforces uniqueness on `threads.id`:

```
TZ=UTC npx jest app/lib/methods/getThreadName.test.ts
```

The test is red on the previous code with the exact sqlite message, green with the fix.

## Types of changes

## Checklist

## Further comments

The pre-existing "skips creating the thread when another writer created it during the network gap" test changed one assertion: the guarded path now goes through `db.write`, because that is where the check lives.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate threads during simultaneous requests.
  - Ensured messages are updated with the correct thread name when threads are created concurrently.
  - Improved consistency when refreshing thread information after network delays.


OtavioStasiak (approved):
LGTM!

Issue #7341, opened 2026-05-25
Add unit-test coverage for app/sagas/


diegolmello:
## Background

`app/sagas/` has just two existing tests (`videoConf.voipBlock.test.ts`, `deepLinking.test.ts`). Core sagas — `login.js`, `init.js`, `rooms.js`, `messages.js`, `encryption.js`, `state.js`, `selectServer.ts`, `room.js` — have no direct unit-test coverage.

## Motivating example

[NATIVE-1180](https://rocketchat.atlassian.net/browse/NATIVE-1180) / #7333 / #7340: an unconditional `cancel(loginSuccessTask)` in `login.js`'s `root()` watcher imposed a hard 2-second deadline on `handleLoginSuccess` and stranded users on the login screen when permission/enterprise fetches ran long. The bug lived in `root()`'s cancel logic for ~3 weeks before a user reported it.

A small saga harness asserting that `cancel(loginSuccessTask)` only fires when `SERVER.SELECT_REQUEST` wins the race would have failed loudly the moment the regression landed.

## Suggested scope

- Set up a saga test harness under `app/sagas/__tests__/` using `runSaga` or `expectSaga` from `@redux-saga/testing-utils`.
- First pass: cover the `root()` watcher in `login.js` — `SERVER.SELECT_REQUEST` cancels `handleLoginSuccess`; 2s timeout does not.
- Subsequent passes: high-traffic sagas (`init.js`, `selectServer.ts`, `rooms.js`).

## Non-goals

- Not aiming for 100% saga coverage. Pick the seams where regressions are most costly (auth, server switch, deep links, rooms sync).

[NATIVE-1180]: https://rocketchat.atlassian.net/browse/NATIVE-1180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

deepak0x:
Hey @diegolmello can i pick this issue?

Agam00:
Opened #7645 covering rooms.js. Happy to do room.js and encryption.js next if this approach looks right.

PR #7645, opened 2026-09-03
test: add unit tests for the rooms saga watcher

Agam00:
## Proposed changes

`app/sagas/rooms.js` had no tests, so this adds some. 10 tests over the `root()` watcher and `handleRoomsRequest`, using the existing `sagaStore` helpers and the same mocking setup as `login.switchCancel.test.ts`. No new deps, saga itself untouched.

Covers: requests ignored when not logged in, the `allData` refresh path, failures dispatching `ROOMS.FAILURE` without killing the watcher loop, and workspace switch / logout / background each cancelling an in-flight sync.

Sanity checked them by breaking `rooms.js` three ways (removing the `cancel`, forcing the auth check true, flipping `allData`) and each one failed only the tests it should.

Skipped the batching part of `handleRoomsRequest` for now, needs a WatermelonDB fixture.

## Issue(s)

Part of #7341

## How to test or reproduce

`pnpm test app/sagas/__tests__/rooms.test.ts`

## Screenshots

n/a, test only.

## Types of changes

## Checklist

- [x] I have signed the CLA

## Further comments

The 30s timeout test just records current behaviour. `root()` cancels unconditionally after the race, so a sync over 30s gets dropped with no SUCCESS or FAILURE. Same shape as the login cancel bug in #7340. Left it as is since this is test only, can fix separately if you want.

Next: `room.js` and `encryption.js`.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for room synchronization behavior.
  * Verified authentication-gated room requests, successful and incremental syncs, refresh handling, and failure scenarios.
  * Added coverage for cancelling in-progress synchronization when the workspace changes, the user logs out, the app is backgrounded, or the operation exceeds the timeout.

### DIFF
--- a/app/sagas/__tests__/rooms.test.ts
+++ b/app/sagas/__tests__/rooms.test.ts
@@ -1,0 +1,230 @@
+jest.mock('../../lib/methods/subscribeRooms', () => ({
+	subscribeRooms: jest.fn(),
+	unsubscribeRooms: jest.fn()
+}));
+
+jest.mock('../../lib/methods/getRooms', () => ({
+	getRooms: jest.fn()
+}));
+
+jest.mock('../../lib/methods/helpers/mergeSubscriptionsRooms', () => ({
+	__esModule: true,
+	default: jest.fn(async () => [])
+}));
+
+jest.mock('../../lib/methods/helpers/buildMessage', () => ({
+	__esModule: true,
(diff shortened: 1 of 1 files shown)


Agam00:
Added the return type annotations, thanks.

PR #7655, opened 2026-09-08
refactor: centralize jump navigation lifecycle in RoomView

diegolmello:
## Proposed changes

`useRoomNavigation` now owns the whole Jump to Message and Thread-opening lifecycle: jump generations, cancellation, Thread-name lookup, destination selection and loading settlement. The `useJumpToMessage` hook, the `jumpToMessage` service and `pushThreadRoom` are removed, along with the cancellation callback that navigation used to hand to the jump hook through a ref and an effect.

Behavior is unchanged. Generation checks stay at the same asynchronous boundaries (after message lookup, after anchor resolution, after the Fabric commit wait and in the error path), cross-Room and Thread-name lookups still have no supersession check, ordinary Thread opening still dismisses loading after 300 ms, and unmount still cancels only scheduled interactions. Message lookup, Thread-name lookup and anchor resolution remain separate helpers.

Tests move from asserting the private callback wiring to exercising the public entry points (`jumpToMessageByUrl`, `onThreadPress`, `onThreadMessagesLoaded`, route params). Two new suites cover the composed navigation flows and the lifecycle interactions between overlapping jumps, Thread presses, cancellation and unmount.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-34

## How to test or reproduce

- Open a message link that targets a message already loaded in the current room; it should scroll to and highlight it.
- Open a message link for an older message in the same room; it should load around it and jump.
- Open a message link for another room and for a thread in the same room; the destination should open and position on the message.
- Tap a thread from the room list and from its parent message; titles, encrypted-message titles and the loading dismissal should look as before.
- Cancel the loading overlay during a jump; no navigation should happen afterwards.
- Select the same search result twice in a row; the jump should fire both times.

## Types of changes

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Consolidated message-jump and thread-navigation handling within room navigation.
  * Preserved support for in-room, cross-room, and cross-thread navigation, including cancellation and asynchronous loading scenarios.

* **Tests**
  * Added comprehensive coverage for navigation lifecycle, route-target handling, cancellation, race conditions, errors, and thread navigation.
  * Updated existing tests to use the current navigation interfaces.


diegolmello:
# How `useRoomNavigation` works

## 📋 TLDR

This hook is the room's "go somewhere" desk. It owns three jobs: jumping to a specific message inside the current room or thread, opening a thread, and reacting to navigation params (`jumpToMessageId`, `jumpToThreadId`) that arrive when the screen is opened or re-focused. Everything hard about it comes from one fact: a jump is slow and asynchronous, and the user can cancel, start another jump, or leave the screen while it is in flight.

---

## 1. Live refs: reading the latest function without re-subscribing

🧠 **Analogy:** a mailbox with a fixed address. The house behind it can be rebuilt every render, but anyone holding the address always reaches the current house.

**Mechanism:** `useLiveRef(fn)` stores the newest version of `fn` in a ref on every render. Effects capture the ref, not the function, so they can run once and still call up-to-date logic.

```ts
const consumeJumpParamRef = useLiveRef(consumeJumpParam); // ← updated every render
useEffect(() => {
	const task = InteractionManager.runAfterInteractions(() => {
		consumeJumpParamRef.current(...); // ← always the newest closure
	});
	return () => task.cancel();
}, [tmid, consumeJumpParamRef]); // ← stable dep, effect does not churn
```

## 2. Generations: the "is my work still wanted?" ticket

🧠 **Analogy:** a deli counter. You take number 7. If the board already says 8 when your sandwich is ready, someone else placed a newer order and yours is thrown away.

**Mechanism:** every jump increments `jumpGenerationRef` and remembers its own number. After each `await`, the jump checks whether its number is still the current one. Cancel is just "bump the counter", so any in-flight jump quietly stops at its next checkpoint.

```ts
const jumpToMessage = async (messageId) => {
	jumpGenerationRef.current += 1;
	const generation = jumpGenerationRef.current; // ← my ticket
	const message = await getMessageInfo(messageId);
	if (!isCurrentJump(generation)) return; // ← someone newer won, stop
	...
};

const cancelJumpToMessage = () => {
	jumpGenerationRef.current += 1; // ← invalidates every running jump
	listContainerRef.current?.cancelJumpToMessage();
	sendLoadingEvent({ visible: false });
};
```

## 3. Deciding where the message lives

🧠 **Analogy:** you are handed a page number. First question: is it in the book you are holding, or in a different book on the shelf?

**Mechanism:** `executeJump` compares the target message's `rid` and `tmid` with the screen's own. Three outcomes: another room (navigate there with `goRoom`), a thread of this room (push a thread `RoomView`), or right here (scroll). It returns `true` only when the jump finished locally, so the caller knows whether to hide the loading overlay itself or let the next screen own it.

```ts
const inThisThread = !!message.tmid && message.tmid === tmid;
const inThisRoom = !message.tmid && message.rid === rid;
if (!inThisThread && !inThisRoom) {
	message.rid !== rid ? await navToRoom(message) : await navToThread(message); // ← leave this screen
	return false;
}
if (inThisRoom && t === 'thread' && message.id !== tmid) {
	await navToRoom(message); // ← we are inside a thread, target is the parent room
	return false;
}
```

## 4. The local jump: anchor, wait for Fabric, then scroll

🧠 **Analogy:** finding a house on a street that is still being built. You first make sure the house exists (load surrounding messages), then wait for the construction crew to finish the block (layout commit), then drive there.

**Mechanism:** `resolveJumpAnchor` decides whether the message is already rendered or must be fetched, and returns a timestamp used as a scroll anchor. Because Fabric commits layout asynchronously, a fixed 100ms pause lets the newly inserted rows get measured before the list is asked to scroll. Generation checks sit around each pause.

```ts
const inWindow = listContainerRef.current?.isMessageInWindow(message.id) ?? false;
const highTsMs = await resolveJumpAnchor(rid, {...}, inWindow, { loadSurroundingMessages, getLocalAnchorTs });
if (!isCurrentJump(generation)) return false;
await waitForFabricCommit(); // ← 100ms, lets Fabric measure new rows
if (!isCurrentJump(generation)) return false;
await listContainerRef.current?.jumpToMessage(message.id, highTsMs); // ← the list does the scrolling
return true; // ← settled locally, caller hides loading
```

## 5. Opening a thread

🧠 **Analogy:** a footnote. The message might be the footnote itself (has `tmid`), or the sentence the footnote hangs off (has `tlm`, meaning it is a thread root).

**Mechanism:** `navToThread` handles both shapes. A thread root pushes directly. A thread reply first fetches the thread name (with a cancellable loading overlay), then pushes a thread `RoomView` carrying `jumpToMessageId` so the new screen scrolls to that reply. Undecryptable E2E messages get a placeholder title.

```ts
if (!message.tmid) {
	if ('tlm' in message) return navigation.push('RoomView', { rid, tmid: message.id, ... }); // ← thread root
	return;
}
let cancelled = false;
sendLoadingEvent({ visible: true, onCancel: () => { cancelled = true; cancelJumpToMessage(); } });
threadName = await fetchThreadName(rid, message.tmid, jumpToMessageId, knownName);
if (!threadName || cancelled) { sendLoadingEvent({ visible: false }); return; } // ← user backed out
return navigation.push('RoomView', { rid, tmid: message.tmid, name, jumpToMessageId, ... });
```

## 6. Params as one-shot commands

🧠 **Analogy:** a sticky note on your monitor. You read it, do the thing, then throw it away so you do not do it again tomorrow.

**Mechanism:** `jumpToMessageId` arrives via route params. It is copied into `pendingJumpRef` on mount, consumed once, and cleared with `setParams` so a re-render or re-focus does not replay it. `useChangedParam` handles later changes (deep link while the screen is already open) by only firing when the value actually differs from the previous render.

```ts
const consumeJumpParam = (messageId) => {
	pendingJumpRef.current = undefined; // ← throw away the note
	jumpToMessage(messageId);
	navigation.setParams({ jumpToMessageId: undefined }); // ← so it never replays
};

function useChangedParam(value, onChange) {
	const prevRef = useRef(value);
	useEffect(() => {
		if (value && value !== prevRef.current) onChange(value); // ← only on real change
		prevRef.current = value;
	}, [value]);
}
```

## 7. Threads must load before jumping

🧠 **Analogy:** you cannot scroll to page 40 of a book that has not been printed yet.

**Mechanism:** in a room, the jump can start right after interactions finish. In a thread, the message list is empty until the thread messages arrive, so the jump is parked in `pendingJumpRef` and released by `onThreadMessagesLoaded`, which the list calls once. `loadedThreadRef` remembers that this thread already loaded, so later param changes jump immediately.

```ts
useEffect(() => {
	const task = InteractionManager.runAfterInteractions(() => {
		if (pendingJumpRef.current && !tmid) consumeJumpParamRef.current(pendingJumpRef.current); // ← room: go now
	});
	return () => task.cancel();
}, [tmid, consumeJumpParamRef]);

const onThreadMessagesLoaded = () => {
	loadedThreadRef.current = tmid;
	if (pendingJumpRef.current) consumeJumpParam(pendingJumpRef.current); // ← thread: go once loaded
};

const onJumpParamChanged = (messageId) =>
	!tmid || loadedThreadRef.current === tmid ? consumeJumpParam(messageId) : (pendingJumpRef.current = messageId);
```

`jumpToThreadId` follows the same pattern but defers to a pending message jump if both params are present, since the message jump already knows how to open the right thread.

## 8. Public surface

The hook exposes only three things, which keeps `RoomView` thin:

- `onThreadMessagesLoaded`: the list calls it when a thread finishes loading.
- `onThreadPress`: debounced (leading edge, 1s) so a double tap does not push two thread screens.
- `jumpToMessageByUrl`: parses `?msg=` from a message link and runs a jump. When `isFromReply`, a permission error becomes a "room not found" alert instead of a log.

---

## 🎯 Recap

A jump is a race between the network, layout, and the user. The hook wins that race with three tools: live refs so long-lived effects call fresh logic, generation tickets so stale work stops itself, and one-shot param consumption so navigation state never replays. Around that core, `executeJump` routes the target to the right screen, and thread loading gates the scroll until there is something to scroll to.

PR #7481, opened 2026-07-09
chore: upgrade to React Native 0.83.10 and Expo SDK 55

diegolmello:
## Proposed changes

Coordinated merge of the React Native 0.81 → 0.83.10 / Expo SDK 54 → 55 upgrade from the long-lived integration branch `chore/rn-0.83-expo-55` into `develop`. Per the "no half upgrades" model, the whole epic lands in a single merge.

Bundled work:
- Framework bump — RN 0.83.10 + Expo SDK 55, Gradle 9, bridgeless `MainApplication` host, native config re-merge.
- `nitro-modules` 0.35.9 + `react-native-mmkv` 4.3.1 lockstep.
- `patch-package` patches re-validated / regenerated against the target versions.
- `@rocket.chat/mobile-crypto` rebuilt on BouncyCastle.
- LaTeX math rendering re-enabled on a WebView-based KaTeX renderer (`react-native-katex`), replacing the abandoned native library removed during the bump.
- `react-native-image-crop-picker` fork forward-ported.
- `ios/Podfile.lock` finalized (dropped the stale `EXAV` / expo-av pod).
- Post-upgrade audit fixes: `expo-font` added as a direct dependency (peer of `@expo/vector-icons`), duplicate `CADisableMinimumFrameDurationOnPhone` key removed from `Info.plist`. `react-native-keyboard-controller` stays pinned at 1.18.5 — the SDK 55-aligned 1.20.7 breaks hardware-keyboard Tab navigation on Android (e2e shard 14); re-bump tracked as follow-up.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1234

## How to test or reproduce

**Automated gates**
- Native iOS + Android store builds pass (this PR's gated build lanes).
- E2E build + Maestro shards pass on both platforms.

**Manual QA (both iOS + Android unless noted)**

Smoke
- [ ] App boots; login, room list, and messaging work.

KaTeX rendering (`react-native-katex` WebView replaces the removed native lib)
- [ ] Block math `$$...$$` renders typeset (not raw source).
- [ ] Tall / large expressions (e.g. `\Huge`, fractions, `\sum` with limits) are **not clipped** — container resizes to content (WebView height handshake).
- [ ] Array / matrix envs (`\begin{matrix}` … `pmatrix`/`bmatrix`) render without a gray box.
- [ ] Inline math `$x$` intentionally shows **raw source** styled as inline code (not typeset) — confirm acceptable.
- [ ] Renders correctly in both light and dark theme.

[...]

PR #7656, opened 2026-09-08
refactor: replace deep relative imports with application aliases

diegolmello:
## Proposed changes

Use `~/` for imports that traverse two or more parent directories in application code and stories. TypeScript paths define the aliases; Babel and Jest derive their mappings from the same configuration. Keep sibling and single-parent imports relative, and reject new deep relative imports through Oxlint.

Move the theme entry into `app/theme/index.tsx` while preserving its public import path.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1533

## How to test or reproduce

Run `pnpm format-lint` and `TZ=UTC pnpm test`. Build Metro release bundles for iOS and Android. Verify alias imports retain platform-specific resolution and the `.mock.ts` preference under `RUNNING_E2E_TESTS=true`.

Verified locally: 258 Jest suites, 2,347 tests and 416 snapshots passed; full release bundles passed on both platforms. Independent AST comparison verified all 4,107 rewrites and preserved all 3,454 sibling/single-parent references. CI test-map validation passed.

Add a temporary `../../` import under `app/` and confirm Oxlint rejects it. Alias paths must also remain visible to `import/no-cycle`.

## Screenshots

No visual changes.

## Types of changes

## Checklist

- [ ] I have read the CONTRIBUTING doc
- [ ] I have signed the CLA

## Further comments

The migration is split into alias setup and import conversion commits. The RN 0.83 / Expo 55 integration in #7481 will need to account for the rewritten paths and the new lint rule when updating its branch.


OtavioStasiak (approved):
LGTM

PR #7482, opened 2026-07-09
refactor: migrate RoomView to a function component

diegolmello:
## Proposed changes

Converted RoomView from a large class component to a function component, splitting its logic into smaller hooks, stores, and components. Added tests around room state, navigation, and message actions to help preserve existing behavior.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-34

## How to test or reproduce

Use a room end to end (send/edit/quote/react, drafts, autocomplete, threads, jump-to-message, join a not-subscribed room, header actions) on phone and tablet, plus an omnichannel/livechat room. Behavior should match the base branch.

## Screenshots

No visual changes intended.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] Improvement (non-breaking change which improves a current function)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update (if none of the other choices apply)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/RocketChat/Rocket.Chat/blob/develop/.github/CONTRIBUTING.md#contributing-to-rocketchat) doc
- [x] I have signed the [CLA](https://cla-assistant.io/RocketChat/Rocket.Chat.ReactNative)
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works (if applicable)
- [ ] I have added necessary documentation (if applicable)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

None.

PR #7658, opened 2026-09-08
refactor: deepen observed Room reads

diegolmello:
## Proposed changes

Room UI consumers previously had to carry a live Subscription plus an independent `roomUpdate` value, and the first iteration of this PR replaced that pair with an observed Room read whose inner Subscription was still reachable from store state. That inner field was a stale-read hole: a selector could subscribe to the Subscription reference, which never changes when WatermelonDB mutates the cached model in place, so a Room rename left the header title stale.

This revision makes the wrapper opaque and names it the Room Snapshot.

- `RoomSnapshot` is a symbol-keyed envelope carrying the live Subscription or Preview Room. Its identity changes exactly when a tracked field changes or the database supplies a different record. The inner Room is not addressable through a Zustand selector on store state, so the stale path is closed at the type level rather than by convention. A type-level test asserts the selector does not compile.
- One hook, `useRoom()`, returns `{ room, snapshot }`. Call sites destructure the Room and pass the `snapshot` wherever React needs a dependency or a memo prop. A store-argument form, `useRoomFromStore(store)`, serves callers outside the provider, and `getRoom(snapshot)` serves non-React callers such as store commands. The old `RoomState.room` envelope, `RoomRead`, and `useRoomReadFromStore` are gone, along with the "read" vocabulary that collided with Last Seen and Unread.
- The tracked-field list, the field-to-database-column map, and the pure change rule now live in `app/lib/roomObservation.ts` beside the neutral Room contract, with no Zustand and no WatermelonDB imports. The Room store keeps the query, subscription lifecycle, readiness callback, cleanup, and patch application. The composer, Share view, Thread Messages view, and Search Messages view import the contract and hook from the shared location instead of from the Room view.
- JSON-backed tracked fields (`roles`, `tunread`, `tunreadUser`, `tunreadGroup`, `muted`, `unmuted`, `ignored`, `sysMes`) compare by serialised content. The Subscription model memoizes its JSON getters on the raw column string, so a byte-identical sync was already stable; a content-equal payload with a different serialisation still produced a new array and a spurious snapshot. It now produces none.
- The tracked-field copy left store state. It lives in the observer closure, seeded from the initial Room at attach time, so the first identical emission no longer replaces the snapshot and Room open renders once instead of twice. Reattaching observation to the same store re-seeds from the current snapshot, preserving the previous retain-across-cleanup behaviour.

Fixes the stale header title on rename and on topic change. Preserves synchronous initial reads, same-instance mutations, selective and no-op emissions, replacement models, Preview, Invited and Subscribed Room transitions, Direct Message no-subscription behaviour, Thread titles, last-Message-derived Omnichannel state, and per-screen observation cleanup. Initialization and composer ownership are unchanged.

React Compiler note: memoizing on the snapshot is only correct because the snapshot's identity is the change signal. Render-time derivations from the Room take the snapshot as their dependency, never the Subscription, since the Subscription reference is stable across in-place mutation and would freeze a derived value.

## Issue(s)

Follow-up to #7482, targeting its verified head branch `native-34-roomview-hooks` in `RocketChat/Rocket.Chat.ReactNative`.

## How to test or reproduce

- `pnpm exec tsc --noEmit`: passed, no errors.
- `pnpm format-lint`: passed, exit 0. All matched files use the correct format; repository-wide lint warnings only, no errors.
- `TZ=UTC pnpm test --watchman=false`: passed: 319 suites, 2,901 tests, 426 snapshots.

To see the fix by hand, open a Room, rename it from another client, and watch the header title and subtitle follow the change. Then trigger a sync that rewrites `roles` with the same members but a different serialisation, and confirm the Room view does not re-render.

Test boundaries, highest first: rows emitted through the mocked `observeWithColumns` asserted against what a consumer renders (header options, footer state, read-only restriction, Ignored Message placeholder); a real WatermelonDB `Subscription` record through the observer, mutated with a content-equal JSON payload, asserting no new snapshot and a single first paint; table tests over the pure change rule for the retained behaviours; and the type-level assertion that the inner Room is unreachable.

Watchman is disabled because its socket is unavailable in this environment.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] Read repository contribution guidance
- [x] Lint, TypeScript, and unit tests pass locally
- [x] Added outcome tests for the Room Snapshot contract, including a real Subscription record and a type-level test

## Further comments

The neutral Room contract from #7660 is merged in from the composer-ownership branch, not rebased onto, so the shared observation module and the shared hook sit beside it and the consuming screens no longer depend on the Room view.

Integration overlaps: the composer-ownership work touches composer store types, provider props, and related tests; other follow-ups may overlap shared Room definitions and screen wiring.

Out of scope here: copying the Subscription into a plain projection to remove the live model from render, suppressing repeated empty-emission patches, and the Message Window, Jump to Message, and Last Open logic in the Room view.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved room updates so changes such as renamed rooms, permission changes, ignored users, read-only status, and announcement updates appear reliably.
  * Improved message composer upload handling: successful uploads clear related input and attachments, while failed uploads restore text, quotes, and attachments.
  * Preserved composer text and quote state during media transfers, cancellations, and failed sends.
  * Improved draft restoration, autocomplete behavior, and attachment handling in shared and room-based composers.

* **Tests**
  * Expanded coverage for room observation, composer input, media transfers, uploads, sharing, and livechat scenarios.


diegolmello:
## What this PR changes, visually

### State shape

WatermelonDB mutates the Subscription model in place, so its reference never changes. Consumers used to pair the live model with a separate `roomUpdate` snapshot just to get a re-render signal.

```diff
 RoomState
-  room: TSubscriptionModel          # live model, same reference forever
-  roomUpdate: Partial<Subscription> # snapshot of 31 tracked attrs, only there to trigger re-renders
+  room: RoomRead                    # { room: TSubscriptionModel }
+                                    # wrapper identity changes on tracked emission
+                                    # inner .room stays the live model
```

The tracked-attribute list and its DB column mapping moved from `constants.ts` into `RoomStore.ts`. Nothing outside the store knows those exist anymore.

### One database emission

```mermaid
sequenceDiagram
    participant DB as WatermelonDB
    participant Store as RoomStore.observeRoom
    participant UI as useHeader / Composer
    DB->>Store: subscriptions row (same instance)
    Store->>Store: compare 31 tracked attrs vs last observed values
    alt nothing tracked changed
        Store-->>UI: no setState, no re-render
    else tracked attr changed
        Store->>Store: setState({ room: { room: next } })
        Store-->>UI: new RoomRead reference
        UI->>UI: read roomRead.room, re-render
    end
```

Last observed values live in a `WeakMap` keyed by the store instance, so an observer teardown and reattach on the same screen does not emit spuriously.

### Consumers

```diff
 useHeader
-  const room = useStore(roomStore, s => s.room);
-  const roomUpdate = useStore(roomStore, useShallow(s => s.roomUpdate));
-  useLayoutEffect(..., [room, roomUpdate, ...])
+  const roomRead = useStore(roomStore, s => s.room);
+  useLayoutEffect(() => { const room = roomRead.room; ... }, [roomRead, ...])

 RoomStoreContext
-  useRoomWithUpdate()          # subscribed to roomUpdate for the re-render side effect
-  useRoomWithUpdateFromStore()
+  useRoom()                    # returns roomRead.room
+  useRoomFromStore()
+  useRoomReadFromStore()       # for callers that want the changing reference

 ComposerState
-  room: Subscription
-  roomUpdate?: Partial<Subscription>
+  roomRead: RoomRead

 Commands / getters (init, joinRoom, resumeRoom)
-  get().room
+  get().room.room              # still the live model, unchanged semantics
```

app/views/RoomView/components/LeftButtons.tsx
```diff
@@ -30,7 +30,7 @@ const LeftButtons = ({ rid, tmid, roomStore }: ILeftButtonsProps): ReactElement
 	const isMasterDetail = useMasterDetail();
 	const baseUrl = useAppSelector(state => state.server.server);
 	const { id: userId, token } = useAppSelector(getUserSelector);
-	const room = useStore(roomStore, s => s.room);
+	const room = useStore(roomStore, s => s.room.room);
```
diegolmello: `s.room.room` is strange. Can we make it better?

app/views/RoomView/components/RoomProviders.tsx
```diff
@@ -15,8 +15,7 @@ export const RoomProviders = ({
 	rid,
 	t,
 	tmid,
-	room,
-	roomUpdate,
+	roomRead,
```
diegolmello: What's roomRead? It doesn't tell much semantically.

app/views/RoomView/stores/RoomStoreContext.tsx
```diff
@@ -15,13 +15,10 @@ const useRoomStoreApi = (): RoomStore => {
...
 
-export const useRoomWithUpdateFromStore = <S extends { room: unknown; roomUpdate?: unknown }>(store: StoreApi<S>): S['room'] => {
-	useRerenderOnRoomMutatedInPlace(store);
-	return useStore(store, s => s.room);
+export const useRoomFromStore = (store: RoomStore): IRoomViewState['room'] => {
+	return useRoomReadFromStore(store).room;
 };
 
-export const useRoomWithUpdate = (): RoomState['room'] => useRoomWithUpdateFromStore(useRoomStoreApi());
+export const useRoom = (): IRoomViewState['room'] => useRoomFromStore(useRoomStoreApi());
```
diegolmello: Too confusing

PR #7663, opened 2026-09-09
test: e2e encryption banner race and iOS Maestro shard 11 timeout

diegolmello:
## Proposed changes

Two pre-existing CI e2e flakes seen on #7656 and on unrelated branches.

**1. "Save your encryption password" banner never shows after a fresh-install deeplink login (Android shard 14, Keyboard Navigation Components)**

Root cause: `handleEncryptionInit` runs on `LOGIN.SUCCESS` and returns early when neither the cached `servers` row nor `state.settings.E2E_Enable` is set. On a fresh install both are populated by the same un-awaited `getSettings(server)` REST round trip started in `connect()`, so whenever `LOGIN.SUCCESS` wins that race the saga exits and nothing re-runs it once settings land. The banner is lost for the whole session, which is why the 60s `extendedWaitUntil` in the flow never helps. Keyboard Navigation Room only passes because it waits on a room sync first, which gives the settings fetch time to finish.

Fix: when `E2E_Enable` is still unknown from both sources, the saga waits for `SETTINGS.ADD` (looping until the value is defined, since the local-DB `setSettings` also dispatches it without `E2E_Enable` on a fresh install) and then re-evaluates. The wait races against `LOGOUT` and `SERVER.SELECT_REQUEST` so it never dangles across a logout or server switch; `takeLatest` alone does not cover that because those paths dispatch `ENCRYPTION.STOP`, never a second `ENCRYPTION.INIT`. The "servers row OR settings says enabled" semantics are unchanged, and a row that already says `false` skips without waiting.

Regression test `app/sagas/__tests__/encryption.settingsRace.test.ts` drives the real saga and reducers through `createRecordingStore`. The late-settings case fails on `develop` (banner stays empty) and passes with the fix.

**2. iOS shard 11 at the edge of the 30-minute Maestro step**

Per-flow timings from the failing run: Ignore User 7m31s, Jump to message 6m06s, Room Actions 6m24s, Mark as unread 1m03s, 21m04s total. Android runs the same shard in 17m57s and Jump to message takes exactly 6m06s on both platforms, so the cost is inherent to the flows (a 300+ message room with `scrollUntilVisible` hunts, and 50+ tap/wait round trips in Room Actions and Ignore User), with iOS adding roughly a minute per navigation-heavy flow. Shards 3, 6 and 12 also landed at 23 to 26 minutes on the same run once a single flow needed a rerun round.

Changes: Jump to message moves from shard 11 to shard 5 (11m31s iOS, 14m01s Android, no reruns), and the iOS "Run Maestro Tests" step timeout goes from 30 to 45 minutes with the job timeout from 40 to 65 so the step limit, which preserves the log upload, always fires before the job limit (the job also has a 15 minute build step). `assert-maestro-shards.sh` passes.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1534

Failing run: https://github.com/RocketChat/Rocket.Chat.ReactNative/actions/runs/34259766941 (#7656). Same Android failure on https://github.com/RocketChat/Rocket.Chat.ReactNative/actions/runs/34278397719.

## How to test or reproduce

1. Fresh install, login via the auth deeplink, watch for the "Save your encryption password" banner on the rooms list. On `develop` it is missing on a fraction of runs; with this change it appears once settings finish loading.
2. `TZ=UTC pnpm test app/sagas/__tests__/encryption.settingsRace.test.ts`
3. Check the iOS Maestro shard 5 and shard 11 job durations on this PR's run.

## Types of changes

## Checklist

## Further comments

Raising the timeout alone would have hidden the shard imbalance; rebalancing alone still leaves the other 23 to 26 minute iOS shards one rerun away from the old limit, so both changes are needed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved encryption setup when the end-to-end encryption setting loads after initialization.
  - Encryption now waits for a defined setting and stops safely when logging out or switching servers.

- **Tests**
  - Added coverage for delayed, immediate, disabled, and undefined encryption settings.
  - Updated automated test scenario assignments.

- **Chores**
  - Increased time limits for iOS automated test workflows.


OtavioStasiak (approved):
LGTM

PR #7665, opened 2026-09-09
fix: admin panel requiring a new login

diegolmello:
## Proposed changes

The admin panel WebView injected `Meteor.loginWithToken(token)` through `injectedJavaScript`, which react-native-webview evaluates at document-end. On a workspace whose bundle is slow to evaluate, that fires while `document.readyState` is still `loading` and the Meteor client does not exist yet, so the script threw `Can't find variable: Meteor`, no login was ever attempted, and the panel rendered the login form. Instrumenting the WebView on device caught it:

```
{"at":"injected","readyState":"loading","meteor":"undefined","hasToken":true}
{"at":"threw","message":"Can't find variable: Meteor"}
{"at":"after-3s","userId":null,"status":{"status":"connected","connected":true}}
```

Three seconds later Meteor exists and is connected, and nothing retries. This is a race rather than a workspace configuration difference, which is why the report was specific to one server: faster workspaces win it.

The injected script now polls for `Meteor.loginWithToken` before calling it, capped so it cannot spin forever, and the token is serialised with `JSON.stringify` so it cannot break out of the script literal.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1530

## How to test or reproduce

1. Log in to `open.rocket.chat` with an account holding an admin role.
2. Open the sidebar and tap **Admin panel**.

Before: the WebView shows the workspace login form. After: it loads straight into the Workspace / Version screen.

## Screenshots

Before, the WebView rendered the workspace login form. After, it renders the Workspace / Version screen.

## Types of changes

## Checklist

## Further comments

`buildLoginScript` is extracted into its own module so the generated script is unit-testable: `app/views/AdminPanelView/__tests__/buildLoginScript.test.ts` covers Meteor already present, Meteor arriving late, Meteor never arriving, and token escaping. Reverting to the previous one-liner turns three of those four red.

The e2e coverage the issue also asks for is not in this PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Admin Panel sign-in reliability when the page application loads slowly.
  - Sign-in now waits briefly for the login service to become available and stops retrying after a limited period.
  - Login tokens are safely handled to prevent malformed scripts from disrupting sign-in.
  - Embedded Admin Panel sessions now authenticate automatically without displaying the workspace login form.

### DIFF
--- a/.maestro/tests/assorted/admin-panel.yaml
+++ b/.maestro/tests/assorted/admin-panel.yaml
@@ -1,0 +1,49 @@
+appId: ${APP_ID}
+name: Admin Panel
+onFlowStart:
+  - runFlow: '../../helpers/setup.yaml'
+onFlowComplete:
+  - evalScript: ${output.utils.deleteCreatedUsers()}
+tags:
+  - test-3
+
+---
+- runFlow:
+    file: '../../helpers/login-with-deeplink.yaml'
+    env:
+      USERNAME: ${output.account.adminUser}
+      PASSWORD: ${output.account.adminPassword}
+
+# should open the admin panel from the sidebar
+- extendedWaitUntil:
+    visible:
+      id: 'rooms-list-view'
+    timeout: 60000
+- tapOn:
+    id: 'rooms-list-view-sidebar'
+- extendedWaitUntil:
+    visible:
+      id: 'sidebar-view'
+    timeout: 60000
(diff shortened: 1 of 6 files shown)


OtavioStasiak (approved):
LGTM

PR #7666, opened 2026-09-09
refactor: select room fields instead of the whole room in RoomView

diegolmello:
## Proposed changes

RoomStore published a hand-maintained `roomUpdate` snapshot so consumers reading the whole WatermelonDB Room model would re-render at all: the model is mutated in place, so a whole-room selector never trips `Object.is`. Any column missing from the observed list was silently stale, which is why the Omnichannel source icon and the `departmentId` behind return-to-queue and close-chat did not update.

RoomStore now keeps the live model in the `room` slot and publishes every emission of `room.observe()`. Consumers select the fields they use, so Zustand re-renders each reader exactly when its own field changes. `roomUpdate`, `roomObservedFields`, `TRoomObservedField`, `TRoomObservedFields`, the observed column map, `useRoomWithUpdateFromStore`, `useRoomWithUpdate`, `subscribed` and `lastMessageFromAgent` are deleted, along with `getRoomHeaderFields`, `getRoomHeaderProps` and `useComposerRoom`. ComposerStore stops mirroring the Room and takes `roomTitle` and `t` as scalars.

Row lookup mirrors develop: find the Subscription row once by rid, observe the record if present, otherwise observe a query on the rid until a row appears and then switch. On record destroy the observable completes and the store sets `joined: false` for non-DM rooms. Navigation away on removal stays with the `ROOM_REMOVED` handler.

Two live defects fall out of the conversion: the Omnichannel source icon and department are now reactive, and the master-detail avatar in `LeftButtons` updates on rename, which it never did.

## Issue(s)

Room-reads follow-up to #7482, implemented from its head `e2333f49cc`.

## How to test or reproduce

- `pnpm format-lint`
- `TZ=UTC pnpm test`

Passing on this branch: 313 suites, 2850 tests. TypeScript clean.

The store tests cover a present row, DM vs non-DM completion, DM vs non-DM absent row, the query-to-record switch with the query unsubscribed, a synchronous query emission, and the unchanged `init` flow including retry and abort. `RoomStoreContext` gains the invariant this PR rests on: a field selector re-renders when that field is mutated in place and re-emitted, and does not re-render when an unrelated field changes. Fixtures are plain objects typed as `TRoomOrPreview`, mutated between emissions, with no WatermelonDB internals.

Not tested on a device. The re-render count comparison called for by the plan is not in this PR.

## Types of changes

## Checklist

## Further comments

One deliberate exception to the reader rule: `RoomMessageActions` reads `getState().room` during render to hand `MessageActions` the full model. It is safe because the value is the live instance and every room field `MessageActions` uses is read at press time, but it is a render read rather than an imperative one, and it means the rule cannot be enforced by grep alone. Moving the read into `MessageActions` at action time is the clean fix and is out of scope here.

`RoomGate` previously re-rendered on every tracked field, which kept `EncryptedRoom`'s title and `InvitedRoomScreen`'s invitation data fresh as a side effect. Those now come from explicit field selectors; `getInvitationData` was split into a pure text half usable inside a selector and an actions half read imperatively at press time.

`useSubscriptionUnreads` and `useGoRoomActionsView` stopped returning the model and read it imperatively in their press-time handlers instead.

No lint rule forbidding whole-room reads in render is included; the guard is code review, as agreed in the plan.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when room details change across message lists, headers, controls, banners, encryption notices, and footers.
  * Improved handling of Preview, Invited, and Subscribed room states, including room loading and join controls.
  * Improved invitation screens with clearer room information and separate accept/reject actions.
  * Improved message composer placeholders and canned-response availability when room names or types change.
  * Improved room and thread loading, including invitation, media, and notification interactions.
* **Tests**
  * Expanded coverage for room changes, invitations, message updates, and room controls.


app/views/RoomView/components/RoomMessageActions.tsx
```diff
@@ -17,13 +18,17 @@ export const RoomMessageActions = ({
...
+	const isSubscribed = useRoomStore(s => 'id' in s.room);
 	const user = useAppSelector(getUserSelector);
 	const readOnly = useReadOnly();
 
-	if (!('id' in room)) {
+	if (!isSubscribed) {
 		return null;
 	}
+
+	const room = roomStore.getState().room as TSubscriptionModel;
```
diegolmello, line 31: getState() on render?

app/views/RoomView/hooks/useCloseBanner.ts
```diff
@@ -1,17 +1,19 @@
...
-		if ('id' in room) {
-			try {
-				const db = database.active;
-				await db.write(async () => {
-					await room.update(r => {
-						r.bannerClosed = true;
-					});
-				});
-			} catch {}
+		const { room } = roomStore.getState();
```
diegolmello, line 7: getState on render? is this used on header?

app/views/RoomView/hooks/useReadOnly.ts
```diff
@@ -1,17 +1,17 @@
...
-		return false;
-	}
-
-	return isReadOnlySync(room as Partial<ISubscription>, user.username as string, postReadOnlyPermission, user.roles ?? []);
+	return useRoomStore(s => {
+		const { room } = s;
+		if (!('id' in room)) {
+			return false;
+		}
+		return isReadOnlySync(room, user.username as string, postReadOnlyPermission, user.roles ?? []);
```
diegolmello, line 16: Why do we need both read-only and read-only sync?

app/views/ShareView/index.tsx
```diff
@@ -395,7 +396,7 @@ class ShareView extends Component<IShareViewProps, IShareViewState> {
 					<ComposerProvider
 						rid={room.rid}
 						t={room.t}
-						room={room}
+						roomTitle={getRoomTitle(room)}
```
diegolmello, line 399: We call this on every render?

diegolmello:
### What changes, in one picture

Reads used to go through a hand-maintained snapshot beside the live model. Now they go through `room.observe()` and per-field selectors.

```diff
 RoomStore
-  room            # live WatermelonDB model, mutated in place
-  roomUpdate      # hand-built snapshot of an observed column list
-  observe(room.observe(...roomObservedFields))
-    on emission -> rebuild roomUpdate
+  room            # live model, republished on every emission
+  observe(room.observe())
+    on emission -> set({ room })

 consumers
-  useRoomWithUpdate() / useRoomWithUpdateFromStore()
-    -> whole room + roomUpdate, re-render on any observed column
+  useRoomStore(room => room.<field>)
+    -> re-render only when that field changes
```

Why the old shape could not work for a whole-room read: the model is mutated in place, so a selector returning `room` never trips `Object.is`. Freshness came entirely from `roomUpdate`, and any column missing from `roomObservedFields` was silently stale.

```mermaid
sequenceDiagram
    participant DB as WatermelonDB
    participant Store as RoomStore
    participant A as Header
    participant B as Footer
    DB->>Store: room.observe() emission
    Store->>Store: set({ room })
    Store->>A: selector(room => room.fname) changed -> render
    Store->>B: selector(room => room.ro) same -> no render
```

### Row lookup

```text
findSubscriptionByRid(rid)
├── row present  -> record.observe()
└── row absent   -> query(rid).observe()
                      └── first row -> switch to record.observe(), unsubscribe query
on destroy -> observable completes -> joined: false (non-DM)
```

### Deleted with the snapshot

```text
roomUpdate            roomObservedFields        TRoomObservedField(s)
useRoomWithUpdate     useRoomWithUpdateFromStore  observed column map
subscribed            lastMessageFromAgent      getRoomHeaderFields
getRoomHeaderProps    useComposerRoom
```

Two live defects fall out of the conversion: the Omnichannel source icon and `departmentId` are now reactive, and the master-detail avatar in `LeftButtons` updates on rename.

PR #7685, opened 2026-09-14
chore: replace react-native-background-timer with setInterval

diegolmello:
## Proposed changes

Drops `react-native-background-timer` from the video-conference keep-alive timer and uses plain `setInterval`.

The library is unmaintained (last release 2021) and the timer it backs only refreshes the Jitsi timeout every 10s while a call is in progress. `stopBackgroundTimer` / `clearInterval` bookkeeping collapses into a single `clearInterval`.

Behaviour note, needs a call: this is not like-for-like. `react-native-background-timer` deliberately kept the keepalive firing while the app was backgrounded (iOS background task, Android native handler outside RN's Timing module). An RN JS timer is suspended on host pause on Android and stops firing on iOS once the app is backgrounded. The consumer is `JitsiMeetView`, which runs the call in an in-app WebView with `activateKeepAwake()`, so the screen stays foregrounded for the duration of a normal call; but if the user backgrounds the app mid-call the server-side Jitsi timeout is no longer refreshed and the conference can be reaped. The refresh resumes on foreground. The "Jitsi native app installed" path is unaffected: `handleJitsiApp` calls `goBack()`, which unmounts and runs `endVideoConfTimer`.

## Issue(s)

https://rocketchat.atlassian.net/browse/NATIVE-1570

## How to test or reproduce

1. Start a Jitsi video conference in a room.
2. Stay in the call for over 20s and confirm it is not reaped server-side (the timeout keeps refreshing).
3. Leave the call and confirm the refresh stops.

Both platforms.

## Types of changes

## Checklist

## Further comments

Ported from the closed RN 0.83 / Expo 55 upgrade branch so the de-interop work lands independently of the framework bump.

## Summary by CodeRabbit

- **Behavior Changes**
  - Video conference timers now run only while the app is active.
  - Timer updates continue at 10-second intervals during active use.
  - Timers are stopped cleanly when the video conference ends.
  - Timer behavior is now consistent with standard in-app scheduling and does not continue while the app is in the background.


OtavioStasiak (approved):
LGTM!

PR #7699, opened 2026-09-16
feat: apps engine action buttons in the composer action sheet

OtavioStasiak:
## Proposed changes

Brings Apps-Engine **action buttons** to the mobile composer, including the new `ai` category
  used by AI apps.

  **What's new**

  - `app/lib/apps/` — a self-contained module for Apps-Engine UI buttons:
    - `appsStore.ts`: Zustand store holding action buttons + app translations, fed by
      `/api/apps/actionButtons` and `/api/apps/languages`, kept fresh through the `stream-apps`
      DDP stream (`actions/changed`, `app/added|removed|updated`). Subscription is ref-counted, so
      several consumers share a single stream.
    - `filters.ts`: ports the server's `when` conditions — room type (channel/team/discussion/DM/
      livechat), category, roles and permissions.
    - `translations.ts`: resolves `labelI18n` against the app's own language pack, with
      `pt-BR → pt → en` fallback.
    - `useAppActionButtons.ts`: hook that returns the buttons visible for a given context/category/
      room, already filtered and translated.
    - `definitions.ts`: local mirror of `@rocket.chat/apps-engine/definition/ui` (the package isn't a
      dependency here).
  - `ActionsButton.tsx` — the composer action sheet now lists `roomAction` buttons of category `ai`
    (`stars` icon) and `messageBoxAction` buttons (`apps` icon). Message-box actions forward the
    current composer text to the app.
  - UIKit — new `actionButton` interaction type and `tmid` propagation in `triggerAction`, so an
    action triggered from a thread composer lands in the right thread.
  - Unknown modal interaction types no longer throw; they resolve to `unsupported` and surface a
    toast instead (`App_action_unsupported` / `App_action_error`, translated in all locales).

  **Tests**: unit coverage for the filters, translation fallback, the hook, the interaction adapter
  and the composer action sheet.
  
## Issue(s)	

https://rocketchat.atlassian.net/browse/NATIVE-1642

## How to test or reproduce

- Open the app;
- Login in a workspace with apps engine action buttons (like open);
- Click on PLUS icon on composer;
- Tap on a action button;

## Screenshots
| Before | After |
| ----- | ----- |
| Simulator Screenshot - iPhone 16 - 2026-09-17 at 12 25 20 | Simulator Screenshot - iPhone 16 - 2026-09-17 at 12 24 52 |

## Types of changes

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added app and AI action buttons to the message composer action sheet.
  * Actions are filtered by room type, category, permissions, and user roles.
  * Added localized app-action labels and mobile support messages across supported languages.
  * Added support for triggering app actions from messages, threads, rooms, and composer drafts.
* **Improvements**
  * Unsupported or failed app actions now provide clear feedback.
  * UIKit interactions now support action buttons and preserve thread context.
* **Tests**
  * Added coverage for app actions, filtering, translations, and poll interactions.


app/lib/apps/appsStore.ts
```diff
@@ -0,0 +1,115 @@
...
+				return acc;
+			}, {});
+			set({ translations });
+		} catch (e) {
+			set({ translations: {} });
+			log(e);
+		}
+	},
+
+	reset: () => set(initialState)
```
diegolmello, line 55: **`reset()` has no callers, so apps state survives logout and workspace switch**

### What is happening

`reset` is declared and implemented here, but nothing in the app ever calls it. I grepped the whole `app/` tree: the only `.reset()` calls belong to `useCallStore`, `mediaSessionInstance` and `usePeerAutocompleteStore`. `useAppsStore` is referenced only by the action buttons hook, never to reset it.

`actionButtons` and `translations` live in a module level store, so they outlive any screen. Nothing tears them down when the user logs out or switches to another workspace.

### The simple version

The list of app buttons is written on a whiteboard that belongs to the whole app, not to one screen. When you leave one workspace and walk into another, nobody wipes the whiteboard. So for a moment you are looking at the first workspace's buttons while already standing in the second one. Tapping one sends the request to the new workspace, which has never heard of that app.

### Steps to reproduce

1. Log into workspace A, which has an app that contributes a composer action button.
2. Open any room and tap the composer `+` button. The app button is listed.
3. Log out and log into workspace B, which does not have that app installed.
4. Open a room and tap `+` before the first fetch for workspace B lands.
5. Workspace A's button is still listed. Tapping it posts to `/api/apps/ui.interaction/<appId>` on workspace B, for an app id that does not exist there.

### Proposed change

There is already a precedent for this teardown: `mediaSessionInstance.reset()` is called from the disconnect path in `app/lib/services/connect.ts` and from the logout saga in `app/sagas/login.js`. Call the apps store reset alongside it:

```ts
useAppsStore.getState().reset();
```

If we would rather rely on the refetch that the next mount triggers, then delete `reset` instead. Shipping a public store action with zero callers leaves the reader guessing which of the two was intended.

app/lib/apps/appsStore.ts
```diff
@@ -0,0 +1,115 @@
...
+					if (current !== generation || consumers === 0) {
+						subscription?.unsubscribe().catch(log);
+						return;
+					}
+					streamSubscription = subscription ?? null;
+				})
+				.catch(log);
+		} catch (e) {
+			log(e);
+		}
```
diegolmello, line 101: **Nothing retries setup after the first attempt fails, and the comment says the first attempt can fail**

### What is happening

Setup runs exactly once, when the consumer count goes from 0 to 1. Every failure path here only logs:

- `fetchActionButtons().catch(log)` and `fetchTranslations().catch(log)` swallow a failed fetch.
- The `try` wraps `sdk.onStreamData`, which the comment on the line above states throws synchronously while no server is selected.
- `sdk.subscribe(...).catch(log)` swallows a failed subscribe, leaving `streamSubscription` unset.

After any of these, `consumers` is still 1, so the `if (consumers === 1)` gate never opens again. There is no connection callback, retry or backoff that re-enters this block. The store stays empty and unsubscribed for the rest of the session.

### The simple version

The code knocks on the door once. The comment right above admits that nobody may be home yet on the first knock. When nobody answers, it writes the failure in a notebook and never knocks again. Meanwhile the door does open a second later, but the code has already stopped trying.

### Steps to reproduce

1. Cold start the app so the composer mounts before the DDP connection is established.
2. `sdk.onStreamData` throws, the `catch` logs it, and setup stops.
3. Open any room with an app that contributes composer buttons.
4. Tap `+`. No app buttons are listed.
5. They stay missing for the whole session. Pulling to refresh or navigating between rooms does not help, because the composer stays mounted and `consumers` never returns to 0. Only killing and reopening the app recovers.

### Proposed change

Drive setup off the connection signal rather than off first mount, which is how the room stream subscribers already handle this. Keep a flag for whether setup succeeded, separate from the consumer count, and re-enter setup when the connection comes up:

```ts
let subscribed = false;

const setupApps = async (): Promise<void> => {
	if (subscribed || consumers === 0) {
		return;
	}
	const current = generation;
	const { fetchActionButtons, fetchTranslations } = useAppsStore.getState();
	await Promise.all([fetchActionButtons(), fetchTranslations()]);
	streamListener = sdk.onStreamData(APPS_STREAM, handleStreamData);
	const subscription = await sdk.subscribe(APPS_STREAM, APPS_EVENT);
	if (current !== generation || consumers === 0) {
		subscription?.unsubscribe().catch(log);
		return;
	}
	streamSubscription = subscription ?? null;
	subscribed = true;
};
```

Then call `setupApps` both from `subscribeToApps` and from the existing connected handler, and clear `subscribed` in the disposer. The `generation` guard stays as is, it correctly covers a fast unmount and remount while a subscribe is still in flight.

app/containers/MessageComposer/components/Buttons/ActionsButton.tsx
```diff
@@ -95,6 +99,30 @@ export const ActionsButton = () => {
...
+			options.push({
+				title: label,
+				icon: 'apps',
+				danger: button.variant === 'danger',
+				testID: `message-composer-app-action-${id}`,
+				onPress: () => {
+					triggerAppActionButton({ button, rid, tmid, message: getText() ?? '' });
+				}
+			});
+		});
```
diegolmello, line 124: **App action options do not dismiss the action sheet before triggering**

### What is happening

The camera, video and library options in this same list call `hideActionSheet()` inside `onPress` and then defer the real work:

```ts
onPress: () => {
	hideActionSheet();
	setTimeout(() => takePhoto(), 550);
}
```

The app action and AI action options do not. They call `triggerAppActionButton` while the sheet is still on screen. That call can navigate to `ModalBlockView`, so a modal is pushed on top of an open action sheet.

The 550ms timeouts on the neighbouring options exist precisely because presenting something over a sheet that is still animating out does not work reliably. The app actions skip both halves of that pattern.

### The simple version

The other buttons close the drawer before opening a new window. These ones open the window while the drawer is still sticking out, then the drawer tries to close underneath it. Sometimes you end up with the window stuck behind the drawer, or neither one responding to taps.

### Steps to reproduce

1. Install an app whose composer action button responds with a UIKit modal.
2. Open a room and tap the composer `+`.
3. Tap the app's action button.
4. `ModalBlockView` is presented while the action sheet is still open. Depending on timing the modal appears behind the sheet, or the sheet stays on screen over the modal and swallows taps.

`Choose_file` and `Create_Discussion` also skip `hideActionSheet()`, so this is not unprecedented in this file, but neither of those presents a modal on top of the sheet, which is what makes this case different.

### Proposed change

Match the pattern the presenting options already use, for both the AI and the app branch:

```ts
onPress: () => {
	hideActionSheet();
	triggerAppActionButton({ button, rid, tmid, message: getText() ?? '' });
}
```

If the modal turns out to race the sheet dismissal animation, wrap the trigger in the same `setTimeout(..., 550)` the camera and library options use rather than inventing a different delay.

app/containers/MessageComposer/components/Buttons/ActionsButton.tsx
```diff
@@ -95,6 +99,30 @@ export const ActionsButton = () => {
...
 		});
 
+		aiActions.forEach(({ id, label, button }) => {
+			options.push({
+				title: label,
+				icon: 'stars',
+				danger: button.variant === 'danger',
+				testID: `message-composer-ai-action-${id}`,
+				onPress: () => {
+					triggerAppActionButton({ button, rid });
```
diegolmello, line 109: **`tmid` is passed for message box actions but not for room actions. Intentional?**

### What is happening

The two branches build different payloads. This one, for `UIActionButtonContext.ROOM_ACTION`, sends only the room:

```ts
triggerAppActionButton({ button, rid });
```

The message box branch just below sends the thread id and the composer text:

```ts
triggerAppActionButton({ button, rid, tmid, message: getText() ?? '' });
```

### The simple version

Both buttons live in the same menu, and that menu knows which thread you have open. One button passes that information along, the other one drops it. If you are inside a thread and tap the second kind of button, the app on the other end hears "this happened in the room" and never learns which thread.

### Steps to reproduce

1. Open a thread in a room that has an app contributing a room action button.
2. Tap the composer `+` and then the room action button.
3. The interaction payload reaches the app without `tmid`, so any response the app sends lands in the main room rather than the thread.

### Proposed change

If a room action is genuinely room scoped and should never carry a thread, no code change is needed, but please confirm it here so the asymmetry is on the record. The PR description highlights thread id propagation, which is what made this stand out.

If it should carry the thread when one is open, make the two branches symmetric:

```ts
triggerAppActionButton({ button, rid, tmid });
```

app/lib/apps/useAppActionButtons.ts
```diff
@@ -0,0 +1,128 @@
...
+				permissions: permissionRoles
+			});
+		};
+
+		resolve().catch(log);
+
+		return () => {
+			cancelled = true;
+		};
+	}, [hasButtons, rid, permissionsKey, userRolesKey]);
```
diegolmello, line 108: **The previous room's buttons stay on screen until the new room's data resolves**

### What is happening

When `rid` changes, this effect reruns, but `filterContext` is never cleared. It keeps the old room's `t`, `teamMain`, `prid`, `uids`, roles and permissions until `resolve()` finishes two awaits later: `getSubscriptionByRoomId` and `getPermissionRoles`. The cleanup only sets the local `cancelled` flag, it does not reset state.

So between the room change and the resolve, the memo below filters this room's buttons against the previous room's context.

### The simple version

The filter is a pair of glasses that decides which buttons you are allowed to see. When you walk into a different room, you keep wearing the old room's glasses until the new pair arrives from the other end of the building. For that moment you see the buttons for the room you just left.

### Steps to reproduce

1. Find an app button configured with a room type filter, for example one restricted to private channels.
2. Open a private channel and tap the composer `+`. The button is listed, correctly.
3. Go back and immediately open a public channel, then tap `+` before the subscription and permission lookups finish.
4. The private channel's button is still listed, because the filter is still reading the previous room's context. It disappears once the resolve lands.

The same window applies in the other direction, so a button that should be visible in the new room is missing for that moment.

### Proposed change

Clear the context at the top of the effect. The memo below already returns an empty list when the context is null, so this needs no other change:

```ts
useEffect(() => {
	setContext(null);

	if (!hasButtons) {
		return;
	}
	// ...
}, [hasButtons, rid, permissionsKey, userRolesKey]);
```

The action sheet then shows no app buttons for that short window instead of the wrong ones. Showing nothing briefly is the safer failure, since a button shown in a room it was filtered out of can trigger an interaction the filter was meant to prevent.

app/containers/MessageComposer/components/Buttons/ActionsButton.tsx
```diff
@@ -95,6 +99,30 @@ export const ActionsButton = () => {
 			onPress: () => createDiscussion()
 		});
 
+		aiActions.forEach(({ id, label, button }) => {
```
diegolmello, line 102: I can't see this on web. Also is it correct to target directly AI or it should just target app actions?

app/lib/apps/appsStore.ts
```diff
@@ -0,0 +1,115 @@
...
+const handleStreamData = (ddpMessage: { fields?: { args?: [[string, unknown[]]] } }) => {
+	const [event] = ddpMessage?.fields?.args?.[0] || [];
+	const { fetchActionButtons, fetchTranslations } = useAppsStore.getState();
+	if (event === 'actions/changed') {
+		fetchActionButtons().catch(log);
+	}
+	if (event === 'app/added' || event === 'app/removed' || event === 'app/updated') {
+		fetchActionButtons().catch(log);
+		fetchTranslations().catch(log);
+	}
```
diegolmello, line 75: I think when one of these events happen, it's probably related to either action buttons or translations. So try to separate this so you don't have to do both requests when one of them happens

app/lib/apps/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './actions';
+export * from './definitions';
+export * from './appsStore';
+export * from './triggerActions';
+export * from './triggerAppActionButton';
+export * from './useAppActionButtons';
```
diegolmello, line 6: No barrel files.

app/lib/apps/triggerAppActionButton.ts
```diff
@@ -0,0 +1,36 @@
...
+			mid,
+			payload: { context: button.context, ...(message !== undefined ? { message } : {}) }
+		});
+
+		if (result === ModalActions.UNSUPPORTED) {
+			EventEmitter.emit(LISTENER, { message: I18n.t('App_action_unsupported') });
+		}
+	} catch (e) {
+		log(e);
+		EventEmitter.emit(LISTENER, { message: I18n.t('App_action_error') });
```
diegolmello, line 34: Use emitter instead of this https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/ea8bf3f52b605a09f9222d791de2756795c42da9/app/lib/methods/helpers/emitter.ts#L22

app/lib/apps/useAppActionButtons.ts
```diff
@@ -0,0 +1,128 @@
...
+	const hasButtons = buttons.length > 0;
+
+	useEffect(() => {
+		if (!hasButtons) {
+			return;
+		}
+
+		let cancelled = false;
+
+		const resolve = async (): Promise<void> => {
```
diegolmello, line 85: use a more semantic name for this function.

app/lib/services/restApi.ts
```diff
@@ -1281,3 +1282,31 @@ export const mediaCallsStateSignals = async (contractId: string): Promise<{ sign
...
+	if (!host || !currentLogin) {
+		throw new Error('The Apps REST API requires an initialized, authenticated session');
+	}
+	const { userId, authToken } = currentLogin;
+
+	const response = await fetch(`${host}/api/apps/${path}`, {
+		method: 'GET',
+		headers: {
+			...RocketChatSettings.customHeaders,
+			'Content-Type': 'application/json',
```
diegolmello, line 1297: I don't think you need this. https://github.com/RocketChat/Rocket.Chat.ReactNative/blob/ea8bf3f52b605a09f9222d791de2756795c42da9/app/lib/methods/helpers/fetch.ts#L43

diegolmello (changes requested):

app/containers/MessageComposer/components/Buttons/ActionsButton.tsx
OtavioStasiak: category: 'ai' is the right target, it mirrors web. On web these are roomAction buttons filtered by 
  ▎ useApplyButtonFilters('ai') in useAppsRoomStarActions.tsx, rendered as the featured  "AI Actions" menu in the room header 
  ▎ toolbar, which is why you don't see them in the composer there. The web composer toolbar only shows messageBoxAction 
  ▎ (default category). On mobile we surface both in the composer + sheet, since we don't have that header toolbar slot  the filtering itself is identical.

